PR #1, opened 2014-08-21
Jhelbert/enrollment blacklist

jhelbert:


jhelbert:
@mulby @brianhw

edx/analytics/tasks/daily_enrollment_trends.py
```diff
@@ -60,26 +60,65 @@ def requires(self):
...
+    blacklist_date = luigi.Parameter()
+    @property
+    def table_name(self):
+        return 'course_enrollment_blacklist'
+
+    @property
+    def columns(self):
+        return [
+            ('course_id', 'STRING'),
+            ('date', 'STRING'),
```
mulby: So we could actually take care of a Hive feature here, which is that the data doesn't actually have to match the schema.

You should be able to just specify the single course_id column

edx/analytics/tasks/daily_enrollment_trends.py
```diff
@@ -60,26 +60,65 @@ def requires(self):
...
             PARTITION (dt='{partition_date}')
-            SELECT course_id, date, sum(enrollment_delta)
-            OVER (
-                partition by course_id
-                order by date ASC
-                ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW
-            )
-            FROM {incremental_table_name};
+            SELECT e.*
+
```
mulby: extra new line here?

setup.cfg
```diff
@@ -37,6 +37,7 @@ edx.analytics.tasks =
     database-import = edx.analytics.tasks.database_imports:ImportAllDatabaseTablesTask
     location-per-course = edx.analytics.tasks.location_per_course:LastCountryOfUser
     course-facts-to-mysql = edx.analytics.tasks.daily_enrollment_trends:ImportCourseDailyFactsIntoMysql
+    import-enrollment-blacklist = edx.analytics.tasks.daily_enrollment_trends:ImportCourseEnrollmentOffsetsBlacklist
```
mulby: You shouldn't need this...

jhelbert:
@mulby @brianhw

edx/analytics/tasks/daily_enrollment_trends.py
```diff
@@ -60,26 +60,63 @@ def requires(self):
         )
 
 
+class ImportCourseEnrollmentOffsetsBlacklist(ImportIntoHiveTableTask):
+    blacklist_date = luigi.Parameter()
```
mulby: default from config file

edx/analytics/tasks/daily_enrollment_trends.py
```diff
@@ -60,26 +60,63 @@ def requires(self):
...
+        return self.blacklist_date
+
+    @property
+    def table_format(self):
+        """Provides format of Hive external table data."""
+        return "ROW FORMAT DELIMITED FIELDS TERMINATED BY '\\t'"
+
+
 class SumEnrollmentDeltasTask(BaseCourseEnrollmentTaskDownstreamMixin, ImportIntoHiveTableTask):
+    blacklist_date = luigi.Parameter()
```
mulby: same default as above

client.cfg
```diff
@@ -27,3 +27,6 @@ input_format = oddjob.ManifestTextInputFormat
 
 [answer-distribution]
 valid_response_types = choiceresponse,optionresponse,multiplechoiceresponse,numericalresponse,stringresponse,formularesponse
+
+[user-activity]
+blacklist_date = 2014-08-20
```
mulby: This needs to go in edx-ops/analytics-secure, we shouldn't be modifying client.cfg anymore except to add "global defaults" that you likely will not need to change.

Also you should add a similar value to the test.cfg file in this repo, and make sure that no unit tests are actually reading the real file, so set it to some date for which there is no file.

jhelbert: Ok, I need to get access to edx-ops/analytics-secure

mulby: Are you sure you don't have access? I thought the whole team did...

https://github.com/edx-ops/analytics-secure

jhelbert: no access.  clicking that gives me a 404

and this https://github.com/edx-ops says im not a member

brianhw: It had been the same for me -- I had to put in a support ticket to get
access from devops.

edx/analytics/tasks/daily_enrollment_trends.py
```diff
@@ -60,26 +60,71 @@ def requires(self):
...
+
+    @property
+    def table_name(self):
+        return 'course_enrollment_blacklist'
+
+    @property
+    def columns(self):
+        return [
+            ('course_id', 'STRING'),
+            ('offsets', 'INT'),
```
brianhw: What is 'offsets' storing here?  The input file should be documented in the class docstring, possibly with an example.  For example, mentioning that the file is assumed to exist at <table_location>/<partition_date>, and is tab-separated.

config/test.cfg
```diff
@@ -38,6 +38,7 @@ path = s3://fake/manifest/
 
 [user-activity]
 output_root = s3://fake/activity/
+blacklist_date = 2014-08-18
```
brianhw: Not clear why this goes in 'user-activity'.   Seems like it should be more related to enrollment.

brianhw:
LGTM.   :+1:   after squashing, etc., assuming you've also checked PEP8/Pylint.

mulby:
:+1: after squash and quality as well

PR #2, opened 2014-08-21
Build a course_fact table

mulby:
Early sketch of the kind of thing I'm thinking.

To hopefully identify any early misalignment.


mulby:
@jhelbert @brianhw

edx/analytics/tasks/course.py
```diff
@@ -0,0 +1,246 @@
...
+    def query(self):
+        query_format = textwrap.dedent("""
+            USE {database_name};
+            DROP TABLE IF EXISTS {table_name};
+            CREATE EXTERNAL TABLE {table_name} (
+                date STRING,
+                course_id STRING,
+                gender STRING,
+                age INT,
+                level_of_education INT,
```
brianhw: Why is this an int?  What is the code?

edx/analytics/tasks/course.py
```diff
@@ -0,0 +1,246 @@
...
+    def table_location(self):
+        return url_path_join(self.warehouse_path, self.table_name + '/')
+
+    @property
+    def table_format(self):
+        return "ROW FORMAT DELIMITED FIELDS TERMINATED BY '\t'"
+
+    @property
+    def partition_date(self):
+        return '2014-08-21'
```
brianhw: parametrize?

edx/analytics/tasks/course.py
```diff
@@ -0,0 +1,246 @@
+
```
brianhw: Probably would have preferred if this were called course_fact.py.

edx/analytics/tasks/course.py
```diff
@@ -0,0 +1,246 @@
...
+    def table_name(self):
+        return 'user_course_activity_history'
+
+    @property
+    def columns(self):
+        return [
+            ('date', 'STRING'),
+            ('course_id', 'STRING'),
+            ('user_id', 'INT'),
+            ('active', 'INT'),
```
brianhw: is there a BOOL type that would make better sense to use?

edx/analytics/tasks/course.py
```diff
@@ -0,0 +1,246 @@
...
+            )
+            ROW FORMAT DELIMITED FIELDS TERMINATED BY '\t'
+            LOCATION '{location}';
+
+            INSERT OVERWRITE TABLE {table_name}
+            SELECT
+                date,
+                course_id,
+                user_id,
+                MAX(active) OVER (last_week),
```
brianhw: did you check to see if there were an "OR"?

brianhw:
Looks good! 

Small boundary question.  What happens if a user unenrolls on a day and also has activity on the same day?  With current logic, there would be no entry for the user on that day because at the end of it, they were not enrolled.  So then their activity would not get counted....

edx/analytics/tasks/course.py
mulby: These were just temporary tables I used to emulate the output from your job and @jhelbert's job.

I suppose we could actually use them? In which case, we should parameterize them...

edx/analytics/tasks/course.py
mulby: Seems reasonable, I'll rename this.

edx/analytics/tasks/course.py
mulby: Yeah, I couldn't find the UDAF functions I wanted for the BOOL type. Also, using INT means we can do SUM(), which turns out to be useful.

edx/analytics/tasks/course.py
mulby: Yeah, couldn't find one.

mulby:
Yeah that boundary condition would be problematic.

What if we made the following changes:
- Include an enrollment record if the user was _ever_ enrolled in the course on that day.
- Add a column that is equivalent to "is_active" that indicates if the user was still enrolled in the course at the end of that day.

Additionally, I'd like to change the schema of that table to be more similar to the activity table in that it would have a separate column for each enrollment mode that would either be 0 or 1. That way I can do:

``` sql
SELECT
    uce.date, uce.course_id,
    SUM(uce.enrolled), SUM(uce.mode_verified), SUM(uce.mode_audit)
FROM user_course_enrollment_history uce
GROUP BY uce.date, uce.course_id
```

Which avoids the need to do left join the table with itself, transforming my logic to:

``` sql
            SELECT
                uce.date,
                uce.course_id,
                pr.gender,
                (year(uce.date) - pr.year_of_birth) AS age,
                pr.level_of_education,
                SUM(uce.enrolled),
                SUM(uce.mode_verified),
                SUM(act.active) AS active_prev_week,
                SUM(act.attempted_problem) AS attempted_problem_prev_week,
                SUM(act.played_video) AS played_video_prev_week
            FROM user_course_enrollment_history uce
            LEFT OUTER JOIN auth_userprofile pr
                ON uce.user_id = pr.user_id
            LEFT OUTER JOIN user_course_activity_weekly_history act
                ON  act.date = uce.date
                AND act.course_id = uce.course_id
                AND act.user_id = uce.user_id
            WHERE
                "{start_date}" <= uce.date AND uce.date < "{end_date}"
            GROUP BY uce.date, uce.course_id, pr.gender, (year(uce.date) - pr.year_of_birth), pr.level_of_education;
```

mulby:
To be clear the new schema for user_course_enrollment_history would be:

| date | course_id | user_id | enrolled | mode_verified | mode_audit | mode_honor | mode_prof |
| --- | --- | --- | --- | --- | --- | --- | --- |
| YYYY-mm-dd | OrgX/Course/Run | N | 0 or 1 | 0 or 1 | 0 or 1 | 0 or 1 | 0 or 1 |

mulby:
@brianhw see ideas above

jhelbert:
:thumbsup:

brianhw:
Okay, starting to think about this again.  A couple of random questions:
- The questions raised earlier today on Hipchat about rows vs. columns seem to me to be about the efficiency of avoiding duplication of the "dimensions".  That is, an alternative for the query above would end up writing to a separate course_activity_fact table, looking like:

``` sql
            SELECT
                uce.date,
                uce.course_id,
                pr.gender,
                (year(uce.date) - pr.year_of_birth) AS age,
                pr.level_of_education,
                act.label,
                SUM(act.count) AS active_prev_week
            FROM user_course_enrollment_history uce
            LEFT OUTER JOIN auth_userprofile pr
                ON uce.user_id = pr.user_id
            LEFT OUTER JOIN user_course_activity_weekly_history act
                ON  act.date = uce.date
                AND act.course_id = uce.course_id
                AND act.user_id = uce.user_id
            WHERE
                "{start_date}" <= uce.date AND uce.date < "{end_date}"
            GROUP BY uce.date, uce.course_id, pr.gender, (year(uce.date) - pr.year_of_birth), pr.level_of_education, act.label;
```

But that means not keeping all course facts in the same table, such as the enrollment, etc.  But it would allow the actions to be more easily expanded, without having each new action having to be encoded into the schema itself.
- The issue with enrollment transitions bothers me.  It seems likely that we would have other cases where we want to count state at the end of the day, and yet also account for other states during the day.  The most pressing is enrollment itself, since we are planning on dealing with that as a primary filter for all other stats.  But it seems to me we would end up with problems with other queries if there were similar ties between "current" state and events.  For example, if it mattered whether a user was verified, and there were transitions between audit and verified and back again, how would we model those?  Would we also need a "was verified during this day" flag?  Yet we need to keep the counts defined at some time boundary, like end-of-day, so that the various categories remain mutually exclusive so that the numbers add up correctly.   We can't count a user as being both unenrolled and enrolled.  Anyway, you made a suggestion, but I'm assuming it's not incorporated into your examples -- for that one would rename "enrolled" column to be "is_active".  But I'm fine to leave it as you suggested, as "enrolled" being a flag as to whether they were enrolled at the end of the day.  And the presence of a record with all 0's indicates that they were enrolled at some time during the day (not necessarily the beginning either) but ultimately unenrolled.  But to be more explicit, we might prepend "end_" to each of the columns, to make it clearer (e.g. "end_enrolled", "end_mode_verified", etc.).

mulby:
> But that means not keeping all course facts in the same table, such as the enrollment, etc. But it would allow the actions to be more easily expanded, without having each new action having to be encoded into the schema itself.

I guess I don't really see this value as terribly compelling. I would be surprised if we are frequently changing the types of activity that we care about. I've also found that doing rollups and aggregations is simplified when you don't need a "where" clause and can simply select columns, particularly when joining. Take the example of the enrollment table (where I think it's more clear), having a separate column for the various modes allows for much simplified rollup when compared to separate records which require additional joins.

>  It seems likely that we would have other cases where we want to count state at the end of the day, and yet also account for other states during the day.

This is possible, however, I see it as a bridge we can cross when we get there. I like your suggestion of representing this with a different flag "users_ever_verified" to indicate the user was verified at _some point_ during that day vs. "users_verified" which indicates a user who was in the verified mode at the end of that day. I was planning on tracking other types of transitions explicitly as well "users_change_to_verified" to indicate the number of users who changed mode to verified in the course of that day.

> for that one would rename "enrolled" column to be "is_active"

Yeah I don't really care what we call it, I think we're on the same page about what it represents though.

PR #3, opened 2014-08-22
Jhelbert/daily historical activity

jhelbert:


edx/analytics/tasks/user_activity.py
```diff
@@ -192,6 +194,122 @@ def reducer(self, key, values):
...
+            ('username', 'STRING'),
+            ('active', 'INT'),
+            ('attempted_problem', 'INT'),
+            ('played_video','INT')
+        ]
+
+    @property
+    def table_location(self):
+        output_name = 'user-activity-per-day/'
+        return url_path_join(self.output_root, output_name)
```
mulby: I feel like we should store all of our hive tables in one place. That's why I introduced the concept of the warehouse path in my patch. Then we can just have a single directory for each table.

edx/analytics/tasks/user_activity.py
```diff
@@ -192,6 +194,122 @@ def reducer(self, key, values):
...
+
+    @property
+    def table_format(self):
+        """Provides structure of Hive external table data."""
+        return "ROW FORMAT DELIMITED FIELDS TERMINATED BY '\\t'"
+
+    @property
+    def partition_date(self):
+        return str(self.interval)
```
mulby: A couple of notes about this:
- It's not really a date as much as a date range, so returning that from "partition_date" is counter-intuitive.
- The partition name is "dt" which again indicates a single date, we should probably call the partition something else if, in fact, we are going to partition by date range instead of date.

edx/analytics/tasks/user_activity.py
```diff
@@ -192,6 +194,122 @@ def reducer(self, key, values):
...
+            source=self.source,
+            interval=self.interval,
+            output_root=self.output_root,
+            pattern=self.pattern,
+            n_reduce_tasks=self.n_reduce_tasks
+        )
+
+
+class UserActivityPerDayTask(UserActivityBaseTask):
+    """Make a basic task to gather activity per user for a single time interval."""
```
mulby: This comment is out of date.

edx/analytics/tasks/user_activity.py
```diff
@@ -192,6 +194,122 @@ def reducer(self, key, values):
...
 
 
+class ImportDailyUserActivityToHiveTask(UserActivityBaseTaskDownstreamMixin,
+        EventLogSelectionDownstreamMixin, ImportIntoHiveTableTask):
+    """
+    Creates a Hive Table that points to Hadoop output of CourseEnrollmentChangesPerDay task.
+
+    Parameters are defined by :py:class:`BaseCourseEnrollmentTaskDownstreamMixin`.
+    """
+    n_reduce_tasks = luigi.Parameter(default=10)
```
mulby: why this and not the complete map reduce task mixin?

edx/analytics/tasks/user_activity.py
```diff
@@ -192,6 +194,122 @@ def reducer(self, key, values):
...
+        interval_start = '-'.join(interval_nums[0:3])
+        interval_start_date = datetime.strptime(interval_start,"%Y-%m-%d").date()
+        interval_end = '-'.join(interval_nums[3:6])
+        interval_end_date = datetime.strptime(interval_end,"%Y-%m-%d").date()
+        results = {}
+        interval_index_date = interval_start_date
+
+        # create entries for every date in the interval for each hey
+        # active/problem/video values iniitialized to 0
+        while interval_index_date <= interval_end_date:
```
mulby: This should be "<". Luigi date intervals are open intervals.

edx/analytics/tasks/user_activity.py
```diff
@@ -192,6 +194,122 @@ def reducer(self, key, values):
...
+        return (course_id, username)
+
+    def get_mapper_value(self, _course_id, _username, _date_string, label):
+        return (label, _date_string)
+
+    def reducer(self, key, values):
+        """Outputs labels and usernames for a given course and interval."""
+        course_id, username, date_string = key
+
+        active = int("ACTIVE" in values)
```
mulby: I think these strings are stored in variables somewhere, we should use those variables instead of the string.

edx/analytics/tasks/user_activity.py
```diff
@@ -192,6 +194,122 @@ def reducer(self, key, values):
...
+        # create entries for every date in the interval for each hey
+        # active/problem/video values iniitialized to 0
+        while interval_index_date <= interval_end_date:
+            results[str(interval_index_date)] = [str(interval_index_date),course_id, username,0,0,0]
+            interval_index_date += timedelta(days=1)
+
+        for value in values:
+            label, date_string = value
+
+            labels = ["ACTIVE","ATTEMPTED_PROBLEM", "PLAYED_VIDEO"]
```
mulby: Variables instead of string here too.

edx/analytics/tasks/user_activity.py
```diff
@@ -192,6 +194,122 @@ def reducer(self, key, values):
...
+
+
+    def reducer(self, key, values):
+        """Outputs labels and usernames for a given course and interval."""
+        course_id, username = key
+
+        # create array of results with each date in interval
+        interval_nums = str(self.interval).split('-')
+        interval_start = '-'.join(interval_nums[0:3])
+        interval_start_date = datetime.strptime(interval_start,"%Y-%m-%d").date()
```
mulby: This is going to be really slow, we may want to consider something like this:

``` python
year, month, day = [int(x) for x in interval_start.split('-')]
interval_start_date = datetime.date(year, month, day)
```

Normally I wouldn't recommend performance improvements like this, however, we already know strptime is really slow and the alternative seems pretty simple.

edx/analytics/tasks/user_activity.py
```diff
@@ -192,6 +194,122 @@ def reducer(self, key, values):
...
+        )
+
+    def get_predicate_labels(self, event):
+        return extract_predicate_labels(event)
+
+    def get_mapper_key(self, course_id, username, _date_string):
+        # For daily output, do reduction on all of these.
+        return (course_id, username)
+
+    def get_mapper_value(self, _course_id, _username, _date_string, label):
```
mulby: since `_date_string` is actually used it should simply be `date_string`. The '_' indicates a deliberately unused variable.

edx/analytics/tasks/user_activity.py
```diff
@@ -192,6 +194,122 @@ def reducer(self, key, values):
...
+        while interval_index_date <= interval_end_date:
+            results[str(interval_index_date)] = [str(interval_index_date),course_id, username,0,0,0]
+            interval_index_date += timedelta(days=1)
+
+        for value in values:
+            label, date_string = value
+
+            labels = ["ACTIVE","ATTEMPTED_PROBLEM", "PLAYED_VIDEO"]
+            try:
+                # get the correct location in the yiled result
```
mulby: s/yiled/yield

mulby: Also we should try to figure out a way to tie the `3` below with the column indexes. At a minimum a comment, ideally some other solution that made it more clear that if we change the columns in some way we need to also update this number.

edx/analytics/tasks/user_activity.py
```diff
@@ -213,7 +331,7 @@ class CountUserActivityPerIntervalTask(
     """Counts the number of users for each course/interval/label combination."""
 
     def requires(self):
-        return UserActivityPerIntervalTask(
+        return UserActivityPerDayTask(
```
mulby: I don't think we want this...

edx/analytics/tasks/user_activity.py
jhelbert: yep, mapreducejobtaskmixin is better

jhelbert:
@mulby @brianhw

jhelbert:
@brianhw @mulby  this now creates the updated schema (date, course_id, username, category, count)

edx/analytics/tasks/user_activity.py
```diff
@@ -192,6 +195,118 @@ def reducer(self, key, values):
...
+    def reducer(self, key, values):
+        """Outputs labels and usernames for a given course and interval."""
+        course_id, username, date_string = key
+
+        interval_nums = str(self.interval).split('-')
+        interval_start = '-'.join(interval_nums[0:3])
+        interval_end = '-'.join(interval_nums[3:6])
+
+        if date_string >= interval_start and date_string < interval_end:
+            categories = [ACTIVE_LABEL, PROBLEM_LABEL, PLAY_VIDEO_LABEL]
```
brianhw: Any reason that we filter the categories that this produces?  Seems we could just go through values and count each label, and then output "for category in category_counts".   If we are going to apply some filter, we should probably pull it out as a constant somewhere.  (Perhaps somewhere close to where we define the labels.)  At any rate, I thought we didn't want to output any entry with zero counts, and we're not checking that here.

setup.cfg
```diff
@@ -33,6 +33,7 @@ edx.analytics.tasks =
     last-country = edx.analytics.tasks.user_location:LastCountryForEachUser
     export-events = edx.analytics.tasks.event_exports:EventExportTask
     user-activity = edx.analytics.tasks.user_activity:UserActivityBaseTask
+    import-daily-user-activity-to-hive = edx.analytics.tasks.user_activity:ImportDailyUserActivityToHiveTask
```
brianhw: Is there a reason to add this?  We should probably put a note in at some point indicating that only one entry per file is actually required to get the tasks defined.

edx/analytics/tasks/tests/test_user_activity.py
```diff
@@ -189,6 +190,71 @@ def test_multiple_events_with_dupes(self):
...
+        key_08_12 = (self.course_id, self.username, "2014-08-18")
+        expected = tuple()
+        self._check_output(key_08_12, [], expected)
+
+    def test_single_date(self):
+        key_08_12 = (self.course_id, self.username, "2014-08-12")
+        inputs_08_12 = (ACTIVE_LABEL, PROBLEM_LABEL)
+        expected_08_12 = (('2014-08-12', self.course_id, self.username, ACTIVE_LABEL, 1 ),
+                    ('2014-08-12', self.course_id, self.username, PROBLEM_LABEL, 1 ),
+                    ('2014-08-12', self.course_id, self.username, PLAY_VIDEO_LABEL, 0 ),)
```
brianhw: Again, I thought we weren't outputting entries with zero counts.

edx/analytics/tasks/tests/test_user_activity.py
```diff
@@ -189,6 +190,71 @@ def test_multiple_events_with_dupes(self):
...
+
+    def _get_reducer_output(self, key, values):
+        """Run reducer with provided values hardcoded key."""
+        return tuple(self.task.reducer(key, values))
+
+    def _check_output(self, key, inputs, expected):
+        """Compare generated with expected output."""
+        self.assertEquals(self._get_reducer_output(key, inputs), expected)
+
+    def test_date_out_of_interval(self):
```
brianhw: Probably good to test bounds as well, to exercise the '<' versus the '<='.

edx/analytics/tasks/user_activity.py
jhelbert: i agree, now it is outputting all labels with > 0 counts

jhelbert:
@brianhw @mulby

edx/analytics/tasks/user_activity.py
```diff
@@ -21,12 +23,20 @@
...
 
 
-class UserActivityBaseTask(EventLogSelectionMixin, MapReduceJobTask, UserActivityBaseTaskDownstreamMixin):
+class UserActivityBaseTask(
+        EventLogSelectionMixin,
+        MapReduceJobTask,
+        UserActivityBaseTaskDownstreamMixin,
+        OverwriteOutputMixin):
     """
     Base class to extract events from logs over a selected interval of time.
```
brianhw, line 46: The docstring for this class should be simplified, since it's just a collection of classes.  It should probably just indicate the set of classes that are mixed in, and indicate that no new parameters are added here.

edx/analytics/tasks/user_activity.py
```diff
@@ -21,12 +23,20 @@
 class UserActivityBaseTaskDownstreamMixin(object):
     """Defines output_root parameter with appropriate default."""
```
brianhw: This docstring, in contrast to the one for UserActivityBaseTask, should have doc for the parameters it defines.  These include the new one, the warehouse_path, in addition to output_root (which has been moved here from UserActivityBaseTask).

edx/analytics/tasks/user_activity.py
```diff
@@ -192,6 +202,114 @@ def reducer(self, key, values):
...
 
 
+class ImportDailyUserActivityToHiveTask(
+        EventLogSelectionDownstreamMixin,
+        MapReduceJobTaskMixin,
+        ImportIntoHiveTableTask):
+    """
+    Creates a Hive Table that points to Hadoop output of CourseEnrollmentChangesPerDay task.
+
+    Parameters are defined by :py:class:`BaseCourseEnrollmentTaskDownstreamMixin`.
```
brianhw: I don't understand this comment.  What is this mixin defining here?  It doesn't look like it's mixed in.  And the things that are mixed in don't look like they provide all the arguments to UserActivityPerDayTask. 

Instead, maybe this should include UserActivityBaseTaskDownstreamMixin, and it in turn should include EventLogSelectionMixin, MapReduceJobTaskMixin, and OverwriteOutputMixin as well as the values it defines.

brianhw:
Just some docstring stuff and possible mixin refactoring.  Otherwise it looks good!

edx/analytics/tasks/user_activity.py
```diff
@@ -192,6 +202,114 @@ def reducer(self, key, values):
...
+    """
+
+    warehouse_path = luigi.Parameter(
+        default_from_config={'section': 'hive', 'name': 'warehouse_path'}
+    )
+
+    @property
+    def table_name(self):
+        return 'daily_user_activity'
```
mulby: I think I would prefer simply "user_activity"

edx/analytics/tasks/database_imports.py
```diff
@@ -92,7 +92,8 @@ def query(self):
             col_spec=','.join([' '.join(c) for c in self.columns]),
             location=self.table_location,
             table_format=self.table_format,
-            partition_date=self.partition_date,
+            partition_key=self.partition.keys()[0],
```
mulby: It might be more clear to assign a variable to this so that you don't have to repeat the snippet on the line below.

edx/analytics/tasks/user_activity.py
```diff
@@ -192,6 +202,114 @@ def reducer(self, key, values):
...
+            ('date', 'STRING'),
+            ('course_id', 'STRING'),
+            ('username', 'STRING'),
+            ('category', 'STRING'),
+            ('count', 'INT')
+        ]
+
+    @property
+    def table_location(self):
+        output_name = 'user-activity-per-day/'
```
mulby: I would prefer we keep the folder names the same as the table names.

edx/analytics/tasks/user_activity.py
```diff
@@ -192,6 +202,114 @@ def reducer(self, key, values):
...
+        return url_path_join(self.warehouse_path, output_name)
+
+    @property
+    def table_format(self):
+        """Provides structure of Hive external table data."""
+        return "ROW FORMAT DELIMITED FIELDS TERMINATED BY '\\t'"
+
+    # this needs to be implemented, but we use partition_interval instead
```
mulby: Why is this the case?

edx/analytics/tasks/user_activity.py
```diff
@@ -192,6 +202,114 @@ def reducer(self, key, values):
...
+    @property
+    def partition_date(self):
+        return None
+
+    @property
+    def partition_interval(self):
```
mulby: Is this only used in `partition()` if so I would just inline it...

edx/analytics/tasks/user_activity.py
```diff
@@ -192,6 +202,114 @@ def reducer(self, key, values):
...
+        """Outputs labels and usernames for a given course and interval."""
+        course_id, username, date_string = key
+
+        interval_nums = str(self.interval).split('-')
+        interval_start = '-'.join(interval_nums[0:3])
+        interval_end = '-'.join(interval_nums[3:6])
+
+        if date_string >= interval_start and date_string < interval_end:
+            category_counts = {}
+            for label in values:
```
mulby: I would either use "label" or "category" but not mix the two.

edx/analytics/tasks/user_activity.py
```diff
@@ -192,6 +202,114 @@ def reducer(self, key, values):
...
+                if not category_counts.get(label):
```
mulby:
``` python
category_counts[label] = category_counts.get(label, 0) + 1
```

edx/analytics/tasks/user_activity.py
```diff
@@ -192,6 +202,114 @@ def reducer(self, key, values):
...
+                    category_counts[label] = 0
+                category_counts[label] += 1
+
+            # emit all label counts for his date/course/user
+            for category in category_counts:
```
mulby:
``` python
for category, count in category_counts.iteritems():
    yield ...
```

edx/analytics/tasks/user_activity.py
mulby: Even better, we could just use Counter

http://stackoverflow.com/a/5829377/55460

Assuming it doesn't read the entire list into memory... which may not be a valid assumption now that I think about it.

edx/analytics/tasks/user_activity.py
jhelbert: is it because partition_date raises a notimplementederror? or because we never use it (and override partition to not use it) do we need need to define it

jhelbert:
@brianhw @mulby

PR #4, opened 2014-08-23
Histogram statistics tasks from courseware_studentmodule/sqoop exports

jbau:
@mulby @brianhw 

These tasks compute histogram statistics from courseware_studentmodule dumps.  There are 2 statistics, grade distribution and open count for sequential modules, that Stanford currently calculates directly from c_sm to render a view in the lms instructor panel, that we'd like to replace with aggregated data from the analytics database (for faster loading).

Both grade dist and seq open count have 2 workflows, one from an existing CSV (pre-sqooped) db dump and one that includes running sqoop.  There are acceptance tests covering all 4 of these workflows.

Also, let me lay out the class/mixin hierarchy I was going for.  I wanted to capture the somewhat abstract idea of "calculating histogram statistics from a dump of courseware_studentmodule" and make it easy to build implementations that do something specific with that.  So that abstract concept maps to an abstract base class `HistogramsFromStudentModuleBase`.  Then, two abstract workflows corresponding to "start from CSV"  and "start from Sqoop" (`HistogramFromStudentModuleDumpWorkflowBase` and `HistogramFromStudentModuleSqoopWorkflowBase`) inherit from that abstract base class.  

Then, attributes specific to "grade distribution" versus "sequential open distribution" are mixed in to (composed with) those abstract workflows to form the final concrete workflows.

### DIFF
--- a/edx/analytics/tasks/studentmodule_dist.py
+++ b/edx/analytics/tasks/studentmodule_dist.py
@@ -1,0 +1,303 @@
+"""
+Tasks to compute histogram distribution from the courseware_studentmodule table in edx-platform's mysql database,
+either by producing a new dump with sqoop or using existing dumps, and export them into analytics-data-api's
+mysql database.
+"""
+
+import csv
+from collections import defaultdict
+import logging
+
+import luigi
+
+from edx.analytics.tasks.mapreduce import MapReduceJobTask, MapReduceJobTaskMixin
+from edx.analytics.tasks.sqoop import SqoopImportFromMysql
+from edx.analytics.tasks.util import csv_util
+from edx.analytics.tasks.url import url_path_join, get_target_from_url
+from edx.analytics.tasks.database_exports import FIELD_SIZE_LIMIT, StudentModuleRecord
+from edx.analytics.tasks.mysql_load import MysqlInsertTask
+
+log = logging.getLogger(__name__)
+
+
+# Increase maximum number of characters per field since we have
+# entries that easily exceed the default value of 124 KB.
+csv.field_size_limit(FIELD_SIZE_LIMIT)
+
+
+######################################################################
+###       Abstract Import and Histogram Calculation Section        ###
+######################################################################
+class HistogramTaskFromSqoopParamsMixin(object):
+    """
+    Mixin the parameters for HistogramsFromStudentModule that involve Sqoop
+
+    Parameters:
+        * name: Name of this run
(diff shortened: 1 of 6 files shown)


jbau:
Oh, and I'll squash before merge. I just left the tests and the functional code in different commits for now.

jbau:
Acceptance tests passing here:

http://jenkins-ci.analytics.edx.org/view/ad-hoc/job/edx-analytics-pipeline-acceptance-test-manual/10/parameters/

edx/analytics/tasks/studentmodule_dist.py
```diff
@@ -0,0 +1,430 @@
...
+            table_name=table_name,
+            num_mappers=self.num_mappers,
+            overwrite=True,
+        )
+
+
+######################################################################
+###           Abstract Export to Mysql Workflow Section            ###
+######################################################################
+class DeleteTableBeforeMySqlInsertTask(MysqlInsertTask):
```
mulby: Perhaps we should move this into the parent class and make it aware of "overwrite"?

jbau:
acceptance tests still passing:
http://jenkins-ci.analytics.edx.org/view/ad-hoc/job/edx-analytics-pipeline-acceptance-test-manual/19/console

jbau:
specific acceptance tests passing:
http://jenkins-ci.analytics.edx.org/view/ad-hoc/job/edx-analytics-pipeline-acceptance-test-manual/26/

will kick off the whole suite

jbau:
it seems like now that overwriting the output db table contents is tied into `--overwrite`, `--overwrite` should actually always be set in production runs, because otherwise the output db will contain data over multiple task runs, which is confusing to the API.

jbau:
full suite passing:
http://jenkins-ci.analytics.edx.org/view/ad-hoc/job/edx-analytics-pipeline-acceptance-test-manual/27/

mulby:
I would just hardcode overwrite = True for you task, other than that :+1:

brianhw:
LGTM.  :+1:  It will be good to extend some tests with new-format keys (i.e. after opaque keys) when the time comes, but I expect everything will continue to work.

jbau:
Finally! http://jenkins-ci.analytics.edx.org/view/ad-hoc/job/edx-analytics-pipeline-acceptance-test-manual/78/ passed.  I'll squash and merge today

PR #7, opened 2014-09-02
Hardcode pipeline to use Python 2.7

brianhw:
In anticipation of Opaque Keys integration.

@mulby @benpatterson

### DIFF
--- a/client.cfg
+++ b/client.cfg
@@ -1,5 +1,6 @@
 [hadoop]
 version = apache1
+python-executable=/usr/bin/python2.7
 
 [hive]
 release = apache
--- a/share/task.yml
+++ b/share/task.yml
@@ -20,7 +20,7 @@
     - log_dir: "{{ root_log_dir }}/{{ uuid}}"
     - working_repo_dir: "{{ working_dir }}/repo"
     - working_venv_dir: "{{ working_dir }}/venv"
-    - virtualenv_python: "--python=/usr/bin/python2.7"
+    - virtualenv_python: "/usr/bin/python2.7"
     - task_arguments: ''
     - git_servers:
         # Analytics repositories are currently hosted on github.
@@ -96,7 +96,7 @@
 
     - name: virtualenv created
       command: >
-        virtualenv {{ virtualenv_python }} {{ working_venv_dir }}
+        virtualenv --python={{ virtualenv_python }} {{ working_venv_dir }}
 
     - name: update pip
       command: >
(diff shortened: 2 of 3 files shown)


benpatterson:
Would it make more sense to run a quick python --version test, and fail jobs that are not finding 2.7.x? (This is more directly my reaction to seeing the change in client.cfg...what are the chances we have a different path to the python executable?)

brianhw:
As background, the EMR instances from AWS have python point to python 2.6 by default.  So if we want to use python 2.7 on the clusters, we have to be explicit.

share/task.yml
```diff
@@ -95,7 +96,7 @@
 
     - name: virtualenv created
       command: >
-        virtualenv {{ working_venv_dir }}
+        virtualenv {{ virtualenv_python }} {{ working_venv_dir }}
```
benpatterson: Stylistic: should the virtualenv_python switch appear at the end of the command?

benpatterson:
Got it. Thanks. A stylistic question/comment but :+1:

mulby:
:+1:

client.cfg
```diff
@@ -1,5 +1,6 @@
 [hadoop]
 version = apache1
+python-executable=/usr/bin/python2.7
```
mulby, line 3: thank goodness this exists!

share/task.yml
```diff
@@ -20,6 +20,7 @@
     - log_dir: "{{ root_log_dir }}/{{ uuid}}"
     - working_repo_dir: "{{ working_dir }}/repo"
     - working_venv_dir: "{{ working_dir }}/venv"
+    - virtualenv_python: "--python=/usr/bin/python2.7"
```
mulby: optional, but I would prefer this was simply "/usr/bin/python2.7" and the "--python" part was put into the actual command invocation.

brianhw: My thinking was that if you didn't want to specify a path at all, then you could leave this blank, and you would get whatever was the default python on the machine (wherever it was).   Whereas "--python=  " would break.

mulby: I'm not sure it's worthwhile to support this use case. People can always just pass in /usr/bin/python if they want the default.

I just haven't seen this pattern used before in ansible and suspect it might be confusing to readers.

share/task.yml
brianhw: The documentation for virtualenv, and many other executables, has the options specified first.

benpatterson: Ok got it.

brianhw:
@jbau Just a heads up.  Before you can use this, you would need to update to a later AMI version on EMR than we were using before.  AMI 2.4.2 does not support Python 2.7, but 2.4.5 to 2.4.7 do.

mulby:
:+1:

PR #8, opened 2014-09-04
Integrate with opaque_keys package for checking course_ids.

brianhw:
Requires running Python 2.7 on clusters, since opaque keys
requires that.
- Export opaque keys and dependencies to worker instances in Hadoop.
- Support for checking if course_id is valid.
- Rudimentary support for getting course_id from URL.
- Make sure filenames from course_ids are legal, replacing characters as
  needed.
- Identify organization for each event as close to original algorithm as
  possible on existing data, but working for opaque keys as well.

Acked-by: Brian Wilson brian@edx.org


brianhw:
@mulby For preview.  There are more unit tests to be added in test_opaque_key_util, and new data should be added to acceptance tests.  However, I wanted to get the existing acceptance tests to work, and half of them don't.  Running the branch fails four acceptance tests: two answer dist, one enrollment trends, and one user activity.  Logging indicates that the is_valid_course_id() logic is failing to parse slashified course ids in those cases, but the values presented are very reasonable.   I'm suspicious that there is a loading issue remaining in opaque_keys, despite my pointing to the tip.

Also running a validation task on event_exports, waiting for that to complete before having confidence that the new org-identification algorithm will get the same results as the old one.

mulby:
@brianhw I don't actually see anything particularly controversial here. I'll review in more depth, but at a high level it seems fine.

edx/analytics/tasks/mapreduce.py
```diff
@@ -89,6 +101,78 @@ def requires_hadoop(self):
         return convert_tasks_to_manifest_if_necessary(self.requires())
 
 
+def create_packages_archive_with_egginfo(packages, filename):
```
mulby: We should put in a big comment about where this came from, what modifications we made to it, and why we copied it.

edx/analytics/tasks/mapreduce.py
```diff
@@ -115,6 +199,115 @@ def __init__(self, libjars_in_hdfs=None, input_format=None):
             jobconfs=job_confs,
         )
 
+    def run_job(self, job):
```
mulby: Same thing about commenting on the rationale for why this is here.

edx/analytics/tasks/event_exports.py
```diff
@@ -216,3 +177,82 @@ def _get_recipients(self, org_id):
...
+                    except Exception:  # pylint: disable=broad-except
+                        return None
+            elif item['event_source'] == 'browser':
+                # Note that the context of browser events is ignored for legacy
+                # events.  But check context first in case it's a non-slashified course.
+                course_id = item.get('context', {}).get('course_id')
+                if course_id and '/' not in course_id:
+                    org_id = item.get('context', {}).get('org_id')
+                    if org_id:
+                        return org_id
```
mulby: Won't this result in us emitting more browser events? I thought we had talked about maintaining bit-for-bit compatibility for existing date ranges?

edx/analytics/tasks/mapreduce.py
mulby: Just to put in writing what we've talked about - we should move this (and the changes to run_job) into our fork.

edx/analytics/tasks/event_exports.py
brianhw: This will only emit more browser events for non-slashified course-ids.  Which has no impact on the past, but does going forward.  Actually it seems like a reasonable horizon to allow us to turn on other event exports, such as for workers and the like, for which a course-id might be provided that would allow us to provide more data without breaking backwards consistency.

brianhw:
@mulby This should be done now.  Just waiting on acceptance tests.

edx/analytics/tasks/tests/test_answer_dist.py
```diff
@@ -387,15 +394,23 @@ def test_problem_display_name(self):
...
+        self.initialize_ids()
         self.task = AnswerDistributionPerCourseMixin()
-        self.course_id = "FooX/1.23x/2013_Spring"
-        self.problem_id = "i4x://FooX/1.23x/2013_Spring/problem/PSet1:PS1_Q1"
-        self.answer_id = "i4x-FooX-1_23x-problem-PSet1_PS1_Q1_2_1"
+        #course_key = CourseLocator(org='foo', course='bar', run='baz')
+        #self.course_id = unicode(course_key)
+        #block_id = "9cee77a606ea4c1aa5440e0ea5d0f618"
+        #self.problem_id = unicode(course_key.make_usage_key("problem", block_id))
+        #self.answer_id = "{block_id}_2_1".format(block_id=block_id)
```
mulby: You can remove this commented out code.

mulby:
:+1:

PR #10, opened 2014-09-11
Ensure consistent licensing of ec2.py

mulby:
@nedbat @brianhw @singingwolfboy 

When we discussed "relicensing" I assume this is what was meant. Let me know if any of you had different expectations.


brianhw:
LGTM. :+1:

PR #11, opened 2014-09-12
Compute historical user activity

mulby:
I think this is the bare minimum changeset that will allow us to see historical user activity.


mulby:
obviously needs additional unit tests, and might require the acceptance test to be modified as well

brianhw:
Yup, but the interval math logic looks good to me.

mulby:
@brianhw this is ready for review again

brianhw:
LGTM.  :+1:

PR #13, opened 2014-09-15
Revert to using Python 2.6 for the time being

mulby:
Since we don't yet have wheels and we don't actually need 2.7 yet, I'd like to defer this patch until after the release so that we don't kill the performance of our existing jobs.

### DIFF
--- a/client.cfg
+++ b/client.cfg
@@ -1,6 +1,5 @@
 [hadoop]
 version = apache1
-python-executable=/usr/bin/python2.7
 
 [hive]
 release = apache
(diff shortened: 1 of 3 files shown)


brianhw:
Sigh.  This needs to be done before, or as part of, opaque keys support, which I was hoping to have ready in the next day or so.  (Once I deal with merging out the stuff that's going in before it.)  Are you worried about production runs, or just acceptance tests?

brianhw:
But sure, go ahead.  I'll just add the 2.7 changes into my branch. :+1:

mulby:
Worried about production runs since I'm currently creating a new release.

PR #14, opened 2014-09-15
Use less disk space for acceptance tests

mulby:
@benpatterson @brianhw 

This will create a new virtualenv for each executor, not each test method.


benpatterson:
I see you needed to merge it under duress. FWIW it LGTM. Thanks for jumping on the problem.

PR #15, opened 2014-09-17
Switch to fork of luigi using v1.0.17

brianhw:
@mulby @rocha  Upgrading from 1.0.16 to v1.0.17, in preparation for making changes to Luigi fork.

### DIFF
--- a/client.cfg
+++ b/client.cfg
@@ -1,6 +1,9 @@
 [hadoop]
 version = apache1
 
+[hdfs]
+client = hadoopcli
+
 [hive]
 release = apache
 database = default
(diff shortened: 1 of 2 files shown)


mulby:
:+1:

rocha:
My only suggestion is to use a special version for the luigi fork. I suggest something like 1.0.17-fork-edx-1. On the luigi fork it is necessary to change `setup.py` to use that version number, and it also helps to tag the corresponding fork release.

The main advantage is that it makes easier to find pip problems, since just doing a `pip list` will show you if the right version got installed or not.

brianhw:
@rocha I plan to do that when I actually make a change to Luigi.

rocha:
cool :rocket:

PR #16, opened 2014-09-22
Demographics initial work

mulby:
Note that a lot of the logic in the CourseEnrollmentTask is refactored from the original logic in course_enroll.py

### DIFF
--- a/edx/analytics/tasks/database_imports.py
+++ b/edx/analytics/tasks/database_imports.py
@@ -227,7 +227,7 @@ class ImportStudentCourseEnrollmentTask(ImportMysqlToHiveTableTask):
 
 class ImportAuthUserTask(ImportMysqlToHiveTableTask):
 
-    """Imports course enrollment information from an external LMS DB to a destination directory."""
+    """Imports user information from an external LMS DB to a destination directory."""
 
     @property
     def table_name(self):
@@ -249,6 +249,24 @@ class ImportAuthUserTask(ImportMysqlToHiveTableTask):
         ]
 
 
+class ImportAuthUserProfileTask(ImportMysqlToHiveTableTask):
+
+    """Imports user demographic information from an external LMS DB to a destination directory."""
+
+    @property
+    def table_name(self):
+        return 'auth_userprofile'
+
+    @property
+    def columns(self):
+        return [
+            ('user_id', 'INT'),
+            ('gender', 'STRING'),
+            ('year_of_birth', 'INT'),
+            ('level_of_education', 'STRING'),
+        ]
+
+
 class ImportAllDatabaseTablesTask(DatabaseImportMixin, OverwriteOutputMixin, luigi.WrapperTask):
     """Imports a set of database tables from an external LMS RDBMS."""
     def requires(self):
@@ -263,6 +281,7 @@ class ImportAllDatabaseTablesTask(DatabaseImportMixin, OverwriteOutputMixin, lui
         yield (
             ImportStudentCourseEnrollmentTask(**kwargs),
             ImportAuthUserTask(**kwargs),
+            ImportAuthUserProfileTask(**kwargs),
         )
 
     def output(self):
--- a/edx/analytics/tasks/enrollments.py
+++ b/edx/analytics/tasks/enrollments.py
@@ -1,0 +1,644 @@
+"""Compute metrics related to user enrollments in courses"""
+
+import logging
+import textwrap
+import datetime
+
+import luigi
+
+
+from edx.analytics.tasks.database_imports import ImportAuthUserProfileTask, ImportIntoHiveTableTask
+from edx.analytics.tasks.mapreduce import MapReduceJobTaskMixin, MapReduceJobTask
+from edx.analytics.tasks.pathutil import EventLogSelectionDownstreamMixin, EventLogSelectionMixin
+from edx.analytics.tasks.url import get_target_from_url, url_path_join
+from edx.analytics.tasks.util import eventlog, opaque_key_util
+from edx.analytics.tasks.util.hive import WarehouseMixin
+from edx.analytics.tasks.mysql_load import MysqlInsertTask
+
+
+log = logging.getLogger(__name__)
+DEACTIVATED = 'edx.course.enrollment.deactivated'
+ACTIVATED = 'edx.course.enrollment.activated'
+
+
+class CourseEnrollmentTask(EventLogSelectionMixin, MapReduceJobTask):
+    """Produce a data set that shows which days each user was enrolled in each course."""
+
+    output_root = luigi.Parameter()
+
+    def mapper(self, line):
+        value = self.get_event_and_date_string(line)
+        if value is None:
+            return
+        event, _date_string = value
+
+        event_type = event.get('event_type')
+        if event_type is None:
+            log.error("encountered event with no event_type: %s", event)
+            return
+
+        if event_type not in (DEACTIVATED, ACTIVATED):
+            return
+
+        timestamp = eventlog.get_event_time_string(event)
+        if timestamp is None:
+            log.error("encountered event with bad timestamp: %s", event)
+            return
+
+        event_data = eventlog.get_event_data(event)
+        if event_data is None:
+            return
+
+        course_id = event_data.get('course_id')
+        if course_id is None or not opaque_key_util.is_valid_course_id(course_id):
+            log.error("encountered explicit enrollment event with invalid course_id: %s", event)
+            return
+
+        user_id = event_data.get('user_id')
+        if user_id is None:
+            log.error("encountered explicit enrollment event with no user_id: %s", event)
+            return
+
+        yield (course_id, user_id), (timestamp, event_type)
+
+    def reducer(self, key, values):
+        """Emit records for each day the user was enrolled in the course."""
+        course_id, user_id = key
+
+        event_stream_processor = DaysEnrolledForEvents(course_id, user_id, self.interval, values)
+        for day_enrolled_record in event_stream_processor.days_enrolled():
+            yield day_enrolled_record
+
+    def output(self):
+        return get_target_from_url(self.output_root)
+
+
+class EnrollmentEvent(object):
+    """The critical information necessary to process the event in the event stream."""
+
+    def __init__(self, timestamp, event_type):
+        self.timestamp = timestamp
+        self.datestamp = eventlog.timestamp_to_datestamp(timestamp)
+        self.event_type = event_type
+
+
+class DaysEnrolledForEvents(object):
+    """
+    Determine which days a user was enrolled in a course given a stream of enrollment events.
+
+    Produces a record for each date in which the user was enrolled in the course. Note that the user need not have been
+    enrolled in the course for the entire day. These records will have the following format:
+
+        datestamp (str): The date the user was enrolled in the course during.
+        course_id (str): Identifies the course the user was enrolled in.
+        user_id (int): Identifies the user that was enrolled in the course.
+        enrolled_at_end (int): 1 if the user was still enrolled in the course at the end of the day.
+        change_since_last_day (int): 1 if the user has changed to the enrolled state, -1 if the user has changed
+            to the unenrolled state and 0 if the user's enrollment state hasn't changed.
+
+    If the first event in the stream for a user in a course is an unenrollment event, that would indicate that the user
+    was enrolled in the course before that moment in time. It is unknown, however, when the user enrolled in the course,
+    so we assume that they have been enrolled in the course since the beginning of the interval and records are produced
+    accordingly.
+
+    Similarly if the last event for a user is an enrollment event, that would indicate that the user was still enrolled
+    in the course at the end of the interval, so records are produced from that last enrollment event all the way to the
+    end of the interval.
+
+    Ranges of dates where the user is continuously enrolled will be represented as contiguous records with the first
+    record indicating the change (new enrollment), and the last record indicating the unenrollment. It will look
+    something like this::
+
+        datestamp,enrolled_at_end,change_since_last_day
+        2014-01-01,1,1
+        2014-01-02,1,0
+        2014-01-03,1,0
+        2014-01-04,0,-1
+
+    The above activity indicates that the user enrolled in the course on 2014-01-01 and unenrolled from the course on
+    2014-01-04.
+
+    If a user enrolls and unenrolls from a course on the same day, a record will appear that looks like this::
+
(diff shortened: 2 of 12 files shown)


mulby:
Still a lot of work to do on docstrings, acceptance tests, etc, but I figured I would get this up for early feedback.

edx/analytics/tasks/launchers/local.py
```diff
@@ -28,7 +28,7 @@
 
 def main():
     # In order to see errors during extension loading, you can uncomment the next line.
-    # logging.basicConfig(level=logging.DEBUG)
+    logging.basicConfig(level=logging.DEBUG)
```
brianhw, line 32: Assume this is temporary...

edx/analytics/tasks/tests/test_enrollments.py
```diff
@@ -0,0 +1,309 @@
...
+    """
+    def setUp(self):
+        fake_param = luigi.DateIntervalParameter()
+        self.task = CourseEnrollmentTask(
+            interval=fake_param.parse('2013-12-17'),
+            output_root='/fake/output'
+        )
+        self.task.init_local()
+
+        self.course_id = "MITx/8.02x/2013_Spring"
```
brianhw: I pulled some of this course_id/org_id stuff out into a mixin.

edx/analytics/tasks/enrollments.py
```diff
@@ -0,0 +1,618 @@
...
+        2014-01-03,1,0
+        2014-01-04,0,-1
+
+    The above activity indicates that the user enrolled in the course on 2014-01-01 and unenrolled from the course on
+    2014-01-05.
+
+    If a user enrolls and unenrolls from a course on the same day, a record will appear that looks like this::
+
+        datestamp,enrolled_at_end,change_since_last_day
+        2014-01-01,0,0
```
brianhw, line 129: This all seems adequate for these demographic tasks, where we just need to count those users with at_end = 1.  And it's presumably adequate also for calculating changes in daily enrollments, using the change_since_last_day field.  And for activity, we just use the presence of an entry.  Looks okay.  But just to clarify, if we have an UNENROLL, then a few days, and another ENROLL, we should get a run up to the UNENROLL, then no events, then a second run starting with the ENROLL until the end of the interval.

brianhw, line 129: Reading the code below, it looks like it will do the right thing -- it fills only while ENROLL.

brianhw:
Generally looks good.  Algorithm makes sense.   Now that opaque keys is merged, there will need a little work to incorporate those changes into the tests.

edx/analytics/tasks/launchers/local.py
mulby, line 32: Actually, I'm seriously considering just keeping this in. It doesn't generate too much noise and is very useful.

mulby:
This is ready for more in depth review.

edx/analytics/tasks/database_imports.py
```diff
@@ -249,6 +249,26 @@ def columns(self):
...
+
+    @property
+    def table_name(self):
+        return 'auth_userprofile'
+
+    @property
+    def columns(self):
+        # Fields not included are 'password', 'first_name' and 'last_name'.
+        # In our LMS, the latter two are always empty.
+        return [
```
brianhw, line 262: Comment needs to be updated -- left over from AuthUser.

edx/analytics/tasks/database_imports.py
```diff
@@ -249,6 +249,26 @@ def columns(self):
...
+            ('user_id', 'INT'),
+            ('gender', 'STRING'),
+            ('year_of_birth', 'INT'),
+            ('level_of_education', 'STRING'),
+        ]
```
brianhw, line 267: Do we also envision needing "allow_certificate"?  Concerned we might find it useful (i.e. how many are not permitted certificates?).   We don't permit certificates to those from embargoed countries, according to code comments in lms/djangoapps/certificates/models.py.

Also, do we need to take into account the UserStanding table (i.e. account disabled or enabled)?  Some simple counts of number of students disabled (due to bad behavior) might be possible in future, but I'm guessing it  would never be used to actually limit activity.  (I.e. if they're disabled, they won't have any activity.)

edx/analytics/tasks/enrollments.py
```diff
@@ -0,0 +1,635 @@
...
+from edx.analytics.tasks.mysql_load import MysqlInsertTask
+
+
+log = logging.getLogger(__name__)
+DEACTIVATED = 'edx.course.enrollment.deactivated'
+ACTIVATED = 'edx.course.enrollment.activated'
+
+
+class CourseEnrollmentTask(EventLogSelectionMixin, MapReduceJobTask):
+    """Produce a deta set that shows which days each user was enrolled in each course."""
```
brianhw: deta/delta

edx/analytics/tasks/launchers/local.py
brianhw, line 32: Yeah, okay.

edx/analytics/tasks/enrollments.py
```diff
@@ -0,0 +1,635 @@
...
+        event_stream_processor = DaysEnrolledForEvents(course_id, user_id, self.interval, values)
+        for day_enrolled_record in event_stream_processor.days_enrolled():
+            yield day_enrolled_record
+
+    def output(self):
+        return get_target_from_url(self.output_root)
+
+
+class EnrollmentEvent(object):
+    """All of the information necessary to process the event in the event stream."""
```
brianhw: Well, not all the information, since it's not including the user or course.  But not sure how to word this better.

edx/analytics/tasks/enrollments.py
```diff
@@ -0,0 +1,635 @@
...
+
+    def __init__(self, timestamp, event_type):
+        self.timestamp = timestamp
+        self.datestamp = eventlog.timestamp_to_datestamp(timestamp)
+        self.event_type = event_type
+
+
+class DaysEnrolledForEvents(object):
+    """Determine which days a user was enrolled in a course given a stream of enrollment events.
```
brianhw: Yes, PEP257 says that one can put the summary line on the first line or not.  But our Python guidelines (https://github.com/edx/edx-platform/wiki/Python-Guidelines) says that we should not put it on the first line.  So in the name of consistency, please change these to have the summary line on the next line.

edx/analytics/tasks/enrollments.py
```diff
@@ -0,0 +1,635 @@
...
+        self.user_id = user_id
+        self.interval = interval
+
+        self.sorted_events = sorted(events)
+        # After sorting, we can discard time information since we only care about date transitions.
+        self.sorted_events = [EnrollmentEvent(timestamp, value) for timestamp, value in self.sorted_events]
+        # Since each event looks ahead to see the time of the next event, insert a dummy event at then end that
+        # indicates the end of the requested interval. If the user's last event is an enrollment activation event then
+        # they are assumed to be enrolled up until the end of the requested interval.
+        self.sorted_events.append(EnrollmentEvent(self.interval.date_b.isoformat(), None))
```
brianhw: May be good to note here that the mapper ensures that there are no events with a time after self.interval.date_b.

edx/analytics/tasks/enrollments.py
```diff
@@ -0,0 +1,635 @@
...
+
+        If the first event is an unenrollment event, back fill enrollment records to the beginning of the interval.
+
+        Yields:
+            tuple: An enrollment record for each day the user is assumed to have been enrolled in the course from the
+                start of the interval up to the date where the first enrollment event appears.
+
+        """
+        if self.previous_state == self.ENROLLED:
+            interval_start_datestamp = self.interval.date_a.isoformat()
```
brianhw: I'm assuming that we expect to be running with an interval start of around 12/3/13, and that for all users that deenrolled after that date, they would all be marked as having enrolled on the first date.  I'm wondering if we really want this to start the interval on the day before date_a.  Our data collection is actually including events on date_a, and we wouldn't be able to distinguish these inferred enrollments from real enrollments on the date.

edx/analytics/tasks/enrollments.py
```diff
@@ -0,0 +1,635 @@
...
+    """Load gender breakdowns into MySQL"""
+
+    overwrite = luigi.BooleanParameter(default=True)
+
+    @property
+    def columns(self):
+        return [
+            ('date', 'DATE NOT NULL'),
+            ('course_id', 'VARCHAR(255) NOT NULL'),
+            ('gender', 'VARCHAR(255)'),
```
brianhw: Do we really need 255 for this?  I think it's 6 in LMS.

edx/analytics/tasks/tests/acceptance/test_demographics.py
```diff
@@ -0,0 +1,114 @@
...
+from edx.analytics.tasks.url import url_path_join
+
+
+log = logging.getLogger(__name__)
+
+
+class DemographicsAcceptanceTest(AcceptanceTestCase):
+    """End to end test of demographic trends."""
+
+    INPUT_FILE = 'enrollment_trends_tracking.log'
```
brianhw, line 20: I though enrollment_trends_tracking.log had some newer opaque-key-format course_ids.  I don't see that in any of the data below....

brianhw:
Done with pass.  No major issues -- mostly updating comments.

edx/analytics/tasks/database_imports.py
mulby, line 267: I think we can defer addressing both of these two issues until they are necessary. I don't see either impacting any of our current metrics.

mulby:
@brianhw rebased and addressed comments

mulby:
Acceptance tests pass

http://jenkins-ci.analytics.edx.org/job/edx-analytics-pipeline-acceptance-auto-pr/113

edx/analytics/tasks/enrollments.py
brianhw: So I was suggesting that interval_start_datestamp = (self.interval.date_a -  datetime.timedelta(days=1)).isoformat() here.  But I guess that the value of date_a is so arbitrary to begin with that it doesn't make any difference.

edx/analytics/tasks/enrollments.py
```diff
@@ -0,0 +1,644 @@
...
+        All dates from the start date up to the end date.
+
+        Yields:
+            str: ISO 8601 datestamp for each date from the first date (inclusive) up to the end date (exclusive).
+
+        """
+        current_date = self.parse_date_string(start_date_str)
+        end_date = self.parse_date_string(end_date_str)
+
+        while current_date != end_date:
```
brianhw: Wondering if it's safer to use < here instead of !=.  Just in case it were to get the start and end mixed up, in which case this would take forever to bomb out.

edx/analytics/tasks/tests/test_enrollments.py
```diff
@@ -0,0 +1,308 @@
...
+        self.timestamp = "2013-12-17T15:38:32.805444"
+
+    def _create_event_log_line(self, **kwargs):
+        """Create an event log with test values, as a JSON string."""
+        return json.dumps(self._create_event_dict(**kwargs))
+
+    def _create_event_dict(self, **kwargs):
+        """Create an event log with test values, as a dict."""
+        # Define default values for event log entry.
+        org_id = self.course_id.split('/')[0]
```
brianhw: For the record, this shouldn't be here.  It's an id that is already defined in initialize_ids().

edx/analytics/tasks/tests/test_enrollments.py
```diff
@@ -0,0 +1,308 @@
...
+        event_dict = {
+            "username": "test_user",
+            "host": "test_host",
+            "event_source": "server",
+            "event_type": "edx.course.enrollment.activated",
+            "context": {
+                "course_id": self.course_id,
+                "org_id": org_id,
```
brianhw: should use self.org_id.

edx/analytics/tasks/tests/test_enrollments.py
```diff
@@ -0,0 +1,308 @@
...
+from edx.analytics.tasks.enrollments import (
+    CourseEnrollmentTask,
+    DEACTIVATED,
+    ACTIVATED,
+)
+from edx.analytics.tasks.tests import unittest
+from edx.analytics.tasks.tests.opaque_key_mixins import InitializeOpaqueKeysMixin
+
+
+class CourseEnrollmentTaskMapTest(InitializeOpaqueKeysMixin, unittest.TestCase):
```
brianhw, line 16: This may be overkill, but to test with legacy keys instead of opaque keys, just define a CourseEnrollmentTaskMapLegacyKeyTest(InitializeLegacyKeysMixin, CourseEnrollmentTaskMapTest):  pass.  It will rerun the set of tests with a different course_id.  Perhaps there is a ddt that works at the class level to deal with this kind of testing better.

brianhw:
Looks good!  A few small questions, but otherwise :+1:

mulby:
@brianhw I decided to remove the logic that populates records from the start of the interval up to the first unenrollment event when the first event is an unenrollment.

Since we don't actually know when the user enrolled it seems just as valid to me to assume that they enrolled just before they unenrolled as it does to assume that they have been enrolled since the beginning of time.

Additionally, populating the table with these records that we don't really know are valid feels like it might bite us in the future. Safer to assume that they weren't enrolled I think.

Further, when I went to refactor the location code, I expanded the time interval to the beginning 2012, which resulted in _a lot_ of these spurious records filling up the course_enrollment table and dramatically impacting the execution time of queries against that table.

Lastly, it was causing problems with validation since there were a handful of courses that were "off by one" which turned out to be due to an unenrollment event that showed up recently for a user that had no previous enrollment event. This caused _all_ numbers before that date to be inflated by 1 when compared with our previous calculations.

I'm re-running the validation now with this assumption removed.

brianhw:
I don't fully understand.  If we assumed they enrolled just before they unenrolled, we would also assume to see the enrollment event.    The absence of the enrollment event is considered significant as long as it goes back to 12/3/13.   I don't see how that is "just as valid".  That said, what we don't know is how far back before 12/3 to consider the enrollment to have happened, and putting it at 1/1/12 does seem wasteful.  When we create synthetic events for these, we should use the creation dates of the enrollment records to provide us with guidance for many of them.  For the remainder, they will probably end up on 12/3 proper.

We'll have to revisit this with greater validation effort when we add synthetic events.

mulby:
Acceptance tests pass. Perhaps we should discuss the change in person...

http://jenkins-ci.analytics.edx.org/job/edx-analytics-pipeline-acceptance-auto-pr/116/

PR #17, opened 2014-09-24
Modify EventLogSelectionTask to accept multiple values for source parameter.

brianhw:
Values in config files may be specified on separate lines.  E.g.:

``` code
  [event-logs]
  source = s3://fake/input/
           s3://fake/input2/
```

### DIFF
--- a/config/test.cfg
+++ b/config/test.cfg
@@ -21,6 +21,7 @@ marker = s3://fake/marker/
 pattern = .*tracking.log-(?P<date>\d{8}).*\.gz
 expand_interval = 2 days
 source = s3://fake/input/
+         s3://fake/input2/
 
 [event-export]
 output_root = s3://fake/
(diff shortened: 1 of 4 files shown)


brianhw:
@mulby This also passes acceptance tests.

mulby:
Do we need to update test.cfg as well?

And tests/acceptance/**init**.py maybe?

mulby:
Also, given the validation needed for this, we may want to bundle in the config change to the subdirectory of the tracking log bucket instead of the root.

brianhw:
Test.cfg is already updated as part of this PR, and testing of it is included.  But I like the idea of adding a second source directory to tests/acceptance/init.py, and using that for putting the synthetic events into for acceptance tests.  I think we're going to want to add such events to most if not all tests.  (Even exports, to make sure they are _not_ included.)

mulby:
:+1:

PR #18, opened 2014-09-24
Ensure trends are only computed up to the run date

mulby:


mulby:
acceptance tests passed

brianhw:
I would prefer to not be writing the data for a partial day to the table in the first place, rather than implementing it in the query, because the knowledge of which dates are partial are going to be lost for anyone else that ends up using the underlying data.  

Hopefully that can still get addressed in a later PR, though, while this logic should still hold.

:+1:

mulby:
Agreed. The plan is to replace the underlying table with the demographics code, so this is just a modification to soon-to-be dead code.

PR #19, opened 2014-09-24
Wrap end date in quotes

mulby:
Pretty sure this will fix the bug, going to double check the acceptance tests this time.


brianhw:
LGTM. :+1:

PR #20, opened 2014-09-26
Add support for multiple recipients per org, and multiple orgs per alias.

brianhw:
For now, I'm leaving in support for 'recipient' as a fallback option to 'recipients'.  This can be removed once config files have been upgraded.

The multiple-orgs-per-alias should support VJx.

@mulby @rocha


brianhw:
Passes acceptance tests as well.

mulby:
:+1:

PR #21, opened 2014-09-29
Update acceptance tests to use "recipients" in org configuration.

brianhw:
Makes sure that new format of configuring recipients will work, including with exporter's master branch.

@mulby @rocha


rocha:
:rocket:

mulby:
:+1:

PR #23, opened 2014-10-08
Update get_filename_safe_course_id to be more restrictive.

brianhw:
Instead of hardcoding knowledge about colons, just replace all
but a well-defined set of supported characters.

And fix a PEP8 violation.

Also the start of adding some testing for unicode characters that are permitted in opaque key identifiers.

@mulby @rocha


rocha:
:rocket:

PR #27, opened 2014-10-11
Add handling for answer strings that convert to float inf and nan

jbau:
@mulby @brianhw 

This fixes a very interesting bug we ran into.  The symptom was the following SQL query error on insert

```
1054 (42S22): Unknown column 'inf' in 'field list'
```

which turned out to be caused by the following data:

```
(u'Education/EDUC115-S/Spring2014', u'i4x://Education/EDUC115-S/problem/1008ec10508c4b2c99930e47b1f92532', u'i4x-Education-EDUC115-S-problem-1008ec10508c4b2c99930e47b1f92532_2_1', u'1', 1, None, u'infinity', inf, None, u'Quiz - Fractions: Big Ideas', None)
```

What's going on is that `float(u"infinity")` evaluates to `inf`, which breaks the mysql connector prepared statement.  So this converts the output of any `inf` or `nan` to `None`

### DIFF
--- a/edx/analytics/tasks/answer_dist.py
+++ b/edx/analytics/tasks/answer_dist.py
@@ -2,6 +2,7 @@
 Luigi tasks for extracting problem answer distribution statistics from
 tracking log files.
 """
+import math
 import csv
 import hashlib
 import html5lib
@@ -861,7 +862,11 @@ class AnswerDistributionToMySQLTaskWorkflow(
 ################################
 def try_str_to_float(value_str):
     try:
-        return float(value_str)
+        float_val = float(value_str)
+        # infinity values and NaN actually break mysql-connector (because they look like a string), so make those None
+        if math.isinf(float_val) or math.isnan(float_val):
+            return None
+        return float_val
     except (ValueError, TypeError):
         return None
 
--- a/edx/analytics/tasks/tests/test_answer_dist.py
+++ b/edx/analytics/tasks/tests/test_answer_dist.py
@@ -8,6 +8,7 @@ import hashlib
 import os
 import tempfile
 import shutil
+import math
 
 from mock import Mock, call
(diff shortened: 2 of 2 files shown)


brianhw:
I don't understand the limits of the mysql connector here, but on the surface it seems like there is a loss in information here.  That is, by replacing inf or nan with None, it's as if there were no score.  And one can't distinguish between inf or nan either.   Are these values useful, or are they the result of bugs that we just need to protect ourselves from breaking?  Is it more accurate to translate inf into max_float or somesuch (and -inf to min_float)?

jbau:
hi @brianhw I feel okay with recording `None` because there are actually 2 columns that record answer value (`answer_value_text` and `answer_value_numeric`).  `answer_value_text` is typed `LONGTEXT` and actually records the string representation, and `answer_value_numeric` typed `DOUBLE` that records to results of an attempted (Python) cast, or `null` if there isn't a numeric representation of an answer.  `answer_value_numeric` is "best-effort".

The reason this bug creeps up is that strings like `"infinity"` and `"nan"` can be successfully run through `float()`, but the value they return ultimately gets considered by mysql connector as strings (they look something like `"inf"` and `"NaN"`) which breaks the prepared statement for the `answer_value_numeric` column, whereas anything else that's "float-y" works.

However, there is arguably no information loss, since `answer_value_text` still preserves the actual string like `"NaN"` or `"infinity"`.  Recording a `null` in `answer_value_numeric` denotes that there's no DOUBLE representation for that text, which is actually sorta right.

brianhw:
Thanks for the explanation, Jason.  That makes good sense -- as long as we have the original text, it matters less what goes into the numeric column.

Looks good to me.  :+1:

mulby:
LGTM as well :+1:

PR #30, opened 2014-10-22
Add support for multiple event patterns.

brianhw:
@mulby This was a prerequisite for doing further work on synthetic events.   I'm pulling it out into a separate PR.


mulby:
Part of me wants this to be (relatively simply) extended to support specifying patterns _without_ a date group, so that we can process the 2012 logs.

If a pattern matched but didn't contain a date group we would just accept the log file.

brianhw:
Great -- thanks for the reminder.  I made the date group optional.

mulby:
:+1:

PR #31, opened 2014-10-22
Support validation of enrollment.

brianhw:
Includes:
- Synthesize validation events from Sqoop dumps of student_courseenrollment.
- Compare validation events with enroll/unenroll events.   Includes unit tests.
  Supports additional flags for:
  - Outputting tuples (for debugging) or synthetic events (for production)
  - Ignoring any events before the first validation event (for each user/course pair).  

The latter permits checking for errors in an incremental manner, by using an interval to specify a tighter period of time (e.g. a day or a few days) which includes a set of enrollment validation events.   The run can then test the consistency of events since the enrollment validation, up to the end of the interval (which selects all events included).  This could be set up to run on a daily basis, to help flag inconsistencies more quickly as well as to generate new synthetic events in an incremental manner.

### DIFF
--- a/edx/analytics/tasks/enrollment_validation.py
+++ b/edx/analytics/tasks/enrollment_validation.py
@@ -8,6 +8,7 @@ import os
 import re
 
 import luigi
+from luigi.date_interval import DateInterval
 
 from edx.analytics.tasks.mapreduce import MultiOutputMapReduceJobTask, MapReduceJobTask, MapReduceJobTaskMixin
 from edx.analytics.tasks.pathutil import EventLogSelectionMixin, EventLogSelectionDownstreamMixin
@@ -95,7 +96,10 @@ class CourseEnrollmentValidationTask(
         is_active = event_data.get('is_active')
         created = event_data.get('created')
 
-        yield (course_id, user_id), (timestamp, event_type, is_active, created)
+        # Make sure key values that are strings are encoded strings.
+        # Note, however, that user_id is an int.
+        key = (unicode(course_id).encode('utf-8'), user_id)
+        yield key, (timestamp, event_type, is_active, created)
 
     def reducer(self, key, values):
         """Emit records for each day the user was enrolled in the course."""
@@ -514,12 +518,14 @@ class CreateEnrollmentValidationEventsTask(MultiOutputMapReduceJobTask):
             log.error("Encountered bad input: %s", line)
             return
 
-        (_db_id, user_id, encoded_course_id, mysql_created, mysql_is_active, mode) = fields
+        (_db_id, user_id_string, encoded_course_id, mysql_created, mysql_is_active, mode) = fields
 
-        # 'created' is of the form '2012-07-25 12:26:22.0', coming out of
+        # `created` is of the form '2012-07-25 12:26:22.0', coming out of
         # mysql.  Convert it to isoformat.
         created = self._mysql_datetime_to_isoformat(mysql_created)
+        # `is_active` should be a boolean and `user_id` is an int.
         is_active = (mysql_is_active == "true")
+        user_id = int(user_id_string)
 
         # Note that we do not have several standard properties that we
         # might expect in such an event.  These include a username,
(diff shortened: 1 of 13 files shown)


brianhw:
@mulby Ready for initial review of the approach.

edx/analytics/tasks/enrollment_validation.py
```diff
@@ -0,0 +1,635 @@
...
+from edx.analytics.tasks.util.datetime_util import add_microseconds, mysql_datetime_to_isoformat
+from edx.analytics.tasks.util.event_factory import SyntheticEventFactory
+from edx.analytics.tasks.util.hive import WarehouseMixin
+
+
+log = logging.getLogger(__name__)
+
+DEACTIVATED = 'edx.course.enrollment.deactivated'
+ACTIVATED = 'edx.course.enrollment.activated'
+VALIDATED = 'edx.course.enrollment.validated'
```
mulby, line 28: As was mentioned somewhere else, I think we should try to find a different name for this.

edx/analytics/tasks/enrollment_validation.py
```diff
@@ -0,0 +1,635 @@
...
+SENTINEL = 'edx.course.enrollment.sentinel'
```
mulby: Also, a different name for this.

edx/analytics/tasks/enrollment_validation.py
```diff
@@ -0,0 +1,635 @@
...
+        if course_id is None or not opaque_key_util.is_valid_course_id(course_id):
+            log.error("encountered explicit enrollment event with invalid course_id: %s", event)
+            return
+
+        user_id = event_data.get('user_id')
+        if user_id is None:
+            log.error("encountered explicit enrollment event with no user_id: %s", event)
+            return
+
+        mode = event_data.get('mode')
```
mulby, line 126: If we aren't handling mode yet, do we actually need to extract this?

edx/analytics/tasks/enrollment_validation.py
```diff
@@ -0,0 +1,635 @@
...
+        self.lower_bound_date_string = kwargs.get('lower_bound_date_string')
+
+        if self.event_output:
+            self.factory = SyntheticEventFactory(
+                event_source='server',
+                synthesizer='enrollment_validation',
+            )
+            self.generate_output = self.synthetic_event
+        else:
+            self.generate_output = self.create_tuple
```
mulby: I wonder if we couldn't simplify this by removing the tuple mode. I have no problem curating raw JSON events...

edx/analytics/tasks/enrollment_validation.py
```diff
@@ -0,0 +1,817 @@
...
+
+log = logging.getLogger(__name__)
+
+# Product event_type values:
+DEACTIVATED = 'edx.course.enrollment.deactivated'
+ACTIVATED = 'edx.course.enrollment.activated'
+MODE_CHANGED = 'edx.course.enrollment.mode_changed'
+
+# Validation-event event_type values:
+VALIDATED = 'edx.course.enrollment.validated'
```
mulby, line 28: edx.course.enrollment.snapshot?

mulby:
I don't see any blockers here, OK with leaving the "validated" event

:+1: once we find and fix the bug

edx/analytics/tasks/enrollment_validation.py
brianhw, line 126: Actually, we are handling mode yet.  Originally to make sure that the right mode appeared on the synthetic event being output.  But it's also outputting synthetic mode-change events as well.

edx/analytics/tasks/enrollment_validation.py
brianhw: Done.

edx/analytics/tasks/enrollment_validation.py
brianhw, line 28: I like that.  I'll switch it when the validation work is done, and there's a reasonable point to recalculate the validation events.

PR #32, opened 2014-10-24
Map level of education to more descriptive constants

mulby:
Note that this also groups the various codes for doctorate degrees together.

@brianhw @rocha

### DIFF
--- a/edx/analytics/tasks/enrollments.py
+++ b/edx/analytics/tasks/enrollments.py
@@ -534,6 +534,65 @@ class ImportEnrollmentByBirthYearIntoMysql(CourseEnrollmentTableDownstreamMixin,
         )
 
 
+class EducationLevelCodeMappingTask(luigi.Task):
+    """A static table that maps the education level codes found in auth_userprofile to canonical codes."""
+
+    output_root = luigi.Parameter()
+
+    MAPPING = {
+        'none': 'none',
+        'other': 'other',
+        'el': 'primary',
+        'jhs': 'junior_secondary',
+        'hs': 'secondary',
+        'a': 'associates',
+        'b': 'bachelors',
+        'm': 'masters',
+        'p': 'doctorate',
+        'p_se': 'doctorate',
+        'p_oth': 'doctorate'
+    }
+
+    def run(self):
+        with self.output().open('w') as output_file:
+            for item in self.MAPPING.iteritems():
+                output_file.write('\t'.join(item))
+                output_file.write('\n')
+
+    def output(self):
+        return get_target_from_url(url_path_join(self.output_root, 'mapping.tsv'))
+
+
+class EducationLevelCodeMappingTableTask(WarehouseMixin, ImportIntoHiveTableTask):
+    """A hive table for the code mapping."""
+
+    @property
+    def table_name(self):
+        return 'education_level'
+
+    @property
+    def columns(self):
+        return [
+            ('auth_userprofile_code', 'STRING'),
+            ('education_level_code', 'STRING')
+        ]
+
+    @property
+    def table_location(self):
+        return url_path_join(self.warehouse_path, self.table_name)
+
+    @property
+    def partition_date(self):
+        return 'default'
+
+    @property
+    def table_format(self):
+        return "ROW FORMAT DELIMITED FIELDS TERMINATED BY '\\t'"
+
+    def requires(self):
+        yield EducationLevelCodeMappingTask(output_root=self.partition_location)
+
+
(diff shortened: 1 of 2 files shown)


edx/analytics/tasks/enrollments.py
```diff
@@ -567,6 +627,14 @@ def columns(self):
             ('count', 'INT'),
         ]
 
+    def requires(self):
+        for req in super(EnrollmentByEducationLevelTask, self).requires():
```
brianhw: Does this need a 'flatten' call?  I can't remember whether requires() is necessarily flat for iterating like this, or if it can be an arbitrarily nested dict structure.  May not be an issue here anyway, but in case of cut-and-paste later....

edx/analytics/tasks/enrollments.py
```diff
@@ -534,6 +534,65 @@ def insert_source_task(self):
...
+    @property
+    def partition_date(self):
+        return 'default'
+
+    @property
+    def table_format(self):
+        return "ROW FORMAT DELIMITED FIELDS TERMINATED BY '\\t'"
+
+    def requires(self):
+        yield EducationLevelCodeMappingTask(output_root=self.partition_location)
```
brianhw, line 593: So we would not have a single location?  This would re-create the file in each partition.  An alternative is to write it to the table_location instead.  I guess it depends on how often we expect to change the mapping stored in this file.  It's a tradeoff -- if we change the mapping in code, we would have to delete the data file manually.  Part of the motivation for the question is whether this table would be used for other ad hoc BI queries.  Seems that a table location is more accessible than multiple partition locations.  But I don't really understand the requirements for this BI queries.

brianhw:
The code looks fine.  Two minor issues to touch base about. :+1:

edx/analytics/tasks/enrollments.py
```diff
@@ -534,6 +534,65 @@ def insert_source_task(self):
...
+
+    @property
+    def table_name(self):
+        return 'education_level'
+
+    @property
+    def columns(self):
+        return [
+            ('auth_userprofile_code', 'STRING'),
+            ('code', 'STRING')
```
rocha: Maybe the second column name can be a bit more descriptive, since both columns have the word code.

rocha:
Looks good to me. However, I am bit rusty in some finer details of the pipeline and hive tasks.

edx/analytics/tasks/enrollments.py
mulby: It gets flattened up stream... in fact we might not even have to iterate and yield here...

edx/analytics/tasks/enrollments.py
mulby, line 593: I don't think having multiple partition locations is actually any different than a single table location and it lets us "version" the data.

edx/analytics/tasks/enrollments.py
mulby: Yeah... I struggled with this, maybe "education_level_code"?

edx/analytics/tasks/enrollments.py
```diff
@@ -578,7 +646,7 @@ def columns(self):
         return [
             ('date', 'DATE NOT NULL'),
             ('course_id', 'VARCHAR(255) NOT NULL'),
-            ('education_level', 'VARCHAR(6)'),
+            ('education_level', 'VARCHAR(16)'),
```
brianhw, line 649: Is this cutting things too close?  "junior_secondary" is exactly 16, but doesn't give us much room for expansion.   (On the other hand, there may not be much need....)

edx/analytics/tasks/enrollments.py
brianhw: Clarified in person:  add flatten around the requires() call made here.

edx/analytics/tasks/enrollments.py
brianhw, line 593: Clarified in person: the Hive metastore hides the partition_date logic, so it's all the same to BI queries done against Hive.

edx/analytics/tasks/enrollments.py
mulby, line 649: yeah, we can always just regenerate the data set if we need to make this bigger

brianhw:
LGTM. :+1: (again)  Just squash.

mulby:
acceptance tests pass

http://jenkins-ci.analytics.edx.org/view/pull%20requests/job/edx-analytics-pipeline-acceptance-auto-pr/148/

I'm going to squash and merge

PR #33, opened 2014-10-30
Use an ansible version that is on PyPi

mulby:
For some reason 1.4.4 cannot be found. Not sure if it got pulled down or what.

Coincidentally a new version of the wheel was uploaded to S3. Ansible cannot be installed correctly from wheels, so any new virtualenvs built after the wheel was uploaded will break with the error "ERROR: file is not a legal parameter in an Ansible task or handler".

@brianhw

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -1,7 +1,7 @@
 Jinja2==2.7.3		# BSD License
 MarkupSafe==0.23	# BSD
 PyYAML==3.10		# MIT License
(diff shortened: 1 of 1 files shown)


mulby:
FYI - @e0d

e0d:
We're on Ansible 1.5, that should be a straight-forward upgrade I think.

brianhw:
LGTM.  :+1:

PR #34, opened 2014-10-30
Updated README with note to install requirements via make

clintonb:
- Updated mysql-connector-python to 1.2.3
- Updated README with flag for installing mysql-connector-python from an external source

@mulby @brianhw @rocha

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,9 +13,9 @@ Your machine will need the following in order to run the code in this repository
 
 All of the components above can be installed with your preferred package manager (e.g. apt, yum, [brew](http://brew.sh).
 
-The requirements in requirements/default.txt and requirements/test.txt can be installed with pip:
+The requirements in requirements/default.txt and requirements/test.txt can be installed with pip (via make):
 
-    pip install -U -r requirements/default.txt
+    make requirements
 
 *Known Issues on Mac OS X*
 


requirements/default.txt
```diff
@@ -9,7 +9,7 @@ filechunkio==1.5	# MIT
 html5lib==1.0b3 	# MIT
 httplib2==0.9 		# MIT
 mechanize==0.2.5	# BSD
-mysql-connector-python==1.2.2  	# GPL v2 with FOSS License Exception
+mysql-connector-python==1.2.3  	# GPL v2 with FOSS License Exception
```
rocha: I think you can add the allow-external flag here:

```
--allow-external mysql-connector-python==1.2.3      # GPL v2 with FOSS License Exception
```

That means we can remove the comments above.

rocha:
:rocket: check if you can use the external flag in the requirements file instead.

clintonb:
@rocha `Makefile` actually includes the line. Should I update `README.md` reference `make requirements` instead of the pip command?

rocha:
Well, if you add it the flag to the requirements file then you only have to keep it in one place instead of two.

clintonb:
Updated.

mulby:
Why do we need this change?

clintonb:
Only the README change is needed/desired. I was initially under the impression that 1.2.2 is no longer available as I did not see it on pypi, however that doesn't seem to be the case.

rocha:
Oh for some reason I thought it was related to the db update.

clintonb:
I updated the commit to only modify the README.

rocha:
why not the requirements file ? :disappointed:

PR #35, opened 2014-11-07
Add support for multiple "src" arguments for PathSetTask.

mulby:
Used by answer distribution, course enrollment, user location,
and enrollment reports.

This is the single commit extracted from the synthetic event PR that allows the daily course enrollment job to run with synthetic events.

### DIFF
--- a/edx/analytics/tasks/pathutil.py
+++ b/edx/analytics/tasks/pathutil.py
@@ -39,6 +39,7 @@ class PathSetTask(luigi.Task):
       manifest: a URL pointing to a manifest file location.
     """
     src = luigi.Parameter(
+        is_list=True,
         default_from_config={'section': 'event-logs', 'name': 'source'}
     )
     include = luigi.Parameter(is_list=True, default=('*',))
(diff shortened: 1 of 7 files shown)


mulby:
@brianhw authored this

@rocha can you review it?

edx/analytics/tasks/pathutil.py
```diff
@@ -50,21 +51,22 @@ def __init__(self, *args, **kwargs):
...
-                yield ExternalURL(source)
-        else:
-            # Apply the include patterns to the relative path below the src directory.
-            for dirpath, _dirnames, files in os.walk(self.src):
-                for filename in files:
-                    filepath = os.path.join(dirpath, filename)
-                    relpath = os.path.relpath(filepath, self.src)
-                    if any(fnmatch.fnmatch(relpath, include_val) for include_val in self.include):
-                        yield ExternalURL(filepath)
+        for src in self.src:
```
rocha, line 54: What is `self.src` is not a list? Should we add a check in case someone forgets to put `is_list=True` in the parameter? Alternatively, if it is not a list, then we can put it in one before iterating.

rocha:
Looks good to me. My only comment is that it may be nice to make it work with parameters with or without `is_list=True`.

edx/analytics/tasks/pathutil.py
brianhw, line 54: Another possibility is that we do for this what we did with other params -- move it into a single PathSetTaskMixin, and include that in the other tasks.

mulby, line 54: This feels like a higher standard of compatibility than we maintain elsewhere in this codebase. We generally assume that the parameters for the task will match the specified types, which I feel is a reasonable assumption.

mulby:
Acceptance tests passed for this, so I'm going to merge it:

http://jenkins-ci.analytics.edx.org/view/pull%20requests/job/edx-analytics-pipeline-unit-auto-pr/206/

PR #36, opened 2014-11-17
Resolve job failures due to ansible not being able to find the master host

mulby:
This is the code I'm going to run in prod tonight, hopefully it addresses the no hosts matched problem.

If it does, I'll tidy it up and we can actually review it with some rigor. Right now there are a bunch of magic numbers and unexplained behavior :wink: 

@brianhw @rocha

### DIFF
--- a/edx/analytics/tasks/launchers/remote.py
+++ b/edx/analytics/tasks/launchers/remote.py
@@ -6,6 +6,7 @@ import os
 import pipes
 from subprocess import Popen, PIPE
 import sys
+import time
 import uuid
 
 
@@ -38,9 +39,9 @@ def main():
     uid = arguments.remote_name or str(uuid.uuid4())
     log('Remote name = {0}'.format(uid))
 
-    inventory = get_ansible_inventory()
+    hostname = get_master_hostname(arguments)
     if arguments.shell:
-        return_code = run_remote_shell(inventory, arguments)
+        return_code = run_remote_shell(hostname)
     else:
         return_code = run_task_playbook(arguments, uid)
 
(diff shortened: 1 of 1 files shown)


brianhw:
Looks good so far.

mulby:
Not clear yet if this worked or not. Looking at the logs I don't see failures due to this issue... but I also don't see the retry logic kicking in. I'll give it another night or so to see if it actually triggers the retry logic.

rocha:
:rocket:

mulby:
this doesn't really fix the problem... although it may help?

The root cause is not retrieving the metadata, it's that the clusters fail to provision within the 50 minute timeout.

I'm assuming this is just due to the spot instances taking too long to provision.

I see a couple of options here:

1) we procede with executing the job even if we don't yet have all of the spot instances provisioned, as long as the master is ready. Here we run the risk of executing the job in a severally degraded cluster... it would allow us, however, to diagnose and attempt to resize in the morning (or something). This actually may speed up our jobs somewhat since the listing of S3 takes a really long time and we could be doing that on the master at the same time we are provisioning slave instances.
2) we somehow fall back onto non-spot market pricing if we can't provision a spot instance cluster.

Either way, we should fix the provisioning logic that is not failing when the cluster fails to provision. This makes it look like we can't find the cluster, when in fact it was never provisioned properly in the first place.

@brianhw @rocha

rocha:
I worry that if a job starts with a cluster that it is too small, the job may get stuck for a long time. I saw it when running the answers distributions using only 5 nodes. They got overloaded and some of the nodes were retired.

I am more towards option two. We can also see if changing the price helps before going for the non-spot instances. Do you know how much is the price difference of going from spot to normal instances?

mulby:
Regular node: $0.022 per Hour
Spot node: $0.0081 per Hour

rocha:
Almost three time the cost. It is probably worth it since the failures are not that frequent. 

I remembered Miki mentioned that this was the most cause of errors in the data pipeline of his previous job. It may be good to talk to him and see how they solved it.

brianhw:
I ran into a problem where spot nodes weren't available on a particular subnet, but were readily available on another.   So if we're looking to implement retry behavior, a candidate for this would be to try a different subnet, rather than trying with regular nodes.

In terms of cluster size, adding nodes can often be added later, but some parameters for running a job depend on the number of nodes available at the beginning.  So if it were _too_ severely degraded, this might cripple the resulting job, even if nodes were later added.  Setting a minimum-number-of-nodes would be possible, but I'm not sure how much additional value it would provide.  (I'm guessing that if it is not able to configure all the requested nodes in the timeframe, it's probably not going to configure enough anyway.)

PR #38, opened 2014-11-26
Support running the pipeline in a docker container

mulby:
This is really just about constructing a tweaked configuration file and including first class HDFS support.


mulby:
@pchiu33 @brianhw @rocha

brianhw:
LGTM.  

I assume that any change we make to config files should be made to the new docker.cfg file as well?

rocha:
:rocket:

PR #40, opened 2014-12-02
Incorrect variant for answer distribution bug fix

dylanrhodes:
This request fixes an issue we have noticed in which an answer's variant is not correctly reflected in the pipeline.  The cause of the bug is a conditional which assumes that an answer's variant cannot be changed over the duration of a course and therefore relies on the variant value for the most recent answer to fill in those of the rest.  This bug has been causing issues with inline analytics display (referenced <a href="https://github.com/Stanford-Online/edx-platform/pull/165">here</a>) specifically when questions which had had randomization enabled in the past and then disabled displayed incorrectly, as the older answers to the randomized version were reported to be non-randomized by the pipeline.

The changes include a small removal in the answer_dist task and alteration of a related test case.  There was a comment in the test case regarding the behavior which I have altered, so please let me know if the prior behavior was actually desired anywhere.  As always, feedback is encouraged.  @jbau @dcadams @mulby @sarina

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -6,3 +6,4 @@ Gabe Mulley <gabe@edx.org>
 Jason Bau <jbau@stanford.edu>
 Justin Helbert <jhelbert@edx.org>
 Will Daly <will@edx.org>
+Dylan Rhodes <dylanr@stanford.edu>
--- a/edx/analytics/tasks/answer_dist.py
+++ b/edx/analytics/tasks/answer_dist.py
@@ -279,9 +279,7 @@ class AnswerDistributionPerCourseMixin(object):
         # Now construct answer distribution for this input.
         problem_id = most_recent_answer.get('problem_id')
         problem_display_name = most_recent_answer.get('problem_display_name')
-        most_recent_question = most_recent_answer.get('question', '')
         answer_uses_value_id = ('answer_value_id' in most_recent_answer)
-        answer_uses_variant = (most_recent_answer.get('variant', '') != '')
         answer_dist = {}
         for _timestamp, value_string in reversed(values):
             answer = json.loads(value_string)
@@ -325,19 +323,12 @@ class AnswerDistributionPerCourseMixin(object):
                 answer_value_contains_html = (value_id is not None and value_id != '')
                 answer_value = self.stringify(answer_value, contains_html=answer_value_contains_html)
 
-                # If there is a variant, then the question might not be
-                # the same for all variants presented to students.  So
-                # we take the value (if any) provided in this variant.
-                # If there is no variant, then the question should be
-                # the same, and we want to go with the most recently
-                # defined value.
-                if answer_uses_variant:
-                    question = answer.get('question', '')
-                    variant = answer.get('variant') or ''
-                else:
-                    question = most_recent_question
-                    variant = ''
-
+                # Even if the most recent answer does not have a variant,
+                # the question could have been randomized earlier and then
+                # switched, so the variant must be checked for all answers.
+                question = answer.get('question', '')
+                variant = answer.get('variant') or ''
+                
                 # Key values here should match those used in get_column_order().
                 answer_dist[answer_grouping_key] = {
                     'ModuleID': problem_id,
--- a/edx/analytics/tasks/tests/test_answer_dist.py
+++ b/edx/analytics/tasks/tests/test_answer_dist.py
@@ -667,20 +667,24 @@ class AnswerDistributionPerCourseReduceTest(InitializeOpaqueKeysMixin, unittest.
         input_data_1 = (self.earlier_timestamp, json.dumps(answer_data_1))
         input_data_2 = (self.timestamp, json.dumps(answer_data_2))
         expected_output_2 = self._get_expected_output(answer_data_2)
-        # An older non-submission-based event should inherit some
+        # An older non-submission-based event should not inherit
         # information from a newer submission-based event.
-        # In particular, the Variant, the Question, and Problem Display Name.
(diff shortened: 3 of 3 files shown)


openedx-webhooks:
Thanks for the pull request, @dylanrhodes! I've created [OSPR-250](https://openedx.atlassian.net/browse/OSPR-250) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams. 

- supporting documentation
- edx-code email threads
- timeline information ('this must be merged by XX date', and why that is)
- partner information ('this is a course on edx.org')

All technical communication about the code itself will still be done via the Github pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).

We can't start reviewing your pull request until you've added yourself to the [AUTHORS](https://github.com/edx/edx-analytics-pipeline/blob/master/AUTHORS) file. Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst) file for more information.

sarina:
@dylanrhodes is there a JIRA issue for this bug?

sarina:
Also be sure to add yourself to AUTHORS for any new repo you contribute to.

sarina:
@dylanrhodes for all bugs, you should write a unit test that fails before your patch is applied, to prevent regressions and illustrate that your fix works (and for what situation(s) your fix works)

dylanrhodes:
@sarina I've made an analytics ticket for the bug <a href="https://openedx.atlassian.net/browse/AN-4111">here</a> (although I don't seem to have permission to view it) and linked there from the automatically created one (OSPR-250).  I've also added a test for the case in which an older answer for a problem has a variant set and a newer one does not which fails without the patch and passes with it.  Thanks for helping me get my ducks in a row.

sarina:
@dylanrhodes hopefully you can see the ticket now. Thanks!

sarina:
This looks reasonable to me (has test coverage) - although I can't see the build output. Handing off to the Analytics team for full review.

jbau:
@mulby This was the bug we briefly talked about during our hackathon.  Basically, it makes `answer_distribution` entries that aren't the last accurately report whether they have a "seed" value, which is something we need to make the display in inline-analytics sane.

mulby:
@jbau thanks for putting this together, we were actually just discussing it and @brianhw remembered why we did it that way... I'll let him chime in here to explain.

mulby:
errr thanks @dylanrhodes for putting this together, meant to FYI @jbau

brianhw:
I had responded in https://openedx.atlassian.net/browse/AN-4111, but I'll move the conversation here.

The difficulty here is that the "bug" was a conscious design decision. In many cases we encountered, the problem was initially randomized but the randomization wasn't actually being used. Then when the randomization was turned off, it made sense to roll up the results as if the problem had never been randomized.

So given that both scenarios are possible, it's harder to come up with a solution that works for both. Randomization makes things messy in any case. In talking with Victor, maybe the more reasonable approach is to accept the bug fix and assume that variants should be preserved. But then to add some heuristic that inspects the variants to determine if they actually have any effect, and if they don't seem to, then aggregate the answers and toss the variant information (so it works like the pre-bug-fix version).

jbau:
hi @brianhw thanks for explaining.  I understand the rationale and have thought about how to do that rollup of ineffective randomization myself.  I think it'd be better if the rollup step was an explicit one.  Because I think the current method might not be effective:  See the snippet of an answer distribution csv that I posted: https://gist.github.com/jbau/489beceb81d21ba4246d

This shows that the "initially unused randomization" results for a multiple-choice question are still reported in distinct lines, but now just have no reported "variant", so I think reporting variant is actually a step better.

sarina:
@brianhw my understanding people were "randomizing" their problems so that a "Reset" button would be shown on problems (that was the only way to get the Reset button to show). We should notify teams like this that there are now course-wide and per-problem ability to show the "Reset" button, without randomizing the problems:

in Studio advanced settings (course-wide): 
screen shot 2014-12-03 at 1 15 40 pm

per problem: 
screen shot 2014-12-03 at 1 18 48 pm

jbau:
@sarina that's a good point.  Maybe they could also mention the impact to analytics as a reason to not set randomization unless necessary.

brianhw:
@jbau Ouch.  That gist sure suggests that the rollup I thought was happening is indeed not happening.  
But when you suggest that the rollup step should be an explicit one, can you say more about what you had in mind?  Explicit in that the user makes a choice on a UI to "roll up", or that we perform that rollup automatically in the API, or just that the way we structure the answer-dist code is that it makes a pass to attempt a roll-up and does so if the answers across variants all "align"?  I was suggesting the latter.

jbau:
@brianhw I was suggesting to do it without user involvement, which could be in the API or the Task (maybe it's better to do it in the API so that the raw data is still available?)

jbau:
Hey @brianhw @mulby , so I think the course of action to make discrete chunks of work would be to review and merge this bugfix and then do a roll-up either in api or tasks, do you agree with that?

brianhw:
Yes.  This change looks good to me.  :+1: 

And it looks clear to me why the rollup wasn't working -- we always group with whatever variant value is present, independent of whether it was still randomized.  Would have been easy enough to fix, but this is the right way to go.

mulby:
:+1: agreed we should worry about rollup later (if ever)

sarina:
@brianhw @mulby - Dylan does not have merge permissions, so merge when it looks good to you

mulby:
@sarina thanks for the reminder, we are used to @jbau merging his own stuff in!

PR #41, opened 2014-12-03
Add EnrollmentValidationWorkflow.

brianhw:
Adds EnrollmentValidationWorkflow to combine validation event generation and synthetic event generation.  This includes a trigger to perform a student_courseenrollment dump when the interval ends with today.

Includes tests for enrollment validation event creation, and an acceptance test for the whole thing (validation events, synthetic events, and verifying that no synthetic events are missing).

Also changes event_output to tuple_output, and make event output the default.  (Tuple output is still the primary output format for tests.)


mulby:
:+1:

PR #42, opened 2014-12-05
Allow selecting courses on event exports.

rocha:
Reads the `courses` specifier on the configuration file, and only includes the listed courses in the export.

### DIFF
--- a/edx/analytics/tasks/event_exports.py
+++ b/edx/analytics/tasks/event_exports.py
@@ -2,18 +2,20 @@
 
 import logging
 import os
+import gzip
+from collections import defaultdict
 
 import gnupg
 import luigi
 import luigi.date_interval
 import yaml
-import gzip
 
 from edx.analytics.tasks.encrypt import make_encrypted_file
 from edx.analytics.tasks.mapreduce import MultiOutputMapReduceJobTask
 from edx.analytics.tasks.pathutil import EventLogSelectionMixin
 from edx.analytics.tasks.url import url_path_join, ExternalURL, get_target_from_url
 import edx.analytics.tasks.util.opaque_key_util as opaque_key_util
+
 
 log = logging.getLogger(__name__)
 
@@ -71,43 +73,35 @@ class EventExportTask(EventLogSelectionMixin, MultiOutputMapReduceJobTask):
 
         with self.input_local().open() as config_input:
             config_data = yaml.load(config_input)
-            self.organizations = config_data['organizations']
+            organizations = config_data['organizations']
 
-        # Map org_ids to recipient names, taking in to account org_id aliases. For example, if an org_id Foo is also
-        # known as FooX then two entries will appear in this dictionary ('Foo', 'recipient@foo.org') and
-        # ('FooX', 'recipient@foo.org'). Note that both aliases map to the same recipient.
+        # If org_ids are specified, restrict the processed files to that set of primary org ids.
+        if self.org_id:
+            organizations = {k: v for k, v in organizations.iteritems() if k in self.org_id}
+
         self.recipients_for_org_id = {}
-        self.primary_org_ids_for_org_id = {}
-        for org_id, org_config in self.organizations.iteritems():
-            # If org_ids are specified, restrict the processed files to that set of primary org ids.
-            if len(self.org_id) > 0 and org_id not in self.org_id:
-                continue
+        self.courses_for_org_id = {}
+        self.primary_org_ids_for_org_id = defaultdict(list)
+        self.org_id_whitelist = set()
 
-            recipients = org_config.get('recipients')
-            if not recipients:
-                # provide fallback to legacy parameter name:
-                recipients = [org_config['recipient']]
+        for org_id, org_config in organizations.iteritems():
+            aliases = [org_id] + org_config.get('other_names', [])
+
+            recipients = org_config['recipients']
             self.recipients_for_org_id[org_id] = recipients
-            self.primary_org_ids_for_org_id[org_id] = [org_id]
-            for alias in org_config.get('other_names', []):
-                # This entry is not explicitly used, but by creating it,
-                # we implicitly add the alias to the org whitelist.
-                self.recipients_for_org_id[alias] = recipients
-                # The same alias may be used by multiple organizations.
-                if alias in self.primary_org_ids_for_org_id:
-                    self.primary_org_ids_for_org_id[alias].append(org_id)
-                else:
-                    self.primary_org_ids_for_org_id[alias] = [org_id]
 
-        self.org_id_whitelist = set(self.recipients_for_org_id.keys())
+            self.courses_for_org_id[org_id] = org_config.get('courses')
+
+            for alias in aliases:
+                self.org_id_whitelist.add(alias)
+                self.primary_org_ids_for_org_id[alias].append(org_id)
 
         log.debug('Using org_id whitelist ["%s"]', '", "'.join(self.org_id_whitelist))
 
     def mapper(self, line):
-        value = self.get_event_and_date_string(line)
-        if value is None:
+        event, date_string = self.get_event_and_date_string(line) or (None, None)
+        if event is None:
             return
-        event, date_string = value
 
         if not self.is_valid_input_file():
             return
@@ -116,11 +110,18 @@ class EventExportTask(EventLogSelectionMixin, MultiOutputMapReduceJobTask):
         if org_id not in self.org_id_whitelist:
             log.debug('Unrecognized organization: org_id=%s', org_id or '')
             return
+
         # Check to see if the org_id is one that should be grouped with other org_ids.
         org_ids = self.primary_org_ids_for_org_id[org_id]
 
         for key_org_id in org_ids:
             key = (date_string, key_org_id)
+
+            # Include only requested courses
+            requested_courses = self.courses_for_org_id.get(key_org_id)
+            if requested_courses and self.get_course_id(event) not in requested_courses:
+                continue
+
             # Enforce a standard encoding for the parts of the key. Without this a part of the key
             # might appear differently in the key string when it is coerced to a string by luigi. For example,
             # if the same org_id appears in two different records, one as a str() type and the other a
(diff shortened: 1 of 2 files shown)


rocha:
@brianhw @mulby review please.

edx/analytics/tasks/event_exports.py
```diff
@@ -189,6 +191,20 @@ def _get_recipients(self, org_id):
...
         return recipients
 
+    def get_course_id(self, event):
+        """Gets course_id from event."""
+        # try to get the course from the context
+        course_id = event.get('context', {}).get('course_id')
+        if course_id:
+            return course_id
+
+        # try to get the id from the URL for implicit events
```
brianhw: The get_org_id logic below is pretty hairy and not alway correct, but was left in place because it is what we used in the past (and we didn't want to select different events than we did before).  But this method is not going to pull out the same course_id's that the get_org_id implicitly does.  This version is only going to pull a subset of course_ids out.  And in particular it won't do the right thing for any new-style id's.

edx/analytics/tasks/event_exports.py
```diff
@@ -71,43 +73,36 @@ def init_local(self):
...
+        self.org_id_whitelist = set()
+
+        for org_id, org_config in organizations.iteritems():
+            aliases = [org_id] + org_config.get('other_names', [])
 
-            recipients = org_config.get('recipients')
-            if not recipients:
-                # provide fallback to legacy parameter name:
-                recipients = [org_config['recipient']]
+            # use fallback to legacy recipients parameter name
```
brianhw: We should probably remove the legacy parameter name at this point, since we're well beyond ever using it again.

edx/analytics/tasks/event_exports.py
rocha: I don't know what to do here, since this is for implicit events, that don't use a proper ids. Now that I think about it, I am probably doing the wrong thing here, since I think we are not including the `http://edx.org` part in the urls here.

edx/analytics/tasks/event_exports.py
rocha: Yeah, I was thinking the same thing this morning, but forgot to write it down. I'll make the change.

edx/analytics/tasks/event_exports.py
rocha: Now I understand what you mean. I have to somehow use the function below to find the org and the course_id. Ouch.

mulby:
I agree with brian's comments, I'll wait for the next revision

rocha:
Updated.

edx/analytics/tasks/event_exports.py
mulby: wait, I thought we were going to remove the implicit event inference logic

edx/analytics/tasks/event_exports.py
```diff
@@ -198,65 +225,78 @@ def get_org_id(self, item):
...
-                    if course_key and '/' not in unicode(course_key):
-                        return course_key.org
-                    else:
-                        # It doesn't matter if we found a good deprecated key.
-                        # We need to provide backwards-compatibility.
-                        return get_slash_value(page, 4)
+            if event['event_source'] == 'server':
+                return self._parse_server_event(event)
+            elif event['event_source'] == 'browser':
+                return self._parse_browser_event(event)
```
mulby, line 241: I'm a little nervous about touching this code. I realize it's a straightforward copy-paste, but it feels unnecessarily risky.

edx/analytics/tasks/event_exports.py
rocha: I guess I can remove them, it seemed like an easy thing to still left here while delivering some events that did not have any other equivalent for older courses. I was specifically thinking about forums.

mulby: Won't those events have the course_id in their context?

edx/analytics/tasks/event_exports.py
rocha, line 241: We do have unit tests for this area. I prefer to do a small cleanup for the sake of sanity, but if you think it is too risky then I'll let it as it was.

edx/analytics/tasks/event_exports.py
mulby: Pulling one sample event, I see the course_id in the context....

edx/analytics/tasks/event_exports.py
mulby, line 241: Do we have full branch coverage? There are a lot of wonky edge cases in here...

rocha, line 241: You are not going to like my PR on analytics-exporter :worried:

rocha, line 241: I can check, but @brianhw did add a large number of tests.

rocha, line 241: Yeah, full coverage of those sections.

edx/analytics/tasks/event_exports.py
rocha: AH, I was referring to old events. The courses requested were from 2013, which probably does not have the context. The way I wrote the code it will check first if the field is there, and if not then check the URL.

My thought was that since we are not going to give them all the events (because it is too much work to write a detailed `get_course_id`, at least we give them a large number of implicit ones.

mulby: I see, that sounds reasonable. Let's keep it in.

edx/analytics/tasks/event_exports.py
mulby, line 241: Sounds like we're in good shape then.

rocha:
Test should pass now.

mulby:
:+1:

Given the scope of the change we may want to consider spot checking the export of an org.

mulby:
One second thought, perhaps we could batch up that validation with the the small change to support the mobile events.

rocha:
I was planning on doing that with the other PR (loading the configuration files from `analytics-secure`).

edx/analytics/tasks/event_exports.py
brianhw, line 241: Sheesh -- you guys trust my tests?!

rocha, line 241: But coverage.py!

edx/analytics/tasks/event_exports.py
```diff
@@ -198,65 +224,78 @@ def get_org_id(self, item):
...
-
         except Exception:  # pylint: disable=broad-except
-            log.exception('Unable to determine institution for event: %s', unicode(item).encode('utf8'))
+            log.exception('Unable to determine organization for event: %s', unicode(event).encode('utf8'))
+
+        return None
+
+    def _parse_server_event(self, event):
+        # Different behavior based on type of event source.
+        if event['event_source'] == 'server':
```
brianhw: Just wondering if this redundancy is necessary -- you already tested this before calling the method.

rocha: Good call. Copy-pasta error.

brianhw: ... and you don't make the same test in the browser code.

rocha: Fixed.

edx/analytics/tasks/event_exports.py
```diff
@@ -189,7 +190,32 @@ def _get_recipients(self, org_id):
...
+
+        # Try to get the id from the URL for implicit events which
+        # should only happen for old events, that did not have a
+        # context at the time they were emitted.
+        event_type = event.get('event_type', '')
+        if event_type.startswith('/courses/'):
+            course_id = '/'.join(event_type.split('/')[2:])
+            if opaque_key_util.is_valid_course_id(course_id):
+                return course_id
+
```
brianhw, line 223: What subset of the org events are you trying to include in the per-course events?  It's clearly going to be a subset, since you don't include all the other cases.  And it can't include any that didn't already find the org, because you've already skipped them.  (And we can't change the org code because it "must" be idempotent.)

But in terms of "old" data, why only try for server events that appear as "/courses/<course_id>"?  First of all, I would expect a bunch would have more context following the course_id, but this includes the rest of the url.   It also assumes that it starts with /courses/, which is a subset of events.  And even if you got that (by using [2:5] or something similar), why just look at a subset of old server events and not also get old browser events? 

Why not just use:  course_key = opaque_key_util.get_course_key_from_url(url)  where url = event_type if server and url = page if browser.

brianhw:
I would like to see the per-course logic get updated -- seems arbitrary now.

rocha:
It is fairly arbitrary. There are two extremes and I choose some middle-ground: you can only include the events the have the `course_id` in the context, or you can try to make a function that deals with all the corner cases for events that have course_id in arbitrary locations. I choose some point in the middle with the former and some arbitrary subset of the later. My rationale was that there was less corner cases for server events than browser events.

Your suggestion of using `get_course_key_from_url` however is a lot better. I didn't use it out of ignorance. I'll make the change. Thanks.

rocha:
@brianhw updated with your recommendations. Still not perfect, but a bit better than the previous version.

I am in the process of adding more tests for account for the changes, but I'll be really useful to get your early feedback.

edx/analytics/tasks/event_exports.py
```diff
@@ -189,7 +190,37 @@ def _get_recipients(self, org_id):
...
+            course_key = opaque_key_util.get_course_key_from_url(event_type)
+            if course_key:
+                return unicode(course_key)
+
+        # Try to get the course_id from event page for browser events
+        if event.get('event_source') == 'browser':
+            page = event.get('page', '')
+            course_key = opaque_key_util.get_course_key_from_url(page)
+            if course_key:
+                return unicode(course_key)
```
mulby: What about something like this? (slightly more DRY)

``` python
if event.get('event_source') == 'server':
   url = event.get('event_type', '')
else:
   url = event.get('page', '')
course_key = opaque_key_util.get_course_key_from_url(url)
return unicode(course_key) if course_key else None
```

rocha: Good suggestion. Updated.

rocha:
@mulby  @brianhw updated

mulby:
:+1:

brianhw:
LGTM. :+1:

PR #44, opened 2014-12-05
Define AnswerDistributionWorkflow.

brianhw:
Combines AnswerDistributionOneFilePerCourseTask and AnswerDistributionToMySQLTaskWorkflow into a single task.

@rocha @mulby

### DIFF
--- a/edx/analytics/tasks/answer_dist.py
+++ b/edx/analytics/tasks/answer_dist.py
@@ -18,7 +18,7 @@ from edx.analytics.tasks.mapreduce import MapReduceJobTask, MultiOutputMapReduce
 from edx.analytics.tasks.pathutil import PathSetTask
 from edx.analytics.tasks.url import ExternalURL, IgnoredTarget
 from edx.analytics.tasks.url import get_target_from_url, url_path_join
-from edx.analytics.tasks.mysql_load import MysqlInsertTask
+from edx.analytics.tasks.mysql_load import MysqlInsertTask, MysqlInsertTaskMixin
 import edx.analytics.tasks.util.eventlog as eventlog
 import edx.analytics.tasks.util.opaque_key_util as opaque_key_util
 
@@ -329,7 +329,7 @@ class AnswerDistributionPerCourseMixin(object):
                 # switched, so the variant must be checked for all answers.
                 question = answer.get('question', '')
                 variant = answer.get('variant') or ''
-                
+
                 # Key values here should match those used in get_column_order().
                 answer_dist[answer_grouping_key] = {
                     'ModuleID': problem_id,
@@ -563,7 +563,7 @@ def get_text_from_element(node):
 # Task requires/output definitions
 ##################################
 
-class BaseAnswerDistributionTask(MapReduceJobTask):
+class BaseAnswerDistributionDownstreamMixin(object):
     """
     Base class for answer distribution calculations.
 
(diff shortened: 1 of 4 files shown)


edx/analytics/tasks/answer_dist.py
```diff
@@ -563,7 +563,7 @@ def get_text_from_element(node):
 # Task requires/output definitions
 ##################################
 
-class BaseAnswerDistributionTask(MapReduceJobTask):
+class BaseAnswerDistributionDownstreamMixin(object):
```
rocha, line 566: Maybe just `BaseAnswerDistributionMixin`?

edx/analytics/tasks/answer_dist.py
```diff
@@ -580,16 +580,24 @@ class BaseAnswerDistributionTask(MapReduceJobTask):
...
     include = luigi.Parameter(is_list=True, default=('*',))
-    # A manifest file is required by hadoop if there are too many input paths. It hits an operating system limit on the
-    # number of arguments passed to the mapper process on the task nodes.
+    # A manifest file is required by hadoop if there are too many
+    # input paths. It hits an operating system limit on the number of
+    # arguments passed to the mapper process on the task nodes.
     manifest = luigi.Parameter(default=None)
 
+
+class BaseAnswerDistributionTask(MapReduceJobTask):
```
rocha, line 589: Maybe this inherits also from `BaseAnswerDistributionMixin` above? Do all classes that inherit from `BaseAnswerDistributionTask` also do from `BaseAnswerDistributionDownstreamMixin`?

edx/analytics/tasks/mapreduce.py
```diff
@@ -202,13 +199,17 @@ class MultiOutputMapReduceJobTask(MapReduceJobTask):
     Parameters:
         output_root: a URL location where the split files will be stored.
         delete_output_root: if True, recursively deletes the output_root at task creation.
+        marker:  a URL location where a marker file should be written.
```
rocha: I forgot, what is a maker file? The one that is create when the task has finished?

rocha:
Looks good! So much better. I added some comments that would be nice to address. I don't really understand the use of `downstream` since normally we have use `SomethingMixin` as the parameters for `SomethingTask`. Is there a new connotation that I didn't get?

mulby:
This generally looks good, did your resolve the failure you were seeing yesterday?

edx/analytics/tasks/mapreduce.py
brianhw: Yes, that's what a marker file is.  But I'll clarify the comment.

brianhw:
My change did resolve the failure I was seeing yesterday -- removing the IgnoredTarget used for answer_dist's output and going with the MultiMapReduce implementation with the location as a parameter.

brianhw:
Drat -- I wrote up a long history of Mixin and DownstreamMixin last night, but I must not have published it.   I'll recreate it....

mulby:
This looks good to me as is.

:+1:

brianhw:
So the short version of what I wrote:
- We originally had Mixin for map-reduce stuff that was not needed for a Luigi Task, and then mixed it with one (or more) task implementations.
- We then defined DownstreamMixin for defining Luigi parameters that were needed in downstream task definitions (to define the params to be passed to the required tasks).  
- We then (mostly) stopped using separate Mixins for non-Luigi task methods.  Only then did we begin to use Mixin as shorthand for DownstreamMixin.
- Answer distribution code has a number of Mixin classes already.  So when I added a DownstreamMixin, I used the full name to maintain the distinction.  Note that it's only internal to the file -- we shouldn't need the mixin in other classes.
- If we have a clear nomenclature that we want to adopt, I'm happy to make another pass to refactor this further.  (This and the enrollment reports are the older parts of the code.)

rocha:
@brianhw that makes sense. I probably was not working on the pipeline when we started using `DownstreamMixin`, hence my confusion. Other possible names `ArgumentMixin`, `TaskMixin`, `ParameterMixin`, but that is something for later.

rocha:
:+1:

PR #47, opened 2014-12-11
User activity by day and month

mulby:
Early review of direction, here's the approach:
- Refactor user activity to build a common table that shows which users performed which categories of action on each day in each course
- Generate downstream aggregates from the common table
- Introduce a calendar table (which can later be extended to include complex things like fiscal year/quarter stuff)

If this approach looks good I'll write a bunch of unit tests for the calendar table, and increase the coverage of the user activity computation.

@brianhw @rocha


edx/analytics/tasks/user_activity.py
```diff
@@ -106,159 +99,172 @@ def _encode_tuple(self, values):
...
+                mapreduce_engine=self.mapreduce_engine,
+                n_reduce_tasks=self.n_reduce_tasks,
+                source=self.source,
+                interval=self.interval,
+                pattern=self.pattern,
+                warehouse_path=self.warehouse_path,
+            ),
+            CalendarTableTask(
+                warehouse_path=self.warehouse_path,
+                overwrite=self.hive_overwrite,
```
brianhw, line 193: Where does hive_overwrite come from?

edx/analytics/tasks/user_activity.py
```diff
@@ -106,159 +99,172 @@ def _encode_tuple(self, values):
...
             return values[0].encode('utf8')
 
-    def mapper(self, line):
-        """Default mapper implementation, that always outputs the log line, but with a configurable key."""
-        value = self.get_event_and_date_string(line)
-        if value is None:
-            return
-        event, date_string = value
+    def reducer(self, key, values):
+        num_events = sum(values)
```
brianhw, line 115: So we're just summing this for fun for now -- it's not actually being used anywhere, best I can see.  May be interesting...

brianhw:
Looks real nice!

edx/analytics/tasks/user_activity.py
mulby, line 115: Correct.

edx/analytics/tasks/user_activity.py
mulby, line 193: The parent class. It separates the concept of overwriting the Hive data and the RDBMS data.

mulby:
@brianhw @rocha updated

edx/analytics/tasks/user_activity.py
```diff
@@ -270,37 +278,118 @@ def indexes(self):
...
+        """Given the parameters, compute the first and last date of the interval."""
+        # In order to use the modulus operator we have to ensure months are counted from 0.
+        zero_offset_month = self.end_date.month - 1
+        # Note that the first month to include may be in a previous year.
+        new_month = ((zero_offset_month - self.months) % 12) + 1
+        # Depending on the ending month and the number of months to include we may need to adjust the year.
+        year_diff = (zero_offset_month - self.months) / 12
+        new_year = self.end_date.year + year_diff
+        starting_date = datetime.date(new_year, new_month, 1)
+        ending_date = self.end_date.replace(day=1)
```
brianhw, line 359: This way of calculating starting_date looks complicated.  What happens if you just did:

 starting_date = (ending_date - datetime.timedelta(days=1)).replace(day=1)

brianhw:
LGTM.  Starting_date fix is optional.  (But would need squashing anyway.)  :+1:

edx/analytics/tasks/user_activity.py
brianhw, line 359: Nevermind.  I forgot about self.months.  It could be done recursively, of course, but may not be worth it.

edx/analytics/tasks/calendar.py
```diff
@@ -0,0 +1,104 @@
...
+from edx.analytics.tasks.url import get_target_from_url, url_path_join
+from edx.analytics.tasks.util.hive import HiveTableTask, HivePartition
+from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
+
+log = logging.getLogger(__name__)
+
+try:
+    from isoweek import Week
+except ImportError:
+    log.warning('isoweek module not available')
```
rocha: Why catch this exception? It seems that the class bellow will not work if the import fails.

mulby: Slave nodes need to import this file even though they are not executing any code in it...

rocha: Ah right. Can you add a comment? It is weird when you first read it, you kind of expect a fallback later on.

edx/analytics/tasks/user_activity.py
```diff
@@ -1,86 +1,95 @@
...
 import edx.analytics.tasks.util.eventlog as eventlog
-import edx.analytics.tasks.util.opaque_key_util as opaque_key_util
+from edx.analytics.tasks.util.hive import WarehouseMixin, HiveTableTask, HivePartition, HiveQueryToMysqlTask
 
 log = logging.getLogger(__name__)
 
+try:
+    from isoweek import Week
+except ImportError:
+    log.warning('isoweek module not available')
```
rocha: What if we move this to one place:

Here: `from edx.analytics.tasks.calendar import Week`

There:

```
try:
  from isoweek import Week
except ImportError:
  Week = NotImplemented
```

Another option is to put it in `utils` or do the import inside the function that uses Week.

edx/analytics/tasks/user_activity.py
```diff
@@ -1,86 +1,95 @@
...
-    This class provides a parameterized mapper.  Derived classes must
-    override the predicates applied to events, and the definition of
-    what is a key and what a value from the mapper.  Derived classes
-    also define the reducer that consumes the mapper output.
+    output_root = luigi.Parameter()
 
-    Parameters:
-        output_root: Directory to store the output in.
+    def mapper(self, line):
+        value = self.get_event_and_date_string(line)
```
rocha, line 45: Minor: I feel a bit of deja vu. Maybe the return type of `get_event_and_date_string` when there is no valid event could be `(None, None)`. That way we don't have to do the unpacking below.

It is not a big deal, just a little nicety.

edx/analytics/tasks/user_activity.py
rocha, line 359: We can also use `dateutil`, it has a handly `relativedelta` object that makes month based operations very easy (it takes into account the differences in the number of days, etc).

rocha:
I like it, no major comments from me. :rocket:

I think that one item for discussion later is data for on-going time intervals. I think it is still useful to compute the results, maybe putting them in a separate place, or marking the row with a "is_partial". This could be specially useful for plots, graphs, etc. and could help make our results less stale.

edx/analytics/tasks/user_activity.py
mulby, line 45: Yeah, we do repeat this pattern often. I'll leave it to a separate patch that does a comprehensive refactor to this effect, though. I don't want to muddy this one with a broad change like that.

rocha, line 45: Makes sense.

edx/analytics/tasks/user_activity.py
mulby, line 359: I don't think it's worth the overhead of yet another third party package. If/when we need to do other complex date deltas that aren't handled by datetime.timedelta then we should definitely consider it.

rocha, line 359: Yeah, it was a suggestion. `dateutil` is pretty useful, but not necessary at the moment. I disagree on the overhead, which I don't think it is very much operationally. On the other hand, you would have saved some time coding if you used it instead.

mulby, line 359: It was more that a relative delta of months can mean different things to different people and I would have had to go verify that the third party lib was doing what I expect it to... I was reading that in finance 1 month can mean 30 days regardless of the number of days in particular months. Furthermore, other people wandering through the code base may have the same question (what is the third party lib doing?), and I would just have to have a comment explaining it. I figured that in this case it might be simpler just to include the logic since it's only 2 lines of code.

Without some kind of ISO standard (like there is for weeks) I figured using a tool like relativedelta could actually be _less_ clear to a future reader.

The fact that "months" is not included in timedelta is also indicative of this lack of clarity, see:
https://www.google.com/search?q=python+timedelta+months

mulby, line 359: All of that said, I had meant to add unit tests for this, to capture the _actual_ intent, and forgot to! So I really need to go back in and those!

rocha, line 359: In the code that I have used (and worked before) if you are using 30 days months, you the delta is 30 days. There is no ambiguity there. Another pattern I have seen is that date operations that are not normal, use a prefix as in "business_day", "end_of_financial_month" etc.

The omission from "months" in `datetime` is more an indicator of the deficiencies of the library, than of the ambiguity in defining a month. Any other time related library that I know (`dateutil`, `arrows`, `quantlib`, `pandas`, etc) does not have that problem.

It is very easy to mess up with date calculations (not in this case however) and I normally use libraries when dealing with them.

rocha, line 359: We can leave it as it is.

mulby, line 359: I guess I still think the ambiguity is relevant, but your argument is a good one (someone who was dealing with 30 day months is going to be in the minority). So I guess I'm convinced, I put in dateutil.

I still think there is some ambiguity around leap years, but it's not really worth thinking about too much since the library does what I want it to do and the unit tests express my intent.

rocha, line 359: Thanks! I was not really pushing hard for it, just giving you my arguments. It would have been fine with me if you didn't include it.

`dateutil` should deal with leap years and all those other exceptions. That is basically why I preferred to use some library for date manipulation, those things are very easy to get wrong.

I think you can also replace `isoweek` and use something in `dateutil` but that is not really required for this PR.

mulby, line 359: I was hoping to just use one of them, but I couldn't find anything related to ISO weeks in dateutil (granted I didn't look very hard). I would strongly prefer we stick to ISO weeks since it is very easy to describe and very clearly defined. Lots of different libraries deal with weeks differently so I wanted to stick to a standard.

rocha:
Failing tests. Other than that another :rocket: 

Don't forget to rebase the `fixups!` before merging (that already happened to me).

brianhw:
LGTM.  :+1:

brianhw:
(after a major squash....)

mulby:
user activity acceptance test is finally passing... merging

PR #48, opened 2014-12-11
Fix flaky acceptance test

mulby:
This test occasionally fails due to the SQL query not being specific enough about how the records should be ordered. Note that for country code "TH" it was valid for the database to return the two records in either order.

@brianhw @rocha @benpatterson


benpatterson:
LGTM :+1:

brianhw:
LGTM2 :+1:

rocha:
LGTM3 :+1:

PR #50, opened 2014-12-16
Benp/add total event count

benpatterson:
@rocha @brianhw @mulby 

I think it makes sense to get started on review. For now, could you ignore (or at most, skim) the test files? 

The prod code is worth a closer look at this point, and I welcome any feedback.

### DIFF
--- a/edx/analytics/tasks/overall_events.py
+++ b/edx/analytics/tasks/overall_events.py
@@ -1,0 +1,36 @@
+"""Compute overall event metrics"""
+
+import logging
+
+import luigi
+
+from edx.analytics.tasks.mapreduce import MapReduceJobTask
+from edx.analytics.tasks.pathutil import EventLogSelectionMixin
+from edx.analytics.tasks.url import get_target_from_url
+
+log = logging.getLogger(__name__)
+
+
+class TotalEventsDailyTask(EventLogSelectionMixin, MapReduceJobTask):
+    """Produce a dataset for total events within a given time period."""
+
+    output_root = luigi.Parameter()
+
+    def mapper(self, line):
+        event, date_string = self.get_event_and_date_string(line) or (None, None)
+        if event is None:
+            return
+
+        yield (date_string), 1
+
+    def reducer(self, key, values):
+        """Emit counts for each day."""
+
+        count = sum(values)
+        yield key, count
+
+    # Take advantage of a combiner to optimize network hops
+    combiner = reducer
+
+    def output(self):
+        return get_target_from_url(self.output_root)
--- a/edx/analytics/tasks/tests/test_overall_events.py
+++ b/edx/analytics/tasks/tests/test_overall_events.py
@@ -1,0 +1,165 @@
+"""Tests overall count of events"""
+
+import json
+
+import luigi
+
+from edx.analytics.tasks.tests import unittest
+from edx.analytics.tasks.overall_events import TotalEventsDailyTask
+from edx.analytics.tasks.tests.opaque_key_mixins import InitializeOpaqueKeysMixin
+
+from mock import patch
+
+
+class TotalEventsTaskMapTest(InitializeOpaqueKeysMixin, unittest.TestCase):
+    """Ensure events of various flavors are counted"""
+
+    def setUp(self):
+
+        fake_param = luigi.DateIntervalParameter()
+        self.task = TotalEventsDailyTask(
+            interval=fake_param.parse('2014-12-05'),
+            output_root='/fake/output'
+        )
+
+        self.initialize_ids()
+        self.task.init_local()
+
+        self.event_type = "edx.course.enrollment.activated"
+        self.timestamp = "2014-12-05T22:13:09.691008+00:00"
+        self.user_id = 5
+
+    def _create_event_log_line(self, **kwargs):
+        """Create an event log with test values, as a JSON string."""
+        return json.dumps(self._create_event_dict(**kwargs))
+
+    def _create_event_dict(self, **kwargs):
+        """Create an event log with test values, as a dict."""
+        # Define default values for event log entry.
+        event_dict = {
+            "username": "test_user",
+            "host": "test_host",
+            "session": "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb",
+            "event_source": "server",
+            "name": self.event_type,
+            "event_type": self.event_type,
+            "context": {
+                "course_id": self.course_id,
+                "org_id": self.org_id,
+                "user_id": self.user_id,
+            },
+            "time": "{0}+00:00".format(self.timestamp),
+            "ip": "127.0.0.1",
+            "event": {
+                "course_id": self.course_id,
+                "user_id": self.user_id,
+                "mode": "honor",
+            },
+            "agent": "blah, blah, blah",
+            "page": None
+        }
+        event_dict.update(**kwargs)
+        return event_dict
+
+    def test_explicit_event(self):
+        line = self._create_event_log_line(user_id="")
+        self.assertEquals(tuple(self.task.mapper(line)), (('2014-12-05', 1), ))
+
+    def test_no_timestamp(self):
+        line = self._create_event_log_line(timestamp="")
+        self.assertEquals(tuple(self.task.mapper(line)), (('2014-12-05', 1), ))
+
+    def test_bad_event(self):
+        line = "bad event"
+        self.assertEquals(tuple(self.task.mapper(line)), tuple())
+
+    def test_event_no_ids(self):
+        """
+        Many events (generally implicit events) have typical edx IDs missing or unavailable
+        because of their contexts. This test ensures these events are still counted.
+        """
+        self.empty_ids()
+        line = self._create_event_log_line()
+        self.assertEquals(tuple(self.task.mapper(line)), (('2014-12-05', 1), ))
+
+    def test_implicit_event(self):
+        event = {
+            "username": "",
+            "host": "precise64",
+            "event_source": "server",
+            "event_type": "/jsi18n/",
+            "context": {
+                "user_id": "",
+                "org_id": "",
+                "course_id": "",
+                "path": "/jsi18n/"
+            },
+            "time": "2014-12-05T22:11:29.689805+00:00",
+            "ip": "10.0.2.2",
+            "event": "{\"POST\": {}, \"GET\": {}}",
+            "agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_5) AppleWebKit/537.36 (KHTML, like Gecko) "
+                     "Chrome/39.0.2171.71 Safari/537.36",
+            "page": "null"
+        }
+        line = json.dumps(event)
+        self.assertEquals(tuple(self.task.mapper(line)), (('2014-12-05', 1), ))
+
+    def test_no_time_element(self):
+        event = {
+            "username": "faker",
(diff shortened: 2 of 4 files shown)


edx/analytics/tasks/overall_events.py
```diff
@@ -0,0 +1,47 @@
...
+        if value is None:
+            return
+        event, date_string = value
+
+        event_type = event.get('event_type')
+        if event_type is None:
+            log.error("encountered event with no event_type: %s", event)
+            return
+
+        timestamp = eventlog.get_event_time_string(event)
```
mulby: I don't think we need this check (nor the one for event_type)

edx/analytics/tasks/overall_events.py
```diff
@@ -0,0 +1,47 @@
...
+            return
+
+        yield (date_string), 1
+
+
+    def reducer(self, key, values):
+        """Emit counts for each day."""
+
+        count = sum(int(i) for i in values)
+        yield key, count
```
mulby, line 30: for better or worse luigi calls `eval()` on the strings provided by hadoop, so these will already be integers. You should be able to do: `yield key, sum(values)`

edx/analytics/tasks/tests/test_overall_events.py
```diff
@@ -0,0 +1,174 @@
...
+                "mode": "honor"
+            },
+            "event_type": "edx.course.enrollment.activated"
+        }
+
+        line = json.dumps(event)
+        self.assertEquals(tuple(self.task.mapper(line)), (('2014-12-05', 1), ))
+
+    def test_implicit_event(self):
+#        event = '''{"username": "", "host": "precise64", "event_source": "server", "event_type": "/jsi18n/", "context": {"user_id": "", "org_id": "", "course_id": "", "path": "/jsi18n/"}, "time": "2014-12-05T22:11:29.689805+00:00", "ip": "10.0.2.2", "event": "{\"POST\": {}, \"GET\": {}}", "agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/39.0.2171.71 Safari/537.36", "page": "null"}'''
```
mulby: I would remove this comment

edx/analytics/tasks/overall_events.py
```diff
@@ -0,0 +1,47 @@
...
+
+        timestamp = eventlog.get_event_time_string(event)
+        if timestamp is None:
+            log.error("encountered event with bad timestamp: %s", event)
+            return
+
+        yield (date_string), 1
+
+
+    def reducer(self, key, values):
```
mulby, line 26: Also, we can take advantage of a combiner here to save data over the wire: you just have to put `combiner = reducer` in the class definition and it will use the reducer as a combiner.

edx/analytics/tasks/tests/test_overall_events.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def test_explicit2_event(self):
+        line = self._create_event_log_line(user_id="")
+        self.assertEquals(tuple(self.task.mapper(line)), (('2014-12-05', 1), ))
+
+    def test_no_timestamp(self):
+        line = self._create_event_log_line(timestamp="")
+        self.assertEquals(tuple(self.task.mapper(line)), (('2014-12-05', 1), ))
+
+    def test_blank_event(self):
```
mulby: Do these actually occur? Either way, I'm not sure we actually want to exclude them from the count, since I suspect splunk won't filter them out of its count...

edx/analytics/tasks/tests/test_overall_events.py
```diff
@@ -0,0 +1,174 @@
...
+            "event": {
+                "course_id": "edX/DemoX/Demo_Course",
+                "user_id": 5,
+                "mode": "honor"
+            },
+            "event_type": "edx.course.enrollment.activated"
+        }
+
+        line = json.dumps(event)
+        patcher = patch('luigi.hadoop.JobTask.incr_counter').start()
```
mulby: Is this patch used?

mulby:
General approach looks good, called out a few nits.

edx/analytics/tasks/overall_events.py
benpatterson: Ok, I can test those out

edx/analytics/tasks/tests/test_overall_events.py
benpatterson: The real examples I ran into was when timestamp was missing. Those didn't get indexed by splunk (though, if the timestamp was missing in the event, there were other things wrong with it...)

edx/analytics/tasks/tests/test_overall_events.py
benpatterson: The patch is just sitting in there to keep unit test output clean. incr_counter emits output to stderr when an event is not parseable (IIRC).

edx/analytics/tasks/tests/test_overall_events.py
benpatterson: Agreed.  I'm planning to refactor this test file as part of this PR. There's some repetition in here at this point.

edx/analytics/tasks/overall_events.py
```diff
@@ -0,0 +1,47 @@
...
+from edx.analytics.tasks.mapreduce import MapReduceJobTask
+from edx.analytics.tasks.pathutil import EventLogSelectionMixin
+from edx.analytics.tasks.url import get_target_from_url
+from edx.analytics.tasks.util import eventlog
+
+log = logging.getLogger(__name__)
+
+
+class TotalEventsTask(EventLogSelectionMixin, MapReduceJobTask):
+    """Produce a dataset for total events within a given time period."""
```
rocha, line 15: Maybe change the name to `TotalEventsDailyTask` or update the documentation that the count is done by day.

edx/analytics/tasks/tests/test_overall_events.py
rocha: Can you add a comment about this?

edx/analytics/tasks/overall_events.py
benpatterson, line 15: done

edx/analytics/tasks/overall_events.py
benpatterson, line 26: done. Though, it didn't reduce the number of output files. Does that sound right?

edx/analytics/tasks/overall_events.py
benpatterson, line 30: Done.

edx/analytics/tasks/tests/test_overall_events.py
benpatterson: Done + renamed test method.

edx/analytics/tasks/overall_events.py
mulby, line 26: Yeah it won't reduce the number of output files, it will just run a reduce pass on the data computed by each mapper so instead of transmitting a bunch of "date, 1" strings it will actually pre-aggregate by date before sending to the reducer which will perform the global sum. The output should be identical, it is simply a performance optimization.

edx/analytics/tasks/overall_events.py
```diff
@@ -0,0 +1,39 @@
...
+log = logging.getLogger(__name__)
+
+
+class TotalEventsDailyTask(EventLogSelectionMixin, MapReduceJobTask):
+    """Produce a dataset for total events within a given time period."""
+
+    output_root = luigi.Parameter()
+
+    def mapper(self, line):
+        value = self.get_event_and_date_string(line)
```
brianhw: You can use:

```
event,  date_string = self.get_event_and_date_string(line) or (None, None)
if event is None:
    return
```

edx/analytics/tasks/tests/test_overall_events.py
```diff
@@ -0,0 +1,178 @@
...
+        line = self._create_event_log_line(user_id="")
+        self.assertEquals(tuple(self.task.mapper(line)), (('2014-12-05', 1), ))
+
+    def test_no_timestamp(self):
+        line = self._create_event_log_line(timestamp="")
+        self.assertEquals(tuple(self.task.mapper(line)), (('2014-12-05', 1), ))
+
+    def test_event_no_ids(self):
+        """
+        Many events (generaly implicit events) have typical edx IDs missing or unavailable
```
brianhw: generaly/generally

edx/analytics/tasks/tests/test_overall_events.py
```diff
@@ -0,0 +1,178 @@
...
+
+        self.event_type = "edx.course.enrollment.activated"
+        self.timestamp = "2014-12-05T22:13:09.691008+00:00"
+        self.user_id = 5
+
+    def _create_event_log_line(self, **kwargs):
+        """Create an event log with test values, as a JSON string."""
+        return json.dumps(self._create_event_dict(**kwargs))
+
+    def _create_event_dict(self, **kwargs):
```
brianhw, line 36: Is there already some base class that defines this?

benpatterson:
@mulby @brianhw @rocha I think this is in good order. Anything else?

Plan is to merge this, and (when ready), I'll submit a separate PR for producing a report.

edx/analytics/tasks/tests/test_overall_events.py
benpatterson, line 36: There's not that I saw. This is definitely an area for refactoring. The various test classes that use this implement it in slightly different ways (e.g., different fields are being populated).

mulby:
:+1:

brianhw:
It would be nicer to have the tests that use a modified version of an event use the same create_event starting point, and then modify the result, either by adding args or modifying the returned result.  It's easier to see what's different when the difference is explicitly spelled out, and it's easier to maintain.

But this is not a blocker. :+1:

benpatterson:
Yeah I agree with you on that @brianhw . I'd like to refactor that piece of it at some point, with some common/fixture/helper module.

brianhw:
I'm confused by what piece you mean.   I mean for example, test_implicit_event.

benpatterson:
Sorry, by "piece" I guess I meant the general "test event" or "_create_event" pattern. There are several different flavors, not just in the overall_events tests, but others as well. Would be good to make it DRYer.

edx/analytics/tasks/tests/test_overall_events.py
```diff
@@ -0,0 +1,165 @@
...
+                "course_id": "",
+                "path": "/jsi18n/"
+            },
+            "time": "2014-12-05T22:11:29.689805+00:00",
+            "ip": "10.0.2.2",
+            "event": "{\"POST\": {}, \"GET\": {}}",
+            "agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_5) AppleWebKit/537.36 (KHTML, like Gecko) "
+                     "Chrome/39.0.2171.71 Safari/537.36",
+            "page": "null"
+        }
```
brianhw, line 103: Seems simpler to just use _create_event_dict(), and then delete the 'time' element.

brianhw, line 103: Sorry, the previous comment should have been on the test_no_time_element() test below.  Not sure what this is really testing.

benpatterson, line 103: Ah, ok, I understand. That's a good idea... Hopefully I can refactor that relatively soon.

PR #52, opened 2014-12-23
Benp/total event report

benpatterson:


edx/analytics/tasks/reports/total_events_report.py
```diff
@@ -0,0 +1,107 @@
...
+                date, event_count = line.split('\t')
+                counts.append((date, event_count))
+                date = date.strip()
+                total += int(event_count)
+
+        # Write out the counts as a CSV, in reverse order of date.
+        with self.output().open('w') as output_file:
+            output_file.write(self.create_header())
+            output_file.write('\n')
+            for date, event_count in sorted(counts,
```
benpatterson: I noticed here that if I feed it a list that is unsorted, the last item does not have the \n character. As a result, whichever line that one is on, gets concatenated with the next line. This happens in unit tests, but I'm assuming it will be a problem in the Real World. It happens even when I feed it \n directly. See test_multilines_sorted. If I move the 2015 date to be the last entry, tests would fail.

Any suggestions?

benpatterson:
@brianhw @mulby @rocha 

This should be in decent shape, though I do have a bug I could use some advice on.

Beyond the bug, I'd like your feedback on the PR as well!

edx/analytics/tasks/reports/tests/test_total_events_report.py
```diff
@@ -0,0 +1,87 @@
...
+        """
+        out = self.run_task(source)
+        # Info for the 2015 date should appear first
+        self.assertEqual(out.loc[1]['count'], '7000')
+        self.assertEqual(out.loc[1]['date'], '2015-11-01')
+        self.assertEqual(out.loc[2]['count'], '300')
+        self.assertEqual(out.loc[2]['date'], '2014-10-03')
+
+    def tearDown(self):
+        self.cleanup(self.temp_rootdir)
```
brianhw: My understanding is that it is preferable to instead call self.addCleanup(self.cleanup, self.temp_rootdir) from the setUp() method.   In some of our other tests, we actually put the definition of cleanup() inside setUp(), since it is not used elsewhere.

edx/analytics/tasks/reports/total_events_report.py
```diff
@@ -0,0 +1,107 @@
...
+            date=date, event_count=event_count
+        )
+
+    def run(self):
+        """ Provide default values for when no counts are available. """
+        counts = []
+        date = "UNKNOWN"
+        total = 0
+        with self.input().open('r') as input_file:
+            for line in input_file.readlines():
```
brianhw: just call "for line in input_file:"

edx/analytics/tasks/reports/total_events_report.py
brianhw: I think that if you use "for line in input_file:" above, you won't have any trailing newlines.  Then you don't have to worry that some of your event_count elements have one and others don't.  It just means that you have to explicitly add the newline after writing the csv_entry, just as you did above for writing out the header.

edx/analytics/tasks/reports/total_events_report.py
```diff
@@ -0,0 +1,107 @@
...
+        counts: Location of counts per country. The format is a hadoop
+            tsv file, with fields country, count, and date.
+        report: Location of the resulting report. The output format is a
+            excel csv file with country and count.
+    """
+
+    name = luigi.Parameter()
+    interval = luigi.DateIntervalParameter()
+    src = luigi.Parameter(is_list=True)
+    pattern = luigi.Parameter(is_list=True, default='.*')
```
brianhw: You shouldn't need to add these here.  Instead, if you include EventLogSelectionDownstreamMixin, that should define source, interval, and pattern parameters.  (You'll notice that you don't actually use name anywhere.  It's a holdover from the old report stuff that isn't yet using EventLogSelectionMixin.)

edx/analytics/tasks/reports/total_events_report.py
```diff
@@ -0,0 +1,107 @@
...
+            the construction of output filenames, so each run will have distinct outputs.
+        src:  a URL to the root location of input tracking log files (e.g., s3://my_bucket/foo/)
+        interval: events within this date range are counted.
+
+    Additional optional parameters are passed through to :py:class:`MapReduceJobTask`:
+
+        mapreduce_engine:  'hadoop' (the default) or 'local'.
+        n_reduce_tasks: number of reducer tasks to use in upstream tasks.
+
+    Additional parameters are passed through to :py:class:`UsersPerCountryReport`:
```
brianhw: comments here still need to be updated....

edx/analytics/tasks/reports/total_events_report.py
benpatterson: done

edx/analytics/tasks/reports/total_events_report.py
benpatterson: Done. Thanks for helping me out with it!

edx/analytics/tasks/reports/total_events_report.py
benpatterson: Done

edx/analytics/tasks/reports/tests/test_total_events_report.py
benpatterson: Ok, after some digging, it looks like addCleanup will not work for us. There is an underlying configuration that emits a warning:

```
/Users/benpatterson/workspaces/edx-analytics-pipeline/venv/src/luigi/luigi/hdfs.py:539: UserWarning: Deprecated: Specify 'client: hadoopcli' in config
  warnings.warn("Deprecated: Specify 'client: hadoopcli' in config")
```

This happens during load, which means addCleanup is called immediately, rather than at the end of the tests. I've tried a few different configurations in test.cfg but this setting is not immediately apparent to me (and I'm not sure I'd want to override default hdfs client behavior, either).

So...took a good look but I think we'll be alright with tearDown instead, even if it's suboptimal.

brianhw: A few questions:
- Is there an easy way to remove the warnings, by specifying client:
  hadoopcli as it desires?   It's already in the client.cfg file, which I
  thought was merged with other config entries, so I don't know why the
  warning is even happening on regular runs.  But it should be
  straight-forward to add it to config/test.cfg, just to be sure it's there
  as well.
- And why does a warning like that trigger cleanup?
- And why do other tests that use addCleanup() seem to work fine?  (E.g.
  test_answer_dist, test_enrollment_validation, test_mapreduce,
  test_mysql_load, test_sqoop, test_user_location...)

benpatterson: Done.

edx/analytics/tasks/reports/total_events_report.py
benpatterson: removed `name`.

benpatterson: Removed these parameters and am instantiating EventLogSelectionDownstreamMixin object instead. (Thanks for the suggestion.)

benpatterson:
@brianhw @mulby @rocha Ready for another pass. Thanks for the comments.

edx/analytics/tasks/reports/total_events_report.py
```diff
@@ -0,0 +1,103 @@
...
+                counts.append((date.strip(), event_count.strip()))
+                total += int(event_count)
+
+        # Write out the counts as a CSV, in reverse order by date.
+        with self.output().open('w') as output_file:
+            output_file.write(self.create_header())
+            output_file.write('\n')
+            for date, event_count in sorted(counts,
+                                            reverse=True,
+                                            key=lambda k: datetime.datetime.strptime(k[0], '%Y-%m-%d')):
```
brianhw: Just out of curiosity, do you need to do the conversion here to a datetime?  Just sorting without specifying a key should work properly, as it should sort by the date strings alphanumerically.  Since you know the format is %Y-%m-%d, the alphanumeric sorting will be the same as the datetime sorting.

benpatterson: Sure. IIRC, I went down this path because I needed to cast the field in one way or another (be it str or datetime), so I went with datetime. I don't have a strong preference though, if there is a standard I ought to be following.

edx/analytics/tasks/reports/total_events_report.py
```diff
@@ -0,0 +1,103 @@
...
+
+        source:  a URL to the root location of input tracking log files (e.g., s3://my_bucket/foo/).
+        interval: events within this date range are counted.
+        pattern: regex pattern for files to be included when counting total events.
+        (See :py:class `EventLogSelectionTask` for more details.)
+
+    Additional optional parameters are passed through to :py:class:`MapReduceJobTask`:
+
+        mapreduce_engine:  'hadoop' (the default) or 'local'.
+        n_reduce_tasks: number of reducer tasks to use in upstream tasks.
```
brianhw: I would just leave off the description of these here, and just leave the pointer to the class.  The mapreduce_engine values are out of date (leaving out the emulation).  (And yes, they are out of date also in the place from where you cloned this class. : )

edx/analytics/tasks/reports/total_events_report.py
```diff
@@ -0,0 +1,103 @@
...
+
+    @classmethod
+    def create_csv_entry(cls, date, event_count):
+        """Generate a single entry in CSV format."""
+        return '{date},{event_count}'.format(
+            date=date, event_count=event_count
+        )
+
+    def run(self):
+        """ Provide default values for when no counts are available. """
```
brianhw: This comment doesn't seem to be a good summary for what this method does.  And if there are no counts available, I assume there's no need for default values anyway.

edx/analytics/tasks/reports/total_events_report.py
brianhw: Isn't it already a string?  I assume you can just leave off the key altogether, and it should do the right thing.

brianhw:
The change to the sort key is optional, and otherwise just comments on comments.  Update and squash.  LGTM. :+1:

edx/analytics/tasks/reports/total_events_report.py
benpatterson: Done.

edx/analytics/tasks/reports/total_events_report.py
benpatterson: Removed the datetime business. How's it look now?

edx/analytics/tasks/reports/total_events_report.py
```diff
@@ -0,0 +1,103 @@
...
+                counts.append((date.strip(), event_count.strip()))
+                total += int(event_count)
+
+        # Write out the counts as a CSV, in reverse order by date.
+        with self.output().open('w') as output_file:
+            output_file.write(self.create_header())
+            output_file.write('\n')
+            for date, event_count in sorted(counts,
+                                            reverse=True,
+                                            key=lambda k: k[0]):
```
brianhw: Sorry to belabor this, but my suggestion was just to remove the key argument altogether.

edx/analytics/tasks/reports/total_events_report.py
benpatterson: Makes sense. I changed the comment, and I also added a test for the case of an empty source. (Got rid of "date" = 'UNKNOWN')

benpatterson:
Thanks for the feedback. Updated based on comments. Added a test to cover a use case that came up.

mulby:
:+1:

brianhw:
LGTM.  Just squash.  :+1:

benpatterson:
Squashed. Merging.

PR #54, opened 2014-12-29
Ensure export reports progress

mulby:
If a hadoop task does not either write output or reports status for 10 minutes it is killed. Make sure we check back in periodically with hadoop to ensure it doesn't kill our export processes.

@brianhw @rocha

### DIFF
--- a/edx/analytics/tasks/encrypt.py
+++ b/edx/analytics/tasks/encrypt.py
@@ -9,11 +9,33 @@ import gnupg
 
 from edx.analytics.tasks.util.tempdir import make_temp_directory
 
+
+TRANSFER_BUFFER_SIZE = 1024 * 1024  # 1 MB
 log = logging.getLogger(__name__)
+key_cache = {}  # pylint: disable=invalid-name
+
+
+def get_key_from_target(key_file_target):
+    """Get the contents of the key file pointed to by the target"""
+
+    # Use a key cache since this is often called from reduce tasks which are executed within the same python process if
+    # JVM reuse is enabled.
+    key_file_path = key_file_target.path
+
+    key_content = key_cache.get(key_file_path)
+    if not key_content:
+        with key_file_target.open('r') as gpg_key_file:
+            log.info("Reading keyfile from %s", key_file_path)
+            key_content = gpg_key_file.read()
+        key_cache[key_file_path] = key_content
+    else:
+        log.info("Using cached keyfile for %s", key_file_path)
+
+    return key_content
 
 
 @contextmanager
-def make_encrypted_file(output_file, key_file_targets, recipients=None):
+def make_encrypted_file(output_file, key_file_targets, recipients=None, progress=None):
     """
     Creates a file object to be written to, whose contents will afterwards be encrypted.
 
@@ -21,6 +43,7 @@ def make_encrypted_file(output_file, key_file_targets, recipients=None):
         output_file:  a file object, opened for writing.
         key_file_targets: a list of luigi.Target objects defining the gpg public key files to be loaded.
         recipients:  an optional list of recipients to be loaded.  If not specified, uses all loaded keys.
+        progress:  a function that is called periodically as progress is made.
     """
     with make_temp_directory(prefix="encrypt") as temp_dir:
         # Use temp directory to hold gpg keys.
@@ -31,6 +54,7 @@ def make_encrypted_file(output_file, key_file_targets, recipients=None):
         # Create a temp file to contain the unencrypted output, in the same temp directory.
         with tempfile.NamedTemporaryFile(dir=temp_dir, delete=False) as temp_input_file:
             temp_input_filepath = temp_input_file.name
+            log.info('Writing data to temporary file: %s', temp_input_filepath)
             yield temp_input_file
 
         # Encryption produces a second file in the same temp directory.
@@ -39,7 +63,7 @@ def make_encrypted_file(output_file, key_file_targets, recipients=None):
             recipients = [key['keyid'] for key in gpg.list_keys()]
         with open(temp_input_filepath, 'r') as temp_input_file:
             _encrypt_file(gpg, temp_input_file, temp_encrypted_filepath, recipients)
-        _copy_file_to_open_file(temp_encrypted_filepath, output_file)
+        _copy_file_to_open_file(temp_encrypted_filepath, output_file, progress)
 
 
 def _import_key_files(gpg_instance, key_file_targets):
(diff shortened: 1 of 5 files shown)


mulby:
Added a commit here to fix a performance issue. Looking at the debug log statements, I noticed the following:
1. The reducer task JVM was being reused.
2. When the reducer task JVM is reused it also reuses the python process that is actually executing the reducer.
3. This results in the python process using the same memory for different reduce tasks.
4. The recipient list stored in `self.recipients_for_org_id[org_id]` was being appended to on each successive reduce task.
5. After the 5th reduce task this results in a recipient list that looks something like `[real.recipient@foo.com, master@key.com, master@key.com, master@key.com, master@key.com, master@key.com]`
6. We then would _read_ each of those key files. In this case, re-reading the same key, many times.
7. A read is very expensive since it involves a subprocess call to the hadoop CLI which in turn spins up a new JVM and then does a bunch of S3 API calls and network I/O.

To mitigate the issue I have now converted the recipient list to set and ensured it is only appended to once, in `init_local`. Additionally I added in a key cache so that we can take advantage of the fact that several reduce tasks all execute in a single process.

As an aside, I was able to determine this was happening by adding the additional log messages, logging into a slave machine and running strace on the python reduce process.

mulby:
@rocha @brianhw this explains why we were having trouble scaling up to large time intervals even with small data sizes

edx/analytics/tasks/encrypt.py
```diff
@@ -10,10 +10,30 @@
...
+    """Get the contents of the key file pointed to by the target"""
+
+    # Use a key cache since this is often called from reduce tasks which are executed within the same python process if
+    # JVM reuse is enabled.
+    key_file_path = key_file_target.path
+
+    key_content = key_cache.get(key_file_path)
+    if not key_content:
+        with key_file_target.open('r') as gpg_key_file:
+            log.warn("Reading keyfile from %s", key_file_path)
```
brianhw: Do we have to use warn() for all info()-type log statements?  Seems like the wrong level, and would be better if we could fix output so that info stuff appears if that's what we want to see here.

edx/analytics/tasks/encrypt.py
```diff
@@ -63,14 +85,22 @@ def _encrypt_file(gpg_instance, input_file, encrypted_filepath, recipients):
...
 
 
-def _copy_file_to_open_file(filepath, output_file):
+def _copy_file_to_open_file(filepath, output_file, progress=None):
     """Copies a filepath to a file object already opened for writing."""
+    log.warn('Copying to output: %s', filepath)
     with open(filepath, 'r') as src_file:
         while True:
-            transfer_buffer = src_file.read(1024)
+            transfer_buffer = src_file.read(1024 * 1024)
```
brianhw: Should we pull this out as a constant?

edx/analytics/tasks/encrypt.py
mulby: sadly, yes...

https://github.com/spotify/luigi/blob/master/luigi/mrrunner.py#L75

Luigi hard codes the log level on slave nodes, not sure why.

mulby: At a minimum we could pull this out into a util function so that we hide the ugliness?

edx/analytics/tasks/encrypt.py
mulby: Probably...

edx/analytics/tasks/encrypt.py
mulby: On second thought, we could override the luigi method that sets up the execution... and reconfigure the logger there?

edx/analytics/tasks/event_exports.py
```diff
@@ -88,8 +88,9 @@ def init_local(self):
         for org_id, org_config in organizations.iteritems():
             aliases = [org_id] + org_config.get('other_names', [])
 
-            recipients = org_config['recipients']
-            self.recipients_for_org_id[org_id] = recipients
+            self.recipients_for_org_id[org_id] = set(org_config['recipients'])
```
brianhw: I see -- so the problem here was that there wasn't a copy here, so it was actually appending to org_config['recipients'].  Presumably there's no difference between using set and list here.  If we really wanted to be careful, we could also check if gpg_master_key were already in the list of recipients below before adding it.  We could also use get('recipients', []) in place of org_config['recipients'], so that it handles having no recipients.  (That way it's only exported to the master key.)

edx/analytics/tasks/event_exports.py
```diff
@@ -183,25 +184,25 @@ def multi_output_reducer(self, key, values, output_file):
         Write to the encrypted file by streaming through gzip, which compresses before encrypting
         """
         _date_string, org_id = key
-        recipients = self._get_recipients(org_id)
+        recipients = self.recipients_for_org_id[org_id]
+        log.warn('Encryption recipients: %s', str(recipients))
+
+        def report_progress(num_bytes):
+            self.incr_counter('Multi Output', 'Bytes Written to Output', num_bytes)
```
brianhw: 'Event export' instead of 'Multi Output'?  This is in event_exports.py, not in mapreduce.py.

edx/analytics/tasks/encrypt.py
brianhw: Not all that clean to override.  The main is hardcoded into mrrunner.py, and mrrunner.py would have to be replaced in HadoopJobRunner.run_job.  Might actually be easier to patch luigi directly than to work around....

mulby:
@brianhw I updated the code to override the log level to INFO on the slave nodes

brianhw:
Very nice.  LGTM. :+1:

edx/analytics/tasks/event_exports.py
```diff
@@ -183,25 +184,25 @@ def multi_output_reducer(self, key, values, output_file):
...
-        with make_encrypted_file(output_file, key_file_targets) as encrypted_output_file:
+        with make_encrypted_file(output_file, key_file_targets, progress=report_progress) as encrypted_output_file:
             outfile = gzip.GzipFile(mode='wb', fileobj=encrypted_output_file)
             try:
                 for value in values:
                     outfile.write(value.strip())
                     outfile.write('\n')
+                    # WARNING: This line ensures that Hadoop knows that our process is not sitting in an infinite loop.
+                    # Do not remove it.
+                    self.incr_counter('Multi Output', 'Raw Bytes Written', len(value) + 1)
```
brianhw: This should match the other incr_counter() call.

brianhw:
One more nit, then squash. :+1:

edx/analytics/tasks/event_exports.py
```diff
@@ -183,25 +184,26 @@ def multi_output_reducer(self, key, values, output_file):
...
-        with make_encrypted_file(output_file, key_file_targets) as encrypted_output_file:
+        with make_encrypted_file(output_file, key_file_targets, progress=report_progress) as encrypted_output_file:
             outfile = gzip.GzipFile(mode='wb', fileobj=encrypted_output_file)
             try:
                 for value in values:
                     outfile.write(value.strip())
                     outfile.write('\n')
+                    # WARNING: This line ensures that Hadoop knows that our process is not sitting in an infinite loop.
+                    # Do not remove it.
+                    self.incr_counter('Event Export', 'Raw Bytes Written', len(value) + 1)
```
rocha, line 203: Apparently Hadoop's `reporter.setStatus()` also has the same effect. FYI.

PR #55, opened 2015-01-05
WIP: Support docker

mulby:
Support running the pipeline inside a docker container. Note this allows you to point "remote-task" at docker containers.

Other changes here:
- Add support for a MariaDB result store instead of MySQL
- Add support for database credentials stored directly in config files (no need for external *.json files in S3)

Reviewers: @brianhw @rocha @e0d 
FYI: @feanil @pchiu33


edx/analytics/tasks/launchers/remote.py
```diff
@@ -17,6 +17,7 @@ def main():
     parser = argparse.ArgumentParser()
     parser.add_argument('--job-flow-id', help='EMR job flow to run the task', default=None)
     parser.add_argument('--job-flow-name', help='EMR job flow to run the task', default=None)
+    parser.add_argument('--docker-container-name', help='Docker container to run the task', default=None)
```
mulby: After thinking about this more it may make sense to expose the ansible inventory parameters more directly. That would allow us to run against arbitrary hosts that are SSH accessible. It would also remove all of the docker related complexity from this PR.

Makefile
```diff
@@ -21,7 +21,12 @@ develop: requirements
 system-requirements:
 	sudo apt-get update -q
 	# This is not great, we can't use these libraries on slave nodes using this method.
-	sudo apt-get install -y -q libmysqlclient-dev libatlas3gf-base
+	if (sudo dpkg --get-selections | grep -v deinstall | grep mariadb); \
```
brianhw, line 24: I don't understand this -- one cannot have both mariadb and mysql around?

client.cfg
```diff
@@ -1,5 +1,4 @@
 [hadoop]
-version = apache1
```
brianhw, line 2: If this is taken out here, does it need to go into other configs explicitly?  Like test.cfg?

PR #56, opened 2015-01-13
Pin pip. AN-4322.

benpatterson:
@mulby @rocha @brianhw 

I suggest we implement this and pull it out once the ubuntu/jenkins-plugin virtualenv installs catch up (which could admittedly take awhile.)


mulby:
:+1:

brianhw:
LGTM: :+1:

PR #57, opened 2015-01-22
Support log analysis for performance benchmark

mulby:
This patch includes the following changes:
- Write all luigi log messages to the edx_analytics.log file. This allows us to analyze a single log file.
- Set the luigi log threshold to DEBUG instead of INFO. This makes it much easier to analyze the log.
- Support profiling launch-task using a custom version of pyinstrument (https://github.com/mulby/pyinstrument/tree/gabe/json-output)
- Write a script that can generate reports from the edx_analytics.log file as well as pyinstrument execution traces.

It has no tests (this was a deliberate tactical decision), does some hacky stuff (also deliberate), and is generally thrown together in the interest of getting some useful output.

Since this can now generate pretty (yet self-contained) HTML reports given output from a job, it might be nice to modify the scheduler jobs to do so and save the reports so we can review them.

Sample reports:

Gist: https://gist.github.com/mulby/05ec2fc4d401a852d3bd#file-enrollment-log-html
ImportEnrollmentsIntoMysql report from edx_analytics.log: https://rawgit.com/mulby/05ec2fc4d401a852d3bd/raw/9f3c69b7fedebe3618dedec12de04239d6bb3e84/enrollment-log.html
StartupTimeTestTask report from edx_analytics.log: https://rawgit.com/mulby/05ec2fc4d401a852d3bd/raw/4643c5eadc0c4cc738e46ce7a9b54834c2bfc15e/start-time-log.html
StartupTimeTestTask report from pyinstrument execution trace:  https://rawgit.com/mulby/05ec2fc4d401a852d3bd/raw/fa4f37fa6df37799f217ffa2583f31166e74db67/start-time-trace.html

Also note that the analysis script may want to live somewhere else, it's not inherently tied to edx-analytics-pipeline in any way other than the fact that it makes assumptions about the log file format and the version of pyinstrument that was used to capturet he performance info.

Reviewers: @rocha @benpatterson @brianhw 
FYI: @cpennington

### DIFF
--- a/edx/analytics/tasks/launchers/analyze/main.py
+++ b/edx/analytics/tasks/launchers/analyze/main.py
@@ -1,0 +1,258 @@
+"""Analyze log files produced by launch-task"""
+
+import argparse
+from collections import namedtuple
+import datetime
+import re
+import sys
+
+from edx.analytics.tasks.launchers.analyze.parser import LogFileParser
+from edx.analytics.tasks.launchers.analyze.measure import Measurement
+from edx.analytics.tasks.launchers.analyze.report import text_report, json_report, html_report
+
+
+MESSAGE_START_PATTERN = r'(?P<timestamp>\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2},\d{3}) (?P<level>\w+) (?P<pid>\d+) \[(?P<module>.*?)\] (?P<filename>.*?):(?P<line_no>\d+) - (?P<content>.*)'
+LogMessage = namedtuple('LogMessage', 'timestamp level pid module filename line_no content')  # pylint: disable=invalid-name
+
+
+def analyze():
+    arg_parser = argparse.ArgumentParser(description='Analyze log files.')
+    arg_parser.add_argument(
+        '-o', '--output',
+        help='Save the results of the analysis to a file that can be read later.'
+    )
+    arg_parser.add_argument(
+        '-r', '--report',
+        choices=['text', 'json', 'html'],
+        help='Generate a report in the requested format and stream it to stdout.'
+    )
+    arg_parser.add_argument(
+        '--threshold-percent',
+        type=float,
+        default=None,
+        help="Don't include entries that accounted for less than this amount of time relative to the total execution"
+             " time in the report."
+    )
+    group = arg_parser.add_argument_group('Input File')
+    group_ex = group.add_mutually_exclusive_group()
+    group_ex.add_argument(
+        '-l', '--log',
+        default='edx_analytics.log',
+        help='A log file to analyze. Defaults to "edx_analytics.log" in the current directory.'
+    )
+    group_ex.add_argument(
+        '-i', '--input',
+        help='Reads a previously saved result file.'
+    )
+    group_ex.add_argument(
+        '-t', '--trace',
+        help='Path to an execution trace of the launch-task process captured by pyinstrument and saved as JSON.'
+    )
+
+    args = arg_parser.parse_args()
+
+    if args.input:
+        root = Measurement.from_json(args.input)
+    elif args.trace:
+        root = Measurement.from_pyinstrument_trace(args.trace)
+    else:
+        root = analyze_log_file(args.log)
+
+    if args.output:
+        json_report(root, args.output, pretty=False, threshold_percent=args.threshold_percent)
+
+    if args.report == 'json':
+        json_report(root, threshold_percent=args.threshold_percent)
+    elif args.report == 'text':
+        text_report(root, threshold_percent=args.threshold_percent)
+    elif args.report == 'html':
+        html_report(root, threshold_percent=args.threshold_percent)
+
+
+def analyze_log_file(filename):
+    with open(filename, 'rb') as log_file:
+        parser = LogFileParser(log_file, message_pattern=MESSAGE_START_PATTERN, message_factory=create_log_message)
+        try:
+            return analyze_log(parser)
+        except:
+            sys.stderr.write('Exception on line {0}\n'.format(parser.line_number))
+            raise
+
+
+def create_log_message(matched_groups):
+    timestamp_str = matched_groups['timestamp']
+    matched_groups['timestamp'] = datetime.datetime.strptime(timestamp_str, '%Y-%m-%d %H:%M:%S,%f')
+    for field_name in ('pid', 'line_no'):
+        matched_groups[field_name] = int(matched_groups[field_name])
+
+    return LogMessage(**matched_groups)
+
+
+def analyze_log(parser):
+    root = Measurement('Luigi Worker')
+    scheduling_time = analyze_overall_scheduling(parser)
+    root.add_child(scheduling_time)
+    execution_time = analyze_overall_execution(parser)
+    root.add_child(execution_time)
+    return root
+
+
+def analyze_overall_scheduling(parser):
+    all_scheduling = Measurement('Scheduling Tasks')
+
+    start_scheduling_pattern = r'Checking if (?P<task_id>.*?) is complete'
+    message = True
+    start_scheduling_timestamp = None
+
+    while message:
+        message = parser.next_message()
+
+        if message.content == 'Done scheduling tasks':
+            all_scheduling.set_time_from_range(message.timestamp, start_scheduling_timestamp)
+            return all_scheduling
+
+        start_match = re.match(start_scheduling_pattern, message.content, (re.MULTILINE | re.DOTALL))
+        if start_match:
+            if start_scheduling_timestamp is None:
+                start_scheduling_timestamp = message.timestamp
+
+            measurement = analyze_task_scheduling(message, start_match, parser)
+            if measurement:
+                all_scheduling.add_child(measurement)
+
+
+def analyze_task_scheduling(start_message, start_match, parser):
+    scheduled_pattern = r'Scheduled (?P<task_id>.*?) \((?P<status>\w+)\)'
+
+    task = LuigiTaskDescription.from_string(start_match.group('task_id'))
+    start_timestamp = start_message.timestamp
+    message = start_message
+    while message:
+        message = parser.peek_message()
+
+        if message.content == 'Done scheduling tasks':
+            break
+
+        if re.match(scheduled_pattern, message.content, (re.MULTILINE | re.DOTALL)):
+            break
+
+        parser.next_message()
+
+    if not task.name == 'UncheckedExternalURL':
+        return Measurement('Scheduling {}'.format(task), message.timestamp - start_timestamp)
+    else:
+        return None
+
+
+def analyze_overall_execution(parser):
+    all_execution = Measurement('Executing Tasks')
+
+    pattern = r'.*? Worker Worker.* (?P<state>running|done|failed)\s+(?P<task_id>.*)'
+    message = True
+    overall_start_timestamp = None
+    running_measurement = None
+
+    while message:
+        message = parser.next_message()
+
+        if message.content == 'Done':
+            if overall_start_timestamp:
+                all_execution.set_time_from_range(message.timestamp, overall_start_timestamp)
+            return all_execution
+
+        match = re.match(pattern, message.content, (re.MULTILINE | re.DOTALL))
+        if not match:
+            if 'Running job:' in message.content or 'Starting Job =' in message.content:
+                for measurement in analyze_hadoop_job(message, parser):
+                    running_measurement.add_child(measurement)
+
+            continue
+
+        task = LuigiTaskDescription.from_string(match.group('task_id'))
+        state = match.group('state')
+        if state == 'running':
+            start_timestamp = message.timestamp
+            running_measurement = Measurement('Executing {}'.format(task))
+            if not overall_start_timestamp:
+                overall_start_timestamp = start_timestamp
+        else:
+            running_measurement.set_time_from_range(message.timestamp, start_timestamp)
+            all_execution.add_child(running_measurement)
+
+
+def analyze_hadoop_job(starting_message, parser):
+    match = re.match(r'.*?(?P<job_id>job_\d{12}_\d{4})', starting_message.content)
+    job_id = match.group('job_id')
+    start_timestamp = starting_message.timestamp
+
+    message = starting_message
+    while message:
+        message = parser.next_message()
+
+        if 'Job complete:' in message.content or 'Ended Job = ' in message.content:
+            if 'Job complete:' in message.content:
+                move_measure = analyze_output_move(parser)
+                if move_measure:
+                    yield move_measure
+
+            yield Measurement('Hadoop ' + job_id, message.timestamp - start_timestamp)
+            return
+
+
+def analyze_output_move(parser):
+    output_message = parser.peek_message()
+    if re.match(r'.*?Output:.*-temp-', output_message.content):
+        start_timestamp = output_message.timestamp
+        parser.next_message()
+        next_message = parser.peek_message()
+        return Measurement('Moving Job Output', next_message.timestamp - start_timestamp)
+
+
+class LuigiTaskDescription(object):
+
+    PATTERN = r'(?P<name>\w+)\((?P<params>.*)\)'
+
+    def __init__(self, name, params=None):
+        self.name = name
+        self.params = params or {}
+
+    def __str__(self):
+        param_string = ', '.join(['='.join((k, str(v)[:100])) for k, v in self.params.iteritems()])
+        return '{name}{params}'.format(
+            name=self.name,
+            params='(' + param_string + ')' if len(self.params) > 0 else ''
+        )
+
+    @staticmethod
+    def from_string(id_str):
+        match = re.match(LuigiTaskDescription.PATTERN, id_str, (re.MULTILINE | re.DOTALL))
+        if match:
+            task_name = match.group('name')
+            raw_params = match.group('params')
+            param_parser = default_parameter_parser
+            if task_name == 'HiveTableFromQueryTask':
+                param_parser = hive_parameter_parser
+            if task_name == 'SqoopImportFromMysql':
+                param_parser = sqoop_parameter_parser
+            return LuigiTaskDescription(task_name, param_parser(raw_params))
+        else:
+            raise ValueError('Unable to parse task id "%s"', id_str)
+
+
+def default_parameter_parser(_raw_params):
+    return {}
+
+
+def hive_parameter_parser(raw_params):
+    table_param_match = re.search(r'table=(?P<name>[\w_]+)', raw_params)
+    if table_param_match:
+        return {'table': table_param_match.group('name')}
+
+
+def sqoop_parameter_parser(raw_params):
+    table_param_match = re.search(r'table_name=(?P<name>[\w_]+)', raw_params)
+    if table_param_match:
+        return {'table': table_param_match.group('name')}
+
+if __name__ == '__main__':
+    analyze()
--- a/edx/analytics/tasks/launchers/analyze/measure.py
+++ b/edx/analytics/tasks/launchers/analyze/measure.py
@@ -1,0 +1,159 @@
+
+import datetime
+import json
+from operator import methodcaller
+
+
+class Measurement(object):
+    """
+    Represents a node in a tree of measurements.
+
+    Note that a particular measurement may be the parent for several child measurements. Those child measurements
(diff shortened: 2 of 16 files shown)


edx/analytics/tasks/launchers/analyze/main.py
```diff
@@ -0,0 +1,222 @@
...
+def analyze_log_file():
+    arg_parser = argparse.ArgumentParser(description='Analyze log files.')
+    arg_parser.add_argument('-o', '--output', help='Save the results of the analysis to a file that can be read later.')
+    arg_parser.add_argument('-r', '--report', choices=['text', 'json', 'html'], help='Generate a report in the requested format and stream it to stdout.')
+    arg_parser.add_argument('--threshold-percent', type=float, default=None, help="Don't include entries that accounted for less than this amount of time relative to the total execution time in the report.")
+    group = arg_parser.add_argument_group('Input File')
+    group_ex = group.add_mutually_exclusive_group()
+    group_ex.add_argument('-l', '--log', default='edx_analytics.log', help='A log file to analyze. Defaults to "edx_analytics.log" in the current directory.')
+    group_ex.add_argument('-i', '--input', help='Reads a previously saved result file.')
+    group_ex.add_argument('-t', '--trace', help='Path to an execution trace of the launch-task process captured by pyinstrument and saved as JSON.')
```
rocha: Very minor: lines that a longer than 120 characters are a bit hard to read on github, since there is no horizontal scrollbar.

edx/analytics/tasks/launchers/analyze/main.py
```diff
@@ -0,0 +1,222 @@
...
+    elif args.trace:
+        root = Measurement.from_pyinstrument_trace(args.trace)
+    else:
+        with open(args.log, 'rb') as log_file:
+            parser = LogFileParser(log_file, message_pattern=MESSAGE_START_PATTERN, message_factory=create_log_message)
+            try:
+                root = analyze_log(parser)
+            except:
+                sys.stderr.write('Exception on line {0}\n'.format(parser.line_number))
+                raise
```
rocha: Can you move the block above to a function? Something like `analyze_log_file`.

requirements/default.txt
```diff
@@ -27,5 +27,6 @@ six==1.6.1		# MIT
 stevedore==0.14.1 	# Apache 2.0
 tornado==3.1.1 		# Apache 2.0
 wsgiref==0.1.2 		# ZPL
--e git+https://github.com/edx/luigi.git@v1.0.17-fork.edx.1#egg=luigi 		# Apache License 2.0
--e git+https://github.com/edx/opaque-keys.git@33f2b099e92957046a866a9b6d48a71c484bb475#egg=opaque-keys
+git+https://github.com/edx/luigi.git@v1.0.17-fork.edx.1#egg=luigi 		# Apache License 2.0
+git+https://github.com/edx/opaque-keys.git@33f2b099e92957046a866a9b6d48a71c484bb475#egg=opaque-keys
+git+https://github.com/mulby/pyinstrument.git@a35ff76df4c3d5ff9a2876d859303e33d895e78f#egg=pyinstrument
```
benpatterson: License info? I know this is a fork but the requirements file is really put together well. Would be nice to keep the license goodness going.

edx/analytics/tasks/launchers/analyze/main.py
```diff
@@ -0,0 +1,222 @@
...
+def analyze_log(parser):
+    root = Measurement('Luigi Worker')
+    scheduling_time = analyze_scheduling(parser, threshold=SCHEDULING_TASK_THRESHOLD)
+    root.add_child(scheduling_time)
+    execution_time = analyze_execution(parser)
+    root.add_child(execution_time)
+    return root
+
+
+def analyze_scheduling(parser, threshold=1):
```
benpatterson: No big deal but if you have some time to add a docstring on this method and the next one to better distinguish them, it'd be great

benpatterson: Or maybe this just needs to be renamed to analyze_job_scheduling

edx/analytics/tasks/launchers/analyze/main.py
```diff
@@ -0,0 +1,222 @@
...
+
+        if message.content == 'Done scheduling tasks' or re.match(scheduled_pattern, message.content, (re.MULTILINE | re.DOTALL)):
+            break
+
+        parser.next_message()
+
+    return Measurement('Scheduling {}'.format(task), message.timestamp - start_timestamp)
+
+
+def analyze_execution(parser):
```
benpatterson: "analyze_task_execution" ?

edx/analytics/tasks/launchers/analyze/measure.py
```diff
@@ -0,0 +1,139 @@
+
```
benpatterson, line 1: A quick doc blurb up here?

edx/analytics/tasks/launchers/analyze/measure.py
```diff
@@ -0,0 +1,139 @@
...
+        self.parent = None
+
+    def add_child(self, child):
+        child.parent = self
+        self.children.append(child)
+
+    def time_including_children(self):
+        return self.self_time + self.time_from_children()
+
+    def time_from_children(self):
```
benpatterson: If you have multiple children that are running in parallel, will this figure out the times in parallel? Or will it add it all together (sum, rather than total length of all children)?

edx/analytics/tasks/launchers/analyze/measure.py
```diff
@@ -0,0 +1,139 @@
...
+        root = Measurement(
+            description=serialized['description'],
+            self_time=datetime.timedelta(seconds=serialized['self_time'])
+        )
+        for child_ser in serialized.get('children', []):
+            root.add_child(Measurement.from_serialized(child_ser))
+
+        return root
+
+    def to_json(self, file_ref, pretty=False, threshold_percent=None):
```
benpatterson, line 104: I'm a bit less familiar with how resilient json.dump is, but do you need try/catches in here in case output gets messy & the output can't be JSON'd?

edx/analytics/tasks/launchers/analyze/report.py
```diff
@@ -0,0 +1,59 @@
...
+            time=total_time,
+            percentage=measurement.percentage_of_total(),
+            desc=measurement.description
+        )
+    )
+    visible_children = measurement.sorted_filtered_children(threshold_percent=threshold_percent)
+    last_child_index = len(visible_children) - 1
+    for index, child in enumerate(visible_children):
+        if index < last_child_index:
+            c_indent = child_indent + (u'├─ ' if is_a_tty else '|- ')
```
benpatterson, line 29: :+1: on the tty logic. Nice.

edx/analytics/tasks/launchers/local.py
```diff
@@ -7,7 +7,8 @@
 
 """
 
-import os.path
+from contextlib import contextmanager
+import os
```
benpatterson: nit: put os after logging for alphabetical reasons

benpatterson:
Mainly nitty stuff from me. The heavier question I have is whether or not memory performance would be a concern with the LogFileParser class for particularly long files. 
:+1: overall from my desk

edx/analytics/tasks/launchers/analyze/main.py
```diff
@@ -0,0 +1,222 @@
...
+        else:
+            raise ValueError('Unable to parse task id "%s"', id_str)
+
+
+def default_parameter_parser(raw_params):
+    return {}
+
+
+def hive_parameter_parser(raw_params):
+    m = re.search(r'table=(?P<name>[\w_]+)', raw_params)
```
brianhw: I have really gotten used to having more verbose variable names -- e.g. s/m/matcher/

requirements/default.txt
brianhw: If we're eventually merging this stuff to master, we might want to maintain an edx fork instead of mulby.

cpennington: Yeah, we should get an edx fork, and put both your stuff and the flame-graph stuff in it.

logging.cfg
```diff
@@ -28,8 +28,8 @@ propagate=0
 
 [logger_luigi_interface]
 # Errors from luigi-interface get routed to stdout.
-level=INFO
-handlers=luigiHandler
+level=DEBUG
+handlers=luigiHandler,localHandler
```
brianhw: We talked some about moving this version of logging.cfg into a separate file, because we don't want to always run with Luigi debugging on for non-profiling runs.

edx/analytics/tasks/performance.py
```diff
@@ -0,0 +1,26 @@
+
+import luigi
+
+from edx.analytics.tasks.mapreduce import MapReduceJobTask
+from edx.analytics.tasks.pathutil import EventLogSelectionMixin
+from edx.analytics.tasks.url import get_target_from_url
+
+
+class StartupTimeTestTask(EventLogSelectionMixin, MapReduceJobTask):
+
```
brianhw, line 16: It would be helpful to have a docstring here to indicate what the purpose of this task is.  Also to change the name to reflect that the mapper is doing some non-trivial work.  E.g.  Something like ParseEventLogPerformanceTask.

edx/analytics/tasks/launchers/local.py
```diff
@@ -55,7 +57,23 @@ def main():
...
 
     # Launch Luigi using the default builder
-    luigi.run()
+
+    with profile_if_necessary(os.getenv('WORKFLOW_PROFILER', '')):
+        luigi.run()
+
+
+@contextmanager
+def profile_if_necessary(profiler_name):
```
brianhw: Or just call this "workflow_profiling".  I don't know if the _if_necessary is necessary....

edx/analytics/tasks/launchers/analyze/main.py
mulby: Done.

edx/analytics/tasks/launchers/analyze/main.py
mulby: Done.

edx/analytics/tasks/launchers/analyze/main.py
mulby: Done.

edx/analytics/tasks/launchers/analyze/main.py
mulby: Kinda. I renamed it to match the top level scheduling function.

edx/analytics/tasks/launchers/analyze/main.py
mulby: Done.

edx/analytics/tasks/launchers/analyze/measure.py
mulby, line 1: Done.

edx/analytics/tasks/launchers/analyze/measure.py
mulby: Parallel children aren't supported (yet). That's a great idea, though... ideally it would represent this, and allow you to display various statistics (sum, avg, stddev) for the children that executed in parallel... see outliers (stragglers) etc.

edx/analytics/tasks/launchers/analyze/measure.py
mulby, line 104: That level of reliability is beyond what this code is designed to handle. It is expected to have very predictable input and output. Definitely not a general purpose tool.

edx/analytics/tasks/launchers/analyze/report.py
mulby, line 29: I'd love to take credit for this, but it comes from pyinstrument, indeed very clever.

edx/analytics/tasks/launchers/local.py
mulby: Done.

edx/analytics/tasks/launchers/local.py
mulby: I like the "if_necessary" since it makes it obvious that it won't always run the profiler...

edx/analytics/tasks/performance.py
mulby, line 16: Done.

logging.cfg
mulby: I'm actually not sure we don't always want to run this... whats a few more kilobytes of log messages?

requirements/default.txt
mulby: I put a support ticket for an edx fork

mulby: Also added license.

edx/analytics/tasks/launchers/analyze/measure.py
benpatterson, line 104: Got it. Sounds fine.

mulby:
@benpatterson can you clarify your concern about the memory performance of the log parser? I intended it to only store a single (multi line) log message in memory at a time.

benpatterson:
@mulby I was thinking of that when looking at the while loops in that class. Wasn't clear to me (just based on reading) how large the memory ask could get. Would it be a concern?

logging.cfg
brianhw: I'm happy to defer to you on this, as I don't have firsthand experience with how much extra logging gets added.  My main concern is that the logs be more tractable when viewing console output on Jenkins, for doing a quick determination of why jobs fail (when they do).  As it stands, the logging of path information when calculating the manifest is so verbose that it's difficult to inspect a full console output in those cases when the tail of the log is not enough.   The inconvenience of this may be outweighed by the convenience provided by automatically analyzing the logs.

mulby: Yeah I thought about that. However, I think that problem is separate. This
particular change adds relatively few lines to the file.

I think we should do something like only emit >INFO to stderr and only have
Jenkins cat that one file, perhaps archiving the verbose log in case we
wanted to dive into it and/or analyze it.

mulby:
The intent is for the memory ask to be a single log message, so I don't
expect that to be significant.

Do u see a specific concerning path? It's very possible there could be a
bug in that code...

edx/analytics/tasks/launchers/analyze/main.py
```diff
@@ -0,0 +1,259 @@
+
```
brianhw, line 2: Helpful to add a docstring indicating what this is for (i.e. for analyzing log file output produced by Luigi).

edx/analytics/tasks/launchers/analyze/main.py
```diff
@@ -0,0 +1,259 @@
...
+        if message.content == 'Done scheduling tasks':
+            break
+
+        if re.match(scheduled_pattern, message.content, (re.MULTILINE | re.DOTALL)):
+            break
+
+        parser.next_message()
+
+    if not task.name == 'UncheckedExternalURL':
```
brianhw, line 141: nit: If we thought there may be other tasks to skip, this could be moved into a global list, or even into an argument.

edx/analytics/tasks/launchers/analyze/measure.py
```diff
@@ -0,0 +1,159 @@
...
+import json
+from operator import methodcaller
+
+
+class Measurement(object):
+    """
+    Represents a node in a tree of measurements.
+
+    Note that a particular measurement may be the parent for several child measurements. Those child measurements
+    further breakdown the time spent in the parent. The time spent in the parent itself is referred to as "self_time".
```
brianhw: nit:  s/breakdown/break down/

brianhw:
A few optional nits, but LGTM.  :+1:

mulby:
@rocha @benpatterson @brianhw 

Here is the jenkins job that uses this code: http://scheduler.analytics.edx.org/view/Tools/job/performance/

edx/analytics/tasks/launchers/remote.py
```diff
@@ -60,7 +64,44 @@ def run_task_playbook(arguments, uid):
...
+    log_dir = os.path.join(REMOTE_LOG_DIR, uid)
+
+    env_vars = {}
+    if arguments.workflow_profiler:
+        env_vars['WORKFLOW_PROFILER'] = arguments.workflow_profiler
+        env_vars['WORKFLOW_PROFILER_PATH'] = log_dir
+
+    env_var_string = ' '.join('{0}={1}'.format(k, v) for k, v in env_vars.iteritems())
+
+    command = 'cd {data_dir}/repo && . /home/hadoop/.bashrc && {env_vars}{bg}{data_dir}/venv/bin/launch-task {task_arguments}{end_bg}'.format(
```
brianhw: Does this assume that the user is 'hadoop'?  Yes, now I see below that the user is hardcoded to be 'hadoop'.  So arguments.user above is only for running ansible, and not for running the task.  Might be good to note that.

edx/analytics/tasks/launchers/remote.py
```diff
@@ -156,13 +194,14 @@ def run_ansible(args, verbose, executable='ansible'):
...
 
-def run_remote_shell(inventory, arguments):
+def run_remote_shell(inventory, arguments, shell_command, sudo_user=None):
     """Run a shell command on a hadoop cluster."""
     ansible_group_name = 'mr_{0}_master'.format(arguments.job_flow_id or arguments.job_flow_name)
     hostname = inventory[ansible_group_name][0]
-    shell_command = arguments.shell
-    if arguments.sudo_user:
-        shell_command = 'sudo -u hadoop /bin/sh -c {0}'.format(pipes.quote(arguments.shell))
+    shell_command = shell_command
```
brianhw: What does this line do?

requirements/default.txt
```diff
@@ -27,5 +27,6 @@ six==1.6.1		# MIT
 stevedore==0.14.1 	# Apache 2.0
 tornado==3.1.1 		# Apache 2.0
 wsgiref==0.1.2 		# ZPL
--e git+https://github.com/edx/luigi.git@v1.0.17-fork.edx.1#egg=luigi 		# Apache License 2.0
--e git+https://github.com/edx/opaque-keys.git@33f2b099e92957046a866a9b6d48a71c484bb475#egg=opaque-keys
+git+https://github.com/edx/luigi.git@v1.0.17-fork.edx.1#egg=luigi 		# Apache License 2.0
+git+https://github.com/edx/opaque-keys.git@33f2b099e92957046a866a9b6d48a71c484bb475#egg=opaque-keys
+git+https://github.com/mulby/pyinstrument.git@a35ff76df4c3d5ff9a2876d859303e33d895e78f#egg=pyinstrument     # BSD
```
brianhw: There is a new edx/pyinstrument repo set up.  That can be updated and then pointed to in a separate PR if it's not yet ready.

share/task.yml
```diff
@@ -14,19 +14,17 @@
   vars:
     - repo: git@github.com:edx/edx-analytics-pipeline.git
     - branch: master
-    - root_data_dir: /var/lib/analytics-tasks
-    - root_log_dir: /var/log/analytics-tasks
+    # - root_data_dir: /var/lib/analytics-tasks
+    # - root_log_dir: /var/log/analytics-tasks
```
brianhw, line 18: Should these be commented out or just removed?  (And if they are not defined here, where do the values in the following two lines come from?)

edx/analytics/tasks/launchers/analyze/main.py
mulby, line 2: Done

edx/analytics/tasks/launchers/analyze/main.py
mulby, line 141: Yeah, if/when that is the case I figured we could refactor that

edx/analytics/tasks/launchers/analyze/measure.py
mulby: Done.

edx/analytics/tasks/launchers/remote.py
mulby: Yeah "user" is for the ssh connection, "sudo_user" is the user that is used to actually execute the command. I've refactored this to make that more clear.

edx/analytics/tasks/launchers/remote.py
mulby: Oops. Removed.

share/task.yml
mulby, line 18: They are now set by remote.py, I like to leave them in so that it is clear what variables the script is expecting to be set.

brianhw:
This is such a moving target.  Would you just merge it and put us out of our misery? : )  

:+1:

PR #58, opened 2015-01-27
Ensure the persistent SSH connection does not get dropped due to inactivity

mulby:
I haven't observed this yet, but suspect it is possible.

@brianhw


brianhw:
LGTM. :+1:

PR #60, opened 2015-01-28
Remove enrollment blacklist

mulby:
@brianhw


brianhw:
LGTM.  :+1:   (Interesting that no tests needed to change...)

mulby:
heh, uh, I didn't check the acceptance tests, those probably need to change

PR #61, opened 2015-01-28
Replace join with case statement

mulby:
@brianhw


brianhw:
LGTM. :+1: 

Do we know what time savings this provides?

mulby:
not yet. I'm queuing up a bunch of these to run in succession, but since a run takes >10 hrs it's very slow going...

mulby:
This saved about 25 minutes on a m3.xlarge, baseline job usually takes about 6.5 hrs.

PR #62, opened 2015-01-28
Cap number of reduce tasks used by Hive

mulby:
@brianhw

### DIFF
--- a/edx/analytics/tasks/util/hive.py
+++ b/edx/analytics/tasks/util/hive.py
@@ -45,15 +45,17 @@ class HiveTableTask(WarehouseMixin, OverwriteOutputMixin, HiveQueryTask):
     until they are cleaned up by some external process.
     """
 
-    n_reduce_tasks = luigi.Parameter(default=None, significant=False)
+    n_reduce_tasks = luigi.Parameter(default=25, significant=False)
 
-    def __init__(self, *args, **kwargs):
-        super(HiveTableTask, self).__init__(*args, **kwargs)
+    def hiveconfs(self):
+        jcs = super(HiveTableTask, self).hiveconfs()
 
-        # Convert "n_reduce_tasks" to "reducers_max" since n_reduce_tasks forces hive to use that number of reducers
-        # which can be rather inefficient.
-        self.reducers_max = self.n_reduce_tasks
-        self.n_reduce_tasks = None
+        # Don't force the number of reducer to a particular number, instead set a max number of reducers.
+        # For jobs with small numbers of input groups, this will perform better since it won't have to spin up a lot of
+        # reducers to do a small amount of work.
+        num_reduce_tasks = jcs.pop('mapred.reduce.tasks', self.n_reduce_tasks)
+        jcs.setdefault('hive.exec.reducers.max', num_reduce_tasks)
+        return jcs
 
     def query(self):
         # TODO: Figure out how to clean up old data. This just cleans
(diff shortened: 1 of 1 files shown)


brianhw:
I've not looked carefully, but are there other jobs further "downstream" that need to be plumbed to pass this through?

edx/analytics/tasks/util/hive.py
```diff
@@ -45,6 +45,18 @@ class HiveTableTask(WarehouseMixin, OverwriteOutputMixin, HiveQueryTask):
     until they are cleaned up by some external process.
     """
 
+    n_reduce_tasks = luigi.Parameter(default=25, significant=False)
+
+    def hiveconfs(self):
+        jcs = super(HiveTableTask, self).hiveconfs()
+
+        # Don't force the number of reducer to a particular number, instead set a max number of reducers.
```
brianhw, line 53: s/reducer/reducers/

edx/analytics/tasks/util/hive.py
```diff
@@ -45,6 +45,18 @@ class HiveTableTask(WarehouseMixin, OverwriteOutputMixin, HiveQueryTask):
     until they are cleaned up by some external process.
     """
 
+    n_reduce_tasks = luigi.Parameter(default=25, significant=False)
+
+    def hiveconfs(self):
+        jcs = super(HiveTableTask, self).hiveconfs()
```
brianhw, line 51: Please use a more readable variable name than jcs.

edx/analytics/tasks/util/hive.py
```diff
@@ -45,6 +45,18 @@ class HiveTableTask(WarehouseMixin, OverwriteOutputMixin, HiveQueryTask):
     until they are cleaned up by some external process.
     """
 
+    n_reduce_tasks = luigi.Parameter(default=25, significant=False)
```
brianhw, line 48: If this is really max_reduce_tasks, shouldn't it be so named?

edx/analytics/tasks/util/hive.py
```diff
@@ -45,6 +45,18 @@ class HiveTableTask(WarehouseMixin, OverwriteOutputMixin, HiveQueryTask):
...
+
+    def hiveconfs(self):
+        jcs = super(HiveTableTask, self).hiveconfs()
+
+        # Don't force the number of reducer to a particular number, instead set a max number of reducers.
+        # For jobs with small numbers of input groups, this will perform better since it won't have to spin up a lot of
+        # reducers to do a small amount of work.
+        num_reduce_tasks = jcs.pop('mapred.reduce.tasks', self.n_reduce_tasks)
```
brianhw, line 56: max_reduce_tasks?

brianhw:
Okay, best I can tell, all the downstream classes of HiveQueryToMysqlTask are passing in n_reduce_tasks, including classes derived from CourseActivityTask and EnrollmentTask.

mulby:
Yeah I have been just hacking on this to see what the performance impact is. I agree that a) jcs needs to be renamed and b) overloading n_reduce_tasks like this is confusing and we should do something more clear OR just use it the way luigi intended and have it set a fixed number of reducers instead of the max.

mulby:
@brianhw

brianhw:
Wasn't sure what the next step on this is.  I had the impression that the performance gain was small, so that you were inclined to not pursue this.  But let me know if that assumption was wrong.

mulby:
That's correct, closing.

PR #63, opened 2015-02-10
Support writing output directly to destination

mulby:
This needs to be paired with a change to luigi core to be useful.

@brianhw

### DIFF
--- a/edx/analytics/tasks/s3_util.py
+++ b/edx/analytics/tasks/s3_util.py
@@ -211,3 +211,21 @@ class S3HdfsTarget(HdfsTarget):
             if not hasattr(self, 's3_client'):
                 self.s3_client = ScalableS3Client()
             return AtomicS3File(safe_path, self.s3_client)
+
+
+class S3HadoopTarget(S3HdfsTarget):
+
+    def __init__(self, path=None, format=Plain, is_tmp=False):
+        super(S3HadoopTarget, self).__init__(path=path, format=format, is_tmp=is_tmp)
+        self.success_marker_path = path + '_SUCCESS'
+        self.stage_output = False
+
+    def marker_exists(self):
+        self.fs.exists(self.success_marker_path)
+
+    def exists(self):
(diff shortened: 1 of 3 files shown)


mulby:
FYI @jab5569

edx/analytics/tasks/s3_util.py
```diff
@@ -211,3 +211,21 @@ def open(self, mode='r'):
             if not hasattr(self, 's3_client'):
                 self.s3_client = ScalableS3Client()
             return AtomicS3File(safe_path, self.s3_client)
+
+
+class S3HadoopTarget(S3HdfsTarget):
+
+    def __init__(self, path=None, format=Plain, is_tmp=False):
+        super(S3HadoopTarget, self).__init__(path=path, format=format, is_tmp=is_tmp)
+        self.success_marker_path = path + '_SUCCESS'
```
brianhw, line 220: don't we need a url_join for this?

brianhw:
Looks good.  Needs docstrings and tests, but nice.

edx/analytics/tasks/s3_util.py
```diff
@@ -211,3 +211,21 @@ def open(self, mode='r'):
...
+
+    def marker_exists(self):
+        self.fs.exists(self.success_marker_path)
+
+    def exists(self):
+        return self.marker_exists()
+
+    def prepare_for_output(self):
+        if not self.marker_exists() and self.fs.exists(self.path):
+            self.remove()
```
mulby, line 231: This makes me a little nervous. I'm pretty sure it's safe? But not 100% sure... can you think of any cases where this is not what we want to do?

brianhw, line 231: Basically it applies to any target that begins with s3:// and ends with /.  I can't yet access S3 from my new home PC, but I assume that anything that we would specify that ends with a / will be Hadoop output.  Does that carry over to Sqoop (I think so), and to Hive (I don't remember)?

PR #64, opened 2015-02-10
Send unit testing coverage stats to datadog.

benpatterson:
@jzoldak @mulby @brianhw @jab5569

This is currently being done on Jenkins...we'd replace the in-job bash with this (for the most part--ignoring the venv setup).

### DIFF
--- a/scripts/report-build-stats.sh
+++ b/scripts/report-build-stats.sh
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+#
+# Using test-metrics, this script will post
+# stats to datadog.
+#
+#
+
+if [ -n "${DATADOG_API_KEY}" ];
+
+    then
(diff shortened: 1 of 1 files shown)


brianhw:
Looks reasonable.  But this is still tied to Jenkins, in that it piggy-backs on the existing coverage.xml output.

benpatterson:
Wise as usual, @brianhw. Although we don't remove Jenkins dependency entirely, it is an incremental step away, by putting this into a script file. 

That way, if we, say, put unit testing on Travis, then we'd just use this script as a post-build task (when coverage.xml is available.)

brianhw:
Okay.  :+1:

benpatterson:
@mulby @jab5569 could one of you fellas give a pass on this change as well?

mulby:
:+1:

PR #65, opened 2015-02-11
This was breaking the acceptance tests

mulby:
I removed the event from the input data when I removed the blacklist but forgot to update this test.

@brianhw @jab5569


brianhw:
I don't remember the input being changed, but I'll trust that this is the right output. :+1:

PR #66, opened 2015-02-12
Remove combiner from user activity job

mulby:
This appears to be causing a significant performance problem. The user activity job on master took >9hrs when it previously took only ~3hrs.

@brianhw


brianhw:
Can you remind me again why this was done in the first place?  The goal was to do the reducing earlier?

mulby:
The more I think about this, I'm wondering if the issue is actually that the map-reduce job is producing way more records than it was previously. There will be a lot more reduce groups since we are storing the data at a much more detailed level of granularity (by user/day, instead of course/week).

I'm going to close this.

PR #67, opened 2015-02-15
Load last country of user into the result store

mulby:
@brianhw @jab5569


brianhw:
LGTM.  Acceptance tests need updating.  Also, do we need to add an index to this table?

brianhw:
LGTM.   :+1:

PR #68, opened 2015-02-18
Add coveralls capability.

benpatterson:
@mulby @brianhw @jab5569 

In a bit of de-scoping or re-scoping, I'm suggesting we implement this independently of any Travis integration. Posting to Coveralls will make it easier to produce the coverage stat on a dashboard, and can be continued if we end up going onto Travis (or not :) ).

### DIFF
--- a/scripts/report-build-stats.sh
+++ b/scripts/report-build-stats.sh
@@ -8,7 +8,7 @@
 if [ -n "${DATADOG_API_KEY}" ];
 
     then
-        echo "Reporting coverage stats to datadog"
+        echo "Reporting coverage stats to datadog..."
 
         rm -rf test-metrics
         git clone https://github.com/wedaly/test-metrics
@@ -23,8 +23,18 @@ if [ -n "${DATADOG_API_KEY}" ];
 END
         python -m metrics.coverage unit_test_groups.json ../coverage.xml
 
+        cd..
+
     else
         echo "Skipping sending stats to datadog. DATADOG_API_KEY not set."
 
 fi
 
+# Report coverage to Coveralls
+if [ -n "${COVERALLS_REPO_TOKEN}" ];
+    then
+    echo "Sending coverage to Coveralls.io..."
+    pip install coveralls
+    coveralls
+
+fi


scripts/report-build-stats.sh
```diff
@@ -23,8 +23,18 @@ if [ -n "${DATADOG_API_KEY}" ];
...
 
+        cd..
+
     else
         echo "Skipping sending stats to datadog. DATADOG_API_KEY not set."
 
 fi
 
+# Report coverage to Coveralls
+if [ -n "${COVERALLS_REPO_TOKEN}" ]:
```
mulby: shouldn't this end with a semicolon?

scripts/report-build-stats.sh
```diff
@@ -23,8 +23,18 @@ if [ -n "${DATADOG_API_KEY}" ];
 END
         python -m metrics.coverage unit_test_groups.json ../coverage.xml
 
+        cd..
```
mulby, line 26: Why is this necessary?

brianhw, line 26: Looks like it undoes the "cd test-metrics" call, so that cwd is unchanged by datadog output.  Not necessary, but cleaner, no?

brianhw:
Apart from semicolon, LGTM. :+1:

brianhw:
Actually, a quick question about virtualenvs:  what virtualenv is this going to be run with?  This is doing two pip install calls that will presumably load the latest version of coveralls and whatever requirements are pulled in from test-metrics.  Probably no real impact, but just noting.

benpatterson:
@brianhw No I'm glad you asked because I agree this is important. As configured in Jenkins, it's using a new venv (in fact, it's a separate Jenkins job), so venv's will be clean between the reporting and the pipeline test job.

If we end up going with Travis, it would be the same virtualenv, but it would be run as a post-build task. So in that case, the virtualenv would be clean, but only because of the order of tasks.

Thx for pointing out the syntax error. Now that I have the repo token I can test this, and found that problem immediately. Fixed and squashed. Will merge shortly.

scripts/report-build-stats.sh
benpatterson, line 26: Re this line, yes that's the intention. Coveralls will default to the .coveragerc in CWD so I wanted to go back to the same dir as at the beginning of the if/fi statement.

mulby, line 26: ah, I was going to recommend a pushd popd kind of thing, but this is fine

PR #70, opened 2015-02-18
Add in missing categories to case statement

mulby:
This fixes the level of education computation on master. The performance optimization broke it.

Reviewer: @brianhw
FYI: @jab5569


mulby:
I should note, I validated the results of running this job against the prod results store and they matched.

brianhw:
LGTM.  :+1:    

Pity that this isn't something we can test as a unit test, yet it doesn't seem appropriate to do all edge cases of this as part of an acceptance test.

PR #72, opened 2015-02-18
Installs should happen outside of any directory-switching.

benpatterson:
Otherwise it causes corrupt virtualenvs in some environments.


benpatterson:
@mulby @brianhw another hotfix...sorry for the noise.

PR #74, opened 2015-02-20
Prevent overwriting with empty data sets

mulby:
Fixes AN-4631

Since Luigi considers the presence of an empty folder as being a valid "complete" output, it is possible for the MySQL load task to read an empty input and delete all content in a table without replacing it with anything. This is a very basic sanity check that prevents that mitigates the risk of that specific edge case causing the table content to be deleted.


brianhw:
LGTM.  :+1:

PR #76, opened 2015-02-27
Ensure answer distribution always overwrites DB records

mulby:
Reviewer: @brianhw

Answer distribution has been running at the same time each day. It has not been removing records from the "table_updates" table and the unique key for the parameters is the hour of the day. Since it runs during the same hour every day, it hasn't been overwriting the records in the database.

This sets the default to overwrite the DB records, which forces it to run even if the parameters haven't changed and their is a record in table_updates.

FYI: @jab5569

### DIFF
--- a/edx/analytics/tasks/answer_dist.py
+++ b/edx/analytics/tasks/answer_dist.py
@@ -809,14 +809,9 @@ class AnswerDistributionToMySQLTaskWorkflow(
     AnswerDistributionDownstreamMixin,
     MapReduceJobTaskMixin
 ):
-    def init_copy(self, connection):
-        """
-        Truncate the table before re-writing
-        """
-        # Use "DELETE" instead of TRUNCATE since TRUNCATE forces an implicit commit before it executes which would
-        # commit the currently open transaction before continuing with the copy.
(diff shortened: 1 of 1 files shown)


edx/analytics/tasks/answer_dist.py
```diff
@@ -809,14 +809,8 @@ class AnswerDistributionToMySQLTaskWorkflow(
...
-    def init_copy(self, connection):
-        """
-        Truncate the table before re-writing
-        """
-        # Use "DELETE" instead of TRUNCATE since TRUNCATE forces an implicit commit before it executes which would
-        # commit the currently open transaction before continuing with the copy.
-        query = "DELETE FROM {table}".format(table=self.table)
-        connection.cursor().execute(query)
+
+    overwrite = luigi.BooleanParameter(default=True)
```
brianhw, line 814: Should add a comment that this value of overwrite here is overriding the default definition (that comes in through MysqlInsertTask).  Other alternative is to actually pass it down from the top-level (where it is defined by OverwriteOutputMixin), in AnswerDistributionWorkflow.

mulby:
@brianhw I added a comment here

brianhw:
LGTM. :+1:

PR #77, opened 2015-02-27
Create better indexes

mulby:
@brianhw This should fix our indexing issue for database created in the future


mulby:
@shnayder FYI

brianhw:
LGTM.  Is it worth a comment as to why the order matters?  I talked briefly with Victor, and I don't expect it is obvious to anyone.

PR #78, opened 2015-03-10
Fix check for zero row insertion.

brianhw:
Production hit this condition, but returned "UnboundLocalError: local variable 'row_count' referenced before assignment".  Fail more gracefully.

@jab5569


mulby:
:+1:

PR #79, opened 2015-03-11
Compute cumulative enrollments

mulby:
In addition to simply computing the number of students currently enrolled in the course, we add support for displaying the number of students who have ever been in the course.

Open to suggestions about removing the left outer join since it's going to create quite a performance hit.

Also - note this change requires a schema change, so the result store database will need to be regenerated in order for this new column to appear.

Reviewer: @brianhw 

FYI: @dylanrhodes @jab5569 @shnayder

### DIFF
--- a/edx/analytics/tasks/enrollments.py
+++ b/edx/analytics/tasks/enrollments.py
@@ -91,8 +91,9 @@ class DaysEnrolledForEvents(object):
     """
     Determine which days a user was enrolled in a course given a stream of enrollment events.
 
-    Produces a record for each date in which the user was enrolled in the course. Note that the user need not have been
-    enrolled in the course for the entire day. These records will have the following format:
+    Produces a record for each date from the date the user enrolled in the course for the first time to the end of the
+    interval. Note that the user need not have been enrolled in the course for the entire day. These records will have
+    the following format:
 
         datestamp (str): The date the user was enrolled in the course during.
         course_id (str): Identifies the course the user was enrolled in.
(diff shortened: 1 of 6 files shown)


edx/analytics/tasks/enrollments.py
```diff
@@ -91,8 +91,8 @@ class DaysEnrolledForEvents(object):
     """
     Determine which days a user was enrolled in a course given a stream of enrollment events.
 
-    Produces a record for each date in which the user was enrolled in the course. Note that the user need not have been
-    enrolled in the course for the entire day. These records will have the following format:
+    Produces a record for each date during which the user enrolled in the course for the first time. Note that the user
```
brianhw: This is confusing to me.  It seems to produce many records, not just one for the first enrollment as this change seems to suggest.

edx/analytics/tasks/enrollments.py
```diff
@@ -345,10 +336,15 @@ def query(self):
                 ce.date,
                 ce.course_id,
                 IF(p.gender != '', p.gender, NULL),
+                COUNT(enrolled.user_id),
```
brianhw: Can we just do SUM(ce.at_end) and skip the whole left outer join?

edx/analytics/tasks/location_per_course.py
```diff
@@ -213,11 +214,15 @@ def query(self):
                 sce.dt,
                 sce.course_id,
                 uc.country_code,
+                count(enrolled.user_id),
```
brianhw: likewise, SUM(sce.is_active)...

edx/analytics/tasks/enrollments.py
mulby: Updated.

edx/analytics/tasks/enrollments.py
mulby: YES! good call, I had a nagging feeling there was a way to do this without a join...

mulby:
This is now passing acceptance tests and is ready for review.

mulby:
@brianhw

brianhw:
LGTM. :+1:

PR #82, opened 2015-04-14
Add unit tests for mapper and problems attempted

mulby:
@brianhw @jab5569

### DIFF
--- a/edx/analytics/tasks/tests/test_student_engagement.py
+++ b/edx/analytics/tasks/tests/test_student_engagement.py
@@ -208,18 +208,25 @@ class StudentEngagementTaskReducerTest(unittest.TestCase):
         )
         self.task.init_local()
 
+    def test_no_events(self):
+        output = self._get_reducer_output([])
+        self.assertEquals(len(output), 0)
+
+    def _get_reducer_output(self, inputs):
+        """Run the reducer and return the ouput"""
+        return tuple(self.task.reducer(self.REDUCE_KEY, inputs))
+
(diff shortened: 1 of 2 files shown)


edx/analytics/tasks/tests/test_student_engagement.py
```diff
@@ -0,0 +1,244 @@
...
+        self.task.init_local()
+
+    def _check_output(self, inputs, column_num, expected_value):
+        """Compare generated with expected output."""
+        output = self._get_reducer_output(inputs)
+        self.assertEquals(len(output), 1)
+        self.assertEquals(output[0][column_num], expected_value)
+
+    def test_no_events(self):
+        self._check_output([], self.WAS_ACTIVE_COLUMN_NUM, 1)
```
brianhw: Why do we expect the output to have a count of 1 when the input contained no events for the given key?

brianhw:
Just one small question. LGTM otherwise.

mulby:
@brianhw I put in a fix for your comment. Ready for re-review.

brianhw:
Missing a test for last-subsection-viewed with no timestamp.

mulby:
I actually couldn't figure out how to reach that code, I suspect this "if" statement can never actually get executed (given that invalid timestamps are filterd out by the event log selection process).

We should probably remove it, however, I like defensive stuff like this, since those other parts of the code might change which would allow invalid timestamps to seep into this area...

brianhw:
I'm fine with that.  :+1:

PR #83, opened 2015-04-16
PREVIEW:  weekly engagement csv calculation

brianhw:
Add interval_type parameter, and recognize "daily", "weekly" and "all".  

Some initial unit tests.

@mulby @jab5569

### DIFF
--- a/edx/analytics/tasks/student_engagement.py
+++ b/edx/analytics/tasks/student_engagement.py
@@ -10,6 +10,7 @@ import re
 
 import luigi
 
+from edx.analytics.tasks.calendar import CalendarTableTask
 from edx.analytics.tasks.database_imports import (
     ImportAuthUserTask, ImportCourseUserGroupTask, ImportCourseUserGroupUsersTask)
 from edx.analytics.tasks.enrollments import CourseEnrollmentTableTask
@@ -17,7 +18,6 @@ from edx.analytics.tasks.mapreduce import MapReduceJobTask, MapReduceJobTaskMixi
 from edx.analytics.tasks.pathutil import EventLogSelectionMixin, EventLogSelectionDownstreamMixin
 from edx.analytics.tasks.url import get_target_from_url, url_path_join
 from edx.analytics.tasks.util import eventlog
-from edx.analytics.tasks.util import Week
 from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
 
 from edx.analytics.tasks.util.hive import WarehouseMixin, HiveTableTask, HivePartition, HiveTableFromQueryTask
@@ -37,7 +37,7 @@ class StudentEngagementTask(EventLogSelectionMixin, MapReduceJobTask):
     SUBSECTION_ACCESSED_PATTERN = r'/courses/[^/+]+(/|\+)[^/+]+(/|\+)[^/]+/courseware/[^/]+/[^/]+/.*$'
 
     output_root = luigi.Parameter()
-    group_by_week = luigi.BooleanParameter(default=False)
+    interval_type = luigi.Parameter(default="daily")
 
     def mapper(self, line):
         value = self.get_event_and_date_string(line)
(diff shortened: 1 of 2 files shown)


edx/analytics/tasks/student_engagement.py
```diff
@@ -88,32 +88,39 @@ def mapper(self, line):
...
         date_grouping_key = date_string
-        if self.group_by_week:
-            week_of_year = self.get_iso_week_containing_date(date_string)
-            start_date = week_of_year.monday()
-            end_date = week_of_year.sunday() + datetime.timedelta(days=1)
-            date_grouping_key = '{0}-{1}'.format(start_date.isoformat(), end_date.isoformat())
 
-        yield ((date_grouping_key, course_id, username), (entity_id, event_type, info))
+        if self.interval_type == 'weekly':
+            last_valid_date = self.interval.date_b - datetime.timedelta(days=1)  # pylint: disable=no-member
```
mulby, line 96: nit: I would prefer "complete" instead of "valid"

edx/analytics/tasks/student_engagement.py
```diff
@@ -34,7 +34,7 @@ class StudentEngagementTask(EventLogSelectionMixin, MapReduceJobTask):
     SUBSECTION_ACCESSED_PATTERN = r'/courses/[^/+]+(/|\+)[^/+]+(/|\+)[^/]+/courseware/[^/]+/[^/]+/.*$'
 
     output_root = luigi.Parameter()
-    group_by_week = luigi.BooleanParameter(default=False)
+    interval_type = luigi.Parameter(default="daily")
```
mulby, line 40: Is "daily" handled below? Shouldn't this be "all"?

edx/analytics/tasks/student_engagement.py
```diff
@@ -88,32 +88,39 @@ def mapper(self, line):
...
+            end_of_week_date = event_date + datetime.timedelta(days=days_until_end)
+            date_grouping_key = end_of_week_date.isoformat()
+
+        elif self.interval_type == 'all':
+            # If gathering all data for a given user, use the last valid day of the interval
+            # for joining with enrollment.
+            last_valid_date = self.interval.date_b - datetime.timedelta(days=1)  # pylint: disable=no-member
+            date_grouping_key = last_valid_date.isoformat()
+
+        yield ((date_grouping_key, course_id, username), (entity_id, event_type, info, date_string))
```
mulby, line 116: I'm thinking we should json encode info to be really clear about what's happening here, we can do this in a follow on PR if you want...

edx/analytics/tasks/student_engagement.py
```diff
@@ -186,7 +198,8 @@ def output(self):
 
 class StudentEngagementTableDownstreamMixin(WarehouseMixin, EventLogSelectionDownstreamMixin, MapReduceJobTaskMixin):
     """All parameters needed to run the StudentEngagementTableTask task."""
-    pass
+
+    interval_type = luigi.Parameter(default="daily")
```
mulby: same comment as above about "daily"

edx/analytics/tasks/student_engagement.py
mulby, line 40: ah, nevermind, I see that date_grouping_key = date_string in this case...

edx/analytics/tasks/student_engagement.py
```diff
@@ -338,9 +375,16 @@ def requires(self):
             ImportCourseUserGroupUsersTask(**kwargs_for_db_import),
             CourseEnrollmentTableTask(**kwargs_for_enrollment),
         )
+        # TODO: If this doesn't work, then we can just always add it in above.
```
mulby, line 381: did it work?

mulby:
I don't see any blockers here, the only things I would change is stuff not really related to this PR (directly):
1. JSON encode the "info" dict
2. Use ddt more aggressively in the test suite (where it's trivial to do so)

Both actions are optional and we could certainly take care of them later. :+1:

edx/analytics/tasks/student_engagement.py
brianhw, line 381: Yes, it did.  This comment should be removed.

PR #85, opened 2015-04-16
Use json for packing info between map and reduce.

brianhw:
Based on comments from earlier PR.

@mulby @jab5569


mulby:
I tacked another ddt conversion on to this.

:+1:

PR #86, opened 2015-04-21
Initial version of DailyStudentEngagementAcceptanceTest class.

johnalbaker:
Additions Include:
- acceptance/student_engagment.py
  - Tests student_engagement_daily_\* files in HDFS
  - Tests are run on two dimensions: 
    - Attributes
    - Records
- load_student_engagment.sql file

Fyi @mulby @brianhw


edx/analytics/tasks/tests/acceptance/test_student_engagement.py
```diff
@@ -10,116 +11,133 @@
...
+
+                #
+                # Validate various comparisons within a given row. Eg:
+                # 1. problems correct gte to problems_attempted
+                # 2.
+                #
+                self.validate_within_rows(df)
+
+
+            except IOError as exc:
```
mulby: nit: I would wrap as little code in this try block as possible, I suspect this "except" clause can be moved up near the top of the function

edx/analytics/tasks/tests/acceptance/test_student_engagement.py
```diff
@@ -10,116 +11,133 @@
...
+        self.assertRegexpMatches(course_id,'^.{1,}\/.{1,}\/.{1,}$')
+
+    def validate_username_string_format(self,user_name):
+        self.assertRegexpMatches(user_name,'^.{1,}$')
+
+    def validate_email_string_format(self,email):
+        self.assertRegexpMatches(email,'^([^@|\s]+@[^@]+\.[^@|\s]+)$')
+
+    def validate_cohort_format(self,cohort):
+        self.assertRegexpMatches(str(cohort),'^.*$')
```
mulby: this won't ever fail, will it?

mulby:
This looks good, a great start for us to build off of. I think we can merge it as is, other stuff I'd like to check at this level:
- column headings are what we expect
- the actual set of files are what we expect (including hashed directory names)

In addition we should probably deep dive on one or more files:
- validate the contents of the file are exactly what we expect (provide a hardcoded data frame that the read file should match exactly)

All three of those checks can be added in subsequent PRs.

:+1:

edx/analytics/tasks/tests/acceptance/test_student_engagement.py
johnalbaker: Great point. I'll have to update this regex to attempt to match the last-section format or nothing at all.

brianhw:
It would be helpful to fix the pep8 errors before merging.  I don't think we are as fussy about pylint errors in acceptance test files (i.e. docstrings), but you might want to look down those errors as well, to see if there are any problems that should be fixed.

edx/analytics/tasks/tests/acceptance/test_student_engagement.py
```diff
@@ -10,116 +11,133 @@
...
 
-        self.upload_data()
+        # There are 15 student_engagement files in the test data directory
+        self.assertEqual(len(outputs), 15)
 
-    def upload_data(self):
-       src = os.path.join(self.data_dir, 'input', self.INPUT_FILE)
-       dst = url_path_join(self.test_src, self.INPUT_FILE)
+        # Check that the results have data
+        for output in outputs: # 'file' is a builtin type, 'name' is a less-ambiguous variable name.
```
brianhw: nit:  I don't understand the comment -- you suggest using 'name' but then use 'f'.  I'd probably suggest "srcfile" or something similar.

edx/analytics/tasks/tests/acceptance/test_student_engagement.py
```diff
@@ -10,116 +11,133 @@
...
-            '--output-root', self.test_out,
-            '--include', '"*"',
-            '--manifest', url_path_join(self.test_root, 'manifest.txt'),
-            '--base-input-format', self.INPUT_FORMAT,
-            '--lib-jar', self.oddjob_jar,
-            '--n-reduce-tasks', str(self.NUM_REDUCERS),
-        ])
-        self.validate_output()
+                #
+                # Ensure course_id on each row is of the format characters/characters/characters
```
brianhw: nit:  I would move the comments down to be docstrings on the validiation functions.  Otherwise, if we change the validation, we then have to remember to update these comments as well.

edx/analytics/tasks/tests/acceptance/test_student_engagement.py
```diff
@@ -10,116 +11,133 @@
...
+                #
+                for course_id in df["course_id"]:
+                    self.validate_course_id_string_format(course_id)
 
-    def validate_output(self):
-        outputs = self.s3_client.list(self.test_out)
-        outputs = [url_path_join(self.test_out, p) for p in outputs]
+                #
+                # Ensure user_name on each row is of the format characters/characters/characters
```
brianhw: cut-and-paste error.  (That's not the format of username.)

edx/analytics/tasks/tests/acceptance/test_student_engagement.py
```diff
@@ -10,116 +11,133 @@
...
+                #
+                for user_name in df["username"]:
+                    self.validate_username_string_format(user_name)
 
-        #self.fail(outputs)
-        #self.fail("foobarbaz")
+                #
+                # Ensure email address on each row is of the format characters/characters/characters
```
brianhw: ditto.

edx/analytics/tasks/tests/acceptance/test_student_engagement.py
brianhw: Better yet, can you just get rid of this altogether?  If you want to permit an empty directory, then just check for it and continue.  Otherwise, just put this up after the read_csv call.

edx/analytics/tasks/tests/acceptance/test_student_engagement.py
```diff
@@ -10,116 +11,133 @@
...
+
+
+    def validate_number_of_columns(self,num_columns):
+        self.assertTrue(num_columns==15,msg="Number of columns not equal to 15")
+
+    def validate_date_cell_format(self,date):
+        self.assertRegexpMatches(date,'^\d\d\d\d-\d\d-\d\d$')
+
+    def validate_course_id_string_format(self,course_id):
+        self.assertRegexpMatches(course_id,'^.{1,}\/.{1,}\/.{1,}$')
```
brianhw: This should use edx.analytics.tasks.util.opaque_key_util.is_valid_course_id().

edx/analytics/tasks/tests/acceptance/test_student_engagement.py
```diff
@@ -10,116 +11,133 @@
...
+
+    def validate_username_string_format(self,user_name):
+        self.assertRegexpMatches(user_name,'^.{1,}$')
+
+    def validate_email_string_format(self,email):
+        self.assertRegexpMatches(email,'^([^@|\s]+@[^@]+\.[^@|\s]+)$')
```
brianhw: nit:  you don't need parens in the pattern.

edx/analytics/tasks/tests/acceptance/test_student_engagement.py
brianhw: Or just test "if cohort:" before the assert (and change the pattern of the assert).   You could import the regexp used to find the last-section -- is that what you meant?

edx/analytics/tasks/tests/acceptance/test_student_engagement.py
```diff
@@ -10,116 +11,133 @@
...
+                # Ensure email address on each row is of the format characters/characters/characters
+                #
+                for email in df["email"]:
+                    self.validate_email_string_format(email)
 
-#
-#         # There are 3 courses in the test data
-        self.assertEqual(len(outputs), 3)
+                #
+                # Ensure problems,videos,forum and texbook column values are greater than 0
```
brianhw: greater than or equal to

brianhw:
I've done making a pass.  We're not merging to master, so my comments could be addressed in a separate PR if that makes more sense.

edx/analytics/tasks/tests/acceptance/test_student_engagement.py
johnalbaker: Moved this up under the pd.read_csv() code.

edx/analytics/tasks/tests/acceptance/test_student_engagement.py
jab5569: Yes, thanks Brian, that's what I was looking for: the regexp used to find last-section.

edx/analytics/tasks/tests/acceptance/test_student_engagement.py
johnalbaker: Removed.

edx/analytics/tasks/tests/acceptance/test_student_engagement.py
johnalbaker: Done

edx/analytics/tasks/tests/acceptance/test_student_engagement.py
johnalbaker: Fixed.

edx/analytics/tasks/tests/acceptance/test_student_engagement.py
johnalbaker: Fixed.

edx/analytics/tasks/tests/acceptance/test_student_engagement.py
johnalbaker: Done.

edx/analytics/tasks/tests/acceptance/test_student_engagement.py
johnalbaker: Done.

PR #87, opened 2015-04-23
Add logging events and checking for zero counts to acceptance tests.

brianhw:
@jab5569 Here's the current progress I've made.  A couple of things simplify the code, which we can talk about in person.

I think somewhere in the process I broke the cohort assignment, but that can be fixed in a separate PR.  I can try to get that working tomorrow.


johnalbaker:
This looks great @brianhw.

PR #90, opened 2015-04-27
Check zero counts for all files without matching fixtures.

brianhw:
@jab5569 This is what I had in mind when we talked about removing the enumeration.  

I also moved some validation to apply to all csv's.


johnalbaker:
Looks brilliant to me :+1:

mulby:
:+1:

PR #93, opened 2015-05-08
Change "session" to "viewing".

brianhw:
Change terminology: a "session" is now a "viewing".

Also contains fix for finding final_segment.

@mulby


mulby:
:+1:

PR #94, opened 2015-05-08
Video timeline bug fix

mulby:
@brianhw


brianhw:
:+1:

brianhw:
Actually, this is broken.  The classname is wrong in the super() call.  But I noticed that the unit tests for this branch are broken.   I'm guessing that it's from a lot of the work on credentials and hosts and the like to support docker work.  I'm creating a new branch off of master that only has the video-related stuff in it.  And I'll include this change.  But I can't actually run it without the Google API key, I assume.

PR #96, opened 2015-05-12
WIP: Handle videos with no duration metadata

brianhw:
@mulby

### DIFF
--- a/edx/analytics/tasks/video.py
+++ b/edx/analytics/tasks/video.py
@@ -11,7 +11,7 @@ from luigi import configuration
 
 from edx.analytics.tasks.mapreduce import MapReduceJobTask, MapReduceJobTaskMixin
 from edx.analytics.tasks.pathutil import EventLogSelectionMixin, EventLogSelectionDownstreamMixin
-from edx.analytics.tasks.url import get_target_from_url, ExternalURL
+from edx.analytics.tasks.url import get_target_from_url, ExternalURL, url_path_join
 from edx.analytics.tasks.util import eventlog
 from edx.analytics.tasks.util.hive import WarehouseMixin, HivePartition, HiveTableTask, HiveQueryToMysqlTask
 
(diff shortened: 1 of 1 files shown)


edx/analytics/tasks/video.py
```diff
@@ -301,22 +314,17 @@ def reducer(self, key, viewings):
...
-            viewings_by_duration.setdefault(video_duration, []).append((username, start_offset, end_offset))
+            username, start_offset, end_offset, duration = viewing
 
-        video_duration = 0
-        max_count = -1
-        for duration, indexed_viewings in viewings_by_duration.iteritems():
-            current_video_viewing_count = len(indexed_viewings)
-            if current_video_viewing_count > max_count:
+            # Find the maximum actual video duration, but indicate that
+            # it's unknown if any viewing was of a video with unknown duration.
```
mulby, line 292: Why?

brianhw, line 292: If an early video had a known duration, but later videos did not, and the non-known duration was longer and viewings extended past the known duration, it would be really confusing and wrong.  Because any video with unknown duration may permit viewings to extend beyond known durations, the safest approach is to assume that a module with a video of unknown duration should be labeled as having an unknown duration.

This could be relaxed if there are no attested viewings that extend beyond a proposed video duration.

mulby, line 292: That makes sense - lets leave it as is. We will just infer the duration below to figure out the number of users who watched "to the end".

mulby:
This generally looks good to me. I'm assuming @dsjen is OK with us putting in a duration of -1?

dsjen:
Thanks for the info!  That sounds good to me.  I will have Insights ignore -1 for now.

mulby:
@dsjen it will be more subtle than that... if it's -1 we would want to use the max(segment) as the duration

dsjen:
Thanks for the added details.  If duration is -1, I'll calculate a best guess on the client and perhaps add a note on the UI.  Otherwise, I'll use the supplied duration as the official one.

PR #99, opened 2015-05-14
Make video mapper more robust

brianhw:
Changes include:
- Add check for missing event_data and missing encoded_module_id.
- Skip unrelated events more quickly in mapper.
- Catch type errors when casting current_time to a float.

@mulby


edx/analytics/tasks/video.py
```diff
@@ -47,11 +50,23 @@ def init_local(self):
         self.api_key = configuration.get_config().get('google', 'api_key', None)
 
     def mapper(self, line):
+        # Add a filter here to permit quicker rejection of unrelated events.
+        if VIDEO_EVENT_MINIMUM_STRING not in line:
```
mulby, line 54: This is a great idea.

mulby:
:+1:

PR #102, opened 2015-05-15
Video usage tests

mulby:
@brianhw

### DIFF
--- a/edx/analytics/tasks/tests/test_video.py
+++ b/edx/analytics/tasks/tests/test_video.py
@@ -3,10 +3,11 @@
 import json
 
 from mock import patch, MagicMock
-from ddt import ddt, data
+from ddt import ddt, data, unpack
 
-from edx.analytics.tasks.tests.config import with_luigi_config
-from edx.analytics.tasks.video import UserVideoViewingTask
+from edx.analytics.tasks.video import (
+    UserVideoViewingTask, VideoUsageTask, VIDEO_VIEWING_SECONDS_PER_SEGMENT, VIDEO_UNKNOWN_DURATION
+)
 from edx.analytics.tasks.tests import unittest
 from edx.analytics.tasks.tests.opaque_key_mixins import InitializeOpaqueKeysMixin, InitializeLegacyKeysMixin
 from edx.analytics.tasks.tests.map_reduce_mixins import MapperTestMixin, ReducerTestMixin
@@ -14,11 +15,8 @@ from edx.analytics.tasks.tests.map_reduce_mixins import MapperTestMixin, Reducer
 
 @ddt
 class UserVideoViewingTaskMapTest(InitializeOpaqueKeysMixin, MapperTestMixin, unittest.TestCase):
-    """Base class for test analysis of detailed student engagement"""
+    """Test video viewing mapper"""
 
-    DEFAULT_USER_ID = 10
-    DEFAULT_TIMESTAMP = "2013-12-17T15:38:32.805444"
-    DEFAULT_DATE = "2013-12-17"
     UTF8_BYTE_STRING = 'I\xd4\x89\xef\xbd\x94\xc3\xa9\xef\xbd\x92\xd0\xbb\xc3\xa3'
 
     task_class = UserVideoViewingTask
@@ -183,9 +181,7 @@ class UserVideoViewingTaskMapTest(InitializeOpaqueKeysMixin, MapperTestMixin, un
         event = json.dumps(payload)
         self.assert_no_map_output_for(self.create_event_log_line(template_name='pause_video', event=event))
 
-    @data(
-        'old_time', 'new_time'
-    )
+    @data('old_time', 'new_time')
     def test_seek_video_invalid_time(self, field_to_remove):
         payload = {
             "id": self.video_id,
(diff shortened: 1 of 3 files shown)


edx/analytics/tasks/video.py
```diff
@@ -135,17 +135,16 @@ def mapper(self, line):
...
-
-            # Some events have ridiculous (and dangerous) values for time.
-            if time_value > VIDEO_MAXIMUM_DURATION:
-                log.warn('Video event with huge time-offset value: {0}'.format(line))
-                return None
+        try:
+            time_value = float(time_value)
+        except (TypeError, ValueError):
+            log.warn('Video event with invalid time-offset type: {0}'.format(line))
+            return None
```
brianhw, line 142: If we pass in None, this will log a warning saying it an invalid type.  We may not need the try/except in reality, if we are properly catching the None cases elsewhere.  On the other hand, if we are always returning errors on None, then that could be moved in here (but handled explicitly).

mulby:
@brianhw this is ready for "real" review now

edx/analytics/tasks/video.py
mulby, line 142: Agh, I forgot to revert this, I'll do so now

edx/analytics/tasks/tests/test_video.py
```diff
@@ -250,13 +244,59 @@ def test_unicode_code(self):
...
+    @data(
+        ('play_video', 922337203685.4775),
+        ('pause_video', 922337203685.4775),
+        ('stop_video', 922337203685.4775),
+        ('play_video', 'lskdjflskdj'),
+        ('pause_video', 'lskdjflskdj'),
+        ('stop_video', 'lskdjflskdj'),
+        ('play_video', None),
+        ('pause_video', None),
+        ('stop_video', None),
```
brianhw, line 258: These test for "currentTime": null in the payload.  I don't know if that is attested, but it's certainly in the seek events (below).  But should we also have a way of testing when currentTime is not present at all?

edx/analytics/tasks/tests/test_video.py
```diff
@@ -286,14 +327,14 @@ def test_simple_viewing(self):
...
-            Columns.VIDEO_MODULE_ID: self.VIDEO_MODULE_ID,
-            Columns.VIDEO_DURATION: -1,
-            Columns.START_TIMESTAMP: '2013-12-17T00:00:00+00:00',
-            Columns.START_OFFSET: 0,
-            Columns.END_OFFSET: 3,
-            Columns.REASON: 'pause_video'
+            ViewingColumns.USERNAME: self.USERNAME,
+            ViewingColumns.COURSE_ID: self.COURSE_ID,
+            ViewingColumns.VIDEO_MODULE_ID: self.VIDEO_MODULE_ID,
+            ViewingColumns.VIDEO_DURATION: -1,
```
brianhw, line 374: Use the constant instead of -1.

edx/analytics/tasks/tests/test_video.py
```diff
@@ -521,16 +564,380 @@ def test_pause_after_end_of_video(self):
...
+        self.prepare_youtube_api_mock_raw(response_string)
+        inputs = [
+            ('2013-12-17T00:00:00.00000Z', 'play_video', 0, None, '9bZkp7q19f0'),
+            ('2013-12-17T00:00:03.00000Z', 'pause_video', 3, None, None),
+        ]
+        self._check_output(inputs, {
+            ViewingColumns.START_TIMESTAMP: '2013-12-17T00:00:00+00:00',
+            ViewingColumns.START_OFFSET: 0,
+            ViewingColumns.END_OFFSET: 3,
+            ViewingColumns.VIDEO_DURATION: -1,
```
brianhw: Use the constant down here too....

edx/analytics/tasks/tests/test_video.py
```diff
@@ -521,16 +564,380 @@ def test_pause_after_end_of_video(self):
...
+        ])
+
+    def test_view_counts_without_duration(self):
+        inputs = [
+            # These three viewings are in the first segment
+            ('foo', 0, 1, -1),
+            ('foo2', 1.5, 2, -1),
+            ('foo', 4, 4.99, -1),
+
+            # These viewings are in neither the fist nor the last segment
```
brianhw: nit: s/fist/first

edx/analytics/tasks/tests/test_video.py
```diff
@@ -521,16 +564,380 @@ def test_pause_after_end_of_video(self):
...
+        self._check_output(inputs, [
+            {
+                UsageColumns.VIDEO_DURATION: 50,
+                UsageColumns.END_VIEWS: 0,
+            }
+        ])
+
+    def test_an_unknown_durations(self):
+        inputs = [
+            ('foo', 0, 1, -1),
```
brianhw: Is this intended to be unknown, as in using the flag value, or is this a 'bad' value?  If the former, it should use the constant so that's clear.

edx/analytics/tasks/tests/test_video.py
```diff
@@ -521,16 +564,380 @@ def test_pause_after_end_of_video(self):
...
+    def test_end_view_with_duration(self):
+        inputs = [
+            ('foo', 6, 8, 9.2),
+        ]
+        self._check_output(inputs, [
+            {
+                UsageColumns.VIDEO_DURATION: 9,
+                UsageColumns.END_VIEWS: 1,
+            }
+        ])
```
brianhw: nit: add newline.

edx/analytics/tasks/video.py
brianhw, line 142: This code is not optimal -- we are double-warning, once for the specific problem and once to say that it's None.  I think we are requiring all these values, so we could move the warning about None into this and do it just the once.  And just take out the try/except because we would no longer get TypeErrors.   But I'm also open to keeping the ValueError check, to make the code more robust, even if it's not attested.

edx/analytics/tasks/video.py
```diff
@@ -135,17 +135,19 @@ def mapper(self, line):
...
-                time_value = float(time_value)
-            except TypeError:
-                log.warn('Video event with invalid time-offset type: {0}'.format(line))
-                return None
-
-            # Some events have ridiculous (and dangerous) values for time.
-            if time_value > VIDEO_MAXIMUM_DURATION:
-                log.warn('Video event with huge time-offset value: {0}'.format(line))
-                return None
+        if time_value is None:
```
brianhw, line 134: This is where I'm thinking we should warn, instead of in the individual calls, so that we don't double-warn on all the cases below this.  Moving the warning from the caller down into here means losing information about the particular event type, but that could be passed in with the line, if needed.

brianhw:
Tests look good.  Further changes to video.py than what is here could be moved to another PR.

brianhw:
Squash now or later.  :+1:

edx/analytics/tasks/tests/test_video.py
mulby, line 374: Done

edx/analytics/tasks/tests/test_video.py
mulby, line 258: Done

edx/analytics/tasks/tests/test_video.py
mulby: Done

edx/analytics/tasks/tests/test_video.py
mulby: Done

edx/analytics/tasks/tests/test_video.py
mulby: Done

edx/analytics/tasks/tests/test_video.py
mulby: Done

mulby:
@brianhw I revised the _check_time_offset function again, pushed the warning down into it and removed the TypeError catch, I don't think this will double warn now

PR #105, opened 2015-05-20
Define s3util command-line tool.

brianhw:
@mulby @jab5569 Finally creating a PR for this.


johnalbaker:
:+1:

mulby:
nit: I would prefer this was moved to the launchers package where we put the other main entry points

other than that :+1:

brianhw:
Rather than figuring out a different place to put utilities, I'm fine for now with putting them in launchers.

mulby:
k, seems reasonable to me. :+1:

PR #106, opened 2015-05-20
Redefine start_views and end_views.

brianhw:
Change start_views and end_views to be the number of users
who watched the start and the end of the video, respectively.
Before this change, these represented the number of total views
of the video in those segments.

@mulby @jab5569


brianhw:
Acceptance tests pass, and new data_api and dashboards changes are reported to work with data generated from this branch.

Note that I also removed partial_views, since it was not being used by the API.

johnalbaker:
:+1:  LGTM

mulby:
:+1:

PR #111, opened 2015-06-04
Implement cybersource download.

brianhw:
Main code is feature complete.  Still needs tests, but would like review of approach.

@mulby @jab5569

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -4,19 +4,28 @@ PyYAML==3.10		# MIT License
 ansible==1.4.5		# GPL v3 License
 argparse==1.2.1 	# Python Software Foundation License
 boto==2.22.1 		# MIT
+cffi==1.1.0             # MIT
 ciso8601==1.0.1         # MIT
+cryptography==0.9       # BSD or Apache 2.0
 ecdsa==0.11 		# MIT
+enum34==1.0.4           # BSD
 filechunkio==1.5	# MIT
 html5lib==1.0b3 	# MIT
 httplib2==0.9 		# MIT
+idna==2.0               # BSD-like
 isoweek==1.3.0		# BSD
 mechanize==0.2.5	# BSD
 mysql-connector-python==1.2.2  	# GPL v2 with FOSS License Exception
+ndg-httpsclient==0.4.0  # BSD
 numpy==1.8.0 		# BSD
 pandas==0.13.0 		# BSD
 paramiko==1.14.0 	# LGPL
+paypalrestsdk==1.9.0    # Paypal SDK License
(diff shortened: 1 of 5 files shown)


johnalbaker:
:+1:  Very nice!

edx/analytics/tasks/reports/payments.py
```diff
@@ -0,0 +1,197 @@
...
+            'warehouse_path': self.warehouse_path,
+            'overwrite': self.overwrite,
+        }
+
+        current_date = start_date
+        while current_date < end_date:
+            args['run_date'] = current_date
+            task = DailyProcessFromCybersourceTask(**args)
+            # To cut down on the number of tasks that Luigi has to deal with,
+            # screen out here the tasks that are already complete.
```
mulby: Is this necessary? I thought luigi could handle large volumes of tasks?

requirements/default.txt
```diff
@@ -4,19 +4,28 @@ PyYAML==3.10		# MIT License
...
 httplib2==0.9 		# MIT
+idna==2.0               # BSD-like
 isoweek==1.3.0		# BSD
 mechanize==0.2.5	# BSD
 mysql-connector-python==1.2.2  	# GPL v2 with FOSS License Exception
+ndg-httpsclient==0.4.0  # BSD
 numpy==1.8.0 		# BSD
 pandas==0.13.0 		# BSD
 paramiko==1.14.0 	# LGPL
+paypalrestsdk==1.9.0    # Paypal SDK License
```
mulby, line 23: Can we use this? I'm not sure what the terms of the Paypal SDK license are...

mulby:
:+1: (comments are optional / don't need to be addressed yet)

mulby:
As I'm working on the paypal capture, I'm wondering if we shouldn't name this file "cybersource.py" instead of the more generic "payments.py"

mulby:
Also - I tend to prefer having an explicit output_root for low level tasks like this and only use the WarehouseMixin for tasks that are explicitly dealing with Hive tables etc. At that point they _need_ to put the data in the warehouse, but tasks upstream from there don't actually need to put it anywhere, so I tend to plumb the warehouse path into the output_root of those other tasks. It also means the lower level tasks don't need to know about how Hive organizes the data (at least the folder structure).

requirements/default.txt
brianhw, line 23: I had assumed that eCommerce folks had vetted it.  But I was wrong.  Pending...

edx/analytics/tasks/reports/payments.py
mulby: ping

PR #112, opened 2015-06-12
Added order import tasks for financial report.

johnalbaker:
Additions include importing orders from the LMS and Oscar. The tables for each database were recommended by Clinton. 

@brianhw @mulby 

@brianhw you did indeed review the shopping cart tables before, but I decided to include them in this PR as I made changes to a lot of that code (eg, DECIMAL field types, etc.). To be safe I've included them for re-review.

### DIFF
--- a/edx/analytics/tasks/database_imports.py
+++ b/edx/analytics/tasks/database_imports.py
@@ -43,8 +43,11 @@ class DatabaseImportMixin(object):
     credentials = luigi.Parameter(
         default_from_config={'section': 'database-import', 'name': 'credentials'}
     )
-    import_date = luigi.DateParameter(default=None)
+    database = luigi.Parameter(
+        default_from_config={'section': 'database-import', 'name': 'database'}
+    )
 
+    import_date = luigi.DateParameter(default=None)
     num_mappers = luigi.Parameter(default=None, significant=False)
     verbose = luigi.BooleanParameter(default=False, significant=False)
 
@@ -190,6 +193,7 @@ class ImportMysqlToHiveTableTask(DatabaseImportMixin, ImportIntoHiveTableTask):
             num_mappers=self.num_mappers,
             verbose=self.verbose,
             overwrite=self.overwrite,
+            database=self.database,
             # Hive expects NULL to be represented by the string "\N" in the data. You have to pass in "\\N" to sqoop
             # since it uses that string directly in the generated Java code, so "\\N" actually looks like "\N" to the
             # Java code. In order to get "\\N" onto the command line we have to use another set of escapes to tell the
@@ -250,9 +254,11 @@ class ImportAuthUserTask(ImportMysqlToHiveTableTask):
 
 
 class ImportAuthUserProfileTask(ImportMysqlToHiveTableTask):
+    """
+    Imports user demographic information from an external LMS DB to both a
+    destination directory and a HIVE metastore.
 
-    """Imports user demographic information from an external LMS DB to a destination directory."""
-
+    """
     @property
     def table_name(self):
         return 'auth_userprofile'
@@ -268,8 +274,11 @@ class ImportAuthUserProfileTask(ImportMysqlToHiveTableTask):
 
 
 class ImportCourseUserGroupTask(ImportMysqlToHiveTableTask):
-    """Imports course cohort information from an external LMS DB to a destination directory."""
+    """
+    Imports course cohort information from an external LMS DB to both a
+    destination directory and a HIVE metastore.
 
+    """
     @property
     def table_name(self):
         return 'course_groups_courseusergroup'
@@ -285,8 +294,11 @@ class ImportCourseUserGroupTask(ImportMysqlToHiveTableTask):
 
 
 class ImportCourseUserGroupUsersTask(ImportMysqlToHiveTableTask):
-    """Imports user cohort information from an external LMS DB to a destination directory."""
+    """
+    Imports user cohort information from an external LMS DB to both a
+    destination directory and a HIVE metastore.
 
+    """
     @property
     def table_name(self):
         return 'course_groups_courseusergroup_users'
@@ -300,8 +312,11 @@ class ImportCourseUserGroupUsersTask(ImportMysqlToHiveTableTask):
 
 
 class ImportShoppingCartOrder(ImportMysqlToHiveTableTask):
-    """Imports orders from an external LMS DB shopping cart table to a destination directory."""
+    """
+    Imports orders from an external LMS DB shopping cart table to both a
+    destination directory and a HIVE metastore.
 
+    """
     @property
     def table_name(self):
         return 'shoppingcart_order'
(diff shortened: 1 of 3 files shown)


client.cfg
```diff
@@ -31,4 +31,4 @@ threshold = 500
 input_format = oddjob.ManifestTextInputFormat
 
 [answer-distribution]
-valid_response_types = choiceresponse,optionresponse,multiplechoiceresponse,numericalresponse,stringresponse,formularesponse
+valid_response_types = choiceresponse,optionresponse,multiplechoiceresponse,numericalresponse,stringresponse,formularesponse
```
brianhw: should probably not check in this change....

edx/analytics/tasks/database_imports.py
```diff
@@ -299,8 +303,579 @@ def columns(self):
...
+    def table_name(self):
+        return 'catalogue_productattributevalue'
+
+    @property
+    def columns(self):
+        return [
+            ('id', 'INT'),
+            ('value_text', 'STRING'),
+            ('value_integer', 'INT'),
+            ('value_boolean', 'INT'),
```
brianhw: Does Hive have a boolean?

edx/analytics/tasks/database_imports.py
```diff
@@ -299,8 +303,579 @@ def columns(self):
...
+            ('line_price_incl_tax', 'DECIMAL'),
+            ('line_price_excl_tax', 'DECIMAL'),
+            ('line_price_before_discounts_incl_tax', 'DECIMAL'),
+            ('line_price_before_discounts_excl_tax', 'DECIMAL'),
+            ('unit_cost_price', 'DECIMAL'),
+            ('unit_price_incl_tax', 'DECIMAL'),
+            ('unit_price_excl_tax', 'DECIMAL'),
+            ('unit_retail_price', 'DECIMAL'),
+            ('status', 'STRING'),
+            ('est_dispatch_date', 'TIMESTAMP    '),
```
brianhw: nit: remove extra whitespace in 'TIMESTAMP'

edx/analytics/tasks/database_imports.py
```diff
@@ -299,8 +303,579 @@ def columns(self):
...
+        return [
+            ('id', 'INT'),
+            ('status', 'STRING'),
+            ('date_created', 'TIMESTAMP'),
+            ('date_merged', 'TIMESTAMP'),
+            ('date_submitted', 'TIMESTAMP'),
+            ('owner_id', 'INT'),
+        ]
+
+class ImportCurrentRefundRefundState(ImportMysqlToHiveTableTask):
```
brianhw, line 701: nit: Is it enough to call this CurrentRefundState?

edx/analytics/tasks/database_imports.py
```diff
@@ -299,8 +303,579 @@ def columns(self):
         ]
 
 
+class ImportShoppingCartOrder(ImportMysqlToHiveTableTask):
+    """Imports orders from an external LMS DB shopping cart table to a destination directory."""
```
brianhw: nit:  this not only writes such data to a destination directory, but loads it into a table in the current cluster's Hive metastore.   That way one can actually make queries in Hive against the table.  (Goes for all the other tables that use ImportMysqlToHiveTableTask.)

edx/analytics/tasks/database_imports.py
```diff
@@ -299,8 +303,579 @@ def columns(self):
...
+    @property
+    def table_name(self):
+        return 'order_lineprice'
+
+    @property
+    def columns(self):
+        return [
+            ('id', 'INT'),
+            ('quantity', 'INT'),
+            ('price_incl_tax', 'STRING'),
```
brianhw: Others have been DECIMAL.  Are these different?

edx/analytics/tasks/reports/orders_import.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def output(self):
+        return [task.output() for task in self.requires()]
+
+
+class PullFromEcommerceTablesTask(DatabaseImportMixin, OverwriteOutputMixin, luigi.WrapperTask):
+    """Imports a set of ecommerce tables from an external database into a destination directory."""
+
+    destination = luigi.Parameter(
+        default_from_config={'section': 'oscar-database-import', 'name': 'destination'}
```
brianhw: nit:  'oscar' or 'otto'?  (I thought we were going with 'otto'.)

edx/analytics/tasks/database_imports.py
```diff
@@ -299,8 +303,579 @@ def columns(self):
...
+            ('total_credit_excl_tax', 'DECIMAL'),
+            ('status', 'STRING'),
+            ('order_id', 'INT'),
+            ('user_id', 'INT'),
+            ('created', 'TIMESTAMP'),
+            ('currency', 'STRING'),
+            ('modified', 'TIMESTAMP'),
+        ]
+
+class ImportCurrentRefundRefundLineState(ImportMysqlToHiveTableTask):
```
brianhw, line 725: nit: same idea.  The corresponding order doesn't bother to duplicate the OrderOrder.

brianhw:
I'm done.  A few questions about types, mostly.

client.cfg
johnalbaker: I added the newline back.

edx/analytics/tasks/database_imports.py
johnalbaker: It does ('boolean'). Changed.

edx/analytics/tasks/database_imports.py
johnalbaker: fixed.

edx/analytics/tasks/database_imports.py
johnalbaker: Good catch. I changed the comment in all the other methods as well.

edx/analytics/tasks/database_imports.py
johnalbaker: Eagle eye. 

Changed to 'decimal'.

edx/analytics/tasks/database_imports.py
johnalbaker, line 725: Got it. I just wanted to keep it consistent with the table name.

johnalbaker:
@mulby when you have a chance can you please review?

mulby:
Yeah, I'm actively reviewing by running it ;), here's what I have so far:

https://github.com/edx/edx-analytics-pipeline/compare/jab/financial-reporting...gabe/table-imports

johnalbaker:
Ah, okiedoke, thanks!

On Mon, Jun 15, 2015 at 4:00 PM, Gabe Mulley notifications@github.com
wrote:

> Yeah, I'm actively reviewing by running it ;), here's what I have so far:
> 
> jab/financial-reporting...gabe/table-imports
> https://github.com/edx/edx-analytics-pipeline/compare/jab/financial-reporting...gabe/table-imports
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/edx/edx-analytics-pipeline/pull/112#issuecomment-112190458
> .

## 

John Baker
edX | Analytics Architect
141 Portland Street, 8th Floor
Cambridge, MA 02139

john@edx.org mcoyne@edx.org
http://www.edx.org

mulby:
it may make sense to just merge your change, though, and then I can layer my stuff onto it once it's on the branch

mulby:
I actually feel pretty strongly that we should be using "otto" and not "oscar". Using oscar is like using "django" in our code, it refers to the framework, not our application which uses the framework (otto).

brianhw:
Once the "oscar" => "otto" is updated, this LGTM :+1: .  It would be nice to have the names of the tasks more consistently derive from the tables they're importing, but it's less important to me in this case since they're likely to only be called by the workflows defined here.

PR #115, opened 2015-06-17
Restore cybersource task.

brianhw:
The daily cybersource pull failed because the task became undefined.  This restores the task.

@mulby @jab5569


mulby:
:+1:

I forgot this was running every night... I fixed this on my paypal branch as well

PR #117, opened 2015-06-19
Add reconciliation of orders and transactions.

brianhw:
Reconciles cybersource transactions against generic orderitems.

This is a milestone, producing output that has been validated against the 6/1/15 audit results.


brianhw:
This is not the final version of the reconciliation logic, but it may unblock us if we can move it into the feature branch.  Future milestones include:  adding paypal (generic transactions), adding handling for multiple orderitems and multiple transactions per order, and integrating into a larger workflow. 

@mulby @jab5569

PR #118, opened 2015-06-21
Fix lock contention on the MySQL marker table

mulby:
Fixes: https://openedx.atlassian.net/browse/AN-5429

Note this is a fix for the release branch, I'll make a separate PR for fixing this on master.

@brianhw @jab5569

### DIFF
--- a/edx/analytics/tasks/mysql_load.py
+++ b/edx/analytics/tasks/mysql_load.py
@@ -311,6 +311,10 @@ class MysqlInsertTask(MysqlInsertTaskMixin, luigi.Task):
             # create table only if necessary:
             self.create_table(connection)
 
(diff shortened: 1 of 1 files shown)


johnalbaker:
@mulby  I'm assuming you'll be submitting a separate PR to have each job insert into the marker table..?

Otherwise :+1:  LGTM.

mulby:
@jab5569 all jobs already make use of the marker table, so no further changes should be necessary

brianhw:
Makes sense.  LGTM. :+1:

brianhw:
And it's a small and reasonable change to submit back to Luigi at some point.  Though I recall we have some issue with upgrading Luigi anyway because newer versions won't run on EMR.

PR #119, opened 2015-06-22
Download paypal transactions

mulby:
@brianhw @jab5569

### DIFF
--- a/edx/analytics/tasks/reports/cybersource.py
+++ b/edx/analytics/tasks/reports/cybersource.py
@@ -102,6 +102,12 @@ class DailyPullFromCybersourceTask(PullFromCybersourceTaskMixin, luigi.Task):
         return url
 
 
+TRANSACTION_TYPE_MAP = {
+    'ics_bill': 'sale',
+    'ics_credit': 'refund'
+}
+
+
 class DailyProcessFromCybersourceTask(PullFromCybersourceTaskMixin, luigi.Task):
     """
     A task that reads a local file generated from a daily Cybersource pull, and writes to a TSV file.
@@ -143,12 +149,12 @@ class DailyProcessFromCybersourceTask(PullFromCybersourceTaskMixin, luigi.Task):
                     #   batch_id: CyberSource batch in which the transaction was sent.
                     #   payment_processor: code for organization that processes the payment.
                     result = [
+                        # Date
+                        row['batch_date'],
                         # Name of system.
                         'cybersource',
                         # CyberSource merchant ID used for the transaction.
                         row['merchant_id'],
-                        # Date when the batch was sent to the processor.
-                        row['batch_date'],
                         # Merchant-generated order reference or tracking number.
                         # For shoppingcart or otto, this should equal order_id,
                         # though sometimes it is basket_id.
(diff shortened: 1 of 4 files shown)


edx/analytics/tasks/reports/paypal.py
```diff
@@ -0,0 +1,183 @@
...
+            for line in input_file:
+                payment_record = json.loads(line)
+                for trans_with_items in payment_record.get('transactions', []):
+                    for transaction in trans_with_items.get('related_resources', []):
+                        trans_type = transaction.keys()[0]
+                        details = transaction[trans_type]
+                        created_date_string = details['create_time'].split('T')[0]
+
+                        if created_date_string < interval_start_date_string or created_date_string >= interval_end_date_string:
+                            continue
```
brianhw, line 124: Do we expect there to be records before the interval_start_date_string?  If so (i.e. we're hoping to do something incremental with this), I'm concerned that we will miss updates.  We might want to mention this somewhere.

mulby, line 124: Sadly we will likely have to rewrite this entire file if/when we do incremental pulls of this data.

edx/analytics/tasks/reports/paypal.py
```diff
@@ -0,0 +1,183 @@
...
+                            # currency
+                            details['amount']['currency'],
+                            # amount
+                            details['amount']['total'],
+                            # transaction fee
+                            details.get('transaction_fee', {}).get('value', '\\N'),
+                            # transaction type - either "sale" or "refund"
+                            trans_type,
+                            # payment method - currently this is only "instant_transfer"
+                            'instant_transfer',
```
brianhw: Do we want to hardcode this, or use details.get('payment_mode').lower()?

edx/analytics/tasks/reports/paypal.py
```diff
@@ -0,0 +1,183 @@
...
+
+                        record = [
+                            # date
+                            created_date_string,
+                            # payment gateway
+                            'paypal',
+                            # payment gateway account ID
+                            self.client_mode + '-' + self.client_id,
+                            # payment reference number, this is used to join with orders
+                            trans_with_items['invoice_number'],
```
brianhw: Does this always exist?  Would be safer to use a get().  When I look at my test data, there is no invoice_number anywhere.

johnalbaker:
:+1:  LGTM

edx/analytics/tasks/reports/cybersource.py
```diff
@@ -156,16 +162,15 @@ def run(self):
...
+                        # Transaction fee
+                        '\\N',
+                        TRANSACTION_TYPE_MAP[row['transaction_type']],
+                        # We currently only process credit card transactions with Cybersource
+                        'credit_card',
+                        # Type of credit card used
+                        row['payment_method'].lower().replace(' ', '_'),
                         # Identifier for the transaction.
-                        row['request_id'],
+                        row['trans_ref_no'],
```
brianhw: I was wrong.  I asked Clinton in Rogue Initiative, and they're storing the request_id.

edx/analytics/tasks/reports/paypal.py
brianhw, line 124: Then it might be good to log if any are rejected for being before the start date.

brianhw:
Just fix the request_id again, and LGTM. :+1:

edx/analytics/tasks/reports/paypal.py
mulby, line 124: This is actually expected, since there is is a single file that contains _all_ transactions (note this won't scale).

So there will be loads of transactions both before and after the start and end dates.

PR #121, opened 2015-06-22
Produce a CSV in the specified output format

mulby:
@brianhw @jab5569


brianhw:
This looks fine.  However, the transaction_*_per_item values are what should be used in the report, I assume, not the values from the original transaction.  So there will be some post-merge work to combine the two pull requests.

mulby:
I've rebased this onto the head of financial-reporting

mulby:
@brianhw @jab5569

brianhw:
LGTM. :+1:

PR #122, opened 2015-06-23
Handle multiple orderitems and transactions

brianhw:
Fully coded, but not yet fully validated.  Would like eyes on it, at your convenience.

@mulby @jab5569

### DIFF
--- a/edx/analytics/tasks/reports/reconcile.py
+++ b/edx/analytics/tasks/reports/reconcile.py
@@ -1,10 +1,11 @@
 """Perform reconciliation of transaction history against order history"""
 
-from collections import namedtuple
+from collections import namedtuple, defaultdict
 import csv
 import datetime
 from decimal import Decimal
 import logging
+from operator import attrgetter
 
 import luigi
 import luigi.date_interval
@@ -18,10 +19,12 @@ from edx.analytics.tasks.util.opaque_key_util import get_org_id_for_course
 
 log = logging.getLogger(__name__)
 
+# Precision for money is assumed to be two places.
+TWOPLACES = Decimal(10) ** -2       # same as Decimal('0.01')
 
 ORDERITEM_FIELDS = [
     'order_processor',   # "shoppingcart" or "otto"
-    'user_id',
+    'user_id',  # this is the order system's user_id, not the same as auth_user's user_id.
     'order_id',
     'line_item_id',
     'line_item_product_id',  # for "shoppingcart", this is the kind of orderitem table.
@@ -41,15 +44,21 @@ ORDERITEM_FIELDS = [
     'payment_ref_id',  # This is the value to compare with the transactions.
 ]
 
-OrderItemRecord = namedtuple('OrderItemRecord', ORDERITEM_FIELDS)  # pylint: disable=invalid-name
+BaseOrderItemRecord = namedtuple('OrderItemRecord', ORDERITEM_FIELDS)  # pylint: disable=invalid-name
 
-# These are cybersource-specific at the moment, until generalized
-# for Paypal, etc.
-# Generalization will include:
-#  time = timestamp the transaction was recorded (in addition to the date)
-#  transaction_type: needs to be generalized (cybersource-specific terms now).
-#  transaction_fee:  not reported in cybersource reports.
-#
+
+class OrderItemRecord(BaseOrderItemRecord):
+    """Override vanilla namedtuple to redefine types."""
+    def __new__(cls, *args, **kwargs):
+        result = super(OrderItemRecord, cls).__new__(cls, *args, **kwargs)
+        result = result._replace(  # pylint: disable=no-member,protected-access
+            refunded_amount=Decimal(result.refunded_amount),  # pylint: disable=no-member
+            line_item_price=Decimal(result.line_item_price),  # pylint: disable=no-member
+            line_item_unit_price=Decimal(result.line_item_unit_price),  # pylint: disable=no-member
+        )
+        return result
+
+
 TRANSACTION_FIELDS = [
     'date',
     'payment_gateway_id',
@@ -64,7 +73,20 @@ TRANSACTION_FIELDS = [
     'transaction_id',
 ]
 
-TransactionRecord = namedtuple('TransactionRecord', TRANSACTION_FIELDS)  # pylint: disable=invalid-name
+
+BaseTransactionRecord = namedtuple('TransactionRecord', TRANSACTION_FIELDS)  # pylint: disable=invalid-name
+
+
+class TransactionRecord(BaseTransactionRecord):
+    """Override vanilla namedtuple to redefine types."""
+    def __new__(cls, *args, **kwargs):
+        result = super(TransactionRecord, cls).__new__(cls, *args, **kwargs)
+        result = result._replace(  # pylint: disable=no-member,protected-access
+            amount=Decimal(result.amount),  # pylint: disable=no-member
+            transaction_fee=Decimal(result.transaction_fee) if result.transaction_fee is not None else None,  # pylint: disable=no-member
+        )
+        return result
+
 
 LOW_ORDER_ID_SHOPPINGCART_ORDERS = (
     '1556',
@@ -76,6 +98,7 @@ LOW_ORDER_ID_SHOPPINGCART_ORDERS = (
 
 
 class ReconcileOrdersAndTransactionsDownstreamMixin(MapReduceJobTaskMixin):
+    """Define parameters needed downstream for running ReconcileOrdersAndTransactionsTask."""
 
     transaction_source = luigi.Parameter(
         default_from_config={'section': 'payment-reconciliation', 'name': 'transaction_source'}
@@ -125,16 +148,26 @@ class ReconcileOrdersAndTransactionsTask(ReconcileOrdersAndTransactionsDownstrea
 
     def mapper(self, line):
         fields = line.split('\t')
-        # If we put the "payment_ref_id" in the front of all these fields, or
-        # at least always in the same index, then we wouldn't this
-        # ugly heuristic here.  (It would only need to be in the
-        # reducer. :)
-        if len(fields) > 11:
-            # assume it's an order
+        if len(fields) == len(ORDERITEM_FIELDS):
+            # Assume it's an order.
+            record_type = OrderItemRecord.__name__
             key = fields[-1]
-        else:
-            # assume it's a transaction
+            # Convert nulls in 'product_detail', 'refunded_amount', 'refunded_quantity'.
+            if fields[10] == '\\N':
+                fields[10] = ''
+            if fields[16] == '\\N':
+                fields[16] = '0.0'
+            if fields[17] == '\\N':
+                fields[17] = '0'
+
+        elif len(fields) == len(TRANSACTION_FIELDS):
+            # Assume it's a transaction.
+            record_type = TransactionRecord.__name__
             key = fields[3]
+            # Convert nulls in 'transaction_fee'.
+            if fields[6] == '\\N':
+                fields[6] = None
+
             # Edx-only: if the transaction was within a time period when
             # Otto was storing basket-id values instead of payment_ref_ids in
             # its transactions, then apply a heuristic to the transactions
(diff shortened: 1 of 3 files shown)


brianhw:
This is ready for review.  

@mulby @jab5569

mulby:
Meta note - I'm going to make comments here that extend beyond the changes here to include changes I'd like to see before we merge the financial-reporting branch back into master since I'm hoping we won't have to revisit this code before then.

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -129,10 +130,10 @@ def mapper(self, line):
         # at least always in the same index, then we wouldn't this
         # ugly heuristic here.  (It would only need to be in the
         # reducer. :)
-        if len(fields) > 11:
+        if len(fields) == len(ORDERITEM_FIELDS):
```
mulby, line 151: I think I would prefer we use `os.getenv('map_input_file')` to determine the expected format since that will tell us which file the record came from.

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -143,44 +144,36 @@ def mapper(self, line):
             if fields[0] > '2015-05-01' and fields[0] < '2015-06-14':
                 if len(key) <= 4 and key not in LOW_ORDER_ID_SHOPPINGCART_ORDERS:
                     key = 'EDX-{}'.format(int(key) + 100000)
+        else:
+            raise ValueError("ERROR: unrecognized line with {} fields:  {}".format(len(fields), line))
 
         yield key, fields
```
mulby: Maybe we could emit a field here that says whether the record is a payment or an order explicitly? So that the reducer doesn't have to infer.

Also I would prefer we handle the NULL values in the mapper since it's more distributed - I tend to assume all of the data prep/cleaning that can happen in the mapper will happen in the mapper.

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -196,123 +189,291 @@ def reducer(self, _key, values):
...
+            trans_code = 'PURCHASE_SOMETHING' if Decimal(transaction.amount) > 0.0 else 'REFUND_SOMETHING'
+            code = (order_code, 'NO_ORDERITEM', trans_code)
+            yield code, transaction
+
+    def reducer(self, _key, values):
+        """Convert orderitems and transactions into orderitem-transaction records."""
+        orderitems, transactions = self._extract_transactions(values)
+
+        # Check to see that all orderitems belong to the same order.
+        order_ids = set([orderitem.order_id for orderitem in orderitems])
```
mulby: nit: s/order_ids/distinct_order_ids/

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -196,123 +189,291 @@ def reducer(self, _key, values):
...
+        if len(order_ids) > 1:
+            orderitem_ids = [orderitem.line_item_id for orderitem in orderitems]
+            raise Exception("ERROR: orderitems {} encountered with different order_ids {}".format(orderitem_ids, order_ids))
+
+        # Calculate common values.
+        trans_balance = Decimal(0.0)
+        if len(transactions) > 0:
+            trans_balance = sum([Decimal(transaction.amount) for transaction in transactions])
```
mulby: I wonder if we should override the `__new__` function of the namedtuple to coerce the various Decimal typed data into actual Decimal objects so that we aren't casting as frequently.

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -196,123 +189,291 @@ def reducer(self, _key, values):
...
+
+        There are also some professional-ed entries that have
+        transactions in a different account.
+
+        """
+        for orderitem in orderitems:
+            order_status = 'ERROR_ORDER_NOT_BALANCED'
+            orderitem_status = 'ERROR_NO_TRANSACTION'
+            transaction_status = 'NO_TRANSACTION'
+            if self._orderitem_is_professional_ed(orderitem):
```
mulby: Maybe rename this method to something like `_orderitem_is_white_label` since that's a bit more accurate.

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -196,123 +189,291 @@ def reducer(self, _key, values):
...
+                orderitem_status = 'NO_TRANS_PROFESSIONAL'
```
mulby: Similarly this should probably be `NO_TRANS_WHITE_LABEL`

mulby: I'm thinking that we should strive to ensure that only statuses that start with "ERROR_" require tickets to be created, so perhaps change the order status here to simply `ORDER_NOT_BALANCED`

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -196,123 +189,291 @@ def reducer(self, _key, values):
...
+        order pull, or if a shoppingcart basket was charged that was
+        not marked as a purchased order.  In the latter case, if the
+        charge was later refunded and the current balance is zero,
+        then no further action is needed.  Otherwise either the order
+        needs to be updated (to reflect that they did actually receive
+        what they ordered), or the balance should be refunded (because
+        they never received what they were charged for).
+
+        """
+        order_code = 'NO_ORDER_ZERO_BALANCE' if trans_balance == 0 else 'ERROR_NO_ORDER_NONZERO_BALANCE'
```
mulby: nit: we should be consistently naming these audit status code variables either `foo_status` or `foo_code`

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -196,123 +189,291 @@ def reducer(self, _key, values):
...
+        for transaction in transactions:
+            trans_code = 'PURCHASE_SOMETHING' if Decimal(transaction.amount) > 0.0 else 'REFUND_SOMETHING'
```
mulby: Maybe simply `PURCHASE` and `REFUND`.

mulby: Also, IMHO this should be `>= 0.0` since a $0 transaction is a purchase (you've given them something for free)

Side note: not sure it's possible to have $0 purchases...

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -196,123 +189,291 @@ def reducer(self, _key, values):
...
+        transaction_amount = Decimal(transaction.amount)
+        for orderitem in orderitems:
+            if self._get_purchase_status(orderitem, transaction_amount, orderitem_purchases) == status:
+                orderitem_purchases[orderitem].append((transaction, transaction_amount, status))
+                return orderitem
+        return None
+
+    def _map_purchases_to_orderitems(self, orderitems, purchases, orderitem_purchases):
+        """Distributes purchase transactions onto orderitems, splitting as necessary."""
+        sorted_purchases = sorted(purchases, key=lambda x: x.date)
```
mulby: nit: I prefer `attrgetter`

``` python
from operator import attrgetter
sorted_purchases = sorted(purchases, key=attrgettr('date'))
```

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -196,123 +189,291 @@ def reducer(self, _key, values):
...
+    def _get_orderitem_to_purchase(self, status, transaction, orderitems, orderitem_purchases):
+        """Loop through orderitems, and 'purchase' the first one with a matching status."""
+        transaction_amount = Decimal(transaction.amount)
+        for orderitem in orderitems:
+            if self._get_purchase_status(orderitem, transaction_amount, orderitem_purchases) == status:
+                orderitem_purchases[orderitem].append((transaction, transaction_amount, status))
+                return orderitem
+        return None
+
+    def _map_purchases_to_orderitems(self, orderitems, purchases, orderitem_purchases):
```
mulby: nit: for clarity I would prefer we rename `purchases` to `purchase_transactions`

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -196,123 +189,291 @@ def reducer(self, _key, values):
...
+        sorted_purchases = sorted(purchases, key=lambda x: x.date)
+        order_cost = sum([Decimal(orderitem.line_item_price) for orderitem in orderitems])
+
+        for transaction in sorted_purchases:
+            transaction_amount = Decimal(transaction.amount)
+            if transaction_amount == order_cost:
+                # transaction purchases all orderitems
+                for orderitem in orderitems:
+                    orderitem_cost = Decimal(orderitem.line_item_price)
+                    status = 'PURCHASE_AGAIN' if orderitem in orderitem_purchases else 'PURCHASE_ONE'
```
mulby: Isn't `PURCHASE_AGAIN` an error state? Won't this overwrite the data already in orderitem_purchases? Does the original transaction data get "left behind" here?

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -196,123 +189,291 @@ def reducer(self, _key, values):
...
-                    yield self.format_transaction_table_output(code, transaction, orderitem, item_amount)
+                status = 'PURCHASE_MISCHARGE_AGAIN'
+        else:
+            if transaction_amount == orderitem_cost:
+                status = 'PURCHASE_ONE'
+            else:
+                status = 'PURCHASE_MISCHARGE'
+        return status
+
+    def _get_orderitem_to_purchase(self, status, transaction, orderitems, orderitem_purchases):
```
mulby: It was surprising to me to see that this method was actually modifying the `orderitem_purchases` dictionary since its name starts with `_get`. That signaled to me that it was just gathering some data and returning it. I would suggest renaming this method pushing the side effect out to the caller.

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -196,123 +189,291 @@ def reducer(self, _key, values):
...
+                # transaction purchases all orderitems
+                for orderitem in orderitems:
+                    orderitem_cost = Decimal(orderitem.line_item_price)
+                    status = 'PURCHASE_AGAIN' if orderitem in orderitem_purchases else 'PURCHASE_ONE'
+                    orderitem_purchases[orderitem].append((transaction, orderitem_cost, status))
 
+            else:
+                purchase_sequence = ['PURCHASE_ONE', 'PURCHASE_MISCHARGE', 'PURCHASE_AGAIN']
+                found = False
+                for status in purchase_sequence:
```
mulby: This logic was a little difficult to follow.

edx/analytics/tasks/reports/reconcile.py
brianhw: we're finding the transaction_audit_code for the current transaction, by looking to see if that orderitem has already been purchased.  I'm not labeling _any_ of the transaction_audit_code values with any ERROR_ values.  It could be done, but I think they are informative enough in the shorter form.  If it is a second purchase of the orderitem, then there are two transactions in the list associated with the orderitem, the original purchase, and this second purchase.  (Or third purchase, etc., were that to come up.)

edx/analytics/tasks/reports/reconcile.py
brianhw: done

edx/analytics/tasks/reports/reconcile.py
brianhw: done

edx/analytics/tasks/reports/reconcile.py
brianhw: done

edx/analytics/tasks/reports/reconcile.py
brianhw: I decided to be very explicit, so I went with foo_audit_code.   Orderitems have a status (purchased vs. refunded), and I thought that foo_code was not clear enough.  I thought foo_audit_code would do a better job of indicating that the value was assigned by the audit as opposed to being an intrinsic property of the foo.

edx/analytics/tasks/reports/reconcile.py
brianhw: I'm not in agreement here with the general assertion.  I think that ERROR_ should indicate which conditions are unexpected going forward.  The fact that we currently make some choices to ignore some of these errors as being too noisy should not be embedded in the logic creating the codes.

That said, I'm willing to create a third order_status value, "ORDER_NOT_BALANCED", to cover this particular case.  (And changed to white label....)

edx/analytics/tasks/reports/reconcile.py
brianhw: done

edx/analytics/tasks/reports/reconcile.py
brianhw: done

edx/analytics/tasks/reports/reconcile.py
brianhw, line 151: I would prefer not to be applying string tests on filenames -- that seems even more indirect and brittle.

edx/analytics/tasks/reports/reconcile.py
brianhw: done

edx/analytics/tasks/reports/reconcile.py
mulby, line 151: We should think about a more robust way to solve this problem, I agree that the filename is not a great idea either. Perhaps we could leverage the fact that we have a handle to the upstream task in `requires()` (or we will eventually). The upstream task could "tell us" where it stored its data and allow us to do more robust checking of source.

For example:

``` python
for upstream_task in self.requires():
    if url_get_dir(os.getenv('map_input_file')) == upstream_task.output().path:
        record_format_class = upstream_task.output_format  # this could be a namedtuple
        break

record = record_format_class.from_string(line)
yield record.transaction_mapping_key, record.to_tuple()
```

edx/analytics/tasks/reports/reconcile.py
brianhw: Rather than doing this in the **new**, it could be done instead in the mapper, and it would just mean the Decimal values would go through the repr/eval.  Not sure which way is better offhand, but I'm out of time to do this now.

brianhw:
@mulby I responded to most of your comments, and punted on two items, one deliberately.   Feel free to squash and merge, if it's acceptable enough.

edx/analytics/tasks/reports/reconcile.py
brianhw: Yeah, that was stupid.  done.

edx/analytics/tasks/reports/reconcile.py
brianhw: Renamed some, and added some brief comments.

brianhw:
@mulby @jab5569 I think I have responded to all comments.

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -143,186 +176,398 @@ def mapper(self, line):
...
+                total_fees = Decimal(0.0)
                 for index, orderitem in enumerate(orderitems):
-                    code = self._get_code_for_nonmatch(orderitem, trans_balance)
-                    # We need to come up with a value, which is complicated by the
-                    # presence of a refund or a mismatch in order value and transaction.
-                    # Arbitrarily put all the value into one of the order items.
-                    item_amount = trans_balance if index == 1 else Decimal(0.0)
-                    yield self.format_transaction_table_output(code, transaction, orderitem, item_amount)
-
+                    transaction_audit_code = 'PURCHASE_AGAIN' if orderitem in orderitem_purchases else 'PURCHASE_ONE'
```
mulby: Is the ordering of order items gauranteed here? If not, this could potentially assign the "PURCHASE_ONE" and "PURCHASE_AGAIN" codes differently when re-running the algorithm.

mulby: (not a blocker)

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -143,186 +176,398 @@ def mapper(self, line):
...
+        for orderitem in orderitems:
+            purchase_audit_code = self._get_purchase_audit_code(orderitem, transaction.amount, orderitem_purchases)
+            if purchase_audit_code == target_transaction_audit_code:
+                return orderitem
+        return None
+
+    def _map_purchases_to_orderitems(self, orderitems, purchase_transactions):
+        """Distributes purchase transactions onto orderitems, splitting if necessary.
+
+        Returns a dict, with orderitems as keys.  Each orderitem's value is a list
```
mulby, line 430: Side note (not a blocker): I've been leaning towards using the type system to help communicate information about return types and data structures like this one. A named, typed, collection can let us define an interface around it and provide a lot more information about what it is, why those things are grouped together etc. It also enables us to easily track where things are used, how they are passed around and enforce different levels of mutability for the internal parts of the object.

That's a very abstract way of saying that I think we should be returning an object to represent this mapping here instead of a dict. It also may make sense to make order items and transactions mutable objects, that way you could be enriching the object as you go instead of passing around mappings.

We can take this conversation offline too and discuss philosophy at some point ;)

@jab5569 FYI

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -143,186 +176,398 @@ def mapper(self, line):
...
+                    # Calculate transaction_fee_per_item if there is a transaction fee.
+                    transaction_fee_per_item = None
+                    if transaction.transaction_fee is not None:
+                        if index < (len(orderitems) - 1):
+                            proportion = orderitem.line_item_price / transaction.amount
+                            transaction_fee_per_item = (transaction.transaction_fee * proportion).quantize(TWOPLACES)
+                            total_fees += transaction_fee_per_item
+                        else:
+                            # If it's the last orderitem, make sure we don't have roundoff error.
+                            transaction_fee_per_item = (transaction.transaction_fee - total_fees).quantize(TWOPLACES)
```
mulby: Is it possible for the quantization to drop fractions of pennies?

brianhw: I don't believe so.  That is, the two values being subtracted here should both already be quantized to two places, so there should be no further error.

edx/analytics/tasks/reports/reconcile.py
brianhw: A good point, and an easy thing to fix.  That is, the original set of orderitems should be sorted by date and line_item_id.

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -143,186 +176,398 @@ def mapper(self, line):
...
+                        break
+
+                if not found:
+                    # We have a payment that doesn't align with one or all orderitems,
+                    # so just purchase the first orderitem.
+                    # It could be for two out of three orderitems, for example, but that
+                    # should be rarer.  More likely is an overpayment or underpayment, and
+                    # that is a problem that needs to be flagged if it hasn't already
+                    # been addressed.
+                    transaction_audit_code = 'PURCHASE_RANDOMLY'
```
mulby: nit: I would prefer we use a different word than "RANDOMLY" here since it implies some kind of random occurrence, when it is likely not random but rather incorrect. Maybe it shouldn't be differentiated from MISCHARGE?

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -143,186 +176,398 @@ def mapper(self, line):
...
+        """Distributes refund transactions onto orderitems, splitting if necessary.
+
+        Returns a dict, with orderitems as keys.  Each orderitem's value is a list
+        of tuples.  Each tuple in turn contains four objects: the transaction, the
+        value of the transaction that goes towards the refund of the orderitem,
+        the audit_code that annotates the transaction's "refund", and the amount of
+        a transaction_fee going to the orderitem.
+        """
+        orderitem_refunds = defaultdict(list)
+        # sorted_refund_transactions = sorted(refund_transactions, key=attrgetter('date'))
```
mulby: (non-blocker) did this not work? or do you just prefer the lambda style?

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -143,186 +176,398 @@ def mapper(self, line):
...
+                    transaction_fee_per_item = None
+                    if transaction.transaction_fee is not None:
+                        if index < (len(orderitems) - 1):
+                            proportion = orderitem_cost / transaction.amount
+                            transaction_fee_per_item = (transaction.transaction_fee * proportion).quantize(TWOPLACES)
+                            total_fees += transaction_fee_per_item
+                        else:
+                            # If it's the last orderitem, make sure we don't have roundoff error.
+                            transaction_fee_per_item = (transaction.transaction_fee - total_fees).quantize(TWOPLACES)
+                    new_value = (transaction, orderitem_cost, transaction_audit_code, transaction_fee_per_item)
```
mulby: nit: this transaction handling code looks very similar to the code used for purchases, maybe refactor into a function?

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -143,186 +176,398 @@ def mapper(self, line):
...
+                    new_value = (transaction, transaction.amount, transaction_audit_code, transaction_fee)
+                    orderitem_purchases[orderitems[0]].append(new_value)
+
+        return orderitem_purchases
+
+    def _get_refund_audit_code(self, orderitem, orderitem_purchases, orderitem_refunds):
+        """Return a transaction_audit_code string depending on whether the item is already purchased or refunded."""
+        if orderitem in orderitem_purchases:
+            if orderitem in orderitem_refunds:
+                transaction_audit_code = 'REFUND_AGAIN' if orderitem.status == 'refunded' else 'REFUND_UNREFUNDED_AGAIN'
```
mulby: nit: Maybe change "UNREFUNDED" to "PURCHASED"? "UNREFUNDED" feels like a double-negative...

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -143,186 +176,398 @@ def mapper(self, line):
...
+            else:
+                transaction_audit_code = 'REFUND_ONE' if orderitem.status == 'refunded' else 'REFUND_UNREFUNDED'
         else:
-            # for now, just get some transactions into the file, independent of the
-            # orderitems involved.
-            code = "MULTIPLE_ORDERITEMS"
-            for transaction in transactions:
-                yield self.format_transaction_table_output(code, transaction, orderitems[0])
+            transaction_audit_code = 'REFUND_UNPURCHASED'
```
mulby: nit: maybe "NEVER_PURCHASED" instead of "UNPURCHASED"?

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -143,186 +176,398 @@ def mapper(self, line):
...
+            if transaction_amount_per_item is None:
+                transaction_amount_per_item = transaction.amount
+            if transaction_fee_per_item is None and transaction.transaction_fee is not None:
                 transaction_fee_per_item = str(transaction.transaction_fee)
-            else:
-                proportion = Decimal(transaction_amount_per_item) / Decimal(transaction.amount)
-                transaction_fee_per_item = str(Decimal(transaction.transaction_fee) * proportion)
+
+        # proportion = transaction_amount_per_item / transaction.amount
+        # transaction_fee_per_item = str((transaction.transaction_fee * proportion).quantize(TWOPLACES))
```
mulby: Remove commented out code here

edx/analytics/tasks/reports/tests/test_reconcile.py
```diff
@@ -0,0 +1,804 @@
...
+        for ref_id in LOW_ORDER_ID_SHOPPINGCART_ORDERS:
+            trans = self.create_transaction(payment_ref_id=ref_id)
+            line = self._convert_record_to_line(trans)
+            expected_key = ref_id
+            expected_value = ('TransactionRecord', self._convert_record_to_expected_output(trans))
+            self.assert_single_map_output(line, expected_key, expected_value)
+
+
+@ddt
+class ReconciliationTaskReducerTest(ReconciliationTaskMixin, ReducerTestMixin, unittest.TestCase):
```
mulby, line 179: (non-blocker): consider adding a test that randomizes the ordering of the input to the reducer a couple of times and ensures that it produces the same audit codes

edx/analytics/tasks/reports/reconcile.py
brianhw: Don't like "PURCHASED", as most refunds are done on things that were purchased.  This is trying to indicate a problem with the status.  Do you prefer "NOTREFUNDED" over "UNREFUNDED".  (Not sure about the double-negative -- it's a single negative.)

mulby: It is a single negative - but it _feels_ like a double negative - the purchase feels like the base, negating that is the refund, and negating the refund is what?

I guess I'm confused about what these statuses mean.... re-reading this it looks like the following is true here:
1. We are looking at a refund transaction
2. There is a corresponding purchase for this item
3. There is yet another corresponding refund for this item

Basically the item has been double-refunded. However, depending on the status of the item in the database we might mark it "UNREFUNDED"... it's not clear to me what that means... the item has been double refunded but is marked "purchased" in the database. I may be misreading this...

And more importantly - how can we structure this so that it is more obvious what this status represents?

edx/analytics/tasks/reports/reconcile.py
brianhw: Done.

edx/analytics/tasks/reports/reconcile.py
brianhw: More accurate would be to call it "PURCHASE_FIRST", since that's what is actually being done.

edx/analytics/tasks/reports/reconcile.py
brianhw: oops, done.

edx/analytics/tasks/reports/reconcile.py
brianhw: For transactions, the goal of the code is to indicate what decision was made, and what context was relevant.  Purchases begin with "PURCHASE_" and refunds with "REFUND_".   The modifiers that follow try to tell some of the "story" about why the decision was made to assign that transaction to that orderitem.  This includes taking into account whether an orderitem was marked as refunded.  For example, if two orderitems are purchased and one is refunded, we assign the refund to the orderitem that was marked as refunded.  That is why I distinguish between REFUND_UNREFUNDED and REFUND_ONE, and between REFUND_UNREFUNDED_AGAIN and REFUND_AGAIN.    There are a small enough number of transaction codes, so perhaps they just need to be enumerated and documented in docstrings.

mulby: I guess I still think "UNREFUNDED" meaning "marked purchased in the database" is not very clear

Do we need to elevate the visibility of this bit of information? won't this be clear from the order item status?

Could we just say "REFUND_ONE" and "REFUND_AGAIN" like we do for purchases without taking into account the order item status?

mulby:
:+1:

edx/analytics/tasks/reports/reconcile.py
brianhw: Done.  I found the cleanest thing to lift out was the entire loop over orderitems, because of the total_fees logic.

edx/analytics/tasks/reports/reconcile.py
brianhw: Done.

edx/analytics/tasks/reports/reconcile.py
brianhw: Done.

edx/analytics/tasks/reports/reconcile.py
brianhw: We need the information in the internal algorithm, because we make different decisions based on the order item status.  We could then strip that back out at the point that we output them, if it were really confusing.  That depends somewhat on who the audience really ends up being.  One reason for leaving it this way is that it helps with debugging _why_ the mapping was done the way it was done -- I would prefer therefore not to throw that information away.

mulby: I guess my concern is that it's more confusing than helpful. Maybe if we found a different word than "UNREFUNDED" it would tip the balance the other way for me?

edx/analytics/tasks/reports/reconcile.py
brianhw: Funny that I found an SO entry where the question was asked which was better, and the two anecdotal responses were both for the lambda style.  But whatever.  No, this was a remnant of some testing that I missed in cleaning up.  I put it back, and added a second id field to disambiguate entries with the same date.

edx/analytics/tasks/reports/tests/test_reconcile.py
brianhw, line 179: Done.

edx/analytics/tasks/reports/reconcile.py
brianhw: REFUND_STATUSNOTREFUNDED?

brianhw: Okay.  I went with:
- REFUND_ONE:  a regular refund of an item for its cost, with a previous purchase and no previous refunds. 
- REFUND_AGAIN:  a refund of an item for its cost, with a previous purchase and with a previous refund already made.
- REFUND_ONE_STATUS_NOT_REFUNDED:  a regular refund of an item for its cost, with a previous purchase and no previous refunds, but with the status of the orderitem not reflecting that a refund transaction was expected.
- REFUND_AGAIN_STATUS_NOT_REFUNDED:  a refund of an item for its cost, with a previous purchase and with a previous refund already made, but with the status of the orderitem not reflecting that a refund transaction was expected.
- REFUND_NEVER_PURCHASED:  for any refund of an item for which no purchase was made.

edx/analytics/tasks/reports/reconcile.py
brianhw, line 430: It's a slippery slope as to when to make things into objects.  Certainly coming from Java before Python, my original inclination was to make everything objects, but that's not necessarily "pythonic", even when passing things outside of the file in which it's defined.  (This is more internal data.)  

There are two "objects" here.  The first is the orderitem-transaction pairing itself, represented here by the 4-tuple.   The second is the mapping of these entries to the orderitems, represented by the dict.   The combination represents the mapping.  Moving the mapping into the orderitems and the transactions themselves is an interesting idea, in that one could add a list of orderitems to each transaction, as well as add the list of 4-tuples to the orderitem.  But it doesn't seem to simplify much:  there are still complications that won't go away by changing this representation.  (E.g. which orderitems in an order didn't get transactions.)   But it would be a good conversation to have.

PR #123, opened 2015-06-24
Define task to load reconciliation table into Hive.

brianhw:
Changes include:
- Provides separate parameters for order_source and transaction_source, with separate defaults.
- Gets date partition info from interval parameter, and uses this for selecting a particular order_source.
- Cleanup includes removal of ReconciliationOutputTask, and output category.

Goal is to provide a task that can be called without requiring any arguments.  (But that requires defaults defined in separate PR.)

@mulby @jab5569


mulby:
:+1:

PR #124, opened 2015-06-24
Fix paypal

mulby:
Paypal is throwing 503s if I try to request too big of a date range. So instead this code will make a separate request per day. It takes a while...

@brianhw @jab5569

### DIFF
--- a/edx/analytics/tasks/reports/paypal.py
+++ b/edx/analytics/tasks/reports/paypal.py
@@ -4,6 +4,8 @@ import luigi
 import paypalrestsdk
 import json
 
+from luigi.date_interval import DateInterval
+
 from edx.analytics.tasks.url import get_target_from_url
 from edx.analytics.tasks.url import url_path_join
 from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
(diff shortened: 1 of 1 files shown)


edx/analytics/tasks/reports/paypal.py
```diff
@@ -179,15 +183,15 @@ def run(self):
...
-            start_date=self.start_date,
-            client_mode=self.client_mode,
-            client_id=self.client_id,
-            client_secret=self.client_secret,
-            run_date=self.run_date,
-        )
+        for run_date in DateInterval(self.start_date, self.interval.date_b):
+            yield RawPaypalTransactionLogTask(
+                output_root=self.output_root,
+                start_date=self.start_date,
```
brianhw, line 189: nit: I'm assuming we no longer need to pass start_date down, if we're using run_date instead.

johnalbaker:
Same question @brianhw had about start_date vs run_date. Otherwise :+1:  LGTM

edx/analytics/tasks/reports/paypal.py
```diff
@@ -179,15 +183,15 @@ def run(self):
             output_file.write('OK')
 
     def requires(self):
-        yield RawPaypalTransactionLogTask(
-            output_root=self.output_root,
-            overwrite=self.overwrite,
```
brianhw, line 184: nit: Are we deliberately removing overwrite because we don't think it should be possible, or because we've not yet implemented it?

brianhw:
Two optional nits.  LGTM. :+1:

edx/analytics/tasks/reports/paypal.py
mulby, line 184: Worse than that... we _have_ to overwrite all of the raw input files every time we run, since they will change (new refunds added to JSON blobs).

PR #125, opened 2015-06-24
Jamesrowan/course subject integration

jrowan:
Adds pipeline functionality to pull the course subjects information from drupal, store that information on s3, and then load that information into the Vertica data warehouse.  The Vertica luigi target class and Vertica bulkloading luigi task class, modeled off the corresponding mysql tasks, should be usable for other data warehouse loading tasks as well.

@jab5569 @brianhw @mulby


client.cfg
```diff
@@ -1,34 +1,32 @@
...
+destination = s3://<bucket for intermediate hadoop products>/intermediate/database-import
+credentials = s3://<secrets bucket>/edxapp_prod_ro_mysql_creds
 
 [event-logs]
 expand_interval = 2 days
+pattern = .*tracking.log-(?P<date>[0-9]+).*
+source = s3://<bucket to where all tracking logs are synched>/tracking/
+
+[hive]
+warehouse_path = http://s3.amazonaws.com/edx-analytics-scratch/jrowan/hive
```
brianhw: I would respecfully suggest that this be obfuscated like the other S3 paths. :)

edx/analytics/tasks/course_catalog.py
```diff
@@ -0,0 +1,173 @@
+"""Collect the course catalog from Drupal for processing of course metadata like subjects or types"""
```
brianhw: I would strongly suggest that we not use "Drupal" in this file, but rather use the "Course Catalog API".

edx/analytics/tasks/course_catalog.py
```diff
@@ -0,0 +1,173 @@
...
+import json
+import luigi
+
+from edx.analytics.tasks.url import get_target_from_url
+from edx.analytics.tasks.url import url_path_join
+from edx.analytics.tasks.util.hive import WarehouseMixin, HivePartition
+from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
+from edx.analytics.tasks.util.vertica_load import VerticaCopyTask, VerticaCopyTaskMixin
+
+import urllib3.contrib.pyopenssl
```
brianhw, line 16: In the cybersource and paypal code, we added the following comment:

> #Tell urllib3 to switch the ssl backend to PyOpenSSL.
> # see https://urllib3.readthedocs.org/en/latest/security.html#pyopenssl

edx/analytics/tasks/course_catalog.py
```diff
@@ -0,0 +1,173 @@
...
+
+class DailyPullCatalogFromDrupalTask(PullFromDrupalMixin, luigi.Task):
+    """
+    A task that reads the course catalog out of Drupal and writes the result json blob to a file.
+
+
+    Pulls are made daily to keep a full historical record.
+    """
+
+    CATALOG_URL = 'https://www.edx.org/api/catalog/v2/courses'
```
brianhw: This should be a parameter that is pulled from the configuration file -- at least the root of it.  OpenEdx installations should not be pulling from edx.org.

edx/analytics/tasks/course_catalog.py
```diff
@@ -0,0 +1,173 @@
...
+    CATALOG_FORMAT = 'json'
+    run_date = luigi.DateParameter(default=datetime.date.today())
```
brianhw: This probably doesn't need to be redefined here if it's already defined in PullFromDrupalMixin and pulled in from there.

edx/analytics/tasks/course_catalog.py
```diff
@@ -0,0 +1,173 @@
...
+from edx.analytics.tasks.util.vertica_load import VerticaCopyTask, VerticaCopyTaskMixin
+
+import urllib3.contrib.pyopenssl
+urllib3.contrib.pyopenssl.inject_into_urllib3()
+
+
+class PullFromDrupalMixin(OverwriteOutputMixin, WarehouseMixin):
+    """Define common parameters for Drupal course catalog pull and downstream tasks."""
+
+    run_date = luigi.DateParameter(default=datetime.date.today())
```
brianhw: A small but sometimes significant detail:  today() returns the date in local time.  This should probably be datetime.datetime.utcnow().date() for general UTC consistency.

edx/analytics/tasks/course_catalog.py
```diff
@@ -0,0 +1,173 @@
...
+            msg = "Encountered status {} on request to Drupal for {}".format(response.status_code, self.run_date)
+            raise Exception(msg)
+        with self.output().open('w') as output_file:
+            output_file.write(response.content)
+
+    def output(self):
+        """Output is in the form {warehouse_path}/catalog/{CCYY-mm-dd}/catalog.json"""
+        date_string = self.run_date.strftime('%Y%m%d')
+        filename = "catalog.{}".format(self.CATALOG_FORMAT)
+        url_with_filename = url_path_join(self.warehouse_path, "catalog", date_string, filename)
```
brianhw: I would suggest a longer name than just "catalog" here.  Things are getting pretty clogged at root level of warehouse_path.  We might want to add more hierarchy at some point, but I'd lengthen this to "course_catalog_api" or something similar.

client.cfg
```diff
@@ -1,34 +1,32 @@
-[hadoop]
```
mulby, line 1: You should revert all changes to this file. In your local environment you can just create an "override.cfg" file in the root of the repo that will be used by luigi.

This file actually contains the defaults for some variables used by luigi.

edx/analytics/tasks/course_catalog.py
```diff
@@ -0,0 +1,173 @@
...
+        response = get_request(self.CATALOG_URL)
+        if response.status_code != request_codes.ok:  # pylint: disable=no-member
+            msg = "Encountered status {} on request to Drupal for {}".format(response.status_code, self.run_date)
+            raise Exception(msg)
+        with self.output().open('w') as output_file:
+            output_file.write(response.content)
+
+    def output(self):
+        """Output is in the form {warehouse_path}/catalog/{CCYY-mm-dd}/catalog.json"""
+        date_string = self.run_date.strftime('%Y%m%d')
```
mulby: We typically use the ISO8601 date format:

%Y-%m-%d

Or more conveniently: `self.run_date.isoformat()`

mulby: Also we follow the hive partition convention for separating snapshots of things like this by prefixing the date with "dt=". If/when we ever use this as a hive table it will prefer the folder hierarchy to be in this format.

edx/analytics/tasks/course_catalog.py
```diff
@@ -0,0 +1,173 @@
...
+        filename = "catalog.{}".format(self.CATALOG_FORMAT)
```
mulby: I'm OK with harcoding ".json" here, I think the constant is unnecessary unless it's actually being passed to the API or something in a "Accept" header or something similar.

edx/analytics/tasks/course_catalog.py
```diff
@@ -0,0 +1,173 @@
...
+        url_with_filename = url_path_join(self.warehouse_path, "catalog", date_string, filename)
+        return get_target_from_url(url_with_filename)
+
+
+class DailyProcessFromCatalogSubjectTask(PullFromDrupalMixin, luigi.Task):
+    """
+    A task that reads a local file generated from a daily catalog pull, and writes the course id and subject to a csv.
```
mulby: Update this to indicate that you are generating a tsv instead of a csv

edx/analytics/tasks/course_catalog.py
```diff
@@ -0,0 +1,173 @@
...
+                for course in catalog:
+                    course_id = course['course_id']
+                    # This will be a list of dictionaries with keys 'title', 'uri', and 'language'.
+                    # Encode in utf-8 as in general the subjects could be given in other languages.
+                    subjects = course.get('subjects', None)
+                    # It's possible no subjects are given for the course, in which case we record the lack of subjects.
+                    if subjects is None or len(subjects) == 0:
+                        line = [
+                            course_id.encode('utf-8'),
+                            self.run_date,
```
mulby: I would be explicit here about the string format you want for the date.

edx/analytics/tasks/course_catalog.py
```diff
@@ -0,0 +1,173 @@
...
+                            '\N'.encode('utf-8'),  # pylint: disable-msg=W1402
```
mulby: you may want to move the encoding down into a list comprehension, something like:

``` python
output_file.write('\t'.join([v.encode('utf-8') for v in line]))
```

It's a little more DRY.

edx/analytics/tasks/course_catalog.py
```diff
@@ -0,0 +1,173 @@
...
+            # Since the course catalog is of fairly manageable size, we can read it all into memory at once.
+            # If this needs to change, we should be able to parse the catalog course by course.
+            catalog = json.loads(input_file.read())['items']
+            with self.output().open('w') as output_file:
+                for course in catalog:
+                    course_id = course['course_id']
+                    # This will be a list of dictionaries with keys 'title', 'uri', and 'language'.
+                    # Encode in utf-8 as in general the subjects could be given in other languages.
+                    subjects = course.get('subjects', None)
+                    # It's possible no subjects are given for the course, in which case we record the lack of subjects.
```
mulby, line 93: Do we need to record this lack of subjects?

When joining with this table we can just use a left outer join and get null values.

edx/analytics/tasks/course_catalog.py
```diff
@@ -0,0 +1,173 @@
...
+
+    run_date = luigi.DateParameter(default=datetime.date.today())
+
+    @property
+    def insert_source_task(self):
+        return(DailyProcessFromCatalogSubjectTask(run_date=self.run_date))
+
+    @property
+    def table(self):
+        return "experimental.d_course_subjects"
```
mulby: Can we push the schema name into the config file and/or a parameter? Similarly to how we define the hive database we are writing to and the LMS database we are reading from?

edx/analytics/tasks/course_catalog.py
```diff
@@ -0,0 +1,173 @@
...
+
+    @property
+    def auto_primary_key(self):
```
mulby, line 141: Consider not making these properties required if they are commonly going to just be `None` (same for default columns).

edx/analytics/tasks/util/vertica_load.py
```diff
@@ -0,0 +1,275 @@
+"""
```
mulby: I thought we were going to commit this to the edx/luigi fork?

https://github.com/edx/luigi

mulby: Oh, wrong file, I meant to put this comment near the vertica target.

edx/analytics/tasks/course_catalog.py
mulby, line 141: Ah, I see you are overriding the default behavior... why do you want to do that here? Maybe explain in a comment?

edx/analytics/tasks/util/vertica_load.py
```diff
@@ -0,0 +1,275 @@
...
+        Override to provide code for creating the target schema, if not existing.
+
+        By default it will be created using types (optionally) specified in columns.
+
+        If overridden, use the provided connection object for setting
+        up the table in order to create the table and insert data
+        using the same transaction.
+        """
+        # In Vertica, tables are specified as schema_name.table_name, so just the first part of the table is needed.
+        schema = self.table.split('.')[0]
```
mulby: why not `self.database`?

edx/analytics/tasks/util/vertica_load.py
```diff
@@ -0,0 +1,275 @@
...
+
+    def update_id(self):
+        """This update id will be a unique identifier for this insert on this table."""
+        # For MySQL tasks, we take the hash of the task id, but since Vertica does not similarly
+        # limit the size of columns, we can safely use the entire task ID.
+        return str(self)
+
+    def output(self):
+        """
+        Returns a MysqlTarget representing the inserted dataset.
```
mulby: s/Mysql/Vertica/

edx/analytics/tasks/util/vertica_load.py
```diff
@@ -0,0 +1,275 @@
...
+        in the database to keep a rolling window of data available in
+        the table.
+        """
+        # clear table contents
+        self.attempted_removal = True
+        if self.overwrite:
+            # first clear the appropriate rows from the luigi Vertica marker table
+            marker_table = self.output().marker_table  # side-effect: sets self.output_target if it's None
+            try:
+                query = "DELETE FROM {marker_table} where `target_table`='{target_table}'".format(
```
mulby: Not sure what the vertica row locking behavior is, but we have to be very careful to only lock the row we are deleting here. Mysql was locking the whole table which was preventing multiple jobs from being able to write into it at once.

edx/analytics/tasks/util/vertica_load.py
```diff
@@ -0,0 +1,275 @@
...
+                    pass
+                else:
+                    raise
+
+            # Use "DELETE" instead of TRUNCATE since TRUNCATE forces an implicit commit before it executes which would
+            # commit the currently open transaction before continuing with the copy.
+            query = "DELETE FROM {table}".format(table=self.table)
+            connection.cursor().execute(query)
+
+    def copy_data_table_from_target(self, cursor, delimiter="E'\t'", null_sequence="'\\N'"):
```
mulby: This is very cool. We should make the delimeter and null_sequence easier to override, though, maybe as properties that a subclass can override?

I can imagine wanting to read a variety of different file formats that have different conventions for these strings.

mulby:
Finished my pass. Great job.

Blockers for me:
- revert client.cfg
- make use of the `database` parameter to specify the schema instead of using "schema.table_name"

edx/analytics/tasks/util/vertica_load.py
jrowan: @mulby I'm changing this to using self.schema and removing self.database.  I'm not using self.database to refer to the schema since the pyODBC/vertica-python end of the target thinks database refers to a specific other concept related to making connections.

jrowan:
Unit and acceptance tests added, client.cfg reverted, and the schema to use is a configurable parameter now.  

@mulby @brianhw @jab5569

edx/analytics/tasks/course_catalog.py
```diff
@@ -0,0 +1,202 @@
...
+    """
+
+    def requires(self):
+        pass
+
+    def run(self):
+        date_string = "dt=" + self.run_date.strftime('%Y-%m-%d')  # pylint: disable=no-member
+        url_with_filename = url_path_join(self.warehouse_path, "course_catalog_api/catalog",
+                                          date_string, "catalog.json")
+        print "URL_WITH_FILENAME: ", str(url_with_filename)
```
mulby: Remove debugging print statements (throughout this PR)

edx/analytics/tasks/course_catalog.py
```diff
@@ -0,0 +1,202 @@
...
+        if response.status_code != requests.codes.ok:  # pylint: disable=no-member
+            msg = "Encountered status {} on request to API for {}".format(response.status_code, self.run_date)
+            raise Exception(msg)
+        with self.output().open('w') as output_file:
+            output_file.write(response.content)
+
+    def output(self):
+        """Output is in the form {warehouse_path}/course_catalog_api/catalog/{CCYY-mm-dd}/catalog.json"""
+        date_string = "dt=" + self.run_date.strftime('%Y-%m-%d')  # pylint: disable=no-member
+        url_with_filename = url_path_join(self.warehouse_path, "course_catalog_api/catalog",
```
mulby: I would prefer this let `url_path_join` handle all of the separators here

edx/analytics/tasks/course_catalog.py
```diff
@@ -0,0 +1,202 @@
...
+urllib3.contrib.pyopenssl.inject_into_urllib3()
+# Since we block-encode the offending strings anyways, ignore complaints about unicode escapes in '\N' appearances.
+# pylint: disable-msg=W1402
+
+
+class PullFromCatalogAPIMixin(OverwriteOutputMixin, WarehouseMixin):
+    """Define common parameters for the course catalog API pull and downstream tasks."""
+
+    run_date = luigi.DateParameter(default=datetime.datetime.utcnow().date())
+    catalog_url = luigi.Parameter(default_from_config={'section': 'course-catalog', 'name': 'catalog_path'})
```
mulby: I would prefer the variable name and the config file parameter name were the same

edx/analytics/tasks/course_catalog.py
```diff
@@ -0,0 +1,202 @@
...
+        print "CATALOG: ", self.catalog_url
+        response = requests.get(self.catalog_url)
+        if response.status_code != requests.codes.ok:  # pylint: disable=no-member
+            msg = "Encountered status {} on request to API for {}".format(response.status_code, self.run_date)
+            raise Exception(msg)
+        with self.output().open('w') as output_file:
+            output_file.write(response.content)
+
+    def output(self):
+        """Output is in the form {warehouse_path}/course_catalog_api/catalog/{CCYY-mm-dd}/catalog.json"""
```
mulby: This comment looks like it's out of date

edx/analytics/tasks/course_catalog.py
```diff
@@ -0,0 +1,202 @@
...
+
+    Pulls are made daily to keep a full historical record.
+    """
+
+    def requires(self):
+        pass
+
+    def run(self):
+        date_string = "dt=" + self.run_date.strftime('%Y-%m-%d')  # pylint: disable=no-member
+        url_with_filename = url_path_join(self.warehouse_path, "course_catalog_api/catalog",
```
mulby: I think I would prefer simply "course_catalog" instead of "course_catalog_api" since the fact that it is pulled from an API is an implementation detail... in the future we might pull it directly from a database (for example).

edx/analytics/tasks/course_catalog.py
```diff
@@ -0,0 +1,202 @@
...
+        }
+        return DailyPullCatalogFromCatalogAPITask(**kwargs)
+
+    def run(self):
+        # Read the catalog and select just the subjects data for output
+        self.remove_output_on_overwrite()
+        with self.input().open('r') as input_file:
+            # Since the course catalog is of fairly manageable size, we can read it all into memory at once.
+            # If this needs to change, we should be able to parse the catalog course by course.
+            catalog = json.loads(input_file.read()).get('items', None)
```
mulby: `None` is the default return value from `dict.get(key)`, so you do not need to explicitly specify it here.

jrowan:
I performed some modifications to the bulkloader based on infrastructure from the events-to-vertica task

jrowan:
Note that there are updates to the Vertica bulkloading infrastructure that I still need to push here, as we are now using a newer version of vertica-python (forcing function on that was the event loading).

edx/analytics/tasks/course_catalog.py
```diff
@@ -0,0 +1,198 @@
...
+    def output(self):
+        """
+        Output is set up so that it can be read as a Hive table with partitions,
+
+        The form is {warehouse_path}/course_catalog_api/subjects/dt={CCYY-mm-dd}/subjects.tsv.
+        """
+        date_string = self.run_date.strftime('%Y-%m-%d')  # pylint: disable=no-member
+        partition_path_spec = HivePartition('dt', date_string).path_spec
+        filename = "subjects.tsv"
+        url_with_filename = url_path_join(self.warehouse_path, "course_catalog_api/subjects",
```
mulby: I think I put this in previous comments, but I two notes here:

1) I think we should call the folder simply "course_catalog". I don't think the "_api" suffix provides valuable information
2) I would prefer this was simply: `url_path_join(self.warehouse_path, 'course_catalog', 'subjects', ...)` since it keeps all separator logic in `url_path_join`.

edx/analytics/tasks/course_catalog.py
```diff
@@ -0,0 +1,198 @@
...
+            'catalog_path': self.catalog_path
+        }
+        kwargs2.update(kwargs2)
+
+        yield (
+            DailyLoadSubjectsToVerticaTask(**kwargs2),
+        )
+
+
+class CourseCatalogWorkflowFixed(PullCatalogMixin, VerticaCopyTaskMixin, luigi.WrapperTask):
```
mulby: This looks duplicated?

edx/analytics/tasks/tests/acceptance/__init__.py
```diff
@@ -69,7 +71,9 @@ def setUp(self):
         self.test_src = url_path_join(self.test_root, 'src')
         self.test_out = url_path_join(self.test_root, 'out')
 
+        self.catalog_path = 'http://acceptance.test/api/courses/v2'
         database_name = 'test_' + self.identifier
+        schema = 'test'
```
mulby: This should incorporate `self.identifier` to ensure that multiple tests can be run at the same time, I would recommend changing this to `'test_' + self.identifier`.

mulby: Or you could simply change the constructor call for the `VerticaService` to take in the `database_name` value instead of `schema`.

edx/analytics/tasks/tests/acceptance/services/task.py
```diff
@@ -72,6 +72,8 @@ def delete_existing_logs(self):
 
     def override_config(self, config_parser, overrides):
         for section_name, section in overrides.iteritems():
+            print section_name
+            print section
```
mulby: Remove debug print statements

edx/analytics/tasks/tests/acceptance/services/vertica.py
```diff
@@ -0,0 +1,70 @@
...
+
+from contextlib import closing
+from contextlib import contextmanager
+
+import vertica_python
+
+from edx.analytics.tasks.url import get_target_from_url
+
+
+class VerticaService(object):
```
mulby, line 12: This looks very similar to the DatabaseService. I would recommend creating a common base class that implements the shared logic.

We may be able to just override the `connect()` method, the `reset()` method and credentials url and have the rest be common.

edx/analytics/tasks/tests/acceptance/test_course_catalog.py
```diff
@@ -0,0 +1,89 @@
...
+    """Base class for the end-to-end test of course catalog-based tasks."""
+
+    INPUT_FILE = 'catalog.json'
+
+    def setUp(self):
+        super(BaseCourseCatalogAcceptanceTest, self).setUp()
+
+        assert 'oddjob_jar' in self.config
+
+        self.oddjob_jar = self.config['oddjob_jar']
```
mulby: is this actually used anywhere?

edx/analytics/tasks/tests/acceptance/test_course_catalog.py
```diff
@@ -0,0 +1,89 @@
...
+
+        self.upload_data()
+
+    def upload_data(self):
+        """Puts the test course catalog where the processing task would look for it, bypassing calling the actual ALI"""
+        src = os.path.join(self.data_dir, 'input', self.INPUT_FILE)
+        dst = url_path_join(self.warehouse_path, 'course_catalog_api/catalog/dt=2015-06-29', self.INPUT_FILE)
```
mulby: I would prefer to have `url_path_join` handle the separators here

edx/analytics/tasks/tests/acceptance/test_course_catalog.py
```diff
@@ -0,0 +1,89 @@
...
+
+class CourseSubjectsAcceptanceTest(BaseCourseCatalogAcceptanceTest):
+    """End-to-end test of pulling the course subject data into Vertica."""
+
+    def test_course_subjects(self):
+        """Tests the workflow for the course subjects, end to end."""
+
+        self.task.launch([
+            'CourseCatalogWorkflow',
+            '--credentials', self.vertica.vertica_creds_url,
```
mulby: This shouldn't be necessary given that you overrode the appropriate settings in the dynamically generated config file. Same goes for the `--catalog-url`.

edx/analytics/tasks/tests/acceptance/test_course_catalog.py
```diff
@@ -0,0 +1,89 @@
...
+            '--run-date', '2015-06-29',
+            '--catalog-url', self.catalog_path
+        ])
+
+        self.validate_output()
+
+        # Drop the table afterwards so that the next test starts with a clean table.
```
mulby: This shouldn't be necessary since each test run calls `reset()` at the beginning which drops everything.

edx/analytics/tasks/tests/acceptance/test_course_catalog.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def validate_output(self):
+        """Validates the output, comparing it to a csv of all the expected output from this workflow."""
+        with self.vertica.cursor() as cursor:
+            cursor.execute("SELECT COUNT(*) FROM {schema}.d_course_subjects;"
+                           .format(schema=self.vertica.schema_name))
+            total_count = cursor.fetchone()[0]
+
+            expected_total_count = 6
+            self.assertEqual(total_count, expected_total_count)
```
mulby: I'd prefer this read in the expected data and then called `len()` on it to figure out the expected count. That way someone need only change the expected data file without having to also update this code if they add rows.

edx/analytics/tasks/tests/acceptance/test_course_catalog.py
```diff
@@ -0,0 +1,89 @@
...
+            subjects = cursor.fetchall()
+
+            try:
+                for subject_listing in subjects:
+                    class_number = subject_listing[0]
+                    expected_row = expected.loc[class_number]
+                    for col_num in range(1, 6):
+                        found_datum = subject_listing[col_num]
+                        expected_datum = expected_row[col_num - 1]  # We subtract 1 since we removed the row number.
+                        if col_num == 2:  # Pandas doesn't parse the dates as datetime.date objects, so manually cast.
```
mulby: `read_csv()` takes in a `parse_dates` parameter which I believe can be used to solve this problem.

edx/analytics/tasks/tests/acceptance/test_course_catalog.py
```diff
@@ -0,0 +1,89 @@
...
+
+            # Verify that all the subject data loaded into Vertica is correct.
+            expected_output_csv = os.path.join(self.data_dir, 'output', 'expected_subjects_for_acceptance.csv')
+            expected = pandas.read_csv(expected_output_csv, index_col=0)
+
+            cursor.execute("SELECT * FROM {schema}.d_course_subjects;".format(schema=self.vertica.schema_name))
+            subjects = cursor.fetchall()
+
+            try:
+                for subject_listing in subjects:
```
mulby: I would prefer to modify the ingest mechanism such that we end up with two supposedly equal `DataFrame` objects that we can just assert are equal.

For example, you could explicitly list the columns read in the SQL statement so that you don't have to shift values over etc.

edx/analytics/tasks/tests/test_course_subjects.py
```diff
@@ -0,0 +1,151 @@
...
+
+        Args:
+            data is a pandas data frame representing the output of a run of the task.
+            row_num is the row number to check.
+            expected is a dictionary of values we expect to see.
+        """
+        working = True
+        self.assertGreater(data.shape[0], row_num)
+        row = data.iloc[row_num]
+        for expected_key in expected.keys():
```
mulby: Could we just do:

``` python
self.assertEquals(dict(row), expected)
```

edx/analytics/tasks/tests/test_course_subjects.py
```diff
@@ -0,0 +1,151 @@
...
+                working = False
+        return working
+
+    def test_empty_catalog(self):
+        """With an empty catalog, we expect no rows of data."""
+
+        data = self.run_task("{}")
+        self.assertEquals(data.shape[0], 0)
+
+    def test_course_no_sbujects(self):
```
mulby: nit: typo here -> s/sbujects/subjects/

edx/analytics/tasks/tests/test_course_subjects.py
```diff
@@ -0,0 +1,151 @@
...
+        """With a course with no subjects, we expect a row with NULLs."""
+
+        data = self.run_task("{\"items\":[{\"course_id\":\"foo\", \"subjects\":[{}]}]}")
```
mulby: I'd be OK with using `json.dumps()` and native python dict, list objects here to avoid messy backslash land.

edx/analytics/tasks/tests/test_course_subjects.py
```diff
@@ -0,0 +1,151 @@
...
+        # We expect an entry in the list of courses, since there is a course in the catalog.
+        self.assertEquals(data.shape[0], 1)
+        # We expect nulls for the three subject data columns.
+        expected = {'course_id': 'foo',
+                    'subject_uri': '\N', 'subject_title': '\N', 'subject_language': '\N'}  # pylint: disable-msg=W1402
```
mulby: I typically just use `'\\N'` to avoid the pylint warning. If you decide not to do that, that's OK too, but we typically use the symbolic name when disabling pylint messages since it's a lot more clear to readers what exactly is being disabled, in this case it would be "anomalous-unicode-escape-in-string".

edx/analytics/tasks/tests/test_course_subjects.py
```diff
@@ -0,0 +1,151 @@
...
+import pandas
+from edx.analytics.tasks.course_catalog import DailyProcessFromCatalogSubjectTask
+
+from edx.analytics.tasks.tests import unittest
+from edx.analytics.tasks.tests.target import FakeTarget
+
+from mock import MagicMock
+
+
+class TestCourseSubjects(unittest.TestCase):
```
mulby, line 29: We should have a test for the corner case when the API returns a list instead of the expected dictionary.

Also for the case when the API returns invalid JSON, and/or unexpected keys / missing keys etc.

edx/analytics/tasks/tests/test_course_subjects.py
```diff
@@ -0,0 +1,151 @@
...
+    """Tests for the task parsing the catalog into rows of course ids and subjects"""
+
+    def setUp(self):
+        self.temp_rootdir = tempfile.mkdtemp()
+        self.input_dir = os.path.join(self.temp_rootdir, "input")
+        os.mkdir(self.input_dir)
+        self.input_file = os.path.join(self.input_dir, "catalog", "20150625", "catalog.json")
```
mulby: This date should be 'dt=2015-06-25' for consistency with the rest of the pipeline and potential future Hive compatibility.

edx/analytics/tasks/tests/test_vertica_load.py
```diff
@@ -0,0 +1,226 @@
+"""
+Ensure we can write to MySQL data sources.
```
mulby: nit: s/MySQL/Vertica/

mulby: (throughout this file)

edx/analytics/tasks/tests/test_vertica_load.py
```diff
@@ -0,0 +1,226 @@
...
+        self.create_task(credentials='{}').run()
+
+    @with_luigi_config('vertica-export', 'schema', 'foobar')
+    def test_parameters_from_config(self):
+        task = CopyToVerticaDummyTable(credentials=sentinel.credentials)
+        self.assertEquals(task.schema, 'foobar')
+
+    def test_run(self):
+        self.create_task().run()
+        self.assertTrue(self.mock_vertica_connector.connect().cursor().execute.called)
```
mulby: I will often assign a core mock to a variable to make this type of thing slightly less verbose:

``` python
mock_conn = self.mock_vertica_connector.connect()
self.assertFalse(mock_conn.rollback.called)
```

mulby:
finished my pass, getting closer!

edx/analytics/tasks/tests/acceptance/services/vertica.py
jrowan, line 12: I thought about this, but I think there's a sufficiently breaking difference between MySQL and Vertica to warrant not doing this.  My reasoning:

MySQL and Vertica mean very different things by "database," (MySQL databases are roughly equivalent to Vertica schemata, and the vertica connection module uses database as a kwarg but doesn't actually need it to make a connection), and there is a set of implicit_db logic in the DatabaseService that can't be used in Vertica.  If we were to make a base class, we'd have to override almost all of the logic in both places anyways.

edx/analytics/tasks/tests/acceptance/test_course_catalog.py
jrowan: What are you referring to by "this"? There are 15 lines of code highlighted here.

edx/analytics/tasks/tests/acceptance/test_course_catalog.py
jrowan: Fixed with respect to hardcoding the slashes in 'course_catalog_api/catalog/dt=2015-06-29'; I have os.path.join for the src because those files are local to the machine running the acceptance test.

edx/analytics/tasks/tests/acceptance/test_course_catalog.py
mulby: Sorry - could have been more clear here. Are you using `self.oddjob_jar`?

edx/analytics/tasks/tests/acceptance/services/task.py
johnalbaker: Do we need to remove these print statements?

edx/analytics/tasks/tests/acceptance/services/vertica.py
```diff
@@ -0,0 +1,70 @@
...
+    def __init__(self, config, schema_name):
+        self.vertica_creds_url = config['vertica_creds_url']
+        self.schema_name = schema_name
+
+    @property
+    def credentials(self):
+        """The credentials for connecting to the database, read from a URL."""
+        if not hasattr(self, '_credentials'):
+            target = get_target_from_url(self.vertica_creds_url)
+            print "TARGET BUILT: ", target
```
johnalbaker: Does this need to be removed?

edx/analytics/tasks/tests/test_course_subjects.py
```diff
@@ -0,0 +1,151 @@
+"""
+Test the course subjects task which processes from the course catalog.
+
+Testing strategy:
+    Empty catalog (expect empty output)
+    Catalog with one course listed which has no subjects listed (expect output with null values?)
+    Catalog with one course listed which has one subject
+    Catalog with one course listed which has multiple subjects
+    Catalog with multiple courses listed
+    Catalog that is an actual day's catalog, checking that the size of the produced .tsv is good.
```
johnalbaker: Very nice

edx/analytics/tasks/tests/test_vertica_load.py
johnalbaker: 'Vertica' data source?

edx/analytics/tasks/tests/test_vertica_load.py
```diff
@@ -0,0 +1,226 @@
...
+    """
+
+    @property
+    def columns(self):
+        return ['course_id', 'interval_start', 'interval_end', 'label', 'count']
+
+
+class VerticaCopyTaskTest(unittest.TestCase):
+    """
+    Ensure we can connect to and write data to MySQL data sources.
```
johnalbaker: 'Vertica'

edx/analytics/tasks/tests/test_vertica_load.py
```diff
@@ -0,0 +1,226 @@
...
+    """
+
+    def setUp(self):
+        patcher = patch('edx.analytics.tasks.vertica_load.vertica_python.vertica')
+        self.mock_vertica_connector = patcher.start()
+        self.addCleanup(patcher.stop)
+
+    def create_task(self, credentials=None, source=None, overwrite=False, cls=CopyToVerticaDummyTable):
+        """
+         Emulate execution of a generic MysqlTask.
```
johnalbaker: Vertica

edx/analytics/tasks/user_activity.py
```diff
@@ -116,18 +129,15 @@ def reducer(self, key, values):
...
 
-class UserActivityDownstreamMixin(WarehouseMixin, EventLogSelectionDownstreamMixin, MapReduceJobTaskMixin):
-    """All parameters needed to run the UserActivityTableTask task."""
-    pass
+    def run(self):
+        self.remove_output_on_overwrite()
+        super(UserActivityTask, self).run()
 
 
-class UserActivityTableTask(UserActivityDownstreamMixin, HiveTableTask):
```
johnalbaker, line 130: Why don't we need to instantiate UserActivityDownstreamMixin here?

edx/analytics/tasks/util/overwrite.py
```diff
@@ -51,6 +51,6 @@ def remove_output_on_overwrite(self):
         """
         if self.overwrite:
             self.attempted_removal = True
-            if self.output().exists():
+            if self.output().exists() and hasattr(self.output(), 'remove'):
```
johnalbaker: What does this do?

edx/analytics/tasks/util/vertica_target.py
```diff
@@ -0,0 +1,131 @@
...
+    def connect(self, autocommit=False):
+        """
+        Creates a connection to a Vertica database using the supplied credentials.
+
+        :param autocommit: whether the connection should automatically commit.
+        :type autocmommit: bool
+        """
+
+        # vertica-python 0.5.0 changes the code for connecting to databases to use kwargs instead of a dictionary.
+        # The 'database' parameter is included for DBAPI reasons and does not actually affect the session.
```
johnalbaker, line 103: This differentiation is impressive

edx/analytics/tasks/tests/acceptance/test_course_catalog.py
```diff
@@ -0,0 +1,89 @@
...
+        with self.vertica.cursor() as cursor:
+            cursor.execute("SELECT COUNT(*) FROM {schema}.d_course_subjects;"
+                           .format(schema=self.vertica.schema_name))
+            total_count = cursor.fetchone()[0]
+
+            expected_total_count = 6
+            self.assertEqual(total_count, expected_total_count)
+
+            # Verify that all the subject data loaded into Vertica is correct.
+            expected_output_csv = os.path.join(self.data_dir, 'output', 'expected_subjects_for_acceptance.csv')
```
johnalbaker, line 53: url_path_join() instead?

edx/analytics/tasks/tests/acceptance/test_course_catalog.py
```diff
@@ -0,0 +1,89 @@
...
+
+        assert 'oddjob_jar' in self.config
+
+        self.oddjob_jar = self.config['oddjob_jar']
+
+        self.upload_data()
+
+    def upload_data(self):
+        """Puts the test course catalog where the processing task would look for it, bypassing calling the actual ALI"""
+        src = os.path.join(self.data_dir, 'input', self.INPUT_FILE)
```
johnalbaker, line 29: url_path_join() instead?

edx/analytics/tasks/tests/test_course_subjects.py
johnalbaker: url_path_join() instead?

johnalbaker:
I'll defer to @mulby  for any showstoppers, what I found were mostly nits and questions.

Overall this is great work James; very impressive. 

:+1:

edx/analytics/tasks/tests/test_course_subjects.py
jrowan: I use os.path.join in these cases since the test data is stored locally, in the repo (e.g. in tests/acceptance/fixtures) or we're making temporary directories to store temporary files, and it's not worth putting them on s3.  I use url_path_join for all s3 or internet locations and os.path.join for all local locations, as in general os path separators and url path separators could be different.

jrowan:
@mulby @jab5569 I've gone through and believe I have addressed your concerns.  I've added the additional unit tests Gabe suggested, the streamlined unit and acceptance tests are passing, and I'm getting a mostly\* clean bill of code cleanliness.  Sorry for the massive quantity of commits that arose when I was trying to debug an acceptance test; I will of course rebase before merging.

\* nitty NB: the creation of mock course catalog json dictionaries in test_course_subjects.py leads to situations that, at least according to PyCharm, are "pep8 paradoxes."  There are some lines marked as "continuation line is over-indented for visual indent" that, when I go through space by space trying to fix the indentation level, go immediately from "under-indented" to "over-indented," without every passing through a pep8 acceptable state.

edx/analytics/tasks/tests/acceptance/services/vertica.py
mulby, line 12: Got it, sounds reasonable.

edx/analytics/tasks/tests/acceptance/test_course_catalog.py
```diff
@@ -0,0 +1,69 @@
...
+            expected_total_count = len(expected.index)
+            self.assertEqual(total_count, expected_total_count)
+
+            cursor.execute("SELECT * FROM {schema}.d_course_subjects;".format(schema=self.vertica.schema_name))
+            database_subjects = cursor.fetchall()
+            subjects = pandas.DataFrame(database_subjects, columns=['row_number', 'course_id', 'date', 'subject_uri',
+                                                                    'subject_title', 'subject_language'])
+
+            # The equality check gives a pandas dataframe object by default, so we condense it to a single boolean.
+            self.assertTrue(all(subjects == expected))
```
mulby: Nit: instead of checking the count and then row equality... you should just be able to use:

``` python
from pandas.util.testing import assert_frame_equal
assert_frame_equal(data_frame_a, data_frame_b)
```

edx/analytics/tasks/tests/acceptance/__init__.py
```diff
@@ -94,6 +98,13 @@ def setUp(self):
                 'credentials': self.config['credentials_file_url'],
                 'database': export_database_name
             },
+            'vertica-export': {
+                'credentials': self.config['vertica_creds_url'],
+                'schema': schema
+            },
+            'course-catalog': {
+                'catalog_path': self.catalog_path
```
mulby, line 106: Nit: note that by default the acceptance tests use the `config/test.cfg` file, so you shouldn't have to put in this override if you define this static value in that file (which you do above). These dynamic overrides are only here since there are many values that are not static and have to be overwritten at runtime.

mulby:
Great work @jrowan - let's get this on master!

:+1:

edx/analytics/tasks/course_catalog.py
```diff
@@ -0,0 +1,178 @@
...
+        url_with_filename = url_path_join(self.warehouse_path, "course_catalog", "catalog", date_string,
+                                          "catalog.json")
+        return get_target_from_url(url_with_filename)
+
+
+class DailyProcessFromCatalogSubjectTask(PullCatalogMixin, luigi.Task):
+    """
+    A task that reads a local file generated from a daily catalog pull, and writes the course id and subject to a tsv.
+
+
```
brianhw: nit: two blank lines seems excessive....

edx/analytics/tasks/course_catalog.py
```diff
@@ -0,0 +1,178 @@
...
+            # Since the course catalog is of fairly manageable size, we can read it all into memory at once.
+            # If this needs to change, we should be able to parse the catalog course by course.
+            catalog = json.loads(input_file.read()).get('items')
+            with self.output().open('w') as output_file:
+                if catalog is None:
+                    return
+                for course in catalog:
+                    # The course catalog occasionally is buggy and has malformed courses, so just skip those.
+                    if not type(course) == dict:
+                        continue
```
brianhw, line 86: Should we at least output a warning, so we have a way of determining how often this occurs over time?

edx/analytics/tasks/tests/acceptance/test_course_catalog.py
```diff
@@ -0,0 +1,71 @@
...
+            # self.assertEqual(total_count, expected_total_count)
+
+            cursor.execute("SELECT * FROM {schema}.d_course_subjects;".format(schema=self.vertica.schema_name))
+            database_subjects = cursor.fetchall()
+            subjects = pandas.DataFrame(database_subjects, columns=['row_number', 'course_id', 'date', 'subject_uri',
+                                                                    'subject_title', 'subject_language'])
+
+            # The equality check gives a pandas dataframe object by default, so we condense it to a single boolean.
+            assert_frame_equal(subjects, expected)
+            # self.assertTrue(all(subjects == expected))
```
brianhw: this can be removed now...

edx/analytics/tasks/tests/test_course_subjects.py
```diff
@@ -0,0 +1,239 @@
...
+        task.run()
+        results = pandas.read_table(output_target.buffer, sep='\t', header=None,
+                                    names=['course_id', 'date', 'subject_uri', 'subject_title', 'subject_language'])
+        return results
+
+    def check_subject_entry(self, data, row_num, expected):
+        """
+        Checks if the entries in a row of the data are what we expected, and returns whether this is true.
+
+
```
brianhw: nit: extra blank line here and below.

edx/analytics/tasks/tests/test_course_subjects.py
```diff
@@ -0,0 +1,239 @@
...
+        Doesn't do asserts directly to help other methods be order-agnostic.
+
+
+        Args:
+            data is a pandas data frame representing the output of a run of the task.
+            row_num is the row number to check.
```
brianhw: nit: note if it's zero or one-based...

edx/analytics/tasks/tests/test_course_subjects.py
```diff
@@ -0,0 +1,239 @@
...
+        self.assertEquals(data.shape[0], 0)
+
+    def test_course_no_subjects(self):
+        """With a course with no subjects, we expect a row with NULLs."""
+        course_with_no_subjects = {"items": [{"course_id": "foo", "subjects": [{}]}]}
+        data = self.run_task(json.dumps(course_with_no_subjects))
+        # We expect an entry in the list of courses, since there is a course in the catalog.
+        self.assertEquals(data.shape[0], 1)
+        # We expect nulls for the three subject data columns.
+        expected = {'course_id': 'foo', 'date': '2015-06-25',
```
brianhw: nit:  'date' should be on a separate line, at least, and actually, so should 'course_id'.

edx/analytics/tasks/tests/test_course_subjects.py
```diff
@@ -0,0 +1,239 @@
...
+                                          "title": "Testing",
+                                          "language": "py"}]
+                            }
+                           ]
+                      }
+
+        data = self.run_task(json.dumps(input_data))
+        # We expect to see this course with the mock_subject information.
+        self.assertEquals(data.shape[0], 1)
+        expected = {'course_id': 'foo', 'subject_uri': 'testing/uri', 'subject_title': 'Testing', 'date': '2015-06-25',
```
brianhw, line 128: nit: the date here should match the date used in the setup.  Pull it out into a constant to make that clear.

edx/analytics/tasks/tests/test_course_subjects.py
```diff
@@ -0,0 +1,239 @@
...
+                            "subjects":
+                                [{"uri": "testing/uri",
+                                  "title": "Testing",
+                                  "language": "py"}]
+                            },
+                           {"course_id": "bar",
+                            "subjects":
+                                [{"uri": "testing/uri",
+                                  "title": "Testing",
+                                  "language": "py"}]}
```
brianhw, line 171: nit: if it's a different course, it should presumably have different title and uri....

edx/analytics/tasks/tests/test_vertica_load.py
```diff
@@ -0,0 +1,226 @@
...
+            "CREATE TABLE IF NOT EXISTS testing.dummy_table "
+            "(id AUTO_INCREMENT PRIMARY KEY,course_id VARCHAR(255),"
+            "interval_start DATETIME,interval_end DATETIME,label VARCHAR(255),"
+            "count INT,created TIMESTAMP DEFAULT NOW(),PRIMARY KEY (id))"
+        )
+
+    def test_create_table_without_column_definition(self):
+        connection = MagicMock()
+        task = self.create_task(cls=CopyToPredefinedVerticaDummyTable)
+        with self.assertRaises(NotImplementedError):
```
brianhw, line 140: nit: should this also check the error message to clarify the right error is occurring?  (e.g table vs column).  There's a regexp version of assertRaises that does this...

edx/analytics/tasks/tests/test_vertica_load.py
```diff
@@ -0,0 +1,226 @@
...
+        with file_to_copy as sent_data:
+            sent_source = sent_data.read()
+        self.assertEquals(sent_source, expected_source)
+
+    def test_copy_multiple_rows(self):
+        task = self.create_task(source=self._get_source_string(4))
+        cursor = MagicMock()
+        task.copy_data_table_from_target(cursor)
+        query = cursor.copy_file.call_args[0][0]
+        self.assertEquals(query, self._get_expected_query())
```
brianhw, line 191: get_expected_query is called here with only one row, but _get_source_string was called with 4.  What happened to the other three?

edx/analytics/tasks/vertica_load.py
```diff
@@ -0,0 +1,285 @@
...
+    vertica_client_available = False  # pylint: disable-msg=C0103
+
+
+class VerticaCopyTaskMixin(OverwriteOutputMixin):
+    """
+    Parameters for copying a database into Vertica.
+
+        credentials: Path to the external access credentials file.
+        schema:  The schema to which to write.
+        insert_chunk_size:  The number of rows to insert at a time.
```
brianhw: where is this used?  cut-and-paste error?

edx/analytics/tasks/tests/test_vertica_load.py
jrowan, line 191: Since we're using Vertica copies, we don't need to make 4 queries.  The set of all 4 rows is used when we compare the readout of the file object marked as the insert source to the readout of the file object output by the copy command; all 4 rows get returned by call_args[0][1].

brianhw:
LGTM. :+1:

PR #129, opened 2015-07-20
Use the XML reporting API

mulby:
@brianhw @jab5569 this is the paypal discovery work

after much tweaking it appears to produce the same data as the previous method but allows us to pull it incrementally

I don't really like how this code is structured, so I'm treating it as effectively throw-away, but it does prove out the strategy

### DIFF
--- a/edx/analytics/tasks/reports/paypal.py
+++ b/edx/analytics/tasks/reports/paypal.py
@@ -1,10 +1,14 @@
 import datetime
+import xml.etree.cElementTree as ET
+from cStringIO import StringIO
 import logging
-import luigi
-import paypalrestsdk
-import json
+from collections import namedtuple, OrderedDict
+from decimal import Decimal
+import time
 
-from luigi.date_interval import DateInterval
+import luigi
+from luigi.configuration import get_config
+import requests
 
 from edx.analytics.tasks.url import get_target_from_url
 from edx.analytics.tasks.url import url_path_join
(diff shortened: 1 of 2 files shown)


mulby:
Also - I checked one day worth of data against the version currently produced by the financial-reporting branch and it matched up exactly

edx/analytics/tasks/reports/paypal.py
```diff
@@ -18,183 +22,436 @@
...
+
+    @classmethod
+    def params_from_xml(cls, root_node):
+        base_response_node = root_node.find('baseResponse')
+        return {
+            'response_code': int(base_response_node.findtext('responseCode')),
+            'response_message': base_response_node.findtext('responseMsg')
         }
 
-        log.debug('Requesting payments')
```
johnalbaker, line 69: Nit: Remove

johnalbaker:
:+1: LGTM

brianhw:
LGTM.   Some things to look at:  
- There is no timeout on the report polling, which makes me nervous.  If for some reason their report generation gets stuck, then our report retrieval also gets stuck.
- It's pretty elegant to have each response try to find the appropriate class to parse the response, and then instantiate that class based on the params that the class parsed.  But it also seems a little twisted, since we presumably know what kind of response we're expecting, so we should be able to just instantiate it directly by passing the response class name into the send method. 
- I wasn't clear on the decimal processing -- do they really return integers with number of cents?  That seems really weird if the currency is dollars.

mulby:
@brianhw - a few responses to your points
- Definitely... in general the exception handling and non-happy path cases need much more careful review.
- Yeah that was part of the reason why I wanted to refactor this code. There is a very clear request-response protocol that this code doesn't do a good job of communicating / leveraging.
- Yes - they return integer number of cents. I agree this is unintuitive.

PR #130, opened 2015-07-20
WIP: Support python UDFs in hive

mulby:
@brianhw and @jab5569 

Here is a sketch of the type of thing I was thinking for the unique identifiers. It uses a python UDF in Hive to generate the IDs.

Stuff that is suboptimal:
- The "encode_id" function is duplicated in the UDF code and the rest of the pipeline code.
- There is some complexity associated with defining and using python UDFs in Hive.
- I had to pretty much fork the "run" function of the HiveQueryTask in order to inject the UDF setup code when running Hive tasks. This may be a good thing, though, since it could allow us to more elegantly inject other Hive boilerplate to our queries.

Other alternate approaches that we could do:
- Defer creation of these IDs until reconciliation where we are running python code on every row anyway.
- Use a Hive-friendly encoding (one of the SHA hashes for example).
- Add another processing step in after the big Hive query that injects these IDs into the records and stores them in another table.

Let me know if you would prefer any of the other approaches (or if you have another suggestion that I didn't mention above). I'm not attached to this method at all.

That said - I do like the idea of being able to define other simple python UDFs and run them in Hive in the future, but I'm not sure when/if we will need them for other purposes.

### DIFF
--- a/edx/analytics/tasks/util/hive.py
+++ b/edx/analytics/tasks/util/hive.py
@@ -1,11 +1,14 @@
 """Various helper utilities that are commonly used when working with Hive"""
 
 import logging
+import os
+import tempfile
 import textwrap
 
 import luigi
+import luigi.hadoop
 from luigi.configuration import get_config
-from luigi.hive import HiveQueryTask, HivePartitionTarget, HiveQueryRunner
+from luigi.hive import HiveQueryTask, HivePartitionTarget, HiveQueryRunner, load_hive_cmd
 from luigi.parameter import Parameter
 
 from edx.analytics.tasks.url import url_path_join, get_target_from_url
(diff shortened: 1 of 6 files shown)


edx/analytics/tasks/util/hive.py
```diff
@@ -131,8 +134,64 @@ def run_job(self, job):
...
 
-        log.debug('Executing hive query: %s', job.query())
-        return super(OverwriteAwareHiveQueryRunner, self).run_job(job)
+        query = job.query()
+        transform_udfs = getattr(job, 'transform_udfs', [])
+
+        log.debug('Executing hive query: %s', query)
+        self.prepare_outputs(job)
+        try:
+            with tempfile.NamedTemporaryFile() as f:
```
brianhw: should we skip all this first if there are no transform_udfs for this job?

brianhw:
Looks good to me.  The "encode_id" function is pretty darned small.  The duplication should be commented, but I think that's enough.  Adding support for UDFs seems like a good thing, at the moment.  It opens the door to greater flexibility, at the risk of greater complexity, but I'm willing to open that door.  

I like the idea of having reversible unique_ids, and the SHA doesn't provide that.  And I don't find that separate tables provides much value.  So the only alternative I like is the addition of these IDs as part of reconciliation.  That had been my original assumption.  Question is whether we need the unique_id values in the canonical transaction and order tables, when they are loaded into Vertica, or if the reconciliation table is sufficient.

PR #133, opened 2015-07-28
Initial work to load event logs into Vertica.

jrowan:
Provides infrastructure for loading event logs into an HP Vertica data warehouse.  Contains a canonicalization step, a Vertica-loading-specific cleaning step that does user agent parsing, and a task and workflow for pulling days' events into the data warehouse.  This is designed to support product-analytics queries like "what percent of our play_video events happened on mobile devices?"


jrowan:
I'm still doing some work testing this (in particular, the acceptance test doesn't work yet), cleaning up the code, and handling issues arising from the incremental-user-activity branch that was pulled in to provide the canonicalization task, but I figured I should at least put this up for review since we have an end-to-end-workflow built.

@jab5569 @mulby @brianhw @shnayder

edx/analytics/tasks/clean_for_vertica.py
```diff
@@ -0,0 +1,257 @@
...
+    VERSION = 2  # Version 1 was after the canonicalization
+    OUTPUT_BUCKETS = 100
+    MAX_KEY_LENGTH = 256
+
+    n_reduce_tasks = OUTPUT_BUCKETS
+
+    def __init__(self, *args, **kwargs):
+        super(CleanForVerticaTask, self).__init__(*args, **kwargs)
+
+        self.interval = luigi.date_interval.Date.from_date(self.date)
```
mulby: You can remove this attribute, I don't think it's used anymore. This task used to operate on intervals but now operates on individual dates.

mulby: Oops, wrong class, ignore the above comment.

edx/analytics/tasks/canonicalization.py
```diff
@@ -0,0 +1,235 @@
...
+
+    VERSION = 1
+    OUTPUT_BUCKETS = 100
+
+    n_reduce_tasks = OUTPUT_BUCKETS
+
+    def __init__(self, *args, **kwargs):
+        super(CanonicalizationTask, self).__init__(*args, **kwargs)
+
+        self.interval = luigi.date_interval.Date.from_date(self.date)
```
mulby, line 52: You can remove this attribute, I don't think it's used anymore. This task used to operate on intervals but now operates on individual dates.

edx/analytics/tasks/canonicalization.py
```diff
@@ -0,0 +1,235 @@
...
+    to make it possible to parallelize downstream jobs.
+
+    Any events that weren't emitted in the given interval will be dropped. In particular, this includes mobile events
+    that are delivered long after they were emitted.
+
+    Once the job finishes, it writes a _SUCCESS file for each date in the interval. If such a file is present on
+    startup, those days are not processed again unless the overwrite flag is set.
+    """
+
+    interval = luigi.DateIntervalParameter(None)
```
mulby, line 40: I think we can remove this. Also update the comment above.

edx/analytics/tasks/canonicalization.py
```diff
@@ -0,0 +1,235 @@
...
+    output_root = None
+    date = luigi.DateParameter()
+
+    VERSION = 1
+    OUTPUT_BUCKETS = 100
```
mulby, line 45: I would make this a configurable parameter.

mulby:
meta point - I apologize for the ragged edges on the C14N code, I threw that together during a hackathon and it's been evolving very slowly ever since - your efforts productizing it are very appreciated!

edx/analytics/tasks/clean_for_vertica.py
```diff
@@ -0,0 +1,257 @@
...
+        Add a metadata dictionary to the event, with the following fields:
+         - version
+         - last_modified (when the event was processed by this task)
+         - id (hash of the raw event format, i.e. the line)
+         - original_file -- the source file where we found this event
+        """
+        if event is None:
+            return None
+
+        metadata = event.setdefault('metadata', {})
```
mulby, line 88: Pretty much all of this metadata should already be set by the C14N task, why overwrite it? Perhaps just include new metadata for vertica cleaning process? I'm not sure there's much value there though... I guess I would argue we just leave the metadata as is.

edx/analytics/tasks/clean_for_vertica.py
```diff
@@ -0,0 +1,257 @@
...
+            the event, with all of the keys shortened to be at most 256 characters.
+        """
+        if event is None:
+            return None
+
+        new_event = {}
+        for key in event:
+            if len(key) > self.MAX_KEY_LENGTH:
+                shortened_key = key[:self.MAX_KEY_LENGTH]
+                new_event[shortened_key] = self.truncate_keys(event[key]) if type(event[key]) == dict else event[key]
```
mulby, line 131: Interesting side effect of this - if two strings truncate to the same key, the second one will overwrite the first.

At a minimum we should log a warning when this happens so that we can understand the frequency. Ideally we would resolve the conflict by appending a unique value to the end of the trunacted string or something. If you go with the second route -> be sure that the same event will always produce the same truncated key strings, a gotcha here is that `for key in event` does not guarantee a particular traversal order.

edx/analytics/tasks/clean_for_vertica.py
```diff
@@ -0,0 +1,257 @@
...
+        self.remove_output_on_overwrite()
+        super(CleanForVerticaTask, self).run()
+
+    def extra_modules(self):
+        """We need the user agent parser user_agents for the mapper function."""
+        return [ua_parser, user_agents]
+
+    def jobconfs(self):
+        jcs = super(CleanForVerticaTask, self).jobconfs()
+        # Downstream luigi methods won't be able to open a folder of gzipped files from hdfs/s3 properly,
```
mulby, line 251: Hrmm... this should work, but we can leave it out for now if that simplifies things for you.

edx/analytics/tasks/events_to_warehouse.py
```diff
@@ -0,0 +1,143 @@
...
+
+    @property
+    def insert_source_task(self):
+        """The previous task in the workflow is to clean the data for loading into Vertica."""
+        return(CleanForVerticaTask(date=self.run_date, remove_implicit=self.remove_implicit))
+
+    @property
+    def table(self):
+        """We use the table event_logs for this task."""
+        return "event_logs_test"
```
mulby, line 32: remove "_test" from this table name

edx/analytics/tasks/events_to_warehouse.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def requires(self):
+        # Add additional args for VerticaCopyMixin.
+        kwargs2 = {
+            'schema': self.schema,
+            'credentials': self.credentials,
+        }
+        kwargs2.update(kwargs2)
+
+        for day in luigi.DateIntervalParameter().parse(str(self.interval)):
```
mulby, line 140: You should be able to just do:

``` python
for day in self.interval:
```

is `self.interval` overwritten in the constructor or something?

edx/analytics/tasks/pathutil.py
```diff
@@ -1,9 +1,7 @@
 """
 Helper classes to specify file dependencies for input and output.
-
```
mulby, line 3: is there an automated tool that is removing all of these whitespace lines? I think I want to keep some of them... (like the ones that indicate paragraph breaks)

edx/analytics/tasks/url.py
```diff
@@ -59,13 +59,18 @@ def open(self, mode='r'):
 }
 
 
-def get_target_from_url(url):
+def get_target_from_url(url, success_marked=False):
```
mulby, line 62: We can scrap the changes in this file for simplicity's sake... I was just toying with this idea.

edx/analytics/tasks/vertica_load.py
```diff
@@ -173,12 +186,14 @@ def init_copy(self, connection):
         # clear table contents
         self.attempted_removal = True
         if self.overwrite:
+            print "IN overwrite area!"
```
mulby, line 189: don't forget to remove this!

edx/analytics/tasks/canonicalization.py
jrowan, line 40: interval is needed as a parameter because the EventLogSelectionMixin in general works with intervals, not just individual dates.  My current plan is to move the interval logic inside the init method before the EventLogSelectionMixin portion gets __init__ed, so that the EventLogSelectionMixin sees what it needs to but the class as a whole doesn't have needless parameters.

edx/analytics/tasks/calendar_task.py
```diff
@@ -29,16 +29,12 @@ class CalendarDownstreamMixin(OverwriteOutputMixin):
...
-
     This table provides information about every day in every year that is being analyzed. It captures many complex
     details associated with calendars and standardizes references to concepts like "weeks" since they can be defined
     in different ways by various systems.
-
     It is also intended to contain business-specific metadata about dates in the future, such as fiscal year boundaries,
     fiscal quarter boundaries and even holidays or other days of special interest for analysis purposes.
-
     Parameters:
-
```
brianhw, line 41: As mentioned elsewhere, we want the whitespace.  Since these are the only changes, this file should be reverted entirely.

PR #135, opened 2015-07-29
Support bootstrapping and deploying to devstack

mulby:
We also need to remove the default setting of the hadoop version from client.cfg so that it doesn't override /etc/luigi/client.cfg which is now written to the vagrant box during provisioning. This should not impact production clusters since they have the same value in their /etc/luigi/client.cfg.

This creates a new make target "bootstrap" which just installs the dependencies need to run remote-task. This means the pipeline can be used to bootstrap itself in a _much_ more lightweight fashion since it doesn't require all of the complex dependencies that are actually needed to run tasks.

This is ready for review.

@brianhw @jab5569


johnalbaker:
:+1:

mulby:
@jab5569, @brianhw this has been updated

edx/analytics/tasks/launchers/remote.py
```diff
@@ -34,6 +39,8 @@ def main():
     parser.add_argument('--shell', help='execute a shell command on the cluster and exit', default=None)
     parser.add_argument('--sudo-user', help='execute the shell command as this user on the cluster', default='hadoop')
     parser.add_argument('--workflow-profiler', choices=['pyinstrument'], help='profiler to run on the launch-task process', default=None)
+    parser.add_argument('--wheel-url', help='url of the wheelhouse', default=os.getenv('WHEEL_URL'))
+    parser.add_argument('--retry', action='store_true', help='skip setup of the environment, --remote-name also must be specified and point to a valid environment')
```
brianhw: I'm confused why "retry" would mean "skip setup of the environment".  One of these should change to better match the other.

edx/analytics/tasks/launchers/remote.py
```diff
@@ -115,8 +128,12 @@ def convert_args_to_extra_vars(arguments, uid):
         arguments (argparse.Namespace): The arguments that were passed in on the command line.
         uid (str): A unique identifier for this task execution.
     """
+    if arguments.host:
+        name = 'all'
+    else:
+        name = 'mr_{0}_master'.format(arguments.job_flow_id or arguments.job_flow_name)
```
brianhw: nit: should this be pulled out into a separate method?  The logic for constructing the name is repeated once below as well.

brianhw: get_ansible_group_name?

brianhw:
LGTM.  Just a few nits and probably a squash. :+1:

edx/analytics/tasks/launchers/remote.py
mulby: Done

edx/analytics/tasks/launchers/remote.py
mulby: Done

edx/analytics/tasks/launchers/remote.py
```diff
@@ -60,13 +74,12 @@ def run_task_playbook(inventory, arguments, uid):
...
         uid (str): A unique identifier for this task execution.
     """
-    extra_vars = convert_args_to_extra_vars(arguments, uid)
-    args = ['task.yml', '-e', extra_vars]
-    if arguments.user:
-        args.extend(['-u', arguments.user])
-    prep_result = run_ansible(tuple(args), arguments.verbose, executable='ansible-playbook')
-    if prep_result != 0:
-        return prep_result
+    if not arguments.retry:
```
brianhw: Wasn't retry renamed to skip_setup?

edx/analytics/tasks/launchers/remote.py
```diff
@@ -115,8 +130,12 @@ def convert_args_to_extra_vars(arguments, uid):
         arguments (argparse.Namespace): The arguments that were passed in on the command line.
         uid (str): A unique identifier for this task execution.
     """
+    if arguments.host:
+        name = 'all'
+    else:
+        name = get_ansible_inventory_host(arguments)
     extra_vars = {
-        'name': arguments.job_flow_id or arguments.job_flow_name,
+        'name': name,
```
brianhw, line 138: The name at this point is going to contain the "mr_{}_master" form of the name.  Is that going to make a difference?  I don't see where else the flow_id/name is accessed, so it doesn't seem like it should be a problem.

brianhw, line 138: Oh, I see now, it's also in the task.yml.  That looks much cleaner now.

brianhw:
Just needs retry to be fixed, squash, then :+1:

mulby:
@jab5569 I'm waiting on a re-review of this PR, it has changed significantly since your last review I believe

johnalbaker:
:+1:

mulby:
hrmm... looks like I broke something since acceptance tests failed, will investigate later: http://jenkins-ci.analytics.edx.org/view/ad-hoc/job/edx-analytics-pipeline-acceptance-test-manual/406/consoleFull

mulby:
Fixed the acceptance tests, merging.

Makefile
```diff
@@ -15,6 +15,11 @@ uninstall:
 install: requirements uninstall
 	python setup.py install --force
 
+bootstrap: uninstall
+	$(PIP_INSTALL) -U -r requirements/pre.txt
+	$(PIP_INSTALL) -U -r requirements/base.txt
+	python setup.py install --force
```
bradenmacdonald, line 21: @mulby This was causing issues for me. I needed to edit `remote-task` to fix a bug, but found that my changes were having no effect. The reason is that this line installs a second copy of the pipeline into `site-packages`, which was being loaded.

I think you should just use `python setup.py develop` here instead, if the goal is to register the entry points.

bradenmacdonald:
@mulby How should one test changes to the pipeline tasks with this setup? I was trying this earlier but couldn't get it working out of the box, since the `remote-task` launcher creates an independent copy of the pipeline repo on the devstack every time it runs.

I was able to get it to work by modifying the devstack Vagrantfile to share the pipeline repo folder from my host to `/edx/app/pipeline/edx-analytics-pipeline`, and then running `remote-task` with the additional args `--repo /edx/app/pipeline/edx-analytics-pipeline` and `--branch [my-wip-branch]`. But that requires I commit my changes before testing them. Is there an easier way?

mulby:
@bradenmacdonald - that's very reasonable way to do it... we actually typically push our branches to github and then deploy them to the cluster (either vagrant or EMR) from github. This workflow is a hold-over from when we were doing all of our development on an EMR cluster.

We are planning on refactoring the operational workflow for the pipeline to more closely match other edX services, however, we chose to commit the work thus far and then work on any more substantial operational refactoring once we wrap up this milestone.

A very simple change would be to make the `--branch` argument optional and have it skip the "checkout the branch" parts of task.yml if it wasn't specified. That way you wouldn't have to commit!

Makefile
mulby, line 21: Yeah we've always had problems with this... I seem to remember `python setup.py develop` not actually registering the entry points - have you had success with this?

My typical workflow is actually something closer to:

```
make uninstall  # this removes both copies of the pipeline (develop and install)
python setup.py install # ensure the entry points are updated
python setup.py develop # now I can make changes to anything but the entry points
```

If `develop` is correctly registering the entry points than we should definitely use that instead!

bradenmacdonald, line 21: In my experience, `python setup.py develop` always updates the entry points (just make sure you don't have a separate copy of the package installed into `site-packages`). There could be some edge cases I'm missing, but that's what I've always used to refresh entry points, when hacking on this pipeline and also XBlocks and other stuff.

mulby, line 21: I now remember - it wasn't the entry points that were causing the problem, it was the extra data files. task.yml has to be copied into the "/share" path of the virtualenv etc and that doesn't happen unless you run `install`.

bradenmacdonald, line 21: I see. That makes sense. Is that part required for the 'bootstrap' version where we're just running remote tasks over ssh? If not, it would be nice to use 'develop' for the bootstrap target and 'install --force' for the normal target that's used on the EMR cluster / devstack remote.

mulby, line 21: Unfortunately remote-task runs the task.yml playbook so it requires this.
Longer term I'd prefer to pull all of this out of edx-analytics-pipeline.
Maybe move it to edx-analytics-configuration.

bradenmacdonald, line 21: I see. I didn't have any problems with that in my testing because the sequence `python setup.py install --force; pip uninstall edx.analytics.tasks` actually leaves those data files behind in venv/share.

How about this?

```
python setup.py develop
python setup.py install_data
```

I just played around with that and it seems like it should do the trick.

mulby, line 21: Ou, I didn't know about install_data

That would be ideal

PR #127, opened 2015-07-18
Properly escape the names of Hive columns

bradenmacdonald:
**Description**: In [newer versions of Hive](http://mail-archives.apache.org/mod_mbox/hive-commits/201503.mbox/%3C20150310220045.68294AC02E7@hades.apache.org%3E), `date` and `interval` are keywords. This change escapes column names and partition key names when generating Hive DML queries.

Without this fix, the following error can occur when running the pipeline:

```
Failed to recognize predicate 'DATE'. Failed rule: 'identifier' in table name
```

This fix should be fully compatible with any version of Hive.

**JIRA**: [OSPR-718](https://openedx.atlassian.net/browse/OSPR-718)

**Merge deadline**: Not an edX partner, so best effort.

**Reviewers**: @mtyaka and TBD

PR #138, opened 2015-08-04
Revert "Properly escape the names of Hive columns"

mulby:
Reverts edx/edx-analytics-pipeline#127

Not sure why it looks like acceptance tests were passing, but they aren't, reverting for now.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -7,4 +7,3 @@ Jason Bau <jbau@stanford.edu>
 Justin Helbert <jhelbert@edx.org>
 Will Daly <will@edx.org>
 Dylan Rhodes <dylanr@stanford.edu>
-Braden MacDonald <braden@opencraft.com>
--- a/edx/analytics/tasks/database_imports.py
+++ b/edx/analytics/tasks/database_imports.py
@@ -89,7 +89,7 @@ class ImportIntoHiveTableTask(OverwriteOutputMixin, HiveQueryTask):
         query = query_format.format(
             database_name=hive_database_name(),
             table_name=self.table_name,
-            col_spec=','.join(['`{}` {}'.format(*c) for c in self.columns]),
+            col_spec=','.join([' '.join(c) for c in self.columns]),
             location=self.table_location,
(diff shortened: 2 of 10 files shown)


mulby:
@brianhw 

@bradenmacdonald FYI - we have to revisit this, I am seeing this error message from Hive while running the acceptance tests (specifically the enrollment tests):

```
SemanticException [Error 10214]: Invalid partition spec specified dt not found in table's partition spec: {`dt`=2014-08-06}
```

bradenmacdonald:
@mulby Rats. I was worried about something like this but didn't hit any errors in my own testing. Do you have a little more context for this error message to help me debug it? It could be that the syntax I used is not as backward compatible as I thought, or it could be a case of applying escaping twice or in the wrong context. I haven't yet been able to run the acceptance tests myself so it's tricky for me to troubleshoot.

mulby:
@bradenmacdonald - yeah, I understand this will be tricky to resolve without actually running the tests, it might make the most sense for me to just diagnose it over here.

This was the task that failed:

```
CourseEnrollmentTableTask(overwrite=False, source=('REDACTED',), expand_interval=2 days, 0:00:00, pattern=('.*tracking.log-(?P<date>\\d{8}).*\\.gz', '.*tracking.notalog-(?P<date>\\d{8}).*\\.gz'), warehouse_path=REDACTED, interval=2014-08-01-2014-08-06)
```

I think the create table statement is failing? Hive isn't giving me a lot of debug output here unfortunately...

bradenmacdonald:
@mulby FYI, these changes for for compatibility with Hive 1.2.1. I also had to make [a change to luigi](https://github.com/spotify/luigi/commit/1de379b56c346754bd9dbc7e921211ca570219c7) - perhaps it requires that change as well?

If this is too much hassle, we can just use an older Hive version and not bother with these changes - especially now that your new analytics devstack configuration is in the works.

But if there is a simple fix that allows compatibility with the old and newer Hive versions, that would be great.

PR #139, opened 2015-08-05
Fix answer distribution handling of malformed event data

mulby:
There is an event in the tracking logs that has a corrupt answer_id that is causing the answer distribution job to fail on the RC branch.

We should prevent malformed events like this from being analyzed.

@brianhw @jab5569


johnalbaker:
:+1:

brianhw:
Looks reasonable.  :+1:  Would like to catch up later as to why this is only an issue on the RC.

mulby:
@brianhw - it's still unclear why we are only seeing this on the RC

PR #141, opened 2015-08-05
User Profiles Table in Analytics MySQL DB

bradenmacdonald:
In order to support user-specific reports in the Insights dashboard, this PR takes the first step by making user data available within the MySQL DB.

Two [related](https://github.com/edx/edx-analytics-data-api/pull/80) [PRs](https://github.com/edx/edx-analytics-data-api-client/pull/21) update the data API to allow analytics software to access the data, and [a final PR for the Dashboard](https://github.com/edx/edx-analytics-dashboard/pull/338) displays it to the user.

**Changes**:
- Add user's "name" to the existing task that imports data from the LMS `auth_userprofile` table to Hive.
- Adds two new pipeline tasks:
  - `InsertToMysqlUserProfilesTask` which creates the new `user_profile` table in the analytics MySQL DB. This new table contains basic user information (ID, name, email, username, gender, age, education level, login date) extracted and joined from the existing `auth_user` and `auth_userprofile` Hive tables.
  - `EnrollmentSnapshotTask` which creates/updates a new MySQL table (`course_enrollment_snapshot`) to contain the most current enrollment information available.

These changes together allow us to create new Data API methods that can (a) get the profiles of all known users or (b) get the profiles of all users in a specific course.

**Partner Information**: Third-party open edX instance, not an edX partner.

**Test Instructions - OpenCraft devstack**:
One way to test is with the analytics devstack:
1. Setup an [analytics devstack](https://github.com/open-craft/edx-analytics-devstack). 
2. Modify your local LMS devstack to work with the new analytics devstack (See analytics devstack README).
3. Login to the analytics devstack as the `analytics` user.
4. Run the command:
   `
   launch-task InsertToMysqlUserProfilesTask --local-scheduler
   `
5. Wait for it to complete. To check if it worked, run `mysql -u root analytics --execute="SELECT * FROM user_profile;"`. You should get a result set showing all the users on your LMS devstack.

**Test Instructions - Combined devstack**:
Another way is the nascent combined devstack:
1. Setup a [combined devstack](https://gist.github.com/bradenmacdonald/8b3b9736bf2b5ec44bc0)
2. Clear out any existing pipeline repo: `vagrant ssh` then `sudo rm -r /var/lib/analytics-tasks/devstack/repo` (This is necessary since the ansible scripts won't force the use of a different upstream repo during the next step)
3. On your host, in the edx-analytics-pipeline folder (with the venv loaded), run these commands:
   
   ```
[...]

PR #143, opened 2015-08-11
Load the tables d_user and f_user_activity into the warehouse via the pipeline.

jrowan:
Port of two pentaho tasks for loading data warehouse tables.  Goes through the pipeline following the architecture (pipeline computation with results stored in hive | read from foreign database into hive)\* -> (build hive table from hive query) -> (load hive table into the data warehouse).  Contains acceptance tests for the workflows, but note that as more tasks are built out, we will likely eventually change the workflow to have a single entry point and table loading tasks depending on prior table loading tasks.

@mulby @jab5569 @brianhw


edx/analytics/tasks/load_warehouse_d_user.py
```diff
@@ -0,0 +1,139 @@
...
+import luigi
+from edx.analytics.tasks.util.hive import HiveTableFromQueryTask, WarehouseMixin, HivePartition
+
+
+class AggregateDUserTableHive(HiveTableFromQueryTask):
+    """Aggregate the d_user table in Hive."""
+    interval = luigi.DateIntervalParameter()
+    user_country_output = luigi.Parameter()
+    n_reduce_tasks = luigi.Parameter()
+    # source = luigi.Parameter()
```
johnalbaker: remove

johnalbaker: remove

edx/analytics/tasks/load_warehouse_d_user.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def requires(self):
+        """
+        This task reads from auth_user, auth_user_profile, and last_country_of_user, so require that they be
+        loaded into Hive (via MySQL loads into Hive or via the pipeline as needed).
+        """
+        return [ImportAuthUserTask(overwrite=False, destination=self.warehouse_path),
+                ImportAuthUserProfileTask(overwrite=False, destination=self.warehouse_path),
+                ImportLastCountryOfUserToHiveTask(interval=self.interval,
+                                                  # source=(self.source,),
```
johnalbaker: remove

edx/analytics/tasks/load_warehouse_d_user.py
```diff
@@ -0,0 +1,139 @@
...
+        user_country_output: location for intermediate output of location_per_course task.  In use here in this spike
+                             to prevent accidentally overwriting production data, will be removed in final version.
+        n_reduce_tasks: number of reduce tasks
+        overwrite: whether or not to overwrite existing outputs; set to False by default for now
+    """
+    interval = luigi.Parameter()
+    user_country_output = luigi.Parameter()
+    n_reduce_tasks = luigi.Parameter()
+    overwrite = luigi.BooleanParameter(default=False)
+    # source = luigi.Parameter()
```
johnalbaker: remove

edx/analytics/tasks/load_warehouse_d_user.py
```diff
@@ -0,0 +1,139 @@
...
+    @property
+    def insert_source_task(self):
+        return (
+            # Get the location of the Hive table, so it can be opened and read.
+            AggregateDUserTableHive(
+                n_reduce_tasks=self.n_reduce_tasks,
+                interval=luigi.DateIntervalParameter().parse(self.interval),
+                warehouse_path=self.warehouse_path,
+                overwrite=self.overwrite,
+                # source=self.source,
```
johnalbaker: remove

edx/analytics/tasks/load_warehouse_f_user_activity.py
```diff
@@ -0,0 +1,121 @@
...
+    def auto_primary_key(self):
+        """The warehouse schema defines an auto primary key called row_number for this table."""
+        return ('row_number', 'AUTO_INCREMENT')
+
+    @property
+    def foreign_key_mapping(self):
+        """Foreign keys are specified in the warehouse schema."""
+        # none for now for testing
+        return {}
+        # return {'user_id': ('{schema}.d_user'.format(schema=self.schema), 'user_id'),}
```
johnalbaker: remove?

edx/analytics/tasks/tests/acceptance/test_warehouse_f_user_activity.py
```diff
@@ -0,0 +1,68 @@
...
+        # Put up the mock tracking log for user locations.
+        self.upload_tracking_log(self.INPUT_FILE, datetime.datetime(2014, 6, 29))
+
+    def test_warehouse_f_user_activity(self):
+        """Tests the workflow for the f_user_activity table, end to end."""
+
+        self.task.launch([
+            'LoadFUserActivityToWarehouse',
+            '--interval', self.DATE_INTERVAL.to_string(),
+            # '--source', self.test_src,
```
johnalbaker: remove

edx/analytics/tasks/load_warehouse_f_user_activity.py
```diff
@@ -0,0 +1,121 @@
+"""
+Loads the d_user table into the warehouse through the pipeline via Hive.
```
johnalbaker: should be "Loads the f_user_activity..."

edx/analytics/tasks/load_warehouse_f_user_activity.py
```diff
@@ -0,0 +1,121 @@
...
+            , uad.category
+            , uad.count
+            FROM auth_user au
+            JOIN user_activity_daily uad ON au.username = uad.username
+            """
+
+
+class LoadFUserActivityToWarehouse(WarehouseMixin, VerticaCopyTask):
+    """
+    Loads the d_user table from Hive into the Vertica data warehouse.
```
johnalbaker: Should be: "Loads the f_user_activity..."

jrowan:
This now handles the f_user_activity concerns we had earlier.

NB: The current strategy for handling the history table for f_user_activity is to build a view on top of the history table and the table that the task loads.  In the long run, we eventually want to replace this with an incremental approach to maintaining the f_user_activity table.

@mulby @jab5569 @brianhw

edx/analytics/tasks/load_warehouse_d_user.py
```diff
@@ -0,0 +1,135 @@
...
+On the roadmap is to write a task that runs validation queries on the aggregated Hive data pre-load.
+"""
+from edx.analytics.tasks.vertica_load import VerticaCopyTask
+from edx.analytics.tasks.location_per_course import ImportLastCountryOfUserToHiveTask
+from edx.analytics.tasks.database_imports import ImportAuthUserProfileTask, ImportAuthUserTask
+import luigi
+from edx.analytics.tasks.util.hive import HiveTableFromQueryTask, WarehouseMixin, HivePartition
+
+
+class AggregateDUserTableHive(HiveTableFromQueryTask):
```
mulby: General comment, I think I would prefer we exclude the "d_" and "f_" and associated prefixes etc from symbols in the code and only use them on the actual schema objects in the database.

I don't see these prefixes providing enough value to justify including them everywhere in the code.

edx/analytics/tasks/load_warehouse_d_user.py
```diff
@@ -0,0 +1,135 @@
...
+    """Aggregate the d_user table in Hive."""
+    interval = luigi.DateIntervalParameter()
```
mulby: General comment - I think we should move towards exposing fewer parameters at the CLI level and instead reading them from the config file.

This may be a contentious point, though, so I wouldn't hold up the PR on it.

Concrete suggestions would be:
- replace interval with an "end date" parameter and read the start date from the config
- read user country output from the config file

edx/analytics/tasks/load_warehouse_d_user.py
```diff
@@ -0,0 +1,135 @@
...
+    user_country_output = luigi.Parameter()
+    n_reduce_tasks = luigi.Parameter()
```
mulby: Instead of explicitly defining this here I would just use the `MapReduceTaskDownstreamMixin`

edx/analytics/tasks/load_warehouse_d_user.py
```diff
@@ -0,0 +1,135 @@
...
+        """
+        return [ImportAuthUserTask(overwrite=False, destination=self.warehouse_path),
+                ImportAuthUserProfileTask(overwrite=False, destination=self.warehouse_path),
+                ImportLastCountryOfUserToHiveTask(interval=self.interval,
+                                                  user_country_output=self.user_country_output,
+                                                  n_reduce_tasks=self.n_reduce_tasks)]
+
+    @property
+    def table(self):
+        return 'd_user_hive'
```
mulby: the `_hive` here isn't really giving me any useful information, I would simply remove it.

General comment for all of us (@brianhw @jab5569) - the hive warehouse namespace is starting to get quite cluttered with the various types of data we have in there:
1. Tables imported directly from the LMS and other databases (ecom etc)
2. Intermediate tables
3. High level aggregate tables that are soon to be copied into the result store
4. Tables that are soon to be copied into Vertica

Maybe we should divide up those various tables into separate schemas/directories?

edx/analytics/tasks/load_warehouse_f_user_activity.py
```diff
@@ -0,0 +1,176 @@
...
+            cursor = connection.cursor()
+            self.copy_data_table_from_target(cursor)
+
+            # mark as complete in same transaction
+            self.output().touch(connection)
+
+            # We commit only if both operations completed successfully.
+            # Vertica doesn't provide a CREATE VIEW IF NOT EXISTS method for some reason, so do the check manually.
+            try:
+                cursor.execute("SELECT * FROM {schema}.{view} LIMIT 1".format(schema=self.schema, view=self.view))
```
mulby: perhaps we should move this to a downstream task? It feels like we want to have a view target of some kind that creates the view if it doesn't exist yet.

mulby:
Finished my pass - generally looks good!

edx/analytics/tasks/load_warehouse_d_user.py
brianhw: In particular, it would be good to use standard names for such interval parameters.  In enrollments.py, we are already using interval_start and interval_end, and the former gets a default value from the config.

edx/analytics/tasks/load_warehouse_d_user.py
brianhw: I think it would at least be good to move the LMS imports into a subdirectory under the warehouse bucket path, indicating where they were imported from (e.g. LMS, ecom).  Maybe that needs two levels:  imports/LMS and imports/eCom.  
I'm less sure about the other distinctions, as they seem somewhat more fluid:  we might decide to copy intermediate stuff into Vertica and/or the result store later, and stuff that's going to one of Vertica or result store may later also go to the other.

edx/analytics/tasks/load_warehouse_d_user.py
brianhw: On the other hand, I don't think that just calling it AggregateUserTable would be informative enough.  Something like AggregateWarehouseUserTable, perhaps?

edx/analytics/tasks/load_warehouse_d_user.py
```diff
@@ -0,0 +1,135 @@
...
+                  usually be from the beginning of edX to the present day (i.e. through the previous day).
+        user_country_output: location for intermediate output of location_per_course task.  In use here in this spike
+                             to prevent accidentally overwriting production data, will be removed in final version.
+        n_reduce_tasks: number of reduce tasks
+        overwrite: whether or not to overwrite existing outputs; set to False by default for now
+    """
+    interval = luigi.Parameter()
+    user_country_output = luigi.Parameter()
+    n_reduce_tasks = luigi.Parameter()
+    overwrite = luigi.BooleanParameter(default=False)
```
brianhw: There's also an OverwriteMixin class -- is there a reason not to use it?

edx/analytics/tasks/load_warehouse_f_user_activity.py
```diff
@@ -0,0 +1,176 @@
...
+    @property
+    def foreign_key_mapping(self):
+        """Foreign keys are specified in the warehouse schema."""
+        return {}
+
+    @property
+    def columns(self):
+        return [
+            ('user_id', 'INTEGER'),
+            ('course_id', 'VARCHAR(200)'),
```
brianhw: Just curious where these sizes came from?  I think that course_id values in LMS are up to 256.

edx/analytics/tasks/load_warehouse_d_user.py
brianhw: And the filenames:  load_warehouse_user.py and load_warehouse_user_activity.py.

brianhw:
Also finished a pass.

edx/analytics/tasks/load_warehouse_d_user.py
mulby: The word "warehouse" makes me nervous since it's overloaded - it can refer to either the Hive warehouse or the Vertica warehouse. Maybe "internal reporting"?

Ideally I would prefer to just centralize our user model in Hive and then use it to generate whatever downstream stuff needs various columns from it etc. Specifically, this feels very similar to the model that is being built here:

https://github.com/edx/edx-analytics-pipeline/pull/141/files#diff-24400049312cdebf279d92850d435f5bR10

jrowan: I removed the d_ and f_ prefixes from file names, docstrings, and method names except where they refer directly to queries being made to the warehouse (so, for example, the table property of the VerticaCopyTask still says 'd_user' and the csv fixtures for acceptance tests are labelled with f_user_activity and d_user because they are supposed to equal the results of a query to a table with that name).  

I went with "internal reporting" because "warehouse" and even "internal warehouse" do seem ambiguous whether they refer to hive or vertica.

edx/analytics/tasks/load_warehouse_d_user.py
jrowan: I changed these names to internal_reporting_d_user and internal_reporting_f_user_activity for now; if we go through and re-organize the hive warehouse we might be able to remove the internal_reporting tag but I figure this will keep it unambiguous for now.

edx/analytics/tasks/load_warehouse_d_user.py
jrowan: I made user_country_output read from the config by default.

I went through the list of prod jenkins jobs and it seems most of the workflows that we are calling (all but answer_distribution, which asks for a manifest directly, and user_activity) ask for an interval directly on the CLI level.  Many of them have the Jenkins configuration set up with start and end dates, and then build the interval string passed in as a parameter on the command line in the initial script before launching the tasks.  Additionally, interval is the parameter used by the EventLogSelectionDownstreamMixin.  It may be that reworking these arguments is the correct thing to do in the future as part of a larger re-working and re-configuring, but the approach I'm currently taking (build the interval command-line parameter from provided from_date and to_date values) seemed to me like the most how-we-currently-do-things workflow for now.

edx/analytics/tasks/load_warehouse_f_user_activity.py
jrowan: 200-character course_id values is the current length used in the warehouse and the Data Model wiki page.  This seems like a mistake if the LMS course_id values can be as long as 256; I'll go on ahead and lengthen the course_id field here.

jrowan:
Addressed code review comments; handling the creation of the view in a separate task took me longer than I anticipated, so sorry this took so long.  The problem was that OverwriteOutputMixin overrides the complete() method of luigi.Task, so the tests were failing because it kept trying to overwrite the view creation.  This makes my solution to the view creation problem feel marginally hacky, but it seems cleaner than not using mixins.

Unit, pep8, pylint, and the two relevant acceptance tests are passing.

@mulby @brianhw @jab5569

setup.cfg
```diff
@@ -48,6 +48,8 @@ edx.analytics.tasks =
     student_engagement = edx.analytics.tasks.student_engagement:StudentEngagementTask
     video = edx.analytics.tasks.video:InsertToMysqlAllVideoTask
     catalog = edx.analytics.tasks.course_catalog:CourseCatalogWorkflow
+    load-d-user = edx.analytics.tasks.load_internal_reporting_user:LoadInternalReportingUserToWarehouse
```
brianhw, line 51: nit:  I would have gone with "-internal-" here instead of -d- or -f-

brianhw:
LGTM.  :+1:  Please squash before merging.

edx/analytics/tasks/load_internal_reporting_user.py
```diff
@@ -0,0 +1,135 @@
...
+from edx.analytics.tasks.location_per_course import ImportLastCountryOfUserToHiveTask
+from edx.analytics.tasks.database_imports import ImportAuthUserProfileTask, ImportAuthUserTask
+import luigi
+from edx.analytics.tasks.util.hive import HiveTableFromQueryTask, WarehouseMixin, HivePartition
+
+
+class AggregateInternalReportingUserTableHive(HiveTableFromQueryTask):
+    """Aggregate the internal reporting user table in Hive."""
+    interval = luigi.DateIntervalParameter()
+    user_country_output = luigi.Parameter()
```
mulby: nit: pull from config like you do below

edx/analytics/tasks/load_internal_reporting_user.py
```diff
@@ -0,0 +1,135 @@
...
+            """
+
+
+class LoadInternalReportingUserToWarehouse(WarehouseMixin, VerticaCopyTask):
+    """
+    Loads the user table from Hive into the Vertica data warehouse.
+
+    Parameters:
+        interval: a date_interval object containing the interval over which to pull data for user location.  Should
+                  usually be from the beginning of edX to the present day (i.e. through the previous day).
```
mulby: nit: change "edX" to "the Open edX installation"

edx/analytics/tasks/load_internal_reporting_user.py
```diff
@@ -0,0 +1,135 @@
...
+    interval = luigi.Parameter()
+    user_country_output = luigi.Parameter(
+        config_path={'section': 'last-country-of-user', 'name': 'user_country_output'}
+    )
+    n_reduce_tasks = luigi.Parameter()
+
+    @property
+    def partition(self):
+        """The table is partitioned by date."""
+        return HivePartition('dt', self.interval.date_b.isoformat())  # pylint: disable=no-member
```
mulby, line 93: I think you need the interval parameter of this class to be a DateIntervalParameter for this to work properly, it also means you can remove some complexity on line 98.

edx/analytics/tasks/load_internal_reporting_user_activity.py
```diff
@@ -0,0 +1,221 @@
...
+
+    Parameters:
+        interval: a date_interval object containing the interval over which to pull data for user location.  Should
+                  usually be from the beginning of edX to the present day (i.e. through the previous day).
+        n_reduce_tasks: number of reduce tasks
+        overwrite: whether or not to overwrite existing outputs; set to False by default for now
+    """
+    interval = luigi.DateIntervalParameter()
+    n_reduce_tasks = luigi.Parameter()
+    history_schema = luigi.Parameter(default='history')
```
mulby: I think you can remove this

edx/analytics/tasks/load_internal_reporting_user_activity.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+class InternalReportingUserActivityWorkflow(VerticaCopyTaskMixin, WarehouseMixin, luigi.WrapperTask):
+    """Wrapper to provide a single entry point for the user activity table construction and view construction."""
+    interval = luigi.DateIntervalParameter()
+    n_reduce_tasks = luigi.Parameter()
+    history_schema = luigi.Parameter(default='history')
+
+    def requires(self):
+        return [LoadInternalReportingUserActivityToWarehouse(
```
mulby, line 210: Won't this be called automatically since it is an upstream dependency of `BuildInternalReportingUserActivityCombinedView`?

edx/analytics/tasks/tests/acceptance/services/vertica.py
```diff
@@ -62,8 +62,8 @@ def connect(self):
...
 
     def reset(self):
-        """Create a testing schema on the Vertica replacing any existing content with an empty database."""
-        # The testing Vertica user doesn't have create/delete schema privileges, so we pass here.
+        """Create a testing schema on the Vertica database replacing any existing content with an empty database."""
         with self.cursor() as cur:
-            cur.execute('DROP SCHEMA IF EXISTS {0} CASCADE;'.format(self.schema_name))
-            cur.execute('CREATE SCHEMA {0};'.format(self.schema_name))
+            reset_query = 'DROP SCHEMA IF EXISTS {0} CASCADE; CREATE SCHEMA {0}'.format(self.schema_name)
+            print reset_query
```
mulby: nit: remove

mulby:
Just nits left, only blockers for me are the duplicated dependencies on LoadInternalReportingUserActivityToWarehouse and the DateIntervalParameter

mulby:
other than those two things :+1:

edx/analytics/tasks/load_internal_reporting_user_activity.py
jrowan, line 210: Because the existence check for the output of the view-building task is merely an attempt to select from the view (and thus makes no guarantee about the freshness of the data in the table), we will likely need to re-run the warehouse loading task even if the view task has been built, so we require the warehouse load task as well.

mulby, line 210: Ah I see, I would put a comment in with this explanation.

PR #145, opened 2015-08-12
Fix wheel URL default

mulby:
@brianhw and @jab5569 - I broke the acceptance tests on master since the Jenkins boxes specify an Ubuntu wheel URL to use when they are installing the pipeline locally, this was getting picked up and used to install the pipeline remotely as well. Since the jenkins boxes are Ubuntu and EMR boxes are debian based this was causing issues.

We should really only include the WHEEL_URL override if it's passed in explicitly.

Not sure how that one acceptance test run passed with this issue present.


mulby:
@bradenmacdonald - I added a fix to this PR for the `/home/hadoop` issue

bradenmacdonald:
@mulby Cool, thanks :)

mulby:
@jab5569 can you take a look a this PR real quick? I want to merge it to fix the acceptance tests on master...

johnalbaker:
:+1:  LGTM

PR #146, opened 2015-08-14
Ensure the $HOME variable is expanded properly in the sudo command

mulby:
@bradenmacdonald thanks for finding this (and suggesting the fix)! Can you make sure this works for you?

@jab5569 FYI - I tested this on EMR and it works correctly


johnalbaker:
:+1:

bradenmacdonald:
:+1:

PR #148, opened 2015-08-20
Update to vertica-python 0.5.1

jrowan:
Version 0.5.1 added support for streaming copy commands, so we will no longer need to use a fork of the package.  This is a small change which just replaces "copy_stream" with "copy" throughout the code and updates the requirements file.

The acceptance test for the course subjects workflow and the course subjects workflow continue to work as expected, and I've also checked that the Vertica bulkloader still handles unicode in course names/descriptions correctly.

@brianhw @mulby @jab5569


brianhw:
LGTM. :+1:

johnalbaker:
:+1:

PR #150, opened 2015-08-24
Pipeline tasks to load the internal reporting course information table

jrowan:
This workflow replaces the Pentaho Kettle transform Load_Corse_With_Structure.ktr and provides a workflow for building the internal reporting course information table (d_course) purely through the pipeline.  In addition to acceptance and unit tests, I've compared the results of running this workflow on the live data to the output of the existing pentaho job, and they are nearly identical (there are a few changes to courses, but those seem to be from people editing their courses between the times the two jobs were run).

@mulby @jab5569


edx/analytics/tasks/load_internal_reporting_course.py
```diff
@@ -0,0 +1,301 @@
...
+from edx.analytics.tasks.url import url_path_join
+from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
+from edx.analytics.tasks.vertica_load import VerticaCopyTask
+from edx.analytics.tasks.database_imports import ImportStudentCourseEnrollmentTask, ImportIntoHiveTableTask
+from edx.analytics.tasks.util.hive import HiveTableFromQueryTask, WarehouseMixin, HivePartition
+
+# pylint: disable-msg=anomalous-unicode-escape-in-string
+
+
+class LoadDCourseMixin(WarehouseMixin, OverwriteOutputMixin):
```
mulby: Change "DCourse" to "InternalReportingCourse" or something similar

edx/analytics/tasks/load_internal_reporting_course.py
```diff
@@ -0,0 +1,301 @@
...
+    """
+    Mixin to handle parameters common to the tasks involved in loading the internal reporting course table,
+    including calling the course structure API.
+    """
+    run_date = luigi.DateParameter()
+    api_root = luigi.Parameter(
```
mulby: I would make it more clear that this expected to be a URL, maybe "api_root_url" or simply "base_url"?

edx/analytics/tasks/load_internal_reporting_course.py
```diff
@@ -0,0 +1,301 @@
...
+        config_path={'section': 'course-structure', 'name': 'access_token'}, significant=False
+    )
+
+
+class PullCourseStructureAPIData(LoadDCourseMixin, luigi.Task):
+    """Call the course structure API and place the resulting JSON on S3."""
+
+    def run(self):
+        headers = {'authorization': ('Bearer ' + self.api_access_token), 'accept': 'application/json'}
+        api_url = url_path_join(self.api_root, 'api', 'course_structure', 'v0', 'courses', '?page_size=10000')
```
mulby: I would pull out the constant 10000 into a constant defined at the task or module level, or even a config parameter.

edx/analytics/tasks/load_internal_reporting_course.py
```diff
@@ -0,0 +1,301 @@
...
+    )
+    api_access_token = luigi.Parameter(
+        config_path={'section': 'course-structure', 'name': 'access_token'}, significant=False
+    )
+
+
+class PullCourseStructureAPIData(LoadDCourseMixin, luigi.Task):
+    """Call the course structure API and place the resulting JSON on S3."""
+
+    def run(self):
```
mulby, line 42: you should call "remove_output_on_overwrite()" here.

edx/analytics/tasks/load_internal_reporting_course.py
```diff
@@ -0,0 +1,301 @@
...
+        self.remove_output_on_overwrite()
+        with self.input().open('r') as input_file:
+            course_structure = json.load(input_file)
+            with self.output().open('w') as output_file:
+                courses_list = course_structure.get('results')
+                if not courses_list:  # If there are no courses, or 'results' is not a key in the json, output nothing.
+                    return
+                for course in courses_list:
+                    # To maintain robustness, ignore any non-dictionary data that finds its way into the API response.
+                    if not type(course) == dict:
```
mulby: I would change this to "isinstance(course, dict)".

Side note: ideally we wouldn't have to do a type check here. Instead we could rely on duck typing and catching exceptions to make sure that we do the right thing when we encounter data in an unexpected format.

edx/analytics/tasks/load_internal_reporting_course.py
```diff
@@ -0,0 +1,301 @@
...
+                        continue
+                    timestamp_regex = r'^([^T]*)T([^Z]*)Z'
```
mulby: Consider using the ciso8601 which can very rapidly parse ISO 8601 timestamps without the need for a regex like this.

edx/analytics/tasks/load_internal_reporting_course.py
jrowan: I changed it to api_root_url.  Someone now needs to change the corresponding section in the three secure configs in the secure repo to say api_root_url and not api_root.

edx/analytics/tasks/load_internal_reporting_course.py
```diff
@@ -0,0 +1,301 @@
...
+        The form is {warehouse_path}/course_structure/dt={CCYY-mm-dd}/courses.tsv.
+        """
+        date_string = self.run_date.strftime('%Y-%m-%d')  # pylint: disable=no-member
+        partition_path_spec = HivePartition('dt', date_string).path_spec
+        url_with_filename = url_path_join(self.warehouse_path, "course_structure",
+                                          partition_path_spec, "courses.tsv")
+        return get_target_from_url(url_with_filename)
+
+
+class LoadCourseStructureAPIDataIntoHive(LoadDCourseMixin, ImportIntoHiveTableTask):
```
mulby: Use `HiveTableTask` instead of `ImportIntoHiveTableTask`, the latter is deprecated (although not marked as deprecated...)

mulby: See the enrollment.py workflow for a good example of this.

edx/analytics/tasks/load_internal_reporting_course.py
```diff
@@ -0,0 +1,301 @@
...
+        return self.run_date.strftime('%Y-%m-%d')
+
+    @property
+    def columns(self):
+        return [
+            ('course_id', 'STRING'),
+            ('course_org_id', 'STRING'),
+            ('course_number', 'STRING'),
+            ('course_run', 'STRING'),
+            ('course_start_string', 'STRING'),
```
mulby: In Hive it is common to represent dates as strings, so I'd be OK with just calling this "course_start"

edx/analytics/tasks/load_internal_reporting_course.py
```diff
@@ -0,0 +1,301 @@
...
+            SELECT
+              sce.course_id
+            , cs.course_org_id
+            , cs.course_number
+            , cs.course_run
+            , cs.course_start_string
+            , cs.course_end_string
+            , cs.course_name
+            FROM course_structure cs
+            RIGHT OUTER JOIN courses_with_enrollments sce ON cs.course_id = sce.course_id
```
mulby: I tend to prefer setting up the query as a `LEFT OUTER JOIN`

edx/analytics/tasks/tests/test_course_information.py
```diff
@@ -0,0 +1,209 @@
...
+    def test_course_missing_data(self):
+        """With a course with some data missing, we expect a row with null values in some columns."""
+        course_with_missing_data = {"results": [{"id": "foo", "org": "bar"}]}
+        data = self.run_task(json.dumps(course_with_missing_data))
+        # We expect an entry in the list of courses, since there is a course in the list.
+        self.assertEquals(data.shape[0], 1)
+        # We expect nulls for the columns aside from course id and course org id.
+        expected = {'course_id': 'foo',
+                    'course_org_id': 'bar',
+                    'course_number': '\N',  # pylint: disable-msg=anomalous-unicode-escape-in-string
```
mulby: I'm pretty sure you can put this plyint disable flag at the method level.

mulby:
Finished my pass. Generally it's looking good!

edx/analytics/tasks/load_internal_reporting_course.py
jrowan: Deprecated or not, ImportIntoHiveTableTask can do overwriting, while HiveTableTask can't*.  Given that we do in fact use both of the two elsewhere in the code, I'm inclined to stick with ImportIntoHiveTable task for now (we can clean it up when we clean up the other instances if we ever get around to cleaning up our Hive tasks/targets).  

\* Using a HiveTableTask instead produces the following error output, with the params that aren't necessary removed:

15:51:54 2015-08-31 15:51:54,641 ERROR 30382 [luigi-interface] worker.py:304 - [pid 30382] Worker Worker(<parameters>) failed    LoadCourseStructureAPIDataIntoHive(overwrite=True, warehouse_path=<warehouse path>, api_root_url=<api_root_url>, run_date=2015-08-31)
15:51:54 Traceback (most recent call last):
15:51:54   File "/var/lib/analytics-tasks/automation/venv/lib/python2.7/site-packages/luigi/worker.py", line 292, in _run_task
15:51:54     task.run()
15:51:54   File "/var/lib/analytics-tasks/automation/venv/lib/python2.7/site-packages/luigi/hadoop.py", line 567, in run
15:51:54     self.job_runner().run_job(self)
15:51:54   File "/var/lib/analytics-tasks/automation/venv/lib/python2.7/site-packages/edx/analytics/tasks/util/hive.py", line 132, in run_job
15:51:54     job.remove_output_on_overwrite()
15:51:54   File "/var/lib/analytics-tasks/automation/venv/lib/python2.7/site-packages/edx/analytics/tasks/util/overwrite.py", line 56, in remove_output_on_overwrite
15:51:54     self.output().remove()
15:51:54 AttributeError: 'HivePartitionTarget' object has no attribute 'remove'

edx/analytics/tasks/load_internal_reporting_course.py
```diff
@@ -0,0 +1,312 @@
...
+        response = requests.get(url=api_url, headers=headers)
+        if response.status_code != requests.codes.ok:  # pylint: disable=no-member
+            msg = "Encountered status {} on request to API for {}".format(response.status_code, self.run_date)
+            raise Exception(msg)
+        with self.output().open('w') as output_file:
+            output_file.write(response.content)
+
+    def output(self):
+        """Output is in the form {warehouse_path}/courses_raw/dt={CCYY-MM-DD}/course_structure.json"""
+        date_string = "dt=" + self.run_date.strftime('%Y-%m-%d')  # pylint: disable=no-member
```
brianhw: You can actually put the "dt=" in the argument to strftime, rather than appending.

edx/analytics/tasks/load_internal_reporting_course.py
```diff
@@ -0,0 +1,312 @@
...
+        url_with_filename = url_path_join(self.warehouse_path, 'courses_raw', date_string,
+                                          "course_structure.json")
+        return get_target_from_url(url_with_filename)
+
+
+class ProcessCourseStructureAPIData(LoadInternalReportingCourseMixin, luigi.Task):
+    """Process the information from the course structure API and write it to a tsv."""
+    run_date = luigi.DateParameter()
```
brianhw: Is this needed?  Looks like it should already be defined in the mixin.

edx/analytics/tasks/load_internal_reporting_course.py
```diff
@@ -0,0 +1,312 @@
...
+
+    def requires(self):
+        kwargs = {
+            'run_date': self.run_date,
+            'warehouse_path': self.warehouse_path,
+        }
```
brianhw, line 71: Shouldn't this include the api_root_url at least, if not also the access token?  Otherwise they can _only_ use the default values.

brianhw, line 71: And overwrite?

edx/analytics/tasks/load_internal_reporting_course.py
```diff
@@ -0,0 +1,312 @@
...
+# To ensure the API call pulls all courses, set this to be larger than the number of courses
+PAGE_SIZE = 10000
+
+
+class LoadInternalReportingCourseMixin(WarehouseMixin, OverwriteOutputMixin):
+    """
+    Mixin to handle parameters common to the tasks involved in loading the internal reporting course table,
+    including calling the course structure API.
+    """
+    run_date = luigi.DateParameter()
```
brianhw: Should this have a default value of today?

edx/analytics/tasks/load_internal_reporting_course.py
```diff
@@ -0,0 +1,312 @@
...
+        kwargs = {
+            'run_date': self.run_date,
+            'warehouse_path': self.warehouse_path,
+            'overwrite': self.overwrite
+        }
+        return ProcessCourseStructureAPIData(**kwargs)
+
+    @property
+    def table_name(self):
+        return 'course_structure'
```
brianhw, line 137: I can't remember -- were we adding a prefix to table names destined for "internal reporting"?

edx/analytics/tasks/load_internal_reporting_course.py
```diff
@@ -0,0 +1,312 @@
...
+    def table_location(self):
+        return url_path_join(self.warehouse_path, self.table_name)
+
+    @property
+    def partition_date(self):
+        return self.run_date.isoformat()
+
+    # @property
+    # def partition(self):
+    #     return HivePartition('dt', self.run_date.isoformat())  # pylint: disable=no-member
```
brianhw: Shouldn't this be removed?

edx/analytics/tasks/load_internal_reporting_course.py
```diff
@@ -0,0 +1,312 @@
...
+            ('course_start', 'STRING'),
+            ('course_end', 'STRING'),
+            ('course_name', 'STRING')
+        ]
+
+
+class GetCoursesFromStudentCourseEnrollmentTask(LoadInternalReportingCourseMixin, HiveTableFromQueryTask):
+    """
+    Finds all courses from the LMS enrollments table to make sure we haven't left out any courses with enrollments
+    which, for whatever reason, can't be found in the course structure API.
```
brianhw, line 167: Do we know of any reasons?  Migrations, etc?

edx/analytics/tasks/load_internal_reporting_course.py
```diff
@@ -0,0 +1,312 @@
...
+        The form is {warehouse_path}/course_structure/dt={CCYY-mm-dd}/courses.tsv.
+        """
+        date_string = self.run_date.strftime('%Y-%m-%d')  # pylint: disable=no-member
+        partition_path_spec = HivePartition('dt', date_string).path_spec
+        url_with_filename = url_path_join(self.warehouse_path, "course_structure",
+                                          partition_path_spec, "courses.tsv")
+        return get_target_from_url(url_with_filename)
+
+
+class LoadCourseStructureAPIDataIntoHive(LoadInternalReportingCourseMixin, ImportIntoHiveTableTask):
```
brianhw, line 122: To return to the HiveTableTask versus ImportIntoHiveTableTask thread, I looked the classes over, and they are very similar.  The main difference I see is in overwrite -- that HiveTableTask implements overwrite by calling job.remove_output_on_overwrite(), while ImportIntoHiveTableTask explicitly sets self.attempted_removal to True.  Neither seems quite right here, and I'm not sure what is best.  One is to not call remove() (by redefining the remove_output_on_ovewrite() implementation).  Another is to actually define a remove() method on the partition target to actually remove the partition from Hive.  I'm guessing these are out of scope for now.  But perhaps we need a bug created since overwrite doesn't work for HiveTableTask.

edx/analytics/tasks/load_internal_reporting_course.py
```diff
@@ -0,0 +1,312 @@
...
+    """Load the processed course structure API data into Hive."""
+    run_date = luigi.Parameter()
```
brianhw: Again, assuming this is in the mixin.

edx/analytics/tasks/load_internal_reporting_course.py
```diff
@@ -0,0 +1,312 @@
...
+
+    def requires(self):
+        kwargs = {
+            'run_date': self.run_date,
+            'warehouse_path': self.warehouse_path,
+            'overwrite': self.overwrite
+        }
```
brianhw, line 132: again, pass api args.

edx/analytics/tasks/load_internal_reporting_course.py
```diff
@@ -0,0 +1,312 @@
...
+
+    def requires(self):
+        """
+        This task reads from student_coursenrollment as well as course structure API results, so require that they be
+        loaded into Hive (via MySQL loads into Hive or via the pipeline as needed).
+        """
+        return [GetCoursesFromStudentCourseEnrollmentTask(n_reduce_tasks=self.n_reduce_tasks,
+                                                          warehouse_path=self.warehouse_path, run_date=self.run_date),
+                LoadCourseStructureAPIDataIntoHive(run_date=self.run_date, warehouse_path=self.warehouse_path,
+                                                   overwrite=True)]
```
brianhw: I'm not clear on the scope of the overwrite parameter here.  We always want this, and just this, to be overwritten, and to be always overwritten?  I'm not seeing what problem this is solving.

edx/analytics/tasks/load_internal_reporting_course.py
```diff
@@ -0,0 +1,312 @@
...
+            SORT BY sce.course_id ASC
+            """
+
+
+class LoadInternalReportingCourseToWarehouse(LoadInternalReportingCourseMixin, VerticaCopyTask):
+    """
+    Loads the course table from Hive into the Vertica data warehouse.
+    """
+    n_reduce_tasks = luigi.Parameter()
+    overwrite = luigi.BooleanParameter(default=False)
```
brianhw: If this is redefined here for a reason (i.e. to have a default), then add a comment to that effect.   It's overriding the existing definition, which looks pretty much identical to me.  Is this needed?

edx/analytics/tasks/load_internal_reporting_course.py
jrowan, line 71: I don't include overwrite for the pull from api task because we may want to re-run the processing on a later date and don't want to overwrite e.g. Saturday's API pull with Monday's data.

edx/analytics/tasks/load_internal_reporting_course.py
```diff
@@ -0,0 +1,312 @@
...
+
+    @property
+    def auto_primary_key(self):
+        """No automatic primary key is needed; course id is a unique identifier."""
+        return None
+
+    @property
+    def columns(self):
+        return [
+            ('course_id', 'VARCHAR(200) PRIMARY KEY'),
```
brianhw: We should be consistent with sizes, if we can.  I thought we were changing course_id to be 255 or 256....

edx/analytics/tasks/tests/acceptance/test_internal_reporting_course.py
```diff
@@ -0,0 +1,75 @@
...
+        self.upload_data()
+
+    def upload_data(self):
+        """
+        Puts the test course structure information where the processing task would look for it, bypassing
+        calling the actual API
+        """
+        src = os.path.join(self.data_dir, 'input', self.INPUT_FILE)
+        # IMPORTANT: this path should be of the same format as the path that DailyPullCatalogTask uses for output.
+        dst = url_path_join(self.warehouse_path, "courses_raw", "dt={date}".format(date=self.DATE.strftime('%Y-%m-%d')),
```
brianhw: again, you can move the 'dt=' into the strftime.

edx/analytics/tasks/load_internal_reporting_course.py
jrowan, line 137: My understanding is that we add that prefix to tables that are Hive copies of what is about to be loaded into vertica, and not the tables used to generate them (since in principle that information could be used in other places aside from the internal reports warehouse).

edx/analytics/tasks/load_internal_reporting_course.py
jrowan, line 167: I don't know of any specific reasons, and I didn't see in documentation in the kettle transform about why either.  Migrations sounds like it could be the reason.

brianhw:
LGTM.   :+1:

edx/analytics/tasks/load_internal_reporting_course.py
brianhw, line 122: James created AN-5859 to capture this.

PR #151, opened 2015-08-26
Address lock contention problems writing to Vertica

jrowan:
DELETE operations in Vertica acquire exclusive (X) table-level locks.  Since we occasionally will want to use the VerticaCopyTask in an overwriting setting, we need to be able to support deleting rows from the marker table during the execution of that task.  This PR moves the deletion of the marker table rows from before the start of the streaming copy to afterwards, immediately before the insertion of the new row into the marker table.  This avoids the problem of Vertica loading tasks holding exclusive locks on the marker table for too long and causing lock wait timeouts for other Vertica loading tasks trying to concurrently write to the same schema.

@mulby @jab5569

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -7,3 +7,4 @@ Jason Bau <jbau@stanford.edu>
 Justin Helbert <jhelbert@edx.org>
 Will Daly <will@edx.org>
 Dylan Rhodes <dylanr@stanford.edu>
+James Rowan <jrowan@mit.edu>
--- a/edx/analytics/tasks/vertica_load.py
+++ b/edx/analytics/tasks/vertica_load.py
@@ -42,6 +42,15 @@ class VerticaCopyTaskMixin(OverwriteOutputMixin):
 class VerticaCopyTask(VerticaCopyTaskMixin, luigi.Task):
     """
     A task for copying into a Vertica database.
+
+    Note that the default behavior if overwrite is true is to first delete the existing
+    contents of the table being written to and then delete the entire history of table
+    updates corresponding to writes to that table, not just table updates with the same
+    update id (i.e. updates corresponding to writes done by a task with the same exact
+    task name and set of parameters).
+
+    Overwrite init_copy and init_touch if you want a different overwrite behavior in a
+    subclass.
     """
     required_tasks = None
     output_target = None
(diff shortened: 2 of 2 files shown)


mulby:
:+1:

AUTHORS
```diff
@@ -7,3 +7,4 @@ Jason Bau <jbau@stanford.edu>
 Justin Helbert <jhelbert@edx.org>
 Will Daly <will@edx.org>
 Dylan Rhodes <dylanr@stanford.edu>
+James Rowan <jrowan@mit.edu>
```
brianhw, line 10: Finally!!  (good catch...)

edx/analytics/tasks/vertica_load.py
```diff
@@ -187,16 +191,54 @@ def init_copy(self, connection):
...
+        which might in general take longer than the 5 minutes that Vertica is
+        willing to wait for a lock before giving up and throwing an error.
+
         Any code here will be formed in the same transaction as the
         main copy, just prior to copying data. Example use cases
         include truncating the table or removing all data older than X
         in the database to keep a rolling window of data available in
         the table.
+
+        Note that this method acquires an exlusive (X) lock on the table
```
brianhw: nit:  s/exlusive/exclusive/

edx/analytics/tasks/vertica_load.py
```diff
@@ -227,6 +263,8 @@ def init_copy(self, connection):
         # successfully executes the DELETEs), so we flush the message buffer.
         connection.cursor().flush_to_query_ready()
 
+
+
```
brianhw: nit: I assume these two blank lines break pep8.

edx/analytics/tasks/vertica_load.py
```diff
@@ -42,6 +42,10 @@ class VerticaCopyTaskMixin(OverwriteOutputMixin):
 class VerticaCopyTask(VerticaCopyTaskMixin, luigi.Task):
     """
     A task for copying into a Vertica database.
+
+    Note that the default behavior if overwrite is true is to delete the entire table
+    and the entire history of table updates, not just any with the same update id.
```
brianhw: nit:  would suggest a slightly different wording, as I was confused the first time as to which table was deleted.    So "... is to first delete the existing contents of the table being written to ..." in place of the first clause.  And a second sentence for the deletion of the marker-table entries.

brianhw: Also, it might be helpful to say a few words about what an update id is.

brianhw:
Code looks good.  Just a few requests for comment changes, and then :+1:

PR #153, opened 2015-08-28
Use the XML reporting API

mulby:
This is close enough to done that I thought it was worth putting up for review. Feel free to take a pass.

I'm still working on adding unit tests, specifically around the luigi task (I only have the happy path at the moment).

@jab5569 @brianhw

TODO list:
- [x] PEP8 and pylint cleanup
- [x] Lots of docstrings
- [x] Tests for reports that aren't done immediately
- [x] Tests for the task handling errors
- [x] Tests for the timeout behavior

### DIFF
--- a/edx/analytics/tasks/reports/paypal.py
+++ b/edx/analytics/tasks/reports/paypal.py
@@ -1,10 +1,22 @@
-import datetime
-import logging
-import luigi
-import paypalrestsdk
-import json
+"""
+This module uses the PayPal Payflow Gateway Reporting API to gather transaction data from PayPal. Note that we attempted
+to make use of the REST API, but were unable to find a viable solution using that product. Here is the general
+documentation for the API:
 
-from luigi.date_interval import DateInterval
+https://developer.paypal.com/docs/classic/payflow/reporting/
+"""
+
+import datetime
+import xml.etree.cElementTree as ET
+from cStringIO import StringIO
+import logging
+from collections import namedtuple, OrderedDict
+from decimal import Decimal
+import time
+
+import luigi
+from luigi.configuration import get_config
+import requests
 
 from edx.analytics.tasks.url import get_target_from_url
 from edx.analytics.tasks.url import url_path_join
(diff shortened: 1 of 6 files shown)


brianhw:
LGTM.  

One question:  do we want to add some information to our Transactions in general that Paypal provides and Cybersource doesn't (at present)?  In particular, the datetime instead of merely the date.  

Also, the Cybersource and transaction code is very USD-centric, with the assumption of two decimal places.  This seems to make different assumptions about representation, but do we have any idea what will happen if/when we get different currencies?

mulby:
@brianhw I updated this branch with additional tests, docstrings and pylint/pep8 fixes

edx/analytics/tasks/reports/paypal.py
```diff
@@ -18,183 +30,656 @@
...
+    Find the direct descendant of `node` with the tag `child_name`. If the no child is found with that tag, an error is
+    raised.
+
+    Args:
+        node: an ElementTree.Element object representing a node in an XML document that is expected to have a child
+            with the name specified by the `child_name` parameter.
+        child_name: a string containing the name of the tag that is expected to be a direct descendant of `node`.
 
+    Raises:
+        PaypalMalformedResponseException: If the document does not have the expected structure.
```
brianhw: returns an Error now, not an Exception.

edx/analytics/tasks/reports/paypal.py
```diff
@@ -18,183 +30,656 @@
...
 
-    def initialize(self):
-        log.debug('Initializing paypalrestsdk')
-        paypalrestsdk.configure({
-            'mode': self.client_mode,
-            'client_id': self.client_id,
-            'client_secret': self.client_secret
-        })
+    Raises:
+        PaypalMalformedResponseException: If the document does not have the expected structure.
```
brianhw: returns an Error now, not an Exception.

brianhw:
Two doc nits, otherwise LGTM. :+1:

edx/analytics/tasks/reports/paypal.py
mulby: Done

edx/analytics/tasks/reports/paypal.py
mulby: Done

johnalbaker:
:+1:

PR #155, opened 2015-09-04
Precompute total seconds watched for each video

HassanJaveed84:
Added  'total_viewed_seconds' for video table that captures total viewing time in seconds per video across all users. 

@mulby @brianhw @jab5569


mulby:
Looks great!

:+1:

please squash before merging.

PR #156, opened 2015-09-08
Support a single financial reporting entry point

mulby:
I think this is the minimal set of changes needed to get everything working end to end.

Job to run this is here: http://scheduler.analytics.edx.org/job/finance-report/

Outstanding action items that must be completed:
- [x] Ensure the changes to mapreduce.py don't break jobs that don't use a manifest, or solve this problem more elegantly.
- [x] PEP8 and pylint (deferred)
- [x] Ensure unit tests are passing
- [x] Ensure acceptance tests are passing

A few notes:
- I had to move the pyopenssl injection into the local launcher since it isn't needed by the hadoop slaves and we don't ship that module out to the slaves during job execution.
- Since we now import the various cybersource and paypal tasks as part of the hadoop workflow, I'm shipping the requests module out during all jobs to prevent import errors.
- The dynamic manifest replacement task has only previously been used with event logs, never with an actual task that needs to run to produce output. It was substituting those _real_ dependencies on _real_ tasks with just UncheckedExternalURL tasks that point to the output of those tasks. This meant that the payment pulling tasks were never actually run, the job would just expect their output to magically exist - which is obviously incorrect. The hack solution to this was to simply add the output of self.requires() to the list of dependencies... I'm looking into a more comprehensive solution now.
- As discussed in passing, I introduced a pattern where multiple cybersource configuration sections can appear in the config file, they look like "cybersource:<merchant_id>". They are dynamically loaded out of band from the normal parameter loading procedure and can only be changed by changing the config file.
- I also replicated the pattern of allowing an "output-root" to override the warehouse path, but by default everything gets written to the warehouse path.
- The "interval_start", "interval_end" and "interval" pattern is also re-used quite a bit here to ensure we gather all of the logs for the various payment providers starting at some known point in time.

This is ready for first-pass review. I'll be working on the "must" items above, let me know if you find any others.

Reviewers: @jab5569 @brianhw


edx/analytics/tasks/reports/cybersource.py
```diff
@@ -5,40 +5,34 @@
 import logging
 import requests
 import luigi
```
johnalbaker, line 7: Is this needed given the 'from...' lines below.

mulby, line 7: Yes, many of the parameter classes are imported directly from the luigi module: `luigi.Parameter()` etc

edx/analytics/tasks/reports/cybersource.py
```diff
@@ -188,10 +179,11 @@ def output(self):
         return get_target_from_url(url_with_filename)
 
 
-class IntervalPullFromCybersourceTask(PullFromCybersourceTaskMixin, WarehouseMixin, luigi.Task):
+class IntervalPullFromCybersourceTask(PullFromCybersourceTaskMixin, WarehouseMixin, luigi.WrapperTask):
     """Determines a set of dates to pull, and requires them."""
 
-    interval = luigi.DateIntervalParameter()
+    interval = luigi.DateIntervalParameter(default=None)
```
johnalbaker, line 185: Is this needed if you're setting self.interval down on line 196/197?

edx/analytics/tasks/util/hive.py
```diff
@@ -196,7 +196,7 @@ def query(self):
         return create_table_statements + textwrap.dedent(full_insert_query)
 
     def output(self):
-        return get_target_from_url(self.partition_location)
+        return get_target_from_url(self.partition_location + '/')
```
brianhw: What if partition_location already ends with a slash?  Another idea is to pass an optional argument (e.g. is_directory=True or targets_from_directory=True or something) to get_target_from_url to make sure it has the effect of ending with a slash, to make the behavior more explicit and less encoded in the URL with a magic trailing slash.

edx/analytics/tasks/reports/cybersource.py
```diff
@@ -5,40 +5,34 @@
 import logging
 import requests
 import luigi
+from luigi.configuration import get_config
+from luigi import date_interval
 
 from edx.analytics.tasks.url import get_target_from_url
-from edx.analytics.tasks.url import url_path_join
+from edx.analytics.tasks.url import url_path_join, UncheckedExternalURL
```
brianhw: Weird.  I don't see this listed anywhere below as having been added.  So is it not used after all, or was it not finding the import?

mulby: Oops, not used after all... I was using it for testing

edx/analytics/tasks/util/hive.py
mulby: Yeah - I've been thinking about making this change for a while now... I'm concerned it will be rather pervasive. We also have to figure out the issue of CLI parameters which currently are just URLs but sometimes need to be marked as directories or just plain files...

One idea would be to support a query string argument?

s3://foo/bar?type=directory

Not sure what affect that would have on existing jenkins jobs, though... we would need to find all the jobs that are relying on this feature and switch them over to use the query string.

jab5569: How will this affect other tasks which already have the '/' set?

brianhw:
I don't remember any discussion of the cybersource:<merchant_id> approach, but it looks like it addresses the start_date configuration well.

I'm confused about the issue with the dynamic manifest.  It only makes sense to use it when the files pre-exist, I thought, so I figured with the switch to actual dependency tasks, the manifest would no longer be needed.  Hmm.  Having written that, I see that I'm wrong.  The purpose of the manifest is to deal with a large number of inputs, whether they are pre-existing or not.  Hmm.   So this is more complicated because we want to continue to use the dynamic manifest logic.  Hmm.

johnalbaker:
@mulby : generally looks great. The only other nit I have is the use of import_date vs interval vs interval_end. Seems as if we should either pass in interval_start + interval_end  vs just passing in import_date and getting 'interval_start' from config.

johnalbaker:
:+1:

edx/analytics/tasks/reports/cybersource.py
mulby, line 185: Yup, it can be specified to override interval_start, interval_end

mulby:
@brianhw now includes the manifest input target instead of the URLManifestTask

mulby:
I'm currently running the full suite of acceptance tests given the core change to the manifest handling.

edx/analytics/tasks/util/hive.py
brianhw: That seems out of scope.  But I'm wondering if this just needs to be conditional.  Maybe it's enough to just add .rstrip('/') to the partition_location to avoid dupes.

edx/analytics/tasks/util/manifest.py
```diff
@@ -14,63 +13,66 @@
...
+    else:
+        log.debug(
+            'Directly processing files since %d inputs are less than the threshold %d', len(targets), threshold
+        )
+        return targets
+
+
+def create_manifest_target(manifest_id, targets):
+    # If we are running locally, we need our manifest file to be a local file target, however, if we are running on
+    # a real Hadoop cluster, it has to be an HDFS file so that the input format can read it. Luigi, makes it a little
```
brianhw: nit: remove comma after Luigi....

brianhw:
LGTM.  I would just be conservative/defensive about the appended slashes, and then :+1:

PR #159, opened 2015-09-15
Removed end-of-video segments that show a sharp drop in users

HassanJaveed84:
@brianhw  @mulby 
I am not sure if this is the right way to go about it. It basically just looks for a spike in num_users from the end to the start and drops those segments.

### DIFF
--- a/config/test.cfg
+++ b/config/test.cfg
@@ -100,3 +100,6 @@ interval_start = 2015-09-01
 [payment]
 cybersource_merchant_ids = test
                            empty_test
+
+[videos]
+dropoff_threshold = 0.05
--- a/edx/analytics/tasks/tests/test_video.py
+++ b/edx/analytics/tasks/tests/test_video.py
@@ -4,6 +4,7 @@ import json
 
 from mock import patch, MagicMock
 from ddt import ddt, data, unpack
+from mock import sentinel
 
 from edx.analytics.tasks.video import (
     UserVideoViewingTask, VideoUsageTask, VIDEO_VIEWING_SECONDS_PER_SEGMENT, VIDEO_UNKNOWN_DURATION
@@ -11,7 +12,6 @@ from edx.analytics.tasks.video import (
 from edx.analytics.tasks.tests import unittest
 from edx.analytics.tasks.tests.opaque_key_mixins import InitializeOpaqueKeysMixin, InitializeLegacyKeysMixin
 from edx.analytics.tasks.tests.map_reduce_mixins import MapperTestMixin, ReducerTestMixin
-
 
 @ddt
 class UserVideoViewingTaskMapTest(InitializeOpaqueKeysMixin, MapperTestMixin, unittest.TestCase):
@@ -1013,3 +1013,48 @@ class VideoUsageTaskReducerTest(ReducerTestMixin, unittest.TestCase):
                 UsageColumns.USERS_AT_END: 1,
             }
         ])
+
+@ddt
+class GetFinalSegmentTest(unittest.TestCase):
+
+    def setUp(self):
+        self.task = VideoUsageTask(interval=sentinel.ignored, output_root=sentinel.ignored)
+        self.usage_map = self.generate_usage_map(20)
+
+    def generate_segment(self, num_users, num_views):
+        return {
+            'users': ['user{}'.format(i) for i in range(num_users)],
+            'views': num_views
+        }
+
+    def generate_usage_map(self, num_segments):
+        usage_map = {}
+        for i in range(num_segments):
+            usage_map[i] = self.generate_segment(20, 20)
+        return usage_map
+
+    @data(
+        (15, 25, 14),
+        (20, 30, 19),
+    )
+    @unpack
+    def test_with_sudden_dropoff(self, start, end, result):
+        """Sharp drop-off in num_users(20 to 1), triggers the cutoff"""
+        for i in range(start, end):
+            self.usage_map[i] =  self.generate_segment(1,1)
+        self.assertEqual(self.task.get_final_segment(self.usage_map), result)
+
+    def test_without_dropoff(self):
+        """No dropoff"""
+        self.assertEqual(self.task.get_final_segment(self.usage_map), 19)
+
+    def test_with_gradual_dropoff(self):
+        """Gradual dropoff of num_users, does not trigger the cutoff"""
+        for i, j in zip(range(20, 39), range(19, 0, -1)):
+            self.usage_map[i] = self.generate_segment(j, j)
+        self.assertEqual(self.task.get_final_segment(self.usage_map), 38)
+
+    def test_with_dropoff_no_cutoff(self):
+        """Drop-off from 20 to 2 num_users, however, cutoff won't trigger as it does not meet the threshold."""
+        self.usage_map[30] = self.generate_segment(2,2)
+        self.assertEqual(self.task.get_final_segment(self.usage_map), 30)
--- a/edx/analytics/tasks/video.py
+++ b/edx/analytics/tasks/video.py
@@ -333,6 +333,7 @@ class VideoUsageTask(VideoTableDownstreamMixin, MapReduceJobTask):
     """Aggregates usage statistics for video segments across individual user 'viewings'."""
 
     output_root = luigi.Parameter()
+    dropoff_threshold = luigi.FloatParameter(config_path={'section': 'videos', 'name': 'dropoff_threshold'})
 
     def requires(self):
         # Define path so that data could be loaded into Hive, without actually requiring the load to be performed.
@@ -408,7 +409,7 @@ class VideoUsageTask(VideoTableDownstreamMixin, MapReduceJobTask):
         # If we don't know the duration of the video, just use the final segment that was
         # actually viewed to determine users_at_end.
         if video_duration == VIDEO_UNKNOWN_DURATION:
-            final_segment = max(usage_map.keys())
+            final_segment = self.get_final_segment(usage_map)
         else:
             final_segment = self.snap_to_last_segment_boundary(float(video_duration))
 
@@ -429,6 +430,25 @@ class VideoUsageTask(VideoTableDownstreamMixin, MapReduceJobTask):
                 len(stats.get('users', [])),
                 stats.get('views', 0),
             )
+            if video_duration == VIDEO_UNKNOWN_DURATION and segment == final_segment: break
+
+
+    def get_final_segment(self, usage_map):
+        """
+        Identifies the final segment by looking for a sharp drop in number of users per segment.
+        Needed as some events appear after the actual end of videos.
+        """
+        final_segment = last_segment = max(usage_map.keys())
+        last_segment_num_users = len(usage_map[last_segment]['users'])
+        for segment in sorted(usage_map.keys(), reverse=True)[1:]:
+            stats = usage_map[segment]
+            current_segment_num_users = len(stats.get('users', []))
+            if last_segment_num_users <= current_segment_num_users * self.dropoff_threshold:
+                final_segment = segment
+                break
+            last_segment_num_users = current_segment_num_users
+            last_segment = segment
+        return final_segment
 
     def snap_to_last_segment_boundary(self, second):
         """Maps a time_offset to a segment index."""
--- a/config/devstack.cfg
+++ b/config/devstack.cfg
@@ -54,3 +54,7 @@ geolocation_data = hdfs://localhost:9000/edx-analytics-pipeline/geo.dat
 
 [calendar]
 interval = 2012-01-01-2020-01-01
+
+[videos]
+dropoff_threshold = 0.05
+
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -32,6 +32,7 @@ tornado==3.1.1 		# Apache 2.0
 urllib3==1.10.4         # MIT
 vertica-python==0.5.1   # MIT
 wsgiref==0.1.2 		# ZPL
+pip==1.5.6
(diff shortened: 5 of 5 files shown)


edx/analytics/tasks/video.py
```diff
@@ -405,6 +407,18 @@ def reducer(self, key, viewings):
                 users.add(username)
                 stats['views'] = stats.get('views', 0) + 1
 
+        last_segment = max(usage_map.keys())
+        last_segment_num_users = len(usage_map[last_segment]['users'])
+        for segment in sorted(usage_map.keys()[0:-1], reverse=True):
```
mulby: I think you want to apply the slice after the sort, something like:

``` python
sorted(usage_map.keys(), reverse=True)[1:]
```

edx/analytics/tasks/video.py
```diff
@@ -405,6 +407,18 @@ def reducer(self, key, viewings):
...
+            stats = usage_map[segment]
+            current_segment_num_users = len(stats.get('users', []))
+            if last_segment_num_users <= current_segment_num_users * VIDEO_DROPOFF_THRESHOLD:
+                for segment_dropoff in xrange(last_segment, max(usage_map.keys()) + 1):
+                    del usage_map[segment_dropoff]
```
mulby: I think this will fail if there is a gap in the data... for example: 100, 102, 103 etc.

We probably want something like this instead: `usage_map.pop(segement_dropoff, None)` or a try-except that catches the key error.

Meta point - I would include this in your test cases when you write unit tests for this code.

edx/analytics/tasks/video.py
```diff
@@ -405,6 +407,18 @@ def reducer(self, key, viewings):
                 users.add(username)
                 stats['views'] = stats.get('views', 0) + 1
 
+        last_segment = max(usage_map.keys())
+        last_segment_num_users = len(usage_map[last_segment]['users'])
+        for segment in sorted(usage_map.keys()[0:-1], reverse=True):
+            stats = usage_map[segment]
+            current_segment_num_users = len(stats.get('users', []))
+            if last_segment_num_users <= current_segment_num_users * VIDEO_DROPOFF_THRESHOLD:
+                for segment_dropoff in xrange(last_segment, max(usage_map.keys()) + 1):
```
mulby: You may want to consider caching `max(usage_map.keys())` since it is computed earlier and involves a full traversal of a rather large dictionary.

edx/analytics/tasks/video.py
```diff
@@ -405,6 +407,18 @@ def reducer(self, key, viewings):
                 users.add(username)
                 stats['views'] = stats.get('views', 0) + 1
 
+        last_segment = max(usage_map.keys())
```
mulby: I would suggest moving this code to a function and adding a doc string that explains why it is necessary.

mulby:
I like the approach, we should try it out on some real data to see how effective it is at mitigating the issue.

Once we do that we should write unit tests to make sure we cover some of the edge cases here.

edx/analytics/tasks/video.py
```diff
@@ -405,6 +407,18 @@ def reducer(self, key, viewings):
...
+            if last_segment_num_users <= current_segment_num_users * VIDEO_DROPOFF_THRESHOLD:
+                for segment_dropoff in xrange(last_segment, max(usage_map.keys()) + 1):
+                    del usage_map[segment_dropoff]
+                break
+            last_segment_num_users = current_segment_num_users
+            last_segment = segment
+
         # If we don't know the duration of the video, just use the final segment that was
         # actually viewed to determine users_at_end.
         if video_duration == VIDEO_UNKNOWN_DURATION:
```
brianhw, line 411: A good start!  There are two kinds of modifications we want to make to this code.  One is to change how we estimate what the end of the video is.  The second will address how "users_at_end" is calculated below.  I assume this PR is addressing the first.  So I think we only want to do this in cases where we don't know the video's duration.  So I believe the above logic would go below this line.

edx/analytics/tasks/video.py
HassanJaveed84: Yes, I did, thanks!

brianhw: I don't understand the use of [0:-1] here.  You're doing that on the list of keys, before sorting.  Do you know what's at the end of that list?  But I assume the goal is to exclude the last_segment.

edx/analytics/tasks/video.py
HassanJaveed84, line 411: Thanks Brian! I'll move the code to a function as Gabe suggested and make the other improvements.

edx/analytics/tasks/video.py
```diff
@@ -38,6 +38,8 @@
 VIDEO_VIEWING_SECONDS_PER_SEGMENT = 5
 VIDEO_VIEWING_MINIMUM_LENGTH = 0.25  # seconds
 
+VIDEO_DROPOFF_THRESHOLD = 0.05 # percentage of users
```
brianhw: Going forward, I would suggest that this be made a Luigi parameter, with a default value read from config.

edx/analytics/tasks/video.py
brianhw, line 411: It may be a little simplified if we didn't bother with deleting entries.  Instead, we could just calculate a different value for final_segment, and then use final_segment to break from the for-loop below.

edx/analytics/tasks/video.py
brianhw: Just starting to think about how we might validate this approach (i.e. to look backwards for neighboring segments where the users in the second segment fall to 5% of the first segment).  Again, if this were to be applied only to videos with unknown duration, we could validate by checking what values seem to "work" on videos with known duration.  By applying it in such cases, we could determine how much video we might lose.

HassanJaveed84:
@mulby @brianhw 
Only applies the threshold for videos with unknown duration, also changed the way final_segment is calculated.

HassanJaveed84:
@mulby @brianhw 
Added unit test case for get_final_segment function. I have used PyYAML here to load the input data for the test. 
Made dropoff_threshold a luigi parameter for VideoUsageTask. I think it needs to be passed down from InsertToMysqlAllVideoTask ?

mulby:
@HassanJaveed84 - I'm actually OK with not passing down the drop-off parameter. Allowing it to be set in the config file is good enough for me.

config/test.cfg
```diff
@@ -69,3 +69,6 @@ access_token = acceptance
 [vertica-export]
 credentials = s3://fake/vertica_creds.json
 schema = testing
+
+[videos]
+dropoff_threshold = 0.05
```
mulby, line 105: Not sure where you mentioned this, but I agree, I would add this to devstack.cfg as well.

edx/analytics/tasks/tests/test_video.py
```diff
@@ -1013,3 +1016,20 @@ def test_end_view_with_duration(self):
...
             }
         ])
+
+class GetFinalSegmentTest(unittest.TestCase):
+
+    def setUp(self):
+        self.task = VideoUsageTask(dropoff_threshold = 0.10, interval=sentinel.ignored, output_root=sentinel.ignored)
+
+    def test_with_dropoff(self):
+        input = os.path.join(os.path.dirname(__file__), 'fixtures', 'input', 'video_usage_map_with_dropoff.yml')
```
mulby: I would write a small function that returns the full path given just the filename so that you don't have to repeat this logic in the tests. Or perhaps build most of the path in the setup and just append the filename in the actual test.

mulby: I think "input" is a builtin function in python, I would use a different name for this variable.

edx/analytics/tasks/tests/fixtures/input/video_usage_map_with_dropoff.yml
```diff
@@ -0,0 +1,45 @@
+0:
```
mulby: Looking at this input, I'm wondering if you couldn't manufacture "interesting" input that has similar characteristics in the test case. It may actually be more clear what case the input is illustrating.

edx/analytics/tasks/tests/test_video.py
```diff
@@ -1013,3 +1016,20 @@ def test_end_view_with_duration(self):
...
+        with open(input) as data_file:
+            usage_map = yaml.safe_load(data_file)
+        self.assertEqual(self.task.get_final_segment(usage_map), 10)
+
+    def test_without_dropoff(self):
```
mulby, line 1047: I'd like to see a test for the case where there is gradual drop-off. Also one where there is a sharp discontinuity but doesn't trigger the cutoff to illustrate the limitations of the algorithm.

This at least indicates to other developers that they are known limitations. I would put in comments describing those cases where the algorithm will do "the wrong thing" with a note about our duration comparison experiment and the results.

edx/analytics/tasks/tests/fixtures/input/video_usage_map_with_dropoff.yml
mulby: To be more clear, I'm wondering if it might be cleaner to programmatically define the input instead of explicitly declaring it in these files.  Something like:

``` python
def generate_segment(num_users, num_views):
    return {
        'users': ['user{}'.format(i) for i in range(num_users)],
        'views': num_views
    }

usage_map = {}
for j in range(6):
    usage_map[j] = generate_segment(20, 15)

# include a couple segments at the end with dramatically fewer users
usage_map[6] = generate_segment(1, 1)
usage_map[7] = generate_segment(1, 1)
```

I like aspects of the current yaml file - it's very clear and explicit, however, I'm nervous that it will be tedious to create interesting edge cases.

@brianhw do you have an opinion on this subject?

config/test.cfg
brianhw, line 105: and also to our dev/prod config in analytics-secure, when we get to it.

edx/analytics/tasks/tests/fixtures/input/video_usage_map_with_dropoff.yml
brianhw: Sorry to see so much work with the yaml files, but I am more strongly opinionated (or less polite :) about this than Gabe.  I much prefer the use of factory methods over the use of input files for this kind of testing.  It's hard enough to deal with the fixtures in acceptance tests (and I would prefer it if we could do more programmatic work there as well).  Fixtures make sense when one needs to model more complex, real-world type data, and that's not at all the case here.    If you want to get the data values to skip around, then using modulo operators will often give you that.   But I would argue that having a simple base generator to produce a usage_map of a given number of segments, and then others that modify the base, should provide an easier and clearer way to produce the test data.    It looks like there are two sets so far, and the second looks like it just appends to the first.

edx/analytics/tasks/tests/test_video.py
brianhw: Based on my earlier comment, I'm assuming this will be replaced.

edx/analytics/tasks/tests/test_video.py
brianhw, line 1047: Note that one can get a number of the additional cases by running with the same data but a different threshold.  It's much easier to change the one value to get more coverage.  As Gabe noted before, you can run the same tests over with different thresholds and expected final segment values using ddt.  It's pretty handy, and there are examples elsewhere in the pipeline unit tests.

HassanJaveed84:
@mulby @brianhw 
Thank you for the feedback and the guidance, much appreciated!
I updated the test case, though it looks obscure to me, what do you guys think?

edx/analytics/tasks/tests/test_video.py
```diff
@@ -1013,3 +1016,45 @@ def test_end_view_with_duration(self):
...
+            usage_map[i] = self.generate_segment(20, 20)
+        return usage_map
+
+    @data(
+        (15, 25, 14),
+        (20, 30, 19),
+    )
+    @unpack
+    def test_with_sudden_dropoff(self, start, end, result):
+        for i in range(start, end):
```
brianhw, line 1043: Yes, I definitely like this better.  I would just suggest adding a comment on each of these with a brief description of the data.  Or it could be a docstring, as you did for the last one.

edx/analytics/tasks/tests/test_video.py
```diff
@@ -1,9 +1,11 @@
 """Test student engagement metrics"""
 
 import json
+import yaml
```
brianhw: No longer needed....

edx/analytics/tasks/tests/test_video.py
```diff
@@ -12,6 +14,7 @@
 from edx.analytics.tasks.tests.opaque_key_mixins import InitializeOpaqueKeysMixin, InitializeLegacyKeysMixin
 from edx.analytics.tasks.tests.map_reduce_mixins import MapperTestMixin, ReducerTestMixin
 
+import os
```
brianhw: No longer needed

config/test.cfg
brianhw, line 105: I did this already.

HassanJaveed84:
@mulby @brianhw is there anything else that I can improve before we merge it?

brianhw:
Hassan, this looks great!   Next step is to rebase on master, since there seems to be a conflict that will need to be resolved.  And to squash all the intermediate commits in the process so that there's a single commit for this PR.  Let me know if you need help with that!

brianhw:
Oh yeah, for the record:  :+1:

HassanJaveed84:
@brianhw @mulby should I merge it?

mulby:
@HassanJaveed84 - sorry I've been out for the day.

This looks good to me :+1:

You can merge it whenever you are ready.

brianhw:
Yes, feel free to merge.  Or if you would prefer, let us know and we'll hit that button.

PR #160, opened 2015-09-15
Update handling of NULLs in reconcile and ed-services output.

brianhw:
Use \N for values in Hive tables that should be NULL (instead of "").

### DIFF
--- a/edx/analytics/tasks/reports/financial_report/ed_services_financial_report.py
+++ b/edx/analytics/tasks/reports/financial_report/ed_services_financial_report.py
@@ -64,10 +64,10 @@ class BuildEdServicesReportTask(DatabaseImportMixin, MapReduceJobTaskMixin, Hive
         return """
             SELECT
                 courses.course_id,
-                COALESCE(VP_COURSES.mode_slug, ""),
-                COALESCE(VP_COURSES.suggested_prices, ""),
-                COALESCE(VP_COURSES.min_price, ""),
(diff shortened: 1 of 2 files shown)


brianhw:
@mulby @jab5569 This now loads all records into Vertica, with appropriate NULL values in transaction_fee and expiration_datetime.

edx/analytics/tasks/reports/financial_report/ed_services_financial_report.py
```diff
@@ -67,7 +67,7 @@ def insert_query(self):
                 COALESCE(VP_COURSES.mode_slug, ""),
                 COALESCE(VP_COURSES.suggested_prices, ""),
                 COALESCE(VP_COURSES.min_price, ""),
-                COALESCE(CAST(VP_COURSES.expiration_datetime AS STRING), ""),
+                COALESCE(CAST(VP_COURSES.expiration_datetime AS STRING), "\\\\N"),
```
mulby, line 70: This is a user facing report... so it should really have empty strings instead of \N characters IMHO.

mulby, line 70: Scratch that last comment, users will be seeing it in Vertica now, so this is correct. We should probably remove a bunch of the other coalesce statements that are intended to just put in empty strings if the value is null.

brianhw:
Updated to write NULLs in other strings, and confirmed on Vertica.

mulby:
:+1:

PR #161, opened 2015-09-16
Use ABORT ON ERROR in Vertica COPY commands.

brianhw:
Also added pep8 fixes while I was at it.  Verified using CourseCatalogWorkflow.

@mulby 

FYI @jab5569


mulby:
LGTM +1

PR #163, opened 2015-09-17
Run acceptance tests for the finance reports

mulby:
This includes:
- Framework fixes to allow this to run on devstack
- Removal of database imports that we didn't end up actually using

@brianhw - This is ready for a first pass of review.

FYI - @jab5569


edx/analytics/tasks/reports/orders_import.py
```diff
@@ -190,6 +108,11 @@ def columns(self):
     def partition(self):
         return HivePartition('dt', self.import_date.isoformat())  # pylint: disable=no-member
 
+    def hiveconfs(self):
```
mulby, line 113: This is the hack that is required to make the test pass, but really shouldn't be necessary.

edx/analytics/tasks/tests/acceptance/services/vertica.py
```diff
@@ -58,11 +62,16 @@ def connect(self):
         """
         Connect to the Vertica server.
         """
+        if self.disabled:
+            raise RuntimeError('The vertica service is disabled')
```
mulby: This will cause tests that require vertica to fail when credentials haven't been specified. We may want to cause them to be skipped instead.

edx/analytics/tasks/tests/acceptance/__init__.py
```diff
@@ -74,8 +72,9 @@ def setUp(self):
         self.catalog_path = 'http://acceptance.test/api/courses/v2'
         database_name = 'test_' + self.identifier
         schema = 'test_' + self.identifier
-        import_database_name = 'import_' + database_name
-        export_database_name = 'export_' + database_name
+        import_database_name = 'reports_import_' + database_name
+        export_database_name = 'reports_export_' + database_name
```
brianhw: Our nomenclature for import vs. export is not very consistent in general, but this is confusing to me.  Generally, when we talk about imports, we mean loading data into reports (or resultstore).  I thought that when we used 'export' we meant exports mostly from LMS DB.  New additions start to challenge these assumptions:  reading from APIs for course subjects is one example of a different export source.  And we load data into Vertica as well as into the resultstore.  

That said, my issue here is that I don't know what the "reports_" prefix is meaning.  The import is presumably from LMS for this, and has nothing to do with the resultstore.  But maybe "reports_"  simply means that it's a separate database used only for generating reports?

edx/analytics/tasks/tests/acceptance/__init__.py
```diff
@@ -74,8 +72,9 @@ def setUp(self):
...
+        otto_database_name = 'reports_otto_' + database_name
```
brianhw: Oh, similar question about this - what does "reports_" mean?

edx/analytics/tasks/tests/acceptance/test_financial_reports.py
```diff
@@ -0,0 +1,111 @@
...
+    "order_refunded_quantity",
+    "order_user_id",
+    "order_username",
+    "order_user_email",
+    "order_product_class",
+    "order_product_detail",
+    "order_course_id",
+    "order_org_id",
+    "order_processor",
+]
```
brianhw: Does it add value to hard-code this here?  I'm wondering if it makes more sense to determine this list dynamically, rather than cut-paste maintenance.  That is, instantiate a LoadInternalReportingOrderTransactionsToWarehouse object, fetch its columns property, and pull the row names out of that.  We're already checking the expected columns implicitly by maintaining an expected output CSV.

edx/analytics/tasks/reports/orders_import.py
```diff
@@ -99,63 +55,25 @@ def requires(self):
...
-                **kwargs
-            )
+            # Original shopping cart tables.
+            ImportShoppingCartOrder(**kwargs),
+            ImportShoppingCartOrderItem(**kwargs),
+            ImportShoppingCartCertificateItem(**kwargs),
+            ImportShoppingCartPaidCourseRegistration(**kwargs),
+            ImportShoppingCartDonation(**kwargs),
+            ImportShoppingCartCourseRegistrationCodeItem(**kwargs),
+            ImportAuthUserTask(**kwargs),
```
brianhw, line 78: Nit:  add comment to indicate "Other LMS tables" (or similar), since this last one is not a shopping cart table.  I would also remove "Original" from the comment above.

edx/analytics/tasks/reports/orders_import.py
brianhw, line 113: There should definitely be some kind of comment here explaining the issue as best as possible.  I would have no idea looking at this code to know why this configuration parameter is being set, and why here.

edx/analytics/tasks/tests/acceptance/fixtures/input/finance_reports/lms/shoppingcart_courseregcodeitem.sql
```diff
@@ -0,0 +1,12 @@
...
+--
+
+DROP TABLE IF EXISTS `shoppingcart_courseregcodeitem`;
+CREATE TABLE `shoppingcart_courseregcodeitem` (
+  `orderitem_ptr_id` int(11) NOT NULL,
+  `course_id` varchar(128) NOT NULL,
+  `mode` varchar(50) NOT NULL,
+  PRIMARY KEY (`orderitem_ptr_id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
```
brianhw, line 12: Should these have some data too?

edx/analytics/tasks/tests/acceptance/fixtures/input/finance_reports/lms/shoppingcart_donation.sql
```diff
@@ -0,0 +1,12 @@
...
+--
+
+DROP TABLE IF EXISTS `shoppingcart_donation`;
+CREATE TABLE `shoppingcart_donation` (
+  `orderitem_ptr_id` int(11) NOT NULL,
+  `donation_type` varchar(32) NOT NULL,
+  `course_id` varchar(255) NOT NULL,
+  PRIMARY KEY (`orderitem_ptr_id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
```
brianhw, line 12: Should this have some data too?

edx/analytics/tasks/tests/acceptance/fixtures/input/finance_reports/lms/shoppingcart_paidcourseregistration.sql
```diff
@@ -0,0 +1,13 @@
...
+
+DROP TABLE IF EXISTS `shoppingcart_paidcourseregistration`;
+CREATE TABLE `shoppingcart_paidcourseregistration` (
+  `orderitem_ptr_id` int(11) NOT NULL,
+  `course_id` varchar(128) NOT NULL,
+  `mode` varchar(50) NOT NULL,
+  `course_enrollment_id` int(11),
+  PRIMARY KEY (`orderitem_ptr_id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
```
brianhw, line 13: Should this also have some data?

edx/analytics/tasks/tests/acceptance/fixtures/input/finance_reports/otto/catalogue_productattribute.sql
```diff
@@ -0,0 +1,22 @@
...
+CREATE TABLE `catalogue_productattribute` (
+  `id` int(11) NOT NULL AUTO_INCREMENT,
+  `name` varchar(128) NOT NULL,
+  `code` varchar(128) NOT NULL,
+  `type` varchar(20) NOT NULL,
+  `required` tinyint(1) NOT NULL,
+  `option_group_id` int(11) DEFAULT NULL,
+  `product_class_id` int(11),
+  PRIMARY KEY (`id`)
+) ENGINE=InnoDB AUTO_INCREMENT=6 DEFAULT CHARSET=utf8;
```
brianhw: nit: none of the others specify an auto_increment value....

edx/analytics/tasks/tests/acceptance/services/vertica.py
brianhw: Skipping particular acceptance tests when needed infrastructure is not present seems to be a more reasonable way to allow for open development while still allowing our private add-ons, like exporter and vertica.  For the exporter, though, we annotated the particular test:      @unittest.skipIf(os.getenv('EXPORTER') is None, "requires private Exporter code")

So instead of failing, I think that one can: "raise unittest.SkipTest(reason)".  

Is the configuration of this url in acceptance tests a reasonable way to determine if tests using Vertica?  I don't really remember where this config value comes from.  It may make more sense for each test to determine whether to skip or change its scope of testing based on whether the service is disabled, rather than making it a global here.  For example, the end-to-end test could check if vertica were enabled, and if not, could run a different task and check the reconciliation table instead.  (Note that the output should be close to the same, if not identical.)

edx/analytics/tasks/tests/acceptance/services/vertica.py
```diff
@@ -13,7 +13,11 @@ class VerticaService(object):
     """Service object to be used as a member of a class to enable that class to write to and read from Vertica."""
 
     def __init__(self, config, schema_name):
-        self.vertica_creds_url = config['vertica_creds_url']
+        self.vertica_creds_url = config.get('vertica_creds_url', '')
```
brianhw: nit: is there a reason to use '' and len() here?  Instead of just doing self.blah = get('blah') and if self.blah?  Just curious...

edx/analytics/tasks/tests/acceptance/test_financial_reports.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def prepare_database(self, name, database):
+        sql_fixture_base_url = url_path_join(self.data_dir, 'input', 'finance_reports', name)
+        for filename in os.listdir(sql_fixture_base_url):
+            self.execute_sql_fixture_file(url_path_join(sql_fixture_base_url, filename), database=database)
+
+    def test_end_to_end(self):
+        self.task.launch([
+            'BuildFinancialReportsTask',
+            '--import-date', '2015-09-02',
```
brianhw: Is there a reason why this import date is different than IMPORT_DATE?  The difference is confusing.  Maybe rename IMPORT_DATE if it's not accurate...

brianhw:
@mulby Finished a pass.  Very close!

edx/analytics/tasks/reports/orders_import.py
mulby, line 113: Done.

edx/analytics/tasks/tests/acceptance/services/vertica.py
mulby: Fixed the vertica case, we'll have to address the exporter case later...

edx/analytics/tasks/tests/acceptance/__init__.py
mulby: reports_ was an artifact of devstack permissions, I've changed it to "acceptance" to be more clear.

Agreed about import vs. export. They are relative terms and we don't use them consistently.

edx/analytics/tasks/tests/acceptance/__init__.py
mulby: Done.

edx/analytics/tasks/tests/acceptance/test_financial_reports.py
mulby: That's fair... fixing.

edx/analytics/tasks/reports/orders_import.py
mulby, line 78: Done

edx/analytics/tasks/tests/acceptance/fixtures/input/finance_reports/lms/shoppingcart_courseregcodeitem.sql
mulby, line 12: Nope - a given item should appear in exactly one of these tables depending on its type (certificate, coupon etc).

edx/analytics/tasks/tests/acceptance/fixtures/input/finance_reports/otto/catalogue_productattribute.sql
mulby: Oops, good catch. Fixed.

edx/analytics/tasks/tests/acceptance/services/vertica.py
mulby: Not a good one. Changing.

edx/analytics/tasks/tests/acceptance/fixtures/input/finance_reports/lms/shoppingcart_courseregcodeitem.sql
brianhw, line 12: Right, but should our end-to-end include a donation, a coupon, etc. for fuller integration coverage?

edx/analytics/tasks/tests/acceptance/test_financial_reports.py
mulby: Yeah, I've clarified that. One is an open upper bound, the other is the actual day that contains real data.

edx/analytics/tasks/tests/acceptance/fixtures/input/finance_reports/lms/shoppingcart_courseregcodeitem.sql
mulby, line 12: Hrm... that's a reasonable point given that we don't have unit test coverage of these nuances. I'm a little nervous that it could explode the complexity of the testing dataset dramatically if we tried to cover all of the edge cases, though.

Open to your thoughts here...

mulby:
@brianhw addressed all of your comments, running a full acceptance test suite now

edx/analytics/tasks/tests/acceptance/services/vertica.py
```diff
@@ -7,13 +7,18 @@
...
 from edx.analytics.tasks.url import get_target_from_url
+from edx.analytics.tasks.tests import unittest
 
 
 class VerticaService(object):
     """Service object to be used as a member of a class to enable that class to write to and read from Vertica."""
 
     def __init__(self, config, schema_name):
-        self.vertica_creds_url = config['vertica_creds_url']
+        self.vertica_creds_url = config.get('vertica_creds_url')
```
brianhw, line 17: Just a small source of confusion:  It seems that the setUp method for the AcceptanceTestCase (in **init**) requires that a value be set for vertica_creds_url.  So it must be set there, but presumably must be set to a zero-length string.   Does this actually work?  Or do we have to do a strip on it as well?  Or is it something that we want to make optional at the **init** level?  It seems kind of weird to require a value exist but be empty.

edx/analytics/tasks/tests/acceptance/services/vertica.py
```diff
@@ -58,11 +63,16 @@ def connect(self):
         """
         Connect to the Vertica server.
         """
+        if self.disabled:
+            raise unittest.SkipTest('The vertica service is disabled')
```
brianhw, line 67: Will this actually mark the test to be skipped in time?  It seems that the time that the VerticaService is actually connected to is when validating the result of some Vertica load, so I would expect the test to have already failed by the time we get here (by trying to actually load data into a non-existent Vertica server).  It may be necessary instead to move this call into each individual test that would _eventually_ use VerticaService.  So it would have code if self.vertica.disabled: raise unittest.SkipTest(...).

brianhw:
Just two more questions.

edx/analytics/tasks/tests/acceptance/services/vertica.py
mulby, line 17: Yeah, if you leave it empty it does what you expect. I like the idea of making it optional, though, I'll make that change.

edx/analytics/tasks/tests/acceptance/services/vertica.py
mulby, line 67: Not really. It runs the whole test and then skips at the very end when it goes to connect to vertica.

In the future I'd like to add a better pattern that lets you tell the framework whether you are going to use optional features or not (vertica, exporter etc).

I was planning on addressing that in a follow-on story, though, when we solve the devstack more comprehensively.

My intent with this PR is to push the effort forward (largely because it made my development faster), not necessary complete it.

edx/analytics/tasks/tests/acceptance/__init__.py
```diff
@@ -98,6 +98,10 @@ def setUp(self):
                 'credentials': self.config['credentials_file_url'],
                 'database': export_database_name
             },
+            'otto-database-import': {
+                'credentials': self.config['credentials_file_url'],
+                'database': otto_database_name
+            },
             'vertica-export': {
                 'credentials': self.config['vertica_creds_url'],
```
brianhw: If you make it optional for vertica_creds_url to be in config, then this can break ....

mulby: Good catch, fixed.

edx/analytics/tasks/tests/acceptance/__init__.py
```diff
@@ -123,16 +129,21 @@ def setUp(self):
...
         self.hive = hive.HiveService(self.task, self.config, database_name)
+        self.vertica = vertica.VerticaService(self.config, schema)
 
-        self.reset_external_state()
+        if os.getenv('DISABLE_RESET_STATE', 'false').lower() != 'true':
+            self.reset_external_state()
 
     def reset_external_state(self):
-        self.s3_client.remove(self.test_root, recursive=True)
+        root_target = get_target_from_url(self.test_root.replace('s3://', 's3+https://'))
```
mulby, line 148: This is an awful hack - suggestions for improvement are welcome!

On the jenkins box it doesn't have the hadoop client installed, so it can't process "s3://" urls. However, on the devstack it can't use "s3+https" urls since it can't access S3.

At a minimum I suppose I should include a comment about this.

brianhw:
LGTM for the scope of this story. :+1:

PR #164, opened 2015-09-21
Change NO_COST orders to no longer check order status.

brianhw:
Such 'honor' enrollment orders may be refunded when a user unenrolls,
and a zero-cost refund just skips creating a transaction.

@mulby


mulby:
:+1:

PR #165, opened 2015-09-22
Filter out null course_id from ed_services report.

brianhw:
@mulby @jab5569  Small change, but it removes the all-null row.


mulby:
:+1:

PR #166, opened 2015-09-23
Brian/reorder financial files

brianhw:
I flattened the reports directory, but I also removed CourseModeTableTask, as it seemed to add nothing to just running ImportCourseModeTask.    And pushed forward on a few pylint warnings.

@mulby @jab5569


mulby:
:+1:

PR #168, opened 2015-10-01
change num users watching video; ensure duration is set for all videos

HassanJaveed84:
@mulby @brianhw

### DIFF
--- a/edx/analytics/tasks/tests/test_video.py
+++ b/edx/analytics/tasks/tests/test_video.py
@@ -770,7 +770,7 @@ class VideoUsageTaskReducerTest(ReducerTestMixin, unittest.TestCase):
             UsageColumns.PIPELINE_VIDEO_ID: self.COURSE_ID + '|' + self.VIDEO_MODULE_ID,
             UsageColumns.COURSE_ID: self.COURSE_ID,
             UsageColumns.VIDEO_MODULE_ID: self.VIDEO_MODULE_ID,
-            UsageColumns.VIDEO_DURATION: '\\N',
+            UsageColumns.VIDEO_DURATION: 5,
             UsageColumns.SECONDS_PER_SEGMENT: VIDEO_VIEWING_SECONDS_PER_SEGMENT,
             UsageColumns.USERS_AT_START: 1,
             UsageColumns.USERS_AT_END: 1,
@@ -898,7 +898,7 @@ class VideoUsageTaskReducerTest(ReducerTestMixin, unittest.TestCase):
         # identical value in every record. Also note that the middle records are ignored here.
         self._check_output_by_key(inputs, [
             {
-                UsageColumns.VIDEO_DURATION: '\\N',
+                UsageColumns.VIDEO_DURATION: 15,
                 UsageColumns.USERS_AT_START: 2,
                 UsageColumns.USERS_AT_END: 2,
                 UsageColumns.SEGMENT: 0,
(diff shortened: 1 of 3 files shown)


edx/analytics/tasks/tests/test_video.py
```diff
@@ -992,13 +992,27 @@ def test_multiple_known_durations(self):
         ])
 
     def test_an_unknown_durations(self):
+        #duration is always set now
```
brianhw: nit:  our style for comments is to include a space after the hash, and to write as a sentence.  That means duration is capitalized, and the sentence ends with a period.

edx/analytics/tasks/tests/test_video.py
```diff
@@ -992,13 +992,27 @@ def test_multiple_known_durations(self):
         ])
 
     def test_an_unknown_durations(self):
```
brianhw: nit:  agreement.  Either "test_unknown_durations" or "test_an_unknown_duration".  Actually, I would remove the "_an_" either way.

edx/analytics/tasks/video.py
```diff
@@ -410,19 +410,20 @@ def reducer(self, key, viewings):
         # actually viewed to determine users_at_end.
         if video_duration == VIDEO_UNKNOWN_DURATION:
             final_segment = self.get_final_segment(usage_map)
+            video_duration = final_segment * VIDEO_VIEWING_SECONDS_PER_SEGMENT + VIDEO_VIEWING_SECONDS_PER_SEGMENT
```
brianhw: First of all, I think I would write this as (final_segment + 1) \* VIDEO_VIEWING_SECONDS_PER_SEGMENT.

But that seems slightly off.  That is, if I were to calculate the final_segment from this duration value, I don't think you would get the same value you began with.  That is, if I were to call

final_segment = self.snap_to_last_segment_boundary(float(video_duration))

I think it would be off by one.

edx/analytics/tasks/video.py
```diff
@@ -430,8 +431,15 @@ def reducer(self, key, viewings):
                 len(stats.get('users', [])),
                 stats.get('views', 0),
             )
-            if video_duration == VIDEO_UNKNOWN_DURATION and segment == final_segment: break
+            if segment == final_segment: break
 
+    def complete_end_segment(self, duration):
+        """
+        Calculates a complete end segment by cutting off the minimum of 30 seconds and 5% of duration.
+        Needed to cutoff video credits etc.
```
brianhw: nit: cutoff => "cut off".  Also, it's not clear what a "complete end segment" is, in general, and I think this would be a good place to describe what it is.  That is, it's the segment that if a user watches to this point, it is considered that they have watched enough to have watched the "complete" video.  Or something to that effect.

edx/analytics/tasks/video.py
brianhw: So maybe you need to subtract some small amount, like 0.05 s, to get it to work, if I'm right that it doesn't.  (And if I'm wrong, then don't subtract anything. : )

brianhw:
Looks good, Hassan!  Just a few nits.

edx/analytics/tasks/video.py
HassanJaveed84: @brianhw snap_to_last_segment_boundary casts the duration to int, so wouldn't it take care of it?

HassanJaveed84:
@brianhw @mulby updated acceptance test, squashed.

mulby:
:+1:

PR #169, opened 2015-10-02
Make calculated durations be within the final segment.

brianhw:
This change makes sure that the duration is actually within the last segment of data included, by finding the duration of the beginning of the next segment and subtracting one.   (Video duration is output as an int, so that's as close to the end of the segment as we're going to get.)

Also fixed some PEP8 and pylint warnings.

@HassanJaveed84 @mulby


mulby:
:+1:

PR #69, opened 2015-02-18
Make boto try harder when attempting to read credentials

mulby:
This addresses one of the issues that contributed to the outage we experienced yesterday. Boto defaults to 1 attempt to retrieve metadata with a timeout of 1.0 seconds, this increases both of those parameters dramatically. It will try for an hour if necessary.

Reviewer: @brianhw
FYI: @jab5569 @e0d

PR #170, opened 2015-10-07
Event export on a per-course basis

HassanJaveed84:


edx/analytics/tasks/event_exports_by_course.py
```diff
@@ -0,0 +1,83 @@
...
+    environment = luigi.Parameter(
+        config_path={'section': 'event-export', 'name': 'environment'}
+    )
+
+    def mapper(self, line):
+        event, date_string = self.get_event_and_date_string(line) or (None, None)
+        if event is None:
+            return
+
+        cid = eventlog.get_course_id(event, from_url=True)
```
brianhw: small nit:  I would prefer something longer than "cid".  I think it would be clear enough to use course_id, even though there is already a self.course_id

edx/analytics/tasks/event_exports_by_course.py
```diff
@@ -0,0 +1,83 @@
...
+            return event['context']['received_at']
+        except KeyError:
+            return super(EventExportByCourseTask, self).get_event_time(event)
+
+    def output_path_for_key(self, key):
+        date, course_id = key
+        filename_safe_course_id = opaque_key_util.get_filename_safe_course_id(course_id)
+
+        # Remap site name from prod to edx
+        site = "edx" if self.environment == 'prod' else self.environment
```
brianhw: Note that we no longer need this here, nor in the format statement on #69.

edx/analytics/tasks/event_exports_by_course.py
```diff
@@ -0,0 +1,83 @@
...
+
+        return url_path_join(
+            self.output_root,
+            '{course}_{date}.log.gz'.format(
+                course=filename_safe_course_id.lower(),
+                site=site,
```
brianhw: don't need.

edx/analytics/tasks/event_exports_by_course.py
```diff
@@ -0,0 +1,83 @@
...
+
+    output_root = luigi.Parameter(
+        config_path={'section': 'event-export', 'name': 'output_root'}
+    )
+
+    course_id = luigi.Parameter(is_list=True, default=[])
+
+    environment = luigi.Parameter(
+        config_path={'section': 'event-export', 'name': 'environment'}
+    )
```
brianhw: No longer need environment.

edx/analytics/tasks/event_exports_by_course.py
```diff
@@ -0,0 +1,83 @@
...
+log = logging.getLogger(__name__)
+
+
+class EventExportByCourseTask(EventLogSelectionMixin, MultiOutputMapReduceJobTask):
+    """
+    Group events by course and export them for research purposes.
+    """
+
+    output_root = luigi.Parameter(
+        config_path={'section': 'event-export', 'name': 'output_root'}
```
brianhw: We definitely don't want this to use the same default for output_root as the regular event export.   Different section....

brianhw:
Looks good!  Great progress!

edx/analytics/tasks/event_exports_by_course.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def output_path_for_key(self, key):
+        date, course_id = key
+        filename_safe_course_id = opaque_key_util.get_filename_safe_course_id(course_id)
+
+        return url_path_join(
+            self.output_root,
+            '{course}_{date}.log.gz'.format(
+                course=filename_safe_course_id.lower(),
+                date=date,
```
brianhw, line 67: I'm changing my mind on this, as I think about next steps.  I think it works better to have:

<output_root>/<filename_safe_course_id>/events/<filename_safe_course_id>-events-<date>.log.gz

That way we can write events and database data for a given course into the same directory structure.

brianhw:
Changes look good.  One more change in the spec for output files, but hopefully that doesn't mess things up too much.

edx/analytics/tasks/event_exports_by_course.py
HassanJaveed84, line 67: @brianhw can you please clarify a bit on the path ?
Would it be like this ?
output_root/events/course_id_events_date.log.gz

HassanJaveed84:
@brianhw I added test for the new task. Was having trouble making headway with it. Also,
I'm not sure how to test the 'multi_output_reducer' ? Any suggestions for overall improvement ?

In regards to the acceptance test, I know I can't run them on the devstack. Is there any way I could make them work on my machine?

edx/analytics/tasks/event_exports_by_course.py
```diff
@@ -0,0 +1,76 @@
...
+        except KeyError:
+            return super(EventExportByCourseTask, self).get_event_time(event)
+
+    def output_path_for_key(self, key):
+        date, course_id = key
+        filename_safe_course_id = opaque_key_util.get_filename_safe_course_id(course_id)
+
+        return url_path_join(
+            self.output_root,
+            "events",
```
brianhw, line 64: Sorry I wasn't clearer here.  I was suggesting:

self.output_root, filename_safe_course_id, "events", "{course}-events-{date}.log.gz'.format(etc).

edx/analytics/tasks/event_exports_by_course.py
```diff
@@ -0,0 +1,76 @@
...
+            '{course}-events-{date}.log.gz'.format(
+                course=filename_safe_course_id.lower(),
```
brianhw: Hmm.  If we need to lower() this, we should probably do it on #58 when we set the variable.  Especially if we're using it as a directory name as well.

edx/analytics/tasks/tests/test_event_export_by_course.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def test_without_timestamp(self):
+        line = self.create_event_log_line(time='')
+        self.assert_no_map_output_for(line)
+
+    def test_output_for_unwanted_event(self):
+        self.task = EventExportByCourseTask(
+            output_root=None,
+            source=None,
+            course_id='Foo',
```
brianhw: Does this need to be a list?

edx/analytics/tasks/tests/test_event_export_by_course.py
```diff
@@ -0,0 +1,81 @@
...
+        )
+        self.task.init_local()
+        line = self.create_event_log_line()
+        self.assert_no_map_output_for(line)
+
+
+class EventExportByCourseOutputPathTest(EventExportByCourseBaseTest):
+
+    def test_output_path_for_key(self):
+        path = self.task.output_path_for_key((self.DATE, 'CourseX'))
```
brianhw: I think EventExportByCourseBaseTest defines a course_id, so you could use that here.  The value of course_id depends on the OpaqueKeysMixin.  You should then have a second class that uses InitializeLegacyKeysMixin (from the same file).  See test_answer_dist, video, enrollments, user_activity for examples of testing with the old and new forms of identifiers.  And it would be good to put that testing into the test_event_exports, from which it's missing!

edx/analytics/tasks/util/eventlog.py
```diff
@@ -235,11 +235,27 @@ def get_course_id(event):
...
+        if source == 'server':
+            url = event.get('event_type', '')
+        elif source == 'browser':
+            url = event.get('page', '')
+        else:
+            url = ''
+
+        course_key = opaque_key_util.get_course_key_from_url(url)
+        if course_key:
+            return unicode(course_key)
```
brianhw, line 259: I don't think that we need to do a rigorous testing of get_course_key_from_url in the unit tests for this file, but it would be good to make sure that the from_url logic is given coverage.  Some of the tests in the test_event_export could likely be factored over for that purpose.

edx/analytics/tasks/tests/test_event_export_by_course.py
```diff
@@ -0,0 +1,81 @@
...
+    def test_single_output(self):
+        line = self.create_event_log_line()
+        self.assert_single_map_output(line, self.expected_key, line)
+
+    def test_without_timestamp(self):
+        line = self.create_event_log_line(time='')
+        self.assert_no_map_output_for(line)
+
+    def test_output_for_unwanted_event(self):
+        self.task = EventExportByCourseTask(
```
brianhw: I think you can call self.create_task and just pass the argument you want to change (i.e. the course_id).

brianhw: Then again, I can't find an example of where that's been done before, using this version of create_task().  But it seems like it _should_ work. : )

brianhw:
As for testing the 'multi_output_reducer', I found an example in test_answer_dist, but that version won't work with this gzipping version.  So I would just put it off to the acceptance test.
As for running acceptance tests, I don't know of a way to run them locally.   This is one of the few Tasks that doesn't use Hive, so the locally-running non-Hadoop versions of Map-Reduce that we have could actually be used.  But because there are so few acceptance tests like that, I don't know that it's worth the investment to get them to work like that (i.e. without Hadoop).

edx/analytics/tasks/tests/test_event_export_by_course.py
brianhw: Hmm.  You should use the two InitializeXXXKeysMixin classes for testing the bulk of the functionality.  But it might be overkill for this particular *PathTest.  It might be clearer to just do what test_answer_dist does for this case, which is to provide the hard-coded course_ids, in legacy and opaque-keys forms, and checks each of them.

edx/analytics/tasks/tests/test_event_export_by_course.py
HassanJaveed84: I think it only works if the argument passed is in DEFAULT_ARGS, ref [map_reduce_mixins.py](https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/tests/map_reduce_mixins.py#L47)

edx/analytics/tasks/tests/test_event_export_by_course.py
HassanJaveed84: When you mention about putting that testing into test_event_exports, is it only for testing output_path_for key with legacy and new keys?

edx/analytics/tasks/tests/test_event_export_by_course.py
brianhw: That looks right.  But maybe that's worth changing.  So that any values in kwargs that are _not_ in DEFAULT_ARGS simply get added.  The point of the DEFAULT_ARGS is to provide default values for things that need to be set.  It should be okay to add additional args.

edx/analytics/tasks/tests/test_event_export_by_course.py
```diff
@@ -0,0 +1,94 @@
...
+            interval=Date.parse(self.DATE),
+            pattern="*",
+            marker=None
+        )
+        self.task.init_local()
+        line = self.create_event_log_line()
+        self.assert_no_map_output_for(line)
+
+
+class EventExportByCourseLegacyMapTest(InitializeLegacyKeysMixin, EventExportByCourseBaseTest):
```
brianhw: Ah, this should use EventExportByCourseMapTest, not EventExportByCourseBaseTest.  We want to rerun the tests defined in the former, with the override defined by the LegacyKeysMixin.

brianhw: Making that change should result in a greater number of tests being run (as one way to verify its effect).

edx/analytics/tasks/util/tests/test_eventlog.py
```diff
@@ -99,3 +99,37 @@ def test_event_data_with_unknown_type(self):
...
+            'context': {},
+            'event_type': '/courses/edX/Open_DemoX/edx_demo_course/info'
+        }
+        self.assertEquals(eventlog.get_course_id(event, from_url=True), 'edX/Open_DemoX/edx_demo_course')
+
+    def test_course_id_from_browser_url(self):
+        event = {
+            'event_source': 'browser',
+            'context': {},
+            "page": "http://test.edx.org/courses/course-v1:DemoX+DemoX+T1_2014/courseware/interactive_demonstrations"
```
brianhw: nit:  use single quotes consistently here, instead of switching to double quotes...

edx/analytics/tasks/tests/map_reduce_mixins.py
```diff
@@ -44,10 +44,14 @@ def create_task(self, **kwargs):
...
-        for attr in self.DEFAULT_ARGS:
+
+        merged_arguments = self.DEFAULT_ARGS.copy()
+        merged_arguments.update(kwargs)
+
+        for attr in merged_arguments:
             if not hasattr(self.task_class, attr):
                 continue
-            value = kwargs.get(attr, self.DEFAULT_ARGS.get(attr))
+            value = merged_arguments.get(attr)
```
HassanJaveed84, line 54: @brianhw does this change look reasonable ?

edx/analytics/tasks/tests/test_event_export_by_course.py
HassanJaveed84: That was my intent, copy paste -_-. Thanks for pointing it out!

edx/analytics/tasks/util/eventlog.py
brianhw, line 259: Looks good.

HassanJaveed84:
@brianhw added acceptance test.

edx/analytics/tasks/tests/map_reduce_mixins.py
brianhw, line 54: Yes, this looks great!

edx/analytics/tasks/tests/acceptance/test_event_export_by_course.py
```diff
@@ -0,0 +1,82 @@
...
+
+        self.validate_output()
+
+    def validate_output(self):
+        self.validate_output_file('2014-05-15', 'acceptancex_a101_t12014')
+        self.validate_output_file('2014-06-01', 'acceptancex_a101_t12014')
+        self.validate_output_file('2014-05-15', 'acceptancex_demox_t1_2014')
+        self.validate_output_file('2014-04-18', 'edx_open_demox_edx_demo_course')
+        self.validate_output_file('2014-05-15', 'edx_open_demox_edx_demo_course')
+        self.validate_output_file('2014-05-16', 'edx_open_demox_edx_demo_course')
```
brianhw: One observation:  I would like to see this check this more exhaustively -- that is, make sure that all of these files were output, and _only_ these files.  I acknowledge that the current event export tests don't do this, but it seems more important to do so at the course level than at the org level.  So the setup of the downloaded files might be at the top of this method, and shared.  Then download all the files in the output to the downloaded_outputs, and validate each one above.  And then check if there were any downloaded that weren't validated.

edx/analytics/tasks/tests/acceptance/test_event_export_by_course.py
```diff
@@ -0,0 +1,82 @@
...
+            if key is not None:
+                break
+            else:
+                time.sleep(2)
+
+        if key is None:
+            self.fail('Unable to find expected output file {0}'.format(remote_url))
+
+        downloaded_output_path = os.path.join(self.downloaded_outputs, remote_url.split('/')[-1])
+        key.get_contents_to_filename(downloaded_output_path)
```
brianhw: There is common code here between the original event export test and this one.  It seems to me that pulling out the fetching the key and downloading the key to a filename would be a good thing to pull out of both.
If this were to remain S3-specific, then I could imagine the logic moving to s3_util.py or similar.  But I don't think that there is anything in the event-exporting code(s) that enforces that the output be written to S3. 
And in particular, I think that we want to be more agnostic in our use of S3 for acceptance tests.  When running acceptance tests on the analytics-devstack, for instance, one will want the output to be written to local hdfs instead of S3.  I expect the tests would then download from the remote URL depending on the protocol (e.g. s3://).  
I don't necessarily want to pull in all the acceptance-tests-on-analytics-devstack work, but it would be good if we didn't make the task harder and possibly move it forward a little.  @mulby , can you confirm this with your sense of  acceptance-tests-on-analytics-devstack progress and direction?

edx/analytics/tasks/tests/acceptance/test_event_export_by_course.py
brianhw: Looking at test_answer_dist, it would probably be enough to just get a test of the number of files in the output, using some kind of ls (e.g. key.list() for s3).  That function could go in the fs service.  And here we would just confirm that the number is six.  Note that it still might be worth setting up the temp directory once, and copying files once....

HassanJaveed84:
@brianhw 
I updated the acceptance test. I have put the s3 downloading part to s3_util. Not sure if it is the right approach ? If it makes sense, then I can use that function for test_event_export as well. 

I could not find a straight forward way of determining the number of files at a particular folder for S3, the closes thing is list() method on a bucket object, to get the length I would still need to iterate over the list.
Right now, I am downloading all files and just checking if they are 6 in number using `len(os.listdir(self.downloaded_outputs))`

edx/analytics/tasks/s3_util.py
```diff
@@ -35,6 +35,23 @@
...
+def get_file_from_key(s3_conn, url, output_path):
+    for _index in range(30):
+        key = s3_conn.get_key(remote_url)
+        if key is not None:
+            break
+        else:
+            time.sleep(2)
+
+    if key is None:
+        log.error("Unable to find expected output file %s", url)
```
brianhw: This should fail, no?  Otherwise it will call get_contents_to_filename on None.

edx/analytics/tasks/s3_util.py
```diff
@@ -35,6 +35,23 @@
 DEFAULT_KEY_ACCESS_POLICY = 'bucket-owner-full-control'
 
 
+def get_file_from_key(s3_conn, url, output_path):
+    for _index in range(30):
```
brianhw, line 43: Should probably add back the comment about backoff -- or just implement it. : )

edx/analytics/tasks/tests/acceptance/test_event_export_by_course.py
```diff
@@ -0,0 +1,89 @@
...
+
+            remote_url = url_path_join(self.test_out, output_file['course_id'], "events", local_file_name + '.gz')
+
+            downloaded_output_path = get_file_from_key(self.s3_util, remote_url, self.downloaded_outputs)
+
+            decompressed_file_name = downloaded_output_path[:-len(',gz')]
+            fs.decompress_file(downloaded_output_path, decompressed_file_name)
+
+    def validate_output(self):
+        self.assertEqual(len(os.listdir(self.downloaded_outputs)), 6)
```
brianhw: This is a reasonable first test.  But you were also testing the contents of each as well, no?  That seems to have gone away.

HassanJaveed84: I misinterpreted your previous comment regarding checking the number of files in the output, thinking you only suggested having that test. I'll add back the diff one as well.

brianhw: Wait, no.  This doesn't seem to do the right thing.  That is, you create a list of six files as self.output_files, and then you download each of those six files, and then check that you downloaded six files.  That doesn't actually check how many files were in the S3 bucket.  I think you were more right when you noted that you had to use key.list() to get the contents of the remote location.  Once you check that there are six remote files -- and that can be done in the download call -- then it should be enough to then download the six.

edx/analytics/tasks/s3_util.py
```diff
@@ -35,6 +35,23 @@
 DEFAULT_KEY_ACCESS_POLICY = 'bucket-owner-full-control'
 
 
+def get_file_from_key(s3_conn, url, output_path):
```
brianhw: And yes, I would definitely update the current event-export acceptance test to use this.  More DRY.

edx/analytics/tasks/tests/acceptance/test_event_export_by_course.py
HassanJaveed84: @brianhw the code to get the number of objects would look like

```
def get_object_count(s3_conn, bucket, prefix=''):
    count = 0
    bucket = s3_conn.s3.get_bucket(bucket)
    for key in bucket.list(prefix=prefix):
        count += 1

    return count
```

Where prefix limits the list to only keys with that prefix(relative path to bucket). Now how would I get the bucket name? The only place I could find it would be in the env variable for the automation tests `"tasks_output_url": "s3://edx-analytics-test-output/automation/"` but it is not set anywhere in the actual code. I actually need to pass in 'edx-analytics-test-output' as the bucket name.

brianhw: You can get the bucket from the URL.  If you look in s3_util.py, there is a get_s3_bucket_key_names() function that you could use.  Or you could just call generate_s3_sources with a wildcard pattern (patterns = ['_']), since it deals with some of the boundary conditions involving buckets and paths within.  Actually, that ['_'] could be made a default value for the 'patterns' argument.  And then just use len(list(generate_s3_sources(self.s3_util, url))).

HassanJaveed84: @brianhw this line in generate_s3_sources
 `keys = (s.key for s in bucket.list(root_with_slash) if s.size > 0)`
skips directories. For this particular task, it would create output by course such as
`s3://output_path/demox/events/demox-events-2014-09-09.log.gz`.  If I specify 's3://output_path/' as the source parameter, it wouldn't return anything as it skips all the course directories. Thoughts?

edx/analytics/tasks/tests/acceptance/test_event_export_by_course.py
```diff
@@ -0,0 +1,102 @@
...
+            output_file['downloaded_path'] = decompressed_file_name
+            fs.decompress_file(downloaded_output_path, decompressed_file_name)
+
+
+    def validate_output(self):
+        for output_file in self.output_files:
+            local_file_name = '{course}-events-{date}.log'.format(
+                course=output_file['course_id'],
+                date=output_file['date'],
+            )
```
brianhw: Nit: this is now in two places -- I'd pull it out into a function call now.

edx/analytics/tasks/s3_util.py
```diff
@@ -35,6 +35,37 @@
 DEFAULT_KEY_ACCESS_POLICY = 'bucket-owner-full-control'
 
 
+def get_object_count(s3_conn, url):
+    count = 0
+    bucket_name, root = get_s3_bucket_key_names(url)
+    root_with_slash = root if len(root) == 0 or root.endswith('/') else root + '/'
+    bucket = s3_conn.get_bucket(bucket_name)
+    for key in bucket.list(root_with_slash):
```
brianhw: bucket.list() will include all files that are "below" the "root_with_slash".  This will include files that appear to be in subdirectories of subdirectories, so it will include the log.gz files within courses.  But it will also include the subdirectory objects themselves, so those need to be filtered out by testing for size > 0.  
I think if you're reluctant to convert the generator to a list, you can use sum().

edx/analytics/tasks/tests/acceptance/test_event_export_by_course.py
```diff
@@ -0,0 +1,99 @@
...
+            local_file_name = self.generate_file_name(output_file['course_id'], output_file['date'])
+
+            remote_url = url_path_join(self.test_out, output_file['course_id'], "events", local_file_name + '.gz')
+
+            downloaded_output_path = get_file_from_key(self.s3_client, remote_url, self.downloaded_outputs)
+
+            if downloaded_output_path is None:
+                self.fail('Unable to find expected output file {0}'.format(remote_url))
+
+            decompressed_file_name = downloaded_output_path[:-len(',gz')]
```
brianhw: nit:  '.gz' instead of ',gz'

edx/analytics/tasks/tests/acceptance/test_event_export_by_course.py
```diff
@@ -0,0 +1,99 @@
...
+            output_file['downloaded_path'] = decompressed_file_name
+            fs.decompress_file(downloaded_output_path, decompressed_file_name)
+
+    def validate_output(self):
+        for output_file in self.output_files:
+            local_file_name = self.generate_file_name(output_file['course_id'], output_file['date'])
+            shell.run(['diff', output_file['downloaded_path'], os.path.join(self.data_dir, 'output', local_file_name)])
+
+    def generate_file_name(self, course_id, date):
```
brianhw: or you could just pass output_file dict...

brianhw:
Getting to the finish line.  A few nits, and a small refactor of the existing event_export, then we can hand it off to Gabe for a final review.

HassanJaveed84:
@brianhw @mulby

edx/analytics/tasks/event_exports_by_course.py
```diff
@@ -0,0 +1,77 @@
...
+
+        key = (date_string, course_id)
+        yield tuple([value.encode('utf8') for value in key]), line.strip()
+
+    def get_event_time(self, event):
+        # Some events may emitted and stored for quite some time before actually being entered into the tracking logs.
+        # The primary cause of this is mobile devices that go offline for a significant period of time. They will store
+        # events locally and then when connectivity is restored transmit them to the server. We log the time that they
+        # were received by the server and use that to batch them into exports since it is much simpler than trying to
+        # inject them into past exports.
```
brianhw: I think I still agree with the goal of this code, but the logic as expressed in the comment doesn't hold in the same way.  While we use event_export to perform the packaging of events into files for research export on an incremental basis, the use of event_export_by_course is not incremental.  So why do this?  Two reasons:  1.  consistency with what research exports produce.  2.  consistency with the date of the tracking log from which it came.  

So this comment should have a little bit appended to it to clarify that.

brianhw:
LGTM. :+1:

HassanJaveed84:
@brianhw @mulby squashed. Should I merge it in ?

brianhw:
I would like to have thumbs from @mulby first.

mulby:
I'm going to take a pass now.

edx/analytics/tasks/event_exports_by_course.py
```diff
@@ -0,0 +1,78 @@
...
+                date=date,
+            )
+        )
+
+    def multi_output_reducer(self, _key, values, output_file):
+        with gzip.GzipFile(mode='wb', fileobj=output_file) as outfile:
+            try:
+                for value in values:
+                    outfile.write(value.strip())
+                    outfile.write('\n')
```
mulby, line 76: There is a snippet in the other event export class that looks like this:

``` python
                    # WARNING: This line ensures that Hadoop knows that our process is not sitting in an infinite loop.
                    # Do not remove it.
                    self.incr_counter('Event Export', 'Raw Bytes Written', len(value) + 1)
```

We should include that here as well for the same reason detailed in the comment.

mulby:
This generally looks great. :+1: after you make the one change I requested.

Side note: there is still heavy dependence on S3 in the acceptance test. I'm OK cleaning that up when we actually take on the work of removing the dependency, but this test does add to that work a bit. I'd rather just tackle that in follow on work instead of prolonging this task, however.

HassanJaveed84:
Updated.

mulby:
:+1:

PR #171, opened 2015-10-19
Add filesize output to s3util ls.

brianhw:
Small change to tool to supplement 'ls' output. 

@mulby @HassanJaveed84


mulby:
:+1:

PR #173, opened 2015-10-23
(WIP) Using GeoIP2 instead of Pygeoip, and add country_for_private_ip

louyihua:
pygeoip works only on IPv4 or IPv6 with a single database, while in a dual-stack configuration, both IPv4 and IPv6 addresses may occur in the tracking log. In this case, some logs may not be correctly parsed and the location reports will be wrong.

Another issue here is that currently all IP addresses not in the database will be marked as 'UNKNOWN'. However, in some openedx deployments, especially those inside an institute or university, users inside the institute or university may use private adress to access openedx, which will generate a lot of 'UNKNOWN' results with the current code. This PR uses ···country_<name/code>_for_private_ip··· rather than UNKNOWN_COUNTRY for such private addresses.

Tests for these modifications will be provided later...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,9 +24,10 @@ develop: requirements
 	python setup.py develop
 
 system-requirements:
+	sudo add-apt-repository -y ppa:maxmind/ppa
 	sudo apt-get update -q
 	# This is not great, we can't use these libraries on slave nodes using this method.
-	sudo apt-get install -y -q libmysqlclient-dev libatlas3gf-base libpq-dev python-dev libffi-dev libssl-dev libxml2-dev libxslt1-dev
+	sudo apt-get install -y -q libmysqlclient-dev libatlas3gf-base libpq-dev python-dev libffi-dev libssl-dev libxml2-dev libxslt1-dev libmaxminddb-dev
 
 requirements:
 	$(PIP_INSTALL) -U -r requirements/pre.txt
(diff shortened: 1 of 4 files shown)


openedx-webhooks:
Thanks for the pull request, @louyihua! I've created [OSPR-901](https://openedx.atlassian.net/browse/OSPR-901) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

We can't start reviewing your pull request until you've   added yourself to the [AUTHORS](https://github.com/edx/edx-analytics-pipeline/blob/master/AUTHORS) file . Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst) file for more information.

sarina:
@louyihua what does edx-platform use?

louyihua:
@sarina
edx-platform uses pygeoip, too.
In fact, as I have no programming experience with hadoop before, I'm not sure which is the right way to add a new file dependency, so I decided to try geoip2, which allow only one file to provide both IPv4 and IPv6 support. If it is possible for the code to depend on two seperate geoip data files, I think keep using pygeoip is a better solution than switching to geoip2.

sarina:
@louyihua OK I will try to find you an answer, thanks!

sarina:
Closing in favor of #177

PR #174, opened 2015-11-17
Use mixed-case output path for event exports by course.

HassanJaveed84:
@brianhw I couldn't run the acceptance test through jenkins. The edx-analytics-pipeline-acceptance-test-manual build page does not load correctly for me, I'm not sure why.


brianhw:
The PR LGTM. :+1: 

I redeployed the AcceptanceTestCluster (in case that was the problem, since it had recently been terminated).  While I was at it, I added you and Umer's credentials so you guys can log in if you need to debug acceptance tests there.  But it turns out that the Jenkins job had run out of file handles, so I restarted it.  The test off of master was triggered and seems to be running now.

PR #175, opened 2015-11-17
Student module engagement

mulby:
@brianhw here is an early review for preliminary feedback

Todo:
- [x] A LOT of unit tests (pretty much all of record.py and big chunks of engagement.py)
- [x] An acceptance test
- [x] pep8/pylint
- [x] more documentation on the workflow
- [x] figure out overwrite semantics

Questions I have that I would love your opinion/thoughts on:
1. How do you feel about the concept of a "Record" where we represent the schema _once_ and then use it to handle reading and writing tuples that are Hive compatible as well as generate the schema for the other systems? I just spiked this through to see how it felt... let me know what you think. If we like it I'll write an extensive test suite for it. You can see example usage in engagement.py
2. How do you feel about the new Hive utilities? One of the goals was to allow us to keep partitions as output targets, but also allow us to use the partitions as input sources to hadoop jobs. I also wanted to tease apart the concept of "table" and partition to allow us to more flexibly combine them. In this case, allowing us to have many partitions and only one table. Again, there are example usages in engagement.py
3. Should I keep the sparse course engagement stuff in this PR? I feel like it might make more sense to put it into the PR that computes the roster table since that's where it's used primarily.


edx/analytics/tasks/util/hive.py
```diff
@@ -124,6 +124,138 @@ def job_runner(self):
...
+        )
+
+        query = textwrap.dedent(query)
+
+        return query
+
+    @property
+    def partition_by(self):
+        """The partitioning key name. Specify None to create a table that is not partitioned."""
+        raise None
```
brianhw: Can you do this?  What happens?

mulby: Hah oops. Pretty sure this will crash if run in a not-so-helpful way. This is somewhere in between "raise NotImplementedError" and "return None". Will fix.

```
Python 2.7.6 (default, Mar 22 2014, 22:59:56) 
[GCC 4.8.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> raise None
  File "<stdin>", line 1, in <module>
TypeError: exceptions must be old-style classes or derived from BaseException, not NoneType
```

mulby:
@brianhw unit tests are in, this is ready for another round of review while I work on the remaining subtasks called out above

mulby:
@brianhw this is ready for review now

edx/analytics/tasks/engagement.py
```diff
@@ -0,0 +1,380 @@
...
+
+        self.attempted_removal = True
+        if self.overwrite:
+            # first clear the appropriate rows from the luigi mysql marker table
+            marker_table = self.output().marker_table  # side-effect: sets self.output_target if it's None
+            try:
+                query = "DELETE FROM {marker_table} where `update_id`='{update_id}'".format(
+                    marker_table=marker_table,
+                    update_id=self.update_id(),
+                )
```
brianhw: I'm wondering if it's possible to not implement all of this from scratch, but instead pull out and override the particular parts that are different.  The two changes I see that you made were here and the DELETE FROM {table} below, and in particular, the addition or modification of the WHERE clauses.  

And it possibly makes sense to move this into a separate derived class, something like MysqlInsertByDateTask.  Challenge there is getting date in.

edx/analytics/tasks/engagement.py
```diff
@@ -0,0 +1,380 @@
...
+from edx.analytics.tasks.mysql_load import MysqlInsertTask
+
+from edx.analytics.tasks.util.hive import (
+    WarehouseMixin, BareHiveTableTask, HivePartitionTask
+)
+from edx.analytics.tasks.util.record import Record, StringField, IntegerField, DateField
+
+
+class EngagementRecord(Record):
+    """Represents a count of interactions performed by a user on a particular entity (usually a module in a course)"""
```
brianhw: nit: end with a period.

edx/analytics/tasks/engagement.py
```diff
@@ -0,0 +1,380 @@
...
+
+    # Override superclass to disable this parameter
+    interval = None
+
+
+class EngagementTask(EventLogSelectionMixin, OverwriteOutputMixin, MapReduceJobTask):
+    """
+    Process the event log and categorize user engagement with various types of content.
+
+    This emits one record for each type of interaction. Note that this is loosely defined, for problems, for example, it
```
brianhw: nit:  "loosely defined.  For example, for problems it will...."

edx/analytics/tasks/engagement.py
```diff
@@ -0,0 +1,380 @@
...
+    will emit two records if the problem is correct (one for the "attempt" interaction and one for the "correct attempt"
+    interaction).
+
+    This task is intended to be run incrementally and populate a single Hive partition. Although time consuming to
+    bootstrap the table, it results in a significantly cleaner workflow. It is much more clear what the success and
```
brianhw: nit: "more clear" => "clearer"

edx/analytics/tasks/engagement.py
brianhw: It's not really about date, as much as it's the name of the property to partition on (i.e. "date") and the partition_value (i.e. self.date.isoformat()) as used in the Hive table.  Is there an advantage to make the naming consistent across these tables -- i.e. use "date" instead of "dt" for Hive, and then do the same for Mysql "partitions"?

edx/analytics/tasks/engagement.py
```diff
@@ -0,0 +1,380 @@
...
+
+    @property
+    def partition_value(self):
+        """Use a dynamic partition value based on the date parameter."""
+        return self.date.isoformat()  # pylint: disable=no-member
+
+    @property
+    def hive_table_task(self):
+        return EngagementTableTask(
+            warehouse_path=self.warehouse_path
```
brianhw: Just wondering about the location that the Hive table points to, and the output that the data_task writes to.  Seems like the logic for those is spread out.  The table task is getting warehouse_path passed in, but the data_task is getting partition_location passed in.  Should the table actually be getting table_location?

edx/analytics/tasks/engagement.py
```diff
@@ -0,0 +1,380 @@
...
+                entity_type=entity_type,
+                entity_id=entity_id,
+                event=event,
+                count=0
+            )
+            yield (record.to_string_tuple(), 1)
+
+    def reducer(self, key, values):
+        """Replace the count (which is hardcoded to 0 above, with the actual number of records in this group."""
+        yield ('\t'.join(key[:-1]), sum(values))
```
brianhw: I'm thinking it would make more sense to move this funky logic up into the mapper output.  This should be just (key, sum(values)).  Do the [:-1] operation on the tuple, even if it involves tuple(list(record.to_string_tuple())[:-1]).

edx/analytics/tasks/engagement.py
```diff
@@ -0,0 +1,380 @@
...
+            entity_type = 'forum'
+            if event_type.endswith('.created'):
+                if event_type == 'edx.forum.comment.created':
+                    events.append('commented')
+                elif event_type == 'edx.forum.response.created':
+                    events.append('responded')
+                elif event_type == 'edx.forum.thread.created':
+                    events.append('created')
+
+            entity_id = event_data.get('commentable_id')
```
brianhw: This code is only going to grow over time.  Pull it out now into a separate method:  something like get_engagement_events_from_event().  

(A little bit of overloading of "event" seems to be happening here.  Is there a better name for "events = []"?  They are not events as read from event logs, so it's a little confusing.)

edx/analytics/tasks/engagement.py
mulby: done

edx/analytics/tasks/engagement.py
mulby: done

edx/analytics/tasks/engagement.py
mulby: done

edx/analytics/tasks/engagement.py
mulby: done

edx/analytics/tasks/engagement.py
mulby: Good idea. Done.

edx/analytics/tasks/engagement.py
mulby: The mental model is that the Hive layer "knows" where the data lives. Specifically, the table knows where it's location is from the warehouse path, and from that we can derive the partition location. The only place the warehouse path is actually used is in the table location. Does this make sense?

brianhw: I agree that the Hive layer is responsible for the data location.  But we have two parts of the layer:  the table task and the partition task.  If the partition task is telling the data where to write, shouldn't it also tell the table where to read?  It's relying on the partition and table tasks to separately make the same calculation from warehouse_path.  Maybe that's fine, but that's my question.

edx/analytics/tasks/engagement.py
mulby: Mostly done. I like the idea of pulling this logic back into the mysql_load module, which I have done. I also made it slightly more generic.

As for "date" vs. "dt": IIRC hive doesn't like using "date" as a partition name because it is a reserved word. We could go the other way, perhaps? Use dt for the mysql date column?

edx/analytics/tasks/engagement.py
mulby: Ah, I failed to explain this well. The calculation is actually only done once (by the table). The partition has a reference to the table it belongs to, so it just appends it's own information to the path that is generated by the table. warehouse_path -> table_location -> partition_location -> output_root... each layer only builds upon the previous layer.

edx/analytics/tasks/engagement.py
```diff
@@ -0,0 +1,380 @@
...
+    Hadoop is not designed to operate like this and it would be significantly more efficient to process a range of dates
+    at once. The choice was made to stick with the cleaner workflow since the steady-state is the same for both options
+    - in general we will only be processing one day of data at a time.
+    """
+
+    # Required parameters
+    date = luigi.DateParameter()
+    output_root = luigi.Parameter()
+
+    # Override superclass to disable these parameters
```
brianhw: nit: s/these parameters/this parameter/ ?

edx/analytics/tasks/engagement.py
```diff
@@ -0,0 +1,380 @@
...
+    interval = None
+
+    def __init__(self, *args, **kwargs):
+        super(EngagementTask, self).__init__(*args, **kwargs)
+
+        self.interval = date_interval.Date.from_date(self.date)
```
brianhw: Can one just add this to the kwargs before calling super, rather than redefining interval=None?  Oh, we have to do that so it doesn't even appear as a parameter.  Okay.

edx/analytics/tasks/engagement.py
brianhw: Argh.  "dt" is pretty cryptic.  But going the other way by adding characters can be ugly, though less cryptic.  I really want "date", because that's what it is.  "date_value", "date_val", "edate", "event_date", "day" ... any of those acceptable?

edx/analytics/tasks/engagement.py
```diff
@@ -0,0 +1,380 @@
...
+            # commit the currently open transaction before continuing with the copy.
+            query = "DELETE FROM {table} WHERE date='{date}'".format(
+                table=self.table,
+                date=self.date.isoformat()  # pylint: disable=no-member
+            )
+            connection.cursor().execute(query)
+
+    @property
+    def auto_primary_key(self):
+        # We use a rather large compound primary key for this table to take advantage of the clustered index on read.
```
brianhw: It may not be obvious when reading this that it's referring to keys().  It says "we use a ... primary key" but then returns None.  Maybe say that we don't want a primary key added automatically, since we need to explicitly define one in keys() below.  Or something like that.

edx/analytics/tasks/engagement.py
```diff
@@ -0,0 +1,380 @@
...
+class EngagementVerticaTask(EngagementDownstreamMixin, VerticaCopyTask):
+    """
+    Replicate the hive table in Vertica.
+
+    Note that it is updated incrementally.
+    """
+
+    @property
+    def insert_source_task(self):
+        partition_task = EngagementPartitionTask(
```
brianhw: nit:  is there a standard order to define these?    Often insert_source_task is listed last, and here it's listed first.   Would just be neater to pick an order, based on some kind of criteria.

edx/analytics/tasks/engagement.py
brianhw: (Goes for all the others, as well....)

edx/analytics/tasks/pathutil.py
```diff
@@ -186,7 +186,7 @@ def _get_s3_urls(self, source):
 
     def _get_hdfs_urls(self, source):
         """Recursively list all files inside the source directory on the hdfs filesystem."""
-        for source in luigi.hdfs.listdir(source):
+        for source in luigi.hdfs.listdir(source, recursive=True):
```
brianhw, line 189: Excellent -- great to get that to be consistent.

edx/analytics/tasks/tests/acceptance/test_engagement.py
```diff
@@ -0,0 +1,89 @@
+# pylint: disable=line-too-long
+"""
+End to end test of the per-module engagement workflow.
```
brianhw: Is "per-module" what we're calling it now?

edx/analytics/tasks/tests/test_engagement.py
```diff
@@ -0,0 +1,195 @@
...
+            output_root='/fake/output',
+        )
+        self.task.init_local()
+
+    @data(
+        {'time': "2013-12-01T15:38:32.805444"},
+        {'username': ''},
+        {'event_type': None},
+        {'context': {'course_id': 'lskdjfslkdj'}},
+        {'event': 'sdfasdf'}
```
brianhw: nit: don't we normally add trailing commas to the last elements of collections when written in vertical form?  (Also applies to a lot of things below...)

edx/analytics/tasks/tests/test_engagement.py
```diff
@@ -0,0 +1,195 @@
...
+        template['event_type'] = event_type
+        template['name'] = event_type
+        self.assert_single_map_output(
+            json.dumps(template),
+            self.get_expected_output_key('forum', self.forum_id, expected_action),
+            1
+        )
+
+
+class EngagementTaskMapLegacyKeysTest(InitializeLegacyKeysMixin, unittest.TestCase):
```
brianhw: Um, what is this testing?   EngagementTaskMapTest?

edx/analytics/tasks/tests/test_engagement.py
```diff
@@ -0,0 +1,195 @@
...
+    """
+    Tests to verify that engagement data is reduced properly
+    """
+
+    task_class = EngagementTask
+
+    def setUp(self):
+        super(EngagementTaskReducerTest, self).setUp()
+
+        self.reduce_key = (self.COURSE_ID, 'test_user', self.DATE, 'problem', 'foobar', 'completed', '0')
```
brianhw: This will also need to be updated....

edx/analytics/tasks/util/hive.py
```diff
@@ -124,6 +124,166 @@ def job_runner(self):
...
+        )
+
+    def job_runner(self):
+        return OverwriteAwareHiveQueryRunner()
+
+    def remove_output_on_overwrite(self):
+        if self.overwrite:
+            self.attempted_removal = True
+
+        # Note that the query takes care of actually removing the old partition.
```
brianhw: nit: maybe just move the comment above the if....

edx/analytics/tasks/util/hive.py
```diff
@@ -124,6 +124,166 @@ def job_runner(self):
         return OverwriteAwareHiveQueryRunner()
 
 
+class BareHiveTableTask(WarehouseMixin, OverwriteOutputMixin, HiveQueryTask):
```
brianhw, line 128: I assume the "Bare" here is just to distinguish this from the existing HiveTableTask.   Just odd to have "Bare" here and not on the HivePartitionTask.  Oh, but the "Bare" means that it has no data?  Okay.

edx/analytics/tasks/engagement.py
brianhw: I hadn't read/remembered the rest of the pull request.  This makes sense now.

edx/analytics/tasks/util/record.py
```diff
@@ -0,0 +1,357 @@
...
+            fields.sort(key=lambda t: t[1].counter)
+            cls._fields = OrderedDict(fields)
+
+        return cls._fields
+
+    def to_string_tuple(self, null_value=DEFAULT_NULL_VALUE):
+        """
+        Convert the record into a tuple of UTF-8 encoded byte strings.
+
+        This format is convenient for Luigi since it expects tuples of strings as output from reduce functions.
```
brianhw: s/for Luigi/for use with Luigi/

edx/analytics/tasks/util/record.py
```diff
@@ -0,0 +1,357 @@
...
+        return cls.from_string_tuple(tuple(tsv_str.rstrip('\r\n').split('\t')))
+
+    @classmethod
+    def get_sql_schema(cls):
+        """
+        A skeleton schema of the SQL table that could store this data.
+
+        Returns: A list of tuples whose first element is the column name, and the second element is the type to assign
+            to the column. Note that this returns SQL92 compliant types including some modifiers (NOT NULL) etc. It
+            Does not include any index definitions, constraints or relationship declarations.
```
brianhw: nit: s/Does/does/

edx/analytics/tasks/util/record.py
```diff
@@ -0,0 +1,357 @@
...
+    systems as well as interpret data. It is intended to provide structure so that downstream code does not have to
+    handle edge cases related to dynamic typing. It enforces the type and ensures that the data conforms to the
+    declared schema.
+    """
+    counter = 0
+
+    def __init__(self, **kwargs):
+        self.nullable = kwargs.pop('nullable', True)
+
+        # This is a hack that lets us "see" the order the class member variables appear in the class they are declared
```
brianhw: nit: s/is a hack that/counter/
Also s/order the class/order in which the class/.   Or just "order of the class member variables in the class"...

edx/analytics/tasks/util/tests/test_record.py
```diff
@@ -0,0 +1,352 @@
...
+        self.assertEqual(test_record.name, 'baz')
+        self.assertEqual(test_record.bar, 10)
+
+    def test_record_with_no_fields(self):
+
+        class NoFields(Record):
+            """A record without any fields"""
+            pass
+
+        NoFields()
```
brianhw: nothing more to test?

edx/analytics/tasks/util/tests/test_record.py
```diff
@@ -0,0 +1,352 @@
...
+        self.assertEqual(test_record.first, 'a')
+        self.assertEqual(test_record.second, 'b')
+        self.assertEqual(test_record.third, 'c')
+
+    def test_mixed_with_missing(self):
+        with self.assertRaises(TypeError):
+            ThreeFieldRecord('a', second='b')
+
+    def test_extra_kwargs(self):
+        with self.assertRaises(TypeError):
```
brianhw: Should this use RaisesRegex to check the message of the TypeError to confirm it's the right TypeError?

brianhw:
No real issues, just a few nits.

edx/analytics/tasks/engagement.py
mulby: done

edx/analytics/tasks/engagement.py
mulby: Yeah I don't want it to appear when running --help etc.

edx/analytics/tasks/engagement.py
mulby: done

edx/analytics/tasks/engagement.py
mulby: I'm not terribly attached to a standard order. I agree it might be kind of nice, but not worth the effort of enforcing IMHO.

edx/analytics/tasks/tests/acceptance/test_engagement.py
mulby: Heh - I'm struggling to differentiate with the "student engagement" stuff. Thoughts?

edx/analytics/tasks/tests/test_engagement.py
mulby: Yeah - I try to do this, occasionally miss one or two. Also difficult to enforce and keep consistent without a custom pylint rule or something.

edx/analytics/tasks/tests/test_engagement.py
mulby: heh, great catch. Oops. This now is a subclass of EngagementTaskMapTest.

edx/analytics/tasks/tests/test_engagement.py
mulby: I assume this is referring to the change where we lop of the count in the mapper instead of the reducer... if so, then this is done.

edx/analytics/tasks/util/hive.py
mulby: done

edx/analytics/tasks/util/hive.py
mulby, line 128: I was trying to differentiate with the existing class... "bare" was intended to refer to the lack of any partitions.

Once we remove the other one I was planning on renaming this to HiveTableTask.

I'll add a TODO to that effect.

edx/analytics/tasks/util/record.py
mulby: done

edx/analytics/tasks/util/record.py
mulby: done

edx/analytics/tasks/util/record.py
mulby: done and done

edx/analytics/tasks/util/tests/test_record.py
mulby: Good idea, I'll validate a few more of the edge cases here - basically test the other public methods when you have a record without any fields.

edx/analytics/tasks/util/tests/test_record.py
mulby: done (for many of these checks)

mulby:
@brianhw this is ready for another pass.

Changes:
- Addressed most of the nits
- Added some dunder-methods to Record so that it behaves more like namedtuple (repr, hash, ==, != etc). It wasn't abundantly clear to me what the correct behavior was here, so I settled on "if the names and ordering of the fields is the same and the values are the same, they are equal". I also noticed some kind of interesting edge cases with multiple inheritance etc, put in a comment to discourage it - but don't explicitly prevent it. Predicting field order in that case is... tricky.

edx/analytics/tasks/engagement.py
```diff
@@ -0,0 +1,363 @@
...
+    Hadoop is not designed to operate like this and it would be significantly more efficient to process a range of dates
+    at once. The choice was made to stick with the cleaner workflow since the steady-state is the same for both options
+    - in general we will only be processing one day of data at a time.
+    """
+
+    # Required parameters
+    date = luigi.DateParameter()
+    output_root = luigi.Parameter()
+
+    # Override superclass to disable this parameters
```
brianhw: nit: s/parameters/parameter/

edx/analytics/tasks/tests/acceptance/test_engagement.py
brianhw: In engagement.py above, you call it "Measure student engagement with individual pieces of course content".  Not sure what's being differentiated.....

edx/analytics/tasks/engagement.py
```diff
@@ -0,0 +1,363 @@
...
+    This table is appended to every time this workflow is run, so it is expected to grow to be *very* large. For this
+    reason, the records are stored in a clustered index allowing for very fast point queries and date range queries
+    for individual users in particular courses.
+
+    This allows us to rapidly generate activity charts over time for small numbers of users.
+    """
+
+    @property
+    def table(self):
+        return "engagement"
```
brianhw: Were you planning on renaming this to something more descriptive?

edx/analytics/tasks/engagement.py
```diff
@@ -0,0 +1,363 @@
...
+        )
+
+    def requires(self):
+        yield self.data_task
+        yield self.hive_table_task
+
+    @property
+    def data_task(self):
+        """The task that generates the raw data inside this hive partition."""
+        return EngagementTask(
```
brianhw: Rename this to EngagementDataTask, to parallel EngagementPartitionTask and EngagementTableTask....

edx/analytics/tasks/engagement.py
brianhw: Sure.  But this code is likely to be used as a model for other tasks in future, so a lot of them will cut-and-paste-and-modify.  So I suggest just moving this insert_source_task() below the columns() below, as this is what we have generally been doing.

edx/analytics/tasks/util/record.py
```diff
@@ -0,0 +1,431 @@
...
+        if value is None and not self.nullable:
+            validation_errors.append('The field cannot accept null values')
+        return validation_errors
+
+    def serialize_to_string(self, value):
+        """Returns a string representation of a value for this field."""
+        return unicode(value)
+
+    def deserialize_from_string(self, string_value):
+        """Returns a typed representation of the value from it's string representation."""
```
brianhw: nit: s/it's/its/

edx/analytics/tasks/util/tests/test_record.py
```diff
@@ -0,0 +1,474 @@
...
+        class DifferentFieldOrder(Record):
+            """A record with exactly the same fields as TwoFieldRecord but with a different field order"""
+            value = StringField()
+            name = StringField()
+
+        self.assertNotEqual(TwoFieldRecord(name='a', value='b'), DifferentFieldOrder(name='a', value='b'))
+        self.assertNotEqual(TwoFieldRecord(name='a', value='b'), DifferentFieldOrder(name='b', value='a'))
+
+    def test_hash(self):
+        self.assertEqual(hash(SingleFieldRecord('foo')), hash(SingleFieldRecord('foo')))
```
brianhw, line 267: Not critical, but might be reasonable to include test that a different value than 'foo' results in an unequal hash.

mulby:
@brianhw another thought - perhaps we forgo the custom record class and instead use something like SQLAlchemy to model our records and write adapters that know how to convert those models to TSV, Hive schema defs etc?

It would save us from having to worry about a lot of the record logic that is fairly generic and common among model systems like this.

What do you think?

edx/analytics/tasks/util/record.py
```diff
@@ -0,0 +1,431 @@
...
+            arg_strs.append('{name}={value}'.format(name=field_name, value=repr(val)))
+
+        return '{cls_name}({args})'.format(
+            cls_name=self.__class__.__name__,
+            args=', '.join(arg_strs)
+        )
+
+    def __eq__(self, other):
+        if not isinstance(other, Record):
+            return NotImplemented
```
brianhw: return?  or raise?

mulby:
I started with namedtuple and started enriching it... but looking at it now it might be closer to an ORM model than a namedtuple.

edx/analytics/tasks/util/record.py
brianhw, line 173: Add a comment here about NotImplemented.....

edx/analytics/tasks/util/record.py
```diff
@@ -0,0 +1,431 @@
...
+            name = StringField()
+
+        class B(Record):
+            name = StringField()
+
+        class C(A):
+            pass
+
+        class D(Record):
+            name = StringField(nullable=False)
```
brianhw, line 57: It might be worth mentioning here that classes inheriting from Record with additional instance variables that are not of type Field will have those variables ignored when comparisons are made.   Have an example with foo = {} or something like that, in addition to name.

brianhw:
Just a few remaining nits.

edx/analytics/tasks/engagement.py
mulby: Done

edx/analytics/tasks/engagement.py
mulby: I changed to "module_engagement" throughout the code, what do you think?

edx/analytics/tasks/engagement.py
mulby: Good suggestion, done.

edx/analytics/tasks/util/record.py
mulby: done

edx/analytics/tasks/util/tests/test_record.py
mulby, line 267: done

edx/analytics/tasks/util/record.py
mulby, line 173: done

edx/analytics/tasks/util/record.py
mulby, line 57: done

mulby:
@brianhw ready for another pass

edx/analytics/tasks/engagement.py
brianhw: Like it.

brianhw:
LGTM!   :+1:

PR #176, opened 2015-11-18
Use mixed-case output path for event exports by course.

HassanJaveed84:
@brianhw needed to rename(change case) the output files to pass the acceptance test. I did that with my previous PR, but I could not find a way to squash the filenames rename commit. Git apparently does not keep history if you change only the casing of the filename.


brianhw:
LGTM.  :+1:  once it passes the acceptance tests.

PR #177, opened 2015-11-18
Add IPv6 address support

louyihua:
Currently, we assume that the addresses in tracking log are always IPv4 or IPv6 in pipeline. However, in a dual-stack configuration, both IPv4 and IPv6 addresses may occur in the tracking log.
This PR adds support for both IPv4 and IPv6 address by using the "IPv4-Mapped IPv6 Address" feature. An IPv4 address is always converted into its equivalent IPv6 version, so that we can just use the single IPv6 database for processing.
This PR also adds support for specifying country code & name for private addresses (such as "192.168.0.0/16", "172.16.0.0/12" and "10.0.0.0/8").

### DIFF
--- a/edx/analytics/tasks/user_location.py
+++ b/edx/analytics/tasks/user_location.py
@@ -2,6 +2,7 @@
 Determine the number of users in each country.
 """
 import datetime
+import ipaddress
 import tempfile
 
 import luigi
@@ -25,11 +26,25 @@ class GeolocationMixin(object):
 
     Parameters:
         geolocation_data: a URL to the location of country-level geolocation data.
+        geodata_ipv6: Specify whether the database pointed by 'geolocation_data' is an IPv6 database or not.
+        country_name_for_private_ip: The country name for ip addresses inside the private ip range.
+        country_code_for_private_ip: The country code for ip addresses inside the private ip range.
     """
     geolocation_data = luigi.Parameter(
         config_path={'section': 'geolocation', 'name': 'geolocation_data'}
     )
-
+    geodata_ipv6 = luigi.BoolParameter(
+        config_path={'section': 'geolocation', 'name': 'geodata_ipv6'},
+        default=False
+    )
+    country_name_for_private_ip = luigi.Parameter(
+        config_path={'section': 'geolocation', 'name': 'country_name_for_private_ip'},
+        default=None
+    )
+    country_code_for_private_ip = luigi.Parameter(
+        config_path={'section': 'geolocation', 'name': 'country_code_for_private_ip'},
+        default=None
+    )
 
 class BaseUserLocationTask(GeolocationMixin):
     """
@@ -104,8 +119,34 @@ class BaseGeolocation(object):
 
         # This ip address might not provide a country name.
         try:
-            country = self.geoip.country_name_by_addr(last_ip)
-            code = self.geoip.country_code_by_addr(last_ip)
+            ipaddr = ipaddress.ip_address(unicode(last_ip))
+
+            # Raise an exception if 'last_ip' cannot be converted
+            # into a valid IPv4 or IPv6 address
+            if ipaddr is None: raise ValueError()
+
+            # If IPv6 database is used, then convert IPv4 address
+            # its mapped IPv6 version, so that we can use a single
+            # IPv6 database for both IPv4 and IPv6 addresses.
+            if self.geodata_ipv6 and ipaddr.version == 4:
+                last_ip = '::ffff:' + last_ip
+
+            # The IP database doesn't provide geo data for private address
+            # like '192.168.0.1'. However, for an open edx deployment
+            # inside a institution, the private address may exist in
+            # tracking logs. It is more reasonable to use a specific
+            # country name/code for such addresses rather than 'UNKNOWN'.
+            # This feature is enabled only when the two additional
+            # parameters are set.
+            if self.country_name_for_private_ip is not None and ipaddr.is_private:
+                country = self.country_name_for_private_ip
+            else:
+                country = self.geoip.country_name_by_addr(last_ip)
+
+            if self.country_code_for_private_ip is not None and ipaddr.is_private:
+                code = self.country_code_for_private_ip
+            else:
+                code = self.geoip.country_code_by_addr(last_ip)
         except Exception:
             log.exception("Encountered exception getting country:  user '%s', last_ip '%s' on '%s'.",
                           username, last_ip, last_timestamp)
@@ -234,6 +275,8 @@ class UsersPerCountry(MapReduceJobTask, BaseUserLocationTask):
             manifest=self.manifest,
             geolocation_data=self.geolocation_data,
             end_date=self.end_date,
+            country_name_for_private_ip=self.country_name_for_private_ip,
+            country_code_for_private_ip=self.country_code_for_private_ip,
         )
 
     def output(self):
@@ -333,6 +376,7 @@ class UsersPerCountryReportWorkflow(MapReduceJobTaskMixin, UsersPerCountryReport
         manifest: a URL to a file location that can store the complete set of input files.
         end_date: events before or on this date are kept, and after this date are filtered out.
         geolocation_data: a URL to the location of country-level geolocation data.
+        geodata_ipv6: Specify whether the database pointed by 'geolocation_data' is an IPv6 database or not.
 
     Additional optional parameters are passed through to :py:class:`MapReduceJobTask`:
 
@@ -357,6 +401,7 @@ class UsersPerCountryReportWorkflow(MapReduceJobTaskMixin, UsersPerCountryReport
     base_input_format = luigi.Parameter(default=None)
     end_date = luigi.DateParameter()
     geolocation_data = luigi.Parameter()
+    geodata_ipv6 = luigi.BoolParameter()
 
     def requires(self):
         return UsersPerCountry(
@@ -371,4 +416,5 @@ class UsersPerCountryReportWorkflow(MapReduceJobTaskMixin, UsersPerCountryReport
             manifest=self.manifest,
             geolocation_data=self.geolocation_data,
             end_date=self.end_date,
+            geodata_ipv6=self.geodata_ipv6,
         )
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -7,6 +7,7 @@ enum34==1.0.4           # BSD
 filechunkio==1.5	# MIT
 html5lib==1.0b3 	# MIT
 idna==2.0               # BSD-like
+ipaddress==1.0.15   # Python Software Foundation License v2
 isoweek==1.3.0		# BSD
 mechanize==0.2.5	# BSD
 http://cdn.mysql.com/Downloads/Connector-Python/mysql-connector-python-1.2.2.zip  	# GPL v2 with FOSS License Exception
--- a/edx/analytics/tasks/tests/test_location_per_course.py
+++ b/edx/analytics/tasks/tests/test_location_per_course.py
@@ -69,10 +69,23 @@ class LastCountryOfUserReducerTestCase(ReducerTestMixin, unittest.TestCase):
         self.timestamp = "2013-12-17T15:38:32.805444"
         self.earlier_timestamp = "2013-12-15T15:38:32.805444"
         self.task.geoip = FakeGeoLocation()
+        self.task.country_name_for_private_ip = "PRIVATE NAME"
+        self.task.country_code_for_private_ip = "PRIVATE CODE"
+        self.task.geodata_ipv6 = True
         self.reduce_key = self.username
 
     def test_no_ip(self):
         self.assert_no_output([])
+
+    def test_private_ip(self):
+        inputs = [(self.timestamp, "192.168.0.1")]
+        expected = ((("PRIVATE NAME", "PRIVATE CODE"), self.username),)
+        self._check_output_complete_tuple(inputs, expected)
+
+    def test_single_ipv6(self):
+        inputs = [(self.timestamp, FakeGeoLocation.ip_address_3)]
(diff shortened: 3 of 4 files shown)


openedx-webhooks:
Thanks for the pull request, @louyihua! I've created [OSPR-962](https://openedx.atlassian.net/browse/OSPR-962) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

We can't start reviewing your pull request until you've   added yourself to the [AUTHORS](https://github.com/edx/edx-analytics-pipeline/blob/master/AUTHORS) file . Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst) file for more information.

louyihua:
@sarina 
I think I've found an easy way to deal with both IPv4 and IPv6 addresses in a dual stacking configuration environment. My solution is to use the `IPv4-Mapped IPv6 Address` feature defined in RFC4291, and each the IPv4 address is first converted into an IPv4-Mapped IPv6 Address before being processed, so that we can use a single IPv6 database to handle both IPv4 and IPv6 address.
I think this PR is enough to replace the PR in https://github.com/edx/edx-analytics-pipeline/pull/173.

sarina:
OK - I am meeting with the analytics team about a different question today but I'll ask them about your PRs as well to see if they have any guidance.

edx/analytics/tasks/user_location.py
```diff
@@ -101,11 +117,20 @@ def reducer(self, key, values):
...
+        # so that we can use only the IPv6 database
+        if last_ip.find(':') == -1:
+            last_ip = '::ffff:' + last_ip
 
         # This ip address might not provide a country name.
         try:
-            country = self.geoip.country_name_by_addr(last_ip)
-            code = self.geoip.country_code_by_addr(last_ip)
+            ipaddr = ipaddress.ip_address(unicode(last_ip))
+            if ipaddr is not None and ipaddr.is_private:
```
sarina: Can you add comments here explaining what you're doing and why?

requirements/default.txt
```diff
@@ -7,6 +7,7 @@ enum34==1.0.4           # BSD
 filechunkio==1.5	# MIT
 html5lib==1.0b3 	# MIT
 idna==2.0               # BSD-like
+ipaddress=1.0.15    # Python Software Foundation License v2
```
sarina: This should be `==`

sarina:
@louyihua Yes, this looks more reasonable than #173 so let's move forward here.

You will need to add tests. Particularly to make sure existing behavior is not changed. For example what happens on master w/ the address 192.168.0.1 & is the behavior maintained?

edx/analytics/tasks/location_per_course.py
```diff
@@ -56,6 +56,12 @@ def requires_local(self):
     def geolocation_data_target(self):
         return self.input_local()
 
+    def default_country_name(self):
+        return self.country_name_for_private_ip
```
sarina: Where is `self.country_name_for_private_ip` defined?

sarina: n/m in different file

edx/analytics/tasks/user_location.py
```diff
@@ -29,7 +30,14 @@ class GeolocationMixin(object):
     geolocation_data = luigi.Parameter(
         config_path={'section': 'geolocation', 'name': 'geolocation_data'}
     )
-
+    country_name_for_private_ip = luigi.DateParameter(
```
sarina: Add some comments here explaining what each of these are.
(this is important because the edX team is a bit unfamiliar with what you're doing here)

edx/analytics/tasks/user_location.py
```diff
@@ -101,11 +117,20 @@ def reducer(self, key, values):
 
         if not last_ip:
             return
+        # Convert IPv4 address to its compatible IPv6 version,
+        # so that we can use only the IPv6 database
+        if last_ip.find(':') == -1:
```
sarina: It seems there should be a better way of doing this. I don't like how hacky this looks.

louyihua:
@sarina Comments and tests are added.

sarina:
@louyihua great, thanks! This looks good to me. I will hand off to the analytics team to review.

Note that there are a few upcoming US holidays, which will slow our development pace. I appreciate your patience.

mulby:
@louyihua I apologize for the delay getting to this, we are behind on our reviews. We are hoping to be able to get to this in the next few weeks.

mulby:
@louyihua we have scheduled this for review. We should be able to take a pass in the next week or so.

edx/analytics/tasks/user_location.py
mulby: These definitely should not be DateParameter they should simply be `Parameter` instances

edx/analytics/tasks/user_location.py
```diff
@@ -104,8 +124,29 @@ def reducer(self, key, values):
...
+            # IPv4 and IPv6 addresses.
+            if ipaddr.version == 4:
+                last_ip = '::ffff:' + last_ip
+
+            # The IP database doesn't provide geo data for private address
+            # like '192.168.0.1'. However, for an open edx deployment
+            # inside a institution, the private address may exist in
+            # tracking logs. It is more reasonable to use a specific
+            # country name/code for such addresses rather than 'UNKNOWN'.
+            if ipaddr.is_private:
```
mulby: I think I would prefer to always use the old logic if these default parameters are not specified. This would dramatically reduce the risk of this change breaking existing reports.

edx/analytics/tasks/user_location.py
```diff
@@ -104,8 +124,29 @@ def reducer(self, key, values):
...
+
+            # Raise an exception if 'last_ip' cannot be converted
+            # into a valid IPv4 or IPv6 address
+            if ipaddr is None: raise ValueError()
+
+            # Convert IPv4 address to its mapped IPv6 version,
+            # so that we can use a single IPv6 database for both
+            # IPv4 and IPv6 addresses.
+            if ipaddr.version == 4:
+                last_ip = '::ffff:' + last_ip
```
mulby, line 132: Do you have to use a different MaxMind database to support IPv6 addresses?

edx/analytics/tasks/location_per_course.py
```diff
@@ -144,6 +150,8 @@ def requires(self):
             geolocation_data=self.geolocation_data,
             overwrite=self.overwrite,
             user_country_output=self.user_country_output,
+            country_name_for_private_ip=self.country_name_for_private_ip,
+            country_code_for_private_ip=self.country_code_for_private_ip,
```
mulby: I'm OK with not plumbing these parameters all the way up to the command line. Reading them only from the config file is fine.

mulby:
@louyihua Thanks for the contribution. My primary concern is that this new method will produce different results than the old method. Once you address the comments I've already included here, we will need to run some validation to ensure that this new method does not produce different output than the old method when these config values are not set.

edx/analytics/tasks/user_location.py
louyihua, line 132: The database can be downloaded from http://geolite.maxmind.com/download/geoip/database/GeoIPv6.dat.gz

edx/analytics/tasks/location_per_course.py
louyihua: I'm not quite familiar with writing code with hadoop. So here the two lines I added is actually to read the parameters from command line?

edx/analytics/tasks/user_location.py
louyihua: This is required as we need to use single IPv6 database to serve both IPv4 and IPv6 addresses.

louyihua:
I'm not quite familiar with writing code on the hadoop framework, and I don't know what is the right way to make the code work with two versions of GeoIP databases. So my decision is to use the IPv6 version of MaxWind databse instead of the IPv4 version, because the IPv6 database is a superset of the IPv4 one.

edx/analytics/tasks/location_per_course.py
mulby: I think you should be able to revert all changes in this file.

edx/analytics/tasks/user_location.py
```diff
@@ -67,6 +79,14 @@ def geolocation_data_target(self):
         """Defines target from which geolocation data can be read."""
         raise NotImplementedError
 
+    def default_country_name(self):
+        """Defines the default country name when private IP addresses are encountered."""
+        raise NotImplementedError
+
+    def default_country_code(self):
+        """Defines the default country name when private IP addresses are encountered."""
+        raise NotImplementedError
```
mulby: You shouldn't need these functions, in the various methods you can simply use `self.country_code_for_private_ip` directly.

edx/analytics/tasks/user_location.py
mulby: I guess what I'm suggesting is that you structure things such that the two new parameters default to `None`, and this function is structured like:

``` python
if self.country_code_for_private_ip is None:
    # This is the legacy behavior
    country = self.geoip.country_name_by_addr(last_ip)
    code = self.geoip.country_code_by_addr(last_ip)
else:
     # all of your new logic goes here
```

louyihua: I think there may be some misunderstanding. This PR actually provides two new features: 1. IPv6 support, and 2. Private IP support. So line 127~137 provides the support for IPv6 (because you cannot pass an IPv4 address directly to the IPv6 database, which will not produce any result), and line 138~147 provides the support of private IPs. So if we switch to the IPv6 version of MaxWind database, the old logic WILL NOT work for IPv4 addresses.
My logic here is that: as IP addresses not the database (including private addresses) will cause an exception, making the results be set to UNKNOWN_COUNTRY and UNKNOWN_CODE. I choose to set the default value of the two new parameters to UNKNOWN_COUNTRY and UNKNOWN_CODE, which means if these two parameters are not provided, the results of private addresses will be also set to UNKNOWN_COUNTRY and UNKNOWN_CODE (line 145 & 146), while other non-existence addresses will remain to cause an exception.

edx/analytics/tasks/location_per_course.py
louyihua: OK, I see.

edx/analytics/tasks/user_location.py
mulby: I see your point.

I guess I'd like lines 136 and 137 to only be executed if an IPv6 MaxMind database is being used. edX will likely continue to use our IPv4 database without passing in the new parameters, so we will need the code to continue to produce the same result in that scenario.

louyihua:
@mulby I got your concern. So I added another parameter, `geodata_ipv6`, to specify whether the database is IPv6 or IPv4. If this param is not set, the code will continue use the IPv4-only processing procedure.

edx/analytics/tasks/user_location.py
```diff
@@ -104,8 +119,34 @@ def reducer(self, key, values):
 
         # This ip address might not provide a country name.
         try:
-            country = self.geoip.country_name_by_addr(last_ip)
-            code = self.geoip.country_code_by_addr(last_ip)
+            ipaddr = ipaddress.ip_address(unicode(last_ip))
+
+            # Raise an exception if 'last_ip' cannot be converted
+            # into a valid IPv4 or IPv6 address
+            if ipaddr is None: raise ValueError()
```
mulby, line 126: nit: please split this expression into 2 lines for readability

edx/analytics/tasks/user_location.py
```diff
@@ -104,8 +119,34 @@ def reducer(self, key, values):
...
+                last_ip = '::ffff:' + last_ip
+
+            # The IP database doesn't provide geo data for private address
+            # like '192.168.0.1'. However, for an open edx deployment
+            # inside a institution, the private address may exist in
+            # tracking logs. It is more reasonable to use a specific
+            # country name/code for such addresses rather than 'UNKNOWN'.
+            # This feature is enabled only when the two additional
+            # parameters are set.
+            if self.country_name_for_private_ip is not None and ipaddr.is_private:
```
mulby, line 141: I'm wondering if what you really want is a mapping for arbitrary CIDR blocks to particular name+code tuples.

Something like this in the config file:

```
[geolocation-override]
192.168.1.0/24 = My Institution, CODE0
10.5.0.0/16 = Another Place, CODE1
```

And in the code you could then do something like:

``` python
def init_reducer(self):
    # existing functionality goes here

    # parse config section "geolocation-override" into geolocation_overrides
    self.networks = []
    for cidr_block, identifier in geolocation_overrides.items():
        network = ipaddress.ip_network(cidr_block, strict=False)
        country_name, country_code = [v.strip() for v in identifier.split(',')]
        self.networks.append((network, country_name, country_code))

    # existing functionality goes here
    try:
         ipaddr = ipaddress.ip_address(unicode(last_ip))
         # check for ipaddr is None, convert to legacy if IPv6 etc
         country = code = None
         for network, net_country_name, net_country_code in self.networks:
             if ipaddr in network:
                 country = net_country_name
                 code = net_country_code
                 break

         if not country or not code:
             country = self.geoip.country_name_by_addr(last_ip)
             code = self.geoip.country_code_by_addr(last_ip)
```

Note that you can get direct access to the config file using the `luigi.config` module.

Doing this would remove the need for the two new parameters "country_name_for_private_ip" and "country_code_for_private_ip".

@brianhw - I'm also curious to know what you think about the above proposal.

nedbat:
@louyihua are you planning to get back to this?

louyihua:
@nedbat Since @mulby has proposed a new idea, and I think this idea is good too. I need more time to modify the code. But in the recent two or three months, I'll be very busy, so maybe this PR will be suspended for a while.

nedbat:
@louyihua I'll close the pull request for now.  When you have time to come back to it, you can just re-open this pull request.  We'll be glad to help with it when you come back to it.

PR #178, opened 2015-11-24
Umer/event type distribution

UmerF920:
Hassan pointed a few more issues in the unit tests on which I am working. Rest can be reviewed.

### DIFF
--- a/edx/analytics/tasks/event_type_dist.py
+++ b/edx/analytics/tasks/event_type_dist.py
@@ -1,0 +1,97 @@
+"""event_type and event_source values being encountered on each day in a given time interval"""
+
+import logging
+import luigi
+import luigi.task
+from edx.analytics.tasks.mapreduce import MapReduceJobTask
+from edx.analytics.tasks.pathutil import EventLogSelectionMixin
+from edx.analytics.tasks.url import ExternalURL, get_target_from_url, url_path_join
+from edx.analytics.tasks.vertica_load import VerticaCopyTask
+
+log = logging.getLogger(__name__)
+
+
+class EventTypeDistributionTask(EventLogSelectionMixin, MapReduceJobTask):
+    """Task to compute event_type and event_source values being encountered on each day in a given time interval."""
+    output_root = luigi.Parameter()
+    events_list_file_path = luigi.Parameter(default=None)
+
+    def requires_local(self):
+        return ExternalURL(url=self.events_list_file_path)
+
+    def init_local(self):
+        super(EventTypeDistributionTask, self).init_local()
+        if self.events_list_file_path is None:
+            self.known_events = {}
+        else:
+            self.known_events = self.parse_events_list_file()
+
+    def parse_events_list_file(self):
+        """ Read and parse the known events list file and populate it in a dictionary."""
+        parsed_events = {}
+        with self.input_local().open() as f_in:
+            lines = f_in.readlines()
+            for line in lines:
+                if (not line.startswith('#') and len(line.split("\t")) is 3):
+                    parts = line.rstrip('\n').split("\t")
+                    parsed_events[(parts[1], parts[2])] = parts[0]
+        return parsed_events
+
+    def mapper(self, line):
+        value = self.get_event_and_date_string(line)
+        if value is None:
+            return
+        event, event_date = value
+        event_type = event.get('event_type')
+        event_source = event.get('event_source')
+        exported = False
+        if event_source is None or event_type is None or event_date is None:
+            # Ignore if any of the keys is None
+            return
+        if event_type.startswith('/'):
+            # Ignore events that begin with a slash
+            return
+        if (event_source, event_type) in self.known_events:
+            event_category = self.known_events[(event_source, event_type)]
+            exported = True
+        else:
+            event_category = 'unknown'
+        yield (event_date, event_category, event_type, event_source, exported), 1
+
+    def reducer(self, key, values):
+        yield (key), sum(values)
+
+    def output(self):
+        return get_target_from_url(url_path_join(self.output_root, 'event_type_distribution/'))
+
+
+class PushToVerticaEventTypeDistributionTask(VerticaCopyTask):
+    """Push the event type distribution task data to Vertica."""
+    output_root = luigi.Parameter()
+    interval = luigi.DateIntervalParameter()
+    n_reduce_tasks = luigi.Parameter()
+    events_list_file_path = luigi.Parameter(default=None)
+
+    @property
+    def table(self):
+        return "event_type_distribution"
+
+    @property
+    def columns(self):
+        return [
+            ('event_date', 'DATETIME'),
+            ('event_category', 'VARCHAR(255)'),
+            ('event_type', 'VARCHAR(255)'),
+            ('event_source', 'VARCHAR(255)'),
+            ('exported', 'BOOLEAN'),
+            ('event_count', 'INT'),
+        ]
+
+    @property
+    def insert_source_task(self):
+        return EventTypeDistributionTask(
+            output_root=self.output_root,
+            interval=self.interval,
+            n_reduce_tasks=self.n_reduce_tasks,
+            events_list_file_pathF=self.events_list_file_path
+        )
--- a/edx/analytics/tasks/tests/test_event_type_dist.py
+++ b/edx/analytics/tasks/tests/test_event_type_dist.py
@@ -1,0 +1,162 @@
+""" test event type distribution task """
+
+from ddt import ddt, data, unpack
+from edx.analytics.tasks.event_type_dist import EventTypeDistributionTask
+from edx.analytics.tasks.tests import unittest
+from edx.analytics.tasks.tests.map_reduce_mixins import MapperTestMixin, ReducerTestMixin
+from edx.analytics.tasks.tests.opaque_key_mixins import InitializeOpaqueKeysMixin
+from edx.analytics.tasks.tests.target import FakeTarget
+import textwrap
+from mock import MagicMock
+
+
+class EventTypeDistributionTaskMapTest(MapperTestMixin, InitializeOpaqueKeysMixin, unittest.TestCase):
+    """Tests to check if event type distribution task mapper works"""
+
+    def setUp(self):
+        self.task_class = EventTypeDistributionTask
+
+        self.events_list = """
+        admin browser edx.instructor.report.downloaded
+        admin server add-forum-admin
+
+        admin server add-forum-community-TA
+        admin server add-forum-mod
+        admin server add-instructor
+        """
+        input_target = FakeTarget(value=self.reformat(self.events_list))
+
+        super(EventTypeDistributionTaskMapTest, self).setUp()
+        self.task.events_list_file_path = "fake_path"
+        self.task.input_local = MagicMock(return_value=input_target)
+        self.task.init_local()
+
+        self.event_date = '2013-12-17'
+        self.event_type = "test_event"
+        self.event_source = "browser"
+        self.event_category = "unknown"
+
+        self.exported = False
+        self.event_templates = {
+            'event': {
+                "username": "test_user",
+                "host": "test_host",
+                "event_source": self.event_source,
+                "event_type": self.event_type,
+                "context": {
+                    "course_id": "course_id",
+                    "org_id": "org_id",
+                    "user_id": "user_id",
+                },
+                "time": self.event_date,
+                "ip": "127.0.0.1",
+                "event": {
+                    "course_id": "course_id",
+                    "user_id": "user_id",
+                    "mode": "honor",
+                }
+            }
+        }
+
+        self.default_event_template = 'event'
+        self.expected_key = (self.event_date, self.event_category, self.event_type, self.event_source, self.exported)
+
+    def test_no_event(self):
+        line = 'this is garbage'
+        self.assert_no_map_output_for(line)
+
+    def test_event_type_none(self):
+        line = self.create_event_log_line(event_type=None)
+        self.assert_no_map_output_for(line)
+
+    def test_event_date_none(self):
+        line = self.create_event_log_line(time=None)
+        self.assert_no_map_output_for(line)
+
+    def test_event_source_none(self):
+        line = self.create_event_log_line(event_source=None)
+        self.assert_no_map_output_for(line)
+
+    def test_event_type_contains_slash(self):
+        line = self.create_event_log_line(event_type="/event")
+        self.assert_no_map_output_for(line)
+
+    def test_bad_event_date(self):
+        line = self.create_event_log_line(time='this is a bogus time')
+        self.assert_no_map_output_for(line)
+
+    def test_valid_event_type(self):
+        line = self.create_event_log_line()
+        expected_value = 1
+        self.assert_single_map_output(line, self.expected_key, expected_value)
+
+    def test_event_list_file_parsing(self):
+        expected_dict = {
+            ("browser", "edx.instructor.report.downloaded"): "admin",
+            ("server", "add-forum-admin"): "admin",
+            ("server", "add-forum-community-TA"): "admin",
+            ("server", "add-forum-mod"): "admin",
+            ("server", "add-instructor"): "admin",
+        }
+        self.assertEquals(expected_dict, self.task.known_events)
+
+    def test_event_list_dictionary_mapping_category_found(self):
+        """ If event_category is found in the list of events."""
+        line = self.create_event_log_line(event_type="edx.instructor.report.downloaded", event_source=self.event_source)
+        self.expected_key = (self.event_date, "admin", "edx.instructor.report.downloaded", self.event_source, True)
+        self.assert_single_map_output(line, self.expected_key, 1)
+
+    def test_event_list_dictionary_mapping_category_not_found(self):
+        """ If event_category is not found in the list of events."""
+        line = self.create_event_log_line(event_type="test_type", event_source=self.event_source)
+        self.expected_key = (self.event_date, "unknown", "test_type", self.event_source, False)
+        self.assert_single_map_output(line, self.expected_key, 1)
+
+    def reformat(self, string):
+        """
+        Args:
+            string: Input String to be formatted
+
+        Returns:
+            Formatted String i.e., blank spaces replaces with tabs (due to the reason that the events_list file
+            is supposed to be tab separated
+
+        """
+        return textwrap.dedent(string).strip().replace(' ', '\t')
+
+
+@ddt
+class EventTypeDistributionTaskReducerTest(ReducerTestMixin, unittest.TestCase):
+    """Tests to check if event type distribution reducer works"""
+
+    def setUp(self):
+        self.task_class = EventTypeDistributionTask
+        super(EventTypeDistributionTaskReducerTest, self).setUp()
+
+        # Create the task locally, since we only need to check certain attributes
+        self.interval = '2013-01-01'
+        self.event_type = "test_event"
+        self.event_date = "2013-01-01"
+        self.event_source = "browser"
+
+    @data(
+        (('2013-01-01', "test_event", "browser", "unknown", False), [1])
+    )
+    @unpack
+    def test_single_event(self, reduce_key, values):
+        self.reduce_key = reduce_key
+        expected = ((reduce_key, 1),)
+        self._check_output_complete_tuple(values, expected)
+
+    @data(
+        (('2013-01-01', "test_event", "browser", "admin", True), [1, 1, 1, 1]),
+        (('2013-01-01', "test_event", "server", "test_category", True), [1, 1, 1]),
+        (('2013-01-02', "test_event", "browser", "unknown", False), [1, 1, 1, 1, 1]),
+        (('2013-01-03', "test_event", "server", "test_category", True), [1, 1]),
+        (('2013-01-04', "test_event", "mobile", "admin", True), [1, 1, 1, 1]),
+    )
+    @unpack
+    def test_multiple_events(self, reduce_key, values):
+        self.reduce_key = reduce_key
+        expected = ((reduce_key, sum(values)),)
+        self._check_output_complete_tuple(values, expected)
--- a/setup.cfg
+++ b/setup.cfg
@@ -23,6 +23,7 @@ console_scripts =
     s3util = edx.analytics.tasks.launchers.s3util:main
 
 edx.analytics.tasks =
+    event-type-dist = edx.analytics.tasks.event_type_dist:PushToVerticaEventTypeDistributionTask
     enrollments-report = edx.analytics.tasks.reports.enrollments:EnrollmentsByWeek
     total-enrollments-report = edx.analytics.tasks.reports.total_enrollments:WeeklyAllUsersAndEnrollments
     inc-enrollments-report = edx.analytics.tasks.reports.incremental_enrollments:WeeklyIncrementalUsersAndEnrollments


brianhw:
A general comment:  as you're developing your code, and especially when writing unit tests, you will want to make use of our code coverage tools.  In particular, running "make coverage" (or "make coverage-local" if you already have your dependencies in your virtualdev) will not only run tests, but make an effort to determine how much of the changed code has test coverage.  This can be useful to provide guidance on what might need additional tests to be written.   It also provides feedback on pep8 and pylint errors, which should be addressed before a branch is done.

edx/analytics/tasks/event_type_dist.py
```diff
@@ -0,0 +1,67 @@
...
+        log.info(type(event_type))
+        yield (event_date, event_type, event_source), 1
+
+    def reducer(self, key, values):
+        event_date_type_source = key
+        event_count = sum(values)
+        yield (event_date_type_source), event_count
+
+    def output(self):
+            return get_target_from_url(url_path_join(self.output_root, 'LoadEventTypeDistributionToVertica/'))
```
brianhw: Why is vertica mentioned in this task's output path?  Probably enough to call this "EventTypeDistribution", or better yet (to match the form of other outputs), "event_type_distribution".

edx/analytics/tasks/event_type_dist.py
```diff
@@ -0,0 +1,67 @@
...
+from edx.analytics.tasks.mapreduce import MapReduceJobTaskMixin, MapReduceJobTask
+from edx.analytics.tasks.pathutil import EventLogSelectionDownstreamMixin, EventLogSelectionMixin
+from edx.analytics.tasks.url import get_target_from_url, url_path_join
+from edx.analytics.tasks.vertica_load import VerticaCopyTask
+
+log = logging.getLogger(__name__)
+
+
+class EventTypeDistributionTask(EventLogSelectionMixin, MapReduceJobTask):
+
```
brianhw: nit: add docstrings as required.

edx/analytics/tasks/tests/test_event_type_dist.py
```diff
@@ -0,0 +1,73 @@
...
+    def setUp(self):
+        self.task_class = EventTypeDistributionTask
+        super(EventTypeDistributionBaseTest, self).setUp()
+
+        self.event_date = "2013-12-17"
+        self.event_type = "problem_check"
+        self.event_source = "browser"
+        self.reduce_key = (self.event_date, self.event_type, self.event_source)  # pylint: disable=no-member
+
+class EventTypeDistributionTaskMapTest(EventTypeDistributionBaseTest, InitializeOpaqueKeysMixin):
```
brianhw: nit:  I would put the Mixin before the BaseTest in the ordering, here and the next class.

edx/analytics/tasks/tests/test_event_type_dist.py
```diff
@@ -0,0 +1,73 @@
...
+    """ Test if mapper works for event type distribution"""
+
+    def test_non_problem_check_event(self):
+        line = 'this is garbage'
+        self.assert_no_map_output_for(line)
+
+    def test_unparseable_problem_check_event(self):
```
brianhw: This test isn't relevant to this Task.  In answer_dist, we only look at problem_check, so we can grep the event in its string form quickly to decide if it's worth parsing.  We don't have the same problem_check prefilter here.

edx/analytics/tasks/tests/test_event_type_dist.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def test_event_type_should_not_begin_with_a_slash(self):
+        line = self.create_event_log_line(event_type='//event')
+        self.assert_no_map_output_for(line)
+
+    def test_event_type_should_not_be_none(self):
+        line = self.create_event_log_line(event_type=None)
+        self.assert_no_map_output_for(line)
+
+
```
brianhw, line 83: You'll want to add some tests for a successful output here, of course....

edx/analytics/tasks/tests/test_event_type_dist.py
```diff
@@ -0,0 +1,73 @@
...
+from edx.analytics.tasks.answer_dist import ProblemCheckEventMixin
+from edx.analytics.tasks.event_type_dist import (EventTypeDistributionTask)
+
+from edx.analytics.tasks.tests import unittest
+from edx.analytics.tasks.tests.map_reduce_mixins import MapperTestMixin, ReducerTestMixin
+from edx.analytics.tasks.tests.opaque_key_mixins import InitializeOpaqueKeysMixin
+
+
+
+class EventTypeDistributionBaseTest(MapperTestMixin, ReducerTestMixin, unittest.TestCase):
```
brianhw: Is there a reason to combine MapperTestMixin and ReducerTestMixin into the same test?  I thought these were usually tested separately.

edx/analytics/tasks/tests/test_event_type_dist.py
```diff
@@ -0,0 +1,73 @@
+""" Test event type distribution per day in a given time interval """
+
+import json
+
+import luigi
+from edx.analytics.tasks.answer_dist import ProblemCheckEventMixin
```
brianhw: Don't need this anymore...

edx/analytics/tasks/tests/test_event_type_dist.py
```diff
@@ -0,0 +1,73 @@
...
+from edx.analytics.tasks.event_type_dist import (EventTypeDistributionTask)
```
brianhw: nit: doesn't need parentheses here

brianhw:
The task definition looks basically solid.  It might need a few names changed and some docstrings, but it's about ready.  For tests, I would look at test_enrollment and/or test_location_by_course for examples of testing map-reduce tasks.

edx/analytics/tasks/event_type_dist.py
```diff
@@ -0,0 +1,94 @@
...
+from edx.analytics.tasks.mapreduce import MapReduceJobTask
+from edx.analytics.tasks.pathutil import EventLogSelectionMixin
+from edx.analytics.tasks.url import ExternalURL, get_target_from_url, url_path_join
+from edx.analytics.tasks.vertica_load import VerticaCopyTask
+
+log = logging.getLogger(__name__)
+
+
+class EventTypeDistributionTask(EventLogSelectionMixin, MapReduceJobTask):
+    """task to compute event_type and event_source values being encountered on each day in a given time interval"""
```
brianhw: nit:  make a complete sentence.  Capitalize the first word, and add a period to the end.

edx/analytics/tasks/event_type_dist.py
```diff
@@ -0,0 +1,94 @@
...
+    output_root = luigi.Parameter()
+
+    events_list_file_path = "hdfs://localhost:9000/data/event_list.tsv"
```
brianhw: This should become a luigi.Parameter, at any rate.  Should work for hdfs or S3.

edx/analytics/tasks/event_type_dist.py
```diff
@@ -0,0 +1,94 @@
...
+    known_events = {}
+
+    def requires_local(self):
+        return ExternalURL(url=self.events_list_file_path)
+
+    def init_local(self):
+        super(EventTypeDistributionTask, self).init_local()
+        #self.input_local().open() as f_in
```
brianhw: nit: at some point, remove comment....

edx/analytics/tasks/event_type_dist.py
```diff
@@ -0,0 +1,94 @@
...
+            return
+        if (event_source,event_type) in self.known_events:
+            event_category = self.known_events[(event_source,event_type)]
+        else:
+            event_category = 'unknown'
+
+        yield (event_date, event_category, event_type, event_source), 1
+
+    def reducer(self, key, values):
+        event_date_type_source = key
```
brianhw: nit: the name is no longer accurate.  But it doesn't really get used here.  You could just use key.  In fact, you could just write it all on a single line.

edx/analytics/tasks/event_type_dist.py
```diff
@@ -0,0 +1,94 @@
...
+    def output(self):
+        return get_target_from_url(url_path_join(self.output_root, 'event_type_distribution/'))
+
+
+class PushToVerticaEventTypeDistributionTask(VerticaCopyTask):
+    """push the event type distribution task data to vertica"""
+    output_root = luigi.Parameter()
+    interval = luigi.DateIntervalParameter()
+    n_reduce_tasks = luigi.Parameter()
+
```
brianhw: nit: extra newline.

edx/analytics/tasks/event_type_dist.py
```diff
@@ -0,0 +1,94 @@
...
+            ('event_count', 'INT'),
+        ]
+
+    @property
+    def insert_source_task(self):
+        return EventTypeDistributionTask(
+            output_root=self.output_root,
+            interval=self.interval,
+            n_reduce_tasks=self.n_reduce_tasks,
+        )
```
brianhw: nit: missing newline.

edx/analytics/tasks/event_type_dist.py
```diff
@@ -0,0 +1,94 @@
...
+        event_date_type_source = key
+        event_count = sum(values)
+        yield (event_date_type_source), event_count
+
+    def output(self):
+        return get_target_from_url(url_path_join(self.output_root, 'event_type_distribution/'))
+
+
+class PushToVerticaEventTypeDistributionTask(VerticaCopyTask):
+    """push the event type distribution task data to vertica"""
```
brianhw: nit: same as before:  capitalize first word, add period at end.  Capitalize Vertica.

edx/analytics/tasks/tests/test_event_type_dist.py
```diff
@@ -0,0 +1,122 @@
...
+                    "org_id": "org_id",
+                    "user_id": "user_id",
+                },
+                "time": self.event_date,
+                "ip": "127.0.0.1",
+                "event": {
+                    "course_id": "course_id",
+                    "user_id": "user_id",
+                    "mode": "honor",
+
```
brianhw: nit: extra newline.

edx/analytics/tasks/tests/test_event_type_dist.py
```diff
@@ -0,0 +1,122 @@
...
+        }
+        self.default_event_template = 'event'
+        self.expected_key = (self.event_date, self.event_type, self.event_source)
+
+    def test_no_event(self):
+        line = 'this is garbage'
+        self.assert_no_map_output_for(line)
+
+    def test_unparseable_event(self):
+        line = 'this is garbage but contains edx.course.enrollment'
```
brianhw: I don't see what this tests that is different from the previous test.

edx/analytics/tasks/tests/test_event_type_dist.py
```diff
@@ -0,0 +1,122 @@
...
+                "event": {
+                    "course_id": "course_id",
+                    "user_id": "user_id",
+                    "mode": "honor",
+
+                }
+            }
+        }
+        self.default_event_template = 'event'
+        self.expected_key = (self.event_date, self.event_type, self.event_source)
```
brianhw: Should now have a category....

edx/analytics/tasks/tests/test_event_type_dist.py
```diff
@@ -0,0 +1,122 @@
...
+@ddt
+class EventTypeDistributionTaskReducerTest(ReducerTestMixin, unittest.TestCase):
+    """Tests to check if event type distribution reducer works"""
+
+    def setUp(self):
+        self.task_class = EventTypeDistributionTask
+        super(EventTypeDistributionTaskReducerTest, self).setUp()
+
+        # Create the task locally, since we only need to check certain attributes
+        self.create_task()
```
brianhw: I don't understand this.  Is there a need to define this method locally and then call it here?  It defines an interval, which just happens to be a different date.  You could just use that date, or rather self.DATE.

edx/analytics/tasks/tests/test_event_type_dist.py
```diff
@@ -0,0 +1,122 @@
...
+        expected = ((reduce_key, sum(values)),)
+        self._check_output_complete_tuple(values, expected)
+
+    def create_task(self, interval='2013-01-01'):
+        """Create a task for testing purposes."""
+        fake_param = luigi.DateIntervalParameter()
+        self.task = EventTypeDistributionTask(
+            interval=fake_param.parse(interval),
+            output_root="/fake/output",
+            n_reduce_tasks=1,
```
brianhw: This seems the only value that is not defined in the general case in ReducerTestMixin.DEFAULT_ARGS.  I presume that it's optional -- and if not, it probably should be.

edx/analytics/tasks/tests/test_event_type_dist.py
brianhw: Actually, I don't know if it matters here in the reducer what the date or the interval is, since it's only used in the mapper, I believe.

edx/analytics/tasks/event_type_dist.py
```diff
@@ -0,0 +1,94 @@
...
+        #self.input_local().open() as f_in
+        with self.input_local().open() as f_in:
+            lines = filter(None, (line.rstrip() for line in f_in))
+
+        for line in lines:
+            if(not line.startswith('#')):
+                parts = line.split(" ")
+                self.known_events[(parts[1],parts[2])] = parts[0]
+        log.log("==================== Umer Debugging Logs =================")
+        log.log(self.known_events)
```
brianhw: what does log.log do?  I'm familiar with log.info, log.error, log.warning, etc....

edx/analytics/tasks/event_type_dist.py
```diff
@@ -0,0 +1,94 @@
...
+    known_events = {}
+
+    def requires_local(self):
+        return ExternalURL(url=self.events_list_file_path)
+
+    def init_local(self):
+        super(EventTypeDistributionTask, self).init_local()
+        #self.input_local().open() as f_in
+        with self.input_local().open() as f_in:
+            lines = filter(None, (line.rstrip() for line in f_in))
```
brianhw: What does this line do?  Particularly the filter?  You have a loop below.

edx/analytics/tasks/event_type_dist.py
```diff
@@ -0,0 +1,94 @@
...
+
+        for line in lines:
+            if(not line.startswith('#')):
+                parts = line.split(" ")
```
brianhw: Did I really give you a file with a space as a delimiter?  That's not a good idea.   It should be a tab or something.  Not that there are embedded spaces in any of the values right now, but one could imagine the category could at some point in the future.

edx/analytics/tasks/event_type_dist.py
```diff
@@ -0,0 +1,94 @@
...
+        if event_source is None or event_type is None or event_date is None:
+            # Ignore if any of the keys is None
+            return
+        if event_type.startswith('/'):
+            # Ignore events that begin with a slash
+            return
+        if (event_source,event_type) in self.known_events:
+            event_category = self.known_events[(event_source,event_type)]
+        else:
+            event_category = 'unknown'
```
brianhw, line 58: While you are adding the category column, I would like to just get the "exported" column included as well (as mentioned in AN-6253).  It involves setting exported to True if in the known events, and to False otherwise.  And this would be added to the key value below.  

The last task, AN-6252, involves making more of an effort to assign an event_category when it's not an event that is being exported.  That would likely involve applying regex expressions read from another file.  We can put that off.

edx/analytics/tasks/event_type_dist.py
```diff
@@ -0,0 +1,91 @@
...
+                parts = line.split("\t")
+                self.known_events[(parts[1],parts[2])] = parts[0]
+
+    def mapper(self, line):
+        value = self.get_event_and_date_string(line)
+        if value is None:
+            return
+        event, date_string = value
+        event_type = event.get('event_type')
+        event_date = date_string
```
brianhw: this does nothing.  Just put event, event_date = value on line 37.

edx/analytics/tasks/event_type_dist.py
```diff
@@ -0,0 +1,91 @@
...
+            exported = True
+        else:
+            event_category = 'unknown'
+
+        yield (event_date, event_category, event_type, event_source, exported), 1
+
+    def reducer(self, key, values):
+        event_date_type_source_category_exported = key
+        event_count = sum(values)
+        yield (event_date_type_source_category_exported), event_count
```
brianhw: just put "key" on this line.   No need to define event_date_type_source_category_exported at all.

edx/analytics/tasks/tests/test_event_type_dist.py
brianhw: And exported.

edx/analytics/tasks/event_type_dist.py
```diff
@@ -0,0 +1,97 @@
...
+            self.known_events = {}
+        else:
+            self.known_events = self.parse_events_list_file()
+
+    def parse_events_list_file(self):
+        """ Read and parse the known events list file and populate it in a dictionary."""
+        with self.input_local().open() as f_in:
+            lines = filter(None, (line.rstrip() for line in f_in))
+        parsed_events = {}
+        for line in lines:
```
brianhw: I still don't understand why there are two loops, one to find lines and the other to loop over each line in lines.  Isn't it possible to just have this be "for line in f_in:" on line 32 and do the processing on it?

edx/analytics/tasks/tests/test_event_type_dist.py
```diff
@@ -0,0 +1,130 @@
...
+                "event": {
+                    "course_id": "course_id",
+                    "user_id": "user_id",
+                    "mode": "honor",
+                }
+            }
+        }
+        self.default_event_template = 'event'
+        self.expected_key = (self.event_date, self.event_category, self.event_type, self.event_source, self.exported)
+        self.task_class.known_events = {("browser", "test_event"): "hi", }
```
brianhw: I would change "hi" to something like "test-category", so it's clearer what it represents.

edx/analytics/tasks/tests/fixtures/events_list.tsv
```diff
@@ -0,0 +1,5 @@
+admin	browser	edx.instructor.report.downloaded
+admin	server	add-forum-admin
+admin	server	add-forum-community-TA
+admin	server	add-forum-mod
+admin	server	add-instructor
```
brianhw: I would prefer not using a fixture for unit tests.  Starting down this road will lead to a lot of clutter that can be avoided.   If we need to remove the test for now, I would do so.  I suggest instead looking at edx-analytics-pipeline/edx/analytics/tasks/reports/tests/test_incremental_enrollments.py or similar.   Looks to me that it would be enough to create a FakeTarget that contains the file text above, and assign a MagicMock to task_class.input_local with it as its return value.

import textwrap
from mock import MagicMock

events_list = """
admin browser edx.instructor.report.downloaded
admin server add-forum-admin
admin server add-forum-community-TA
admin server add-forum-mod
admin server add-instructor
"""
def reformat(string):
    return textwrap.dedent(string).strip().replace(' ', '\t')
input_target = FakeTarget(value=reformat(events_list))
task.input_local = MagicMock(return_value=input_target)

This needs to be set up before init_local() is called.  I'm guessing offhand that it happens in super().setup() below.

edx/analytics/tasks/tests/test_event_type_dist.py
brianhw: I would get rid of this line, and get the events_list passed in through self.events_list to work.

edx/analytics/tasks/tests/test_event_type_dist.py
```diff
@@ -0,0 +1,158 @@
...
+        admin server add-forum-admin
+
+        admin server add-forum-community-TA
+        admin server add-forum-mod
+        admin server add-instructor
+        """
+        input_target = FakeTarget(value=self.reformat(self.events_list))
+
+        super(EventTypeDistributionTaskMapTest, self).setUp()
+        self.task.input_local = MagicMock(return_value=input_target)
```
brianhw, line 31: In order to get this to work in the normal case, you probably need to put it in front of super().setup().

edx/analytics/tasks/tests/test_event_type_dist.py
```diff
@@ -0,0 +1,158 @@
...
+        expected_dict = {
+            ("browser", "edx.instructor.report.downloaded"): "admin",
+            ("server", "add-forum-admin"): "admin",
+            ("server", "add-forum-community-TA"): "admin",
+            ("server", "add-forum-mod"): "admin",
+            ("server", "add-instructor"): "admin",
+        }
+        actual_parsing = self.task.parse_events_list_file()
+        print("------ actual parsing -------")
+        print(actual_parsing)
```
brianhw: Remove these three.  Check instead that the right value was loaded in the task already.

edx/analytics/tasks/tests/test_event_type_dist.py
```diff
@@ -0,0 +1,158 @@
...
+        self.assertEquals(expected_dict, actual_parsing)
+
+    def test_event_list_dictionary_mapping(self):
+        """ Test if the file file parsing works correct."""
+        line = self.create_event_log_line()
+        self.expected_keys = {(self.event_date, "unknown", self.event_type, self.event_source, False),
+                              (self.event_date, "hi", self.event_type, self.event_source, True), }
+        self.assert_single_map_output(line, self.expected_key, 1)
+
```
brianhw, line 114: Modify events_list so that it matches this.

Also, why does this return two keys?  One with "unknown" and one with "hi"?

edx/analytics/tasks/tests/test_event_type_dist.py
```diff
@@ -0,0 +1,162 @@
...
+            ("server", "add-forum-community-TA"): "admin",
+            ("server", "add-forum-mod"): "admin",
+            ("server", "add-instructor"): "admin",
+        }
+        self.assertEquals(expected_dict, self.task.known_events)
+
+    def test_event_list_dictionary_mapping_category_found(self):
+        """ If event_category is found in the list of events."""
+        line = self.create_event_log_line(event_type="edx.instructor.report.downloaded", event_source="browser")
+        self.expected_key = (self.event_date, "admin", "edx.instructor.report.downloaded", self.event_source, True)
```
brianhw, line 106: If you're expecting self.event_source, you should use it in create_event_log_line, and not override with event_source="browser".

edx/analytics/tasks/tests/test_event_type_dist.py
```diff
@@ -0,0 +1,162 @@
...
+        self.assert_single_map_output(line, self.expected_key, 1)
+
+    def test_event_list_dictionary_mapping_category_not_found(self):
+        """ If event_category is not found in the list of events."""
+        line = self.create_event_log_line(event_type="test_type", event_source="browser")
+        self.expected_key = (self.event_date, "unknown", "test_type", self.event_source, False)
```
brianhw, line 112: same thing here.

edx/analytics/tasks/tests/fixtures/events_list.tsv
brianhw: This fixture still needs to be removed.

brianhw:
Just get rid of the fixture file and two lines from the unit tests, and squash all the commits down into a single one.  Then it should be ready to merge.

brianhw:
LGTM. Go ahead and merge! :+1:

PR #179, opened 2015-11-24
Updated opaque-keys version

HassanJaveed84:
@brianhw @mulby


brianhw:
LGTM. :+1:

HassanJaveed84:
@brianhw just a heads up, the previous tag referred to a commit from Oct '14.

mulby:
:+1:

brianhw:
Good to know.  If it passes unit and acceptance tests, however, I think it's good to go.

brianhw:
Did you run the unit and acceptance tests first?  Looks like the unit tests don't even pass.

brianhw:
It looks like there is another fix needed in Opaque Keys to deal with this.

PR #181, opened 2015-11-25
Added tasks for Database deidentification.

HassanJaveed84:
@brianhw @mulby up for initial review.
TODO: A couple of features: scrubbing/encryption.

Parts on which I would really appreciate feedback/suggestions in particular: 
1. FileCopyTask, is there a better way of doing a platform-agnostic file copy ?
2. Writing back json(DeidentifyMongoDumpsTask), right now I'm using json.dumps with ensure_ascii=False and encoding in utf-8 before writing to the output file.
3. Organization and structure of the Victor's re-mapping algo(Permutations class).

### DIFF
--- a/config/devstack.cfg
+++ b/config/devstack.cfg
@@ -33,6 +33,12 @@ required_path_text = FakeServerGroup
 [event-export-course]
 output_root = hdfs://localhost:9000/edx-analytics-pipeline/event-export-by-course/output/
 
+[data-deidentification]
+output_root = hdfs://localhost:9000/edx-analytics-pipeline/data-deidentification/output/
+
+[id-codec]
+seed_value = 42
+
 [manifest]
 threshold = 500
 input_format = org.edx.hadoop.input.ManifestTextInputFormat
--- a/config/test.cfg
+++ b/config/test.cfg
@@ -37,6 +37,9 @@ required_path_text = FakeServerGroup
 [event-export-course]
 output_root = s3://fake/
 
+[id-codec]
+seed_value = 42
+
 [manifest]
 threshold = 500
 input_format = oddjob.ManifestTextInputFormat
--- a/edx/analytics/tasks/data_deidentification.py
+++ b/edx/analytics/tasks/data_deidentification.py
@@ -1,0 +1,346 @@
+"""Deidentify course state files by removing/stubbing user information."""
+
+import luigi
+import luigi.s3
+import csv
+import json
+import os
+import re
+
+from edx.analytics.tasks.pathutil import PathSetTask
+from edx.analytics.tasks.url import get_target_from_url, url_path_join
+from edx.analytics.tasks.util.id_codec import UserIdRemapperMixin
+import edx.analytics.tasks.util.opaque_key_util as opaque_key_util
+from edx.analytics.tasks.util.file_util import FileCopyMixin
+import edx.analytics.tasks.util.csv_util
+
+import logging
+
+log = logging.getLogger(__name__)
+
+
+class BaseDeidentifyDumpTask(UserIdRemapperMixin, luigi.Task):
+    """
+    Base class for deidentification of state files.
+    """
+
+    course = luigi.Parameter()
+    output_directory = luigi.Parameter()
+    data_directory = luigi.Parameter()
+
+    def output(self):
+        if len(self.input()) == 0:
+            raise IOError("Course File '{filename}' not found for course '{course}'".format(
+                filename=self.file_pattern, course=self.course
+            ))
+
+        output_filename = os.path.basename(self.input()[0].path)
+        return get_target_from_url(url_path_join(self.output_directory, output_filename))
+
+    def requires(self):
+        return PathSetTask([self.data_directory], [self.file_pattern])
+
+    @property
+    def file_pattern(self):
+        """Provides the file pattern for input file."""
+        raise NotImplementedError
+
+
+class DeidentifySqlDumpTask(BaseDeidentifyDumpTask):
+    """Task to deidentify an sql file."""
+
+    def run(self):
+        with self.output().open('w') as output_file:
+            with self.input()[0].open('r') as input_file:
+                writer = csv.writer(output_file, dialect='mysqlpipe')
+                reader = csv.reader(input_file, dialect='mysqlpipe')
+
+                headers = next(reader, None)
+
+                if headers:
+                    writer.writerow(headers)
+
+                for row in reader:
+                    # Found an empty line at the bottom of test file
+                    if row:
+                        filtered_row = self.filter_row(row)
+                        writer.writerow(filtered_row)
+
+    def filter_row(self, row):
+        """Function to deitentify a particular row."""
+        raise NotImplementedError
+
+
+class DeidentifyAuthUserProfileTask(DeidentifySqlDumpTask):
+    """Task to deidentify auth_userprofile dump file."""
+
+    @property
+    def file_pattern(self):
+        return '*-auth_userprofile-*'
+
+    def filter_row(self, row):
+        row[1] = self.remap_id(row[1])  # user_id
+        row[2] = ''  # name
+        row[3] = ''  # language
+        row[4] = ''  # location
+        row[5] = ''  # meta
+        row[6] = ''  # courseware
+        row[8] = ''  # mailing_address
+        row[12] = '1'  # allow_certificate
+        row[14] = ''  # city
+        row[15] = ''  # bio
+
+        return row
+
+
+class DeidentifyAuthUserTask(DeidentifySqlDumpTask):
+    """Task to deidentify auth_user dump file."""
+
+    @property
+    def file_pattern(self):
+        return '*-auth_user-*'
+
+    def filter_row(self, row):
+        row[0] = self.remap_id(row[0])  # id
+        row[1] = "username_{id}".format(id=row[0])
+        row[2] = ''  # first_name
+        row[3] = ''  # last_name
+        row[4] = ''  # email
+        row[5] = ''  # passwords
+
+        for i in xrange(11, len(row)):
+            row[i] = ''
+
+        return row
+
+
+class DeidentifyStudentCourseEnrollmentTask(DeidentifySqlDumpTask):
+    """Task to deidentify student_courseenrollment dump file."""
+
+    @property
+    def file_pattern(self):
+        return '*-student_courseenrollment-*'
+
+    def filter_row(self, row):
+        row[1] = self.remap_id(row[1])  # user_id
+        return row
+
+
+class DeidentifyUserApiUserCourseTagTask(DeidentifySqlDumpTask):
+    """Task to deidentify user_api_usercoursetag dump file."""
+
+    @property
+    def file_pattern(self):
+        return '*-user_api_usercoursetag-*'
+
+    def filter_row(self, row):
+        row[1] = self.remap_id(row[1])  # user_id
+        return row
+
+
+class DeidentifyStudentLanguageProficiencyTask(DeidentifySqlDumpTask):
+    """Task to deidentify student_languageproficiency dump file."""
+
+    @property
+    def file_pattern(self):
+        return '*-student_languageproficiency-*'
+
+    def filter_row(self, row):
+        return row
+
+
+class DeidentifyCoursewareStudentModule(DeidentifySqlDumpTask):
+    """Task to deidentify courseware_studentmodule dump file."""
+
+    @property
+    def file_pattern(self):
+        return '*-courseware_studentmodule-*'
+
+    def filter_row(self, row):
+        row[3] = self.remap_id(row[3])  # student_id
+        #row[4] = ?  # state
+        return row
+
+
+class DeidentifyCertificatesGeneratedCertificate(DeidentifySqlDumpTask):
+    """Task to deidentify certificates_generatedcertificate dump file."""
+
+    @property
+    def file_pattern(self):
+        return '*-certificates_generatedcertificate-*'
+
+    def filter_row(self, row):
+        row[1] = self.remap_id(row[1])  # user_id
+        row[2] = ''  # download_url
+        row[8] = ''  # verify_uuid
+        row[9] = ''  # download_uuid
+        row[10] = ''  # name
+        row[13] = ''  # error_reason
+        return row
+
+
+class DeidentifyTeamsTask(DeidentifySqlDumpTask):
+    """Task to deidentify teams dump file."""
+
+    @property
+    def file_pattern(self):
+        return '*-teams-*'
+
+    def filter_row(self, row):
+        return row
+
+
+class DeidentifyTeamsMembershipTask(DeidentifySqlDumpTask):
+    """Task to deidentify teams_membership dump file."""
+
+    @property
+    def file_pattern(self):
+        return '*-teams_membership-*'
+
+    def filter_row(self, row):
+        row[1] = self.remap_id(row[1])  # user_id
+        return row
+
+
+class DeidentifyVerificationStatusTask(DeidentifySqlDumpTask):
+    """Task to deidentify verify_student_verificationstatus dump file."""
+
+    @property
+    def file_pattern(self):
+        return '*-verify_student_verificationstatus-*'
+
+    def filter_row(self, row):
+        row[4] = self.remap_id(row[4])  # user_id
+        return row
+
+
+class DeidentifyWikiArticleTask(DeidentifySqlDumpTask):
+    """Task to deidentify wiki_article dump file."""
+
+    @property
+    def file_pattern(self):
+        return '*-wiki_article-*'
+
+    def filter_row(self, row):
+        # row[4] = ? # owner_id
+        # row[5] = ? # group_id
+        return row
+
+
+class DeidentifyWikiArticleRevisionTask(DeidentifySqlDumpTask):
+    """Task to deidentify wiki_articlerevision dump file."""
+
+    @property
+    def file_pattern(self):
+        return '*-wiki_articlerevision-*'
+
+    def filter_row(self, row):
+        row[2] = ''  # user_message
+        row[3] = ''  # automatic_log
+        row[4] = ''  # ip_address
+        row[5] = self.remap_id(row[5])  # user_id
+        return row
+
+
+class CourseStructureTask(FileCopyMixin, BaseDeidentifyDumpTask):
+    """Task to copy course_structure dump file, no deidentification needed."""
+
+    @property
+    def file_pattern(self):
+        return '*-course_structure-*'
+
+
+class CourseContentTask(FileCopyMixin, BaseDeidentifyDumpTask):
+    """Task to copy course dump file, no deidentification needed."""
+
+    @property
+    def file_pattern(self):
+        return '*-course-*'
+
+
+class DeidentifyMongoDumpsTask(BaseDeidentifyDumpTask):
+    """Task to deidentify mongo dump file."""
+
+    @property
+    def file_pattern(self):
+        return '*mongo*'
+
+    def run(self):
+        with self.output().open('w') as output_file:
+            with self.input()[0].open('r') as input_file:
+                for line in input_file:
+                    row = json.loads(line)
+                    filtered_row = self.filter_row(row)
+                    output_file.write(json.dumps(filtered_row, ensure_ascii=False).encode('utf-8'))
+                    output_file.write('\n')
+
+    def filter_row(self, row):
+        """Replace/remove sensitive information."""
+        row['author_id'] = str(self.remap_id(row['author_id']))
+        row['author_username'] = "username_{id}".format(id=row['author_id'])
+        # TODO: scrub body
+        # TODO: encrypt votes { up: [ user_ids.... ], down: [ user_ids...] }
+        return row
+
+
+class DeidentifyCourseDumpTask(luigi.WrapperTask):
+    """Wrapper task to deidentify data for a particular course."""
+
+    course = luigi.Parameter()
+    dump_root = luigi.Parameter()
+    output_root = luigi.Parameter()
+
+    def __init__(self, *args, **kwargs):
+        super(DeidentifyCourseDumpTask, self).__init__(*args, **kwargs)
+
+        filename_safe_course_id = opaque_key_util.get_filename_safe_course_id(self.course)
+        auth_userprofile_targets = PathSetTask([url_path_join(self.dump_root, filename_safe_course_id, 'state')],
+                                               ['*auth_userprofile*']).output()
+        # TODO: Refactor out this logic of getting latest file. Right now we expect a date, so we use that
+        dates = [re.search(r"\d{4}-\d{2}-\d{2}", target.path).group() for target in auth_userprofile_targets]
+        latest_date = sorted(dates)[-1]
+        self.data_directory = url_path_join(self.dump_root, filename_safe_course_id, 'state', latest_date)
+        self.output_directory = url_path_join(self.output_root, filename_safe_course_id)
+
+    def requires(self):
+        kwargs = {
+            'course': self.course,
+            'output_directory': self.output_directory,
+            'data_directory': self.data_directory
+        }
+        yield (
+            DeidentifyAuthUserTask(**kwargs),
+            DeidentifyAuthUserProfileTask(**kwargs),
+            DeidentifyStudentCourseEnrollmentTask(**kwargs),
+            DeidentifyUserApiUserCourseTagTask(**kwargs),
+            DeidentifyStudentLanguageProficiencyTask(**kwargs),
+            DeidentifyCoursewareStudentModule(**kwargs),
+            DeidentifyCertificatesGeneratedCertificate(**kwargs),
+            DeidentifyTeamsTask(**kwargs),
+            DeidentifyTeamsMembershipTask(**kwargs),
+            DeidentifyVerificationStatusTask(**kwargs),
+            DeidentifyWikiArticleTask(**kwargs),
+            DeidentifyWikiArticleRevisionTask(**kwargs),
+            CourseContentTask(**kwargs),
+            CourseStructureTask(**kwargs),
+            DeidentifyMongoDumpsTask(**kwargs),
+        )
+
+
+class DataDeidentificationTask(luigi.WrapperTask):
+    """Wrapper task for data deidentification."""
+
+    course = luigi.Parameter(is_list=True)
+    dump_root = luigi.Parameter()
+    output_root = luigi.Parameter(
+        config_path={'section': 'data-deidentification', 'name': 'output_root'}
+    )
+
+    def requires(self):
+        for course in self.course:
+            kwargs = {
+                'dump_root': self.dump_root,
+                'course': course,
+                'output_root': self.output_root,
+            }
+            yield DeidentifyCourseDumpTask(**kwargs)
--- a/edx/analytics/tasks/tests/test_data_deidentification.py
+++ b/edx/analytics/tasks/tests/test_data_deidentification.py
@@ -1,0 +1,266 @@
+"""
+Tests for data deidentification tasks.
+"""
+
+from edx.analytics.tasks.tests import unittest
+from edx.analytics.tasks.tests.target import FakeTarget
+import edx.analytics.tasks.data_deidentification as deid
+import edx.analytics.tasks.util.opaque_key_util as opaque_key_util
+from edx.analytics.tasks.url import url_path_join
+
+from mock import MagicMock
+from mock import sentinel
+
+import json
+import tempfile
+import os
+import shutil
+
+
+class TestDataDeidentification(unittest.TestCase):
+    """Tests for all data deidentification tasks."""
+
+    def run_task(self, task_cls, source):
+        """Runs the task with fake targets."""
+
+        task = task_cls(
+            course=sentinel.ignored,
+            output_directory=sentinel.ignored,
+            data_directory=sentinel.ignored,
+        )
+
+        fake_input = [FakeTarget(value=source)]
+        task.input = MagicMock(return_value=fake_input)
+
+        output_target = FakeTarget()
+        task.output = MagicMock(return_value=output_target)
+
+        task.run()
+        return output_target.buffer.read()
+
+    def reformat(self, data):
+        """Reformat data to make it like a TSV."""
+        return "\n".join(["\t".join(row) for row in data]) + '\n'
+
+    def check_output(self, cls, input_value, expected_value):
+        """Compares input and expected values."""
+        output = self.run_task(task_cls=cls, source=self.reformat(input_value))
+        self.assertEquals(output, self.reformat(expected_value))
+
+    def test_auth_user_deidentification(self):
+        header = ['id', 'username', 'first_name', 'last_name', 'email', 'password', 'is_staff', 'is_active',
+                  'is_superuser', 'last_login', 'date_joined', 'status', 'email_key', 'avatar_type', 'country',
+                  'show_country', 'date_of_birth', 'interesting_tags', 'ignored_tags', 'email_tag_filter_strategy',
+                  'display_tag_filter_strategy', 'consecutive_days_visit_count']
+        data = [
+            header,
+            ['123456', 'JohnDoe', 'John', 'Doe', 'johndoe@edx.org', '', '1', '1',
+             '0', '2015-11-15 22:08:37', '2013-07-08 14:42:50', '', 'NULL', '', '',
+             '0', 'NULL', '', '', '0',
+             '0', '0']
+        ]
+        expected = [
+            header,
+            ['273678626', 'username_273678626', '', '', '', '', '1', '1',
+             '0', '2015-11-15 22:08:37', '2013-07-08 14:42:50', '', '', '', '',
+             '', '', '', '', '',
+             '', '']
+        ]
+        self.check_output(deid.DeidentifyAuthUserTask, data, expected)
+
+    def test_auth_user_profile_deidentification(self):
+        header = ['id', 'user_id', 'name', 'language', 'location',
+                  'meta', 'courseware', 'gender',
+                  'mailing_address', 'year_of_birth', 'level_of_education', 'goals', 'allow_certificate', 'country',
+                  'city', 'bio', 'profile_image_uploaded_at']
+        data = [
+            header,
+            ['123', '123456', 'John Doe', 'English', 'Batcave, USA',
+             '{"old_names": [["old name", "Name change", "2015-09-07T02:30:17.735773+00:00"]]}', 'course.xml', 'm',
+             '4th Street', '1984', 'hs', 'To be someone', '0', 'NA',
+             'ID', 'I like to code', '2015-11-21 22:17:57']
+        ]
+        expected = [
+            header,
+            ['123', '273678626', '', '', '',
+             '', '', 'm',
+             '', '1984', 'hs', 'To be someone', '1', 'NA',
+             '', '', '2015-11-21 22:17:57']
+        ]
+        self.check_output(deid.DeidentifyAuthUserProfileTask, data, expected)
+
+    def test_student_course_enrollment_deidentification(self):
+        header = ['id', 'user_id', 'course_id', 'created', 'is_active', 'mode']
+        data = [
+            header,
+            ['123', '123456', 'course-v1:edX+DemoX+Test_2014', '2015-07-16 19:19:10', '1', 'honor'],
+            ['124', '123457', 'course-v1:edX+DemoX+Test_2014', '2015-07-28 12:41:13', '0', 'verified'],
+        ]
+        expected = [
+            header,
+            ['123', '273678626', 'course-v1:edX+DemoX+Test_2014', '2015-07-16 19:19:10', '1', 'honor'],
+            ['124', '273680674', 'course-v1:edX+DemoX+Test_2014', '2015-07-28 12:41:13', '0', 'verified'],
+        ]
+        self.check_output(deid.DeidentifyStudentCourseEnrollmentTask, data, expected)
+
+    def test_student_language_proficiency_deidentification(self):
+        header = ['id', 'user_profile_id', 'code']
+        data = [
+            header,
+            ['1', '145', 'en'],
+            ['2', '941', 'zh'],
+            ['3', '81724', 'ar'],
+        ]
+        expected = [
+            header,
+            ['1', '145', 'en'],
+            ['2', '941', 'zh'],
+            ['3', '81724', 'ar'],
+        ]
+        self.check_output(deid.DeidentifyStudentLanguageProficiencyTask, data, expected)
+
+    def test_courseware_student_module_deidentification(self):
+        header = ['id', 'module_type', 'module_id', 'student_id',
+                  'state', 'grade', 'created', 'modified', 'max_grade', 'done',
+                  'course_id']
+        data = [
+            header,
+            ['1', 'problem', 'block-v1:edX+DemoX+Test_2014+type@problem+block@123091b4012312r210r120r12r', '123456',
+             '{"attempts": 1, "seed": 1, "done": true}', '0', '2015-10-13 19:22:24', '2015-10-13 19:40:20', '1', 'na',
+             'course-v1:edX+DemoX+Test_2014'],
+        ]
+        expected = [
+            header,
+            ['1', 'problem', 'block-v1:edX+DemoX+Test_2014+type@problem+block@123091b4012312r210r120r12r', '273678626',
+             '{"attempts": 1, "seed": 1, "done": true}', '0', '2015-10-13 19:22:24', '2015-10-13 19:40:20', '1', 'na',
+             'course-v1:edX+DemoX+Test_2014'],
+        ]
+        self.check_output(deid.DeidentifyCoursewareStudentModule, data, expected)
+
+    def test_certificates_generated_certificate_deidentification(self):
+        header = ['id', 'user_id', 'download_url', 'grade', 'course_id', 'key', 'distinction', 'status',
+                  'verify_uuid', 'download_uuid', 'name', 'created_date', 'modified_date',
+                  'error_reason', 'mode']
+        data = [
+            header,
+            ['1', '123456', 'some_url', '0.21', 'course-v1:edX+DemoX+Test_2014', 'key', '0', 'notpassing',
+             'verify_uuid', 'download_uuid', 'John Doe', '2015-10-16 12:53:49', '2015-10-16 12:53:49',
+             'error_reason', 'honor']
+        ]
+        expected = [
+            header,
+            ['1', '273678626', '', '0.21', 'course-v1:edX+DemoX+Test_2014', 'key', '0', 'notpassing',
+             '', '', '', '2015-10-16 12:53:49', '2015-10-16 12:53:49',
+             '', 'honor']
+        ]
+        self.check_output(deid.DeidentifyCertificatesGeneratedCertificate, data, expected)
+
+    def test_teams_deidentification(self):
+        header = ['id', 'team_id', 'name', 'course_id', 'topic_id',
+                  'date_created', 'description', 'country', 'language', 'discussion_topic_id', 'last_activity_at',
+                  'team_size']
+        data = [
+            header,
+            ['1', 'A-Team-8883d3b43094f0e9e6ec7e190e7600e', 'A Team', 'course-v1:edX+DemoX+Test_2014', 'some_topic',
+             '2015-10-13 13:14:41', 'description', 'GB', 'en', 'topic_id', '2015-10-31 21:32:17',
+             '8']
+        ]
+        expected = [
+            header,
+            ['1', 'A-Team-8883d3b43094f0e9e6ec7e190e7600e', 'A Team', 'course-v1:edX+DemoX+Test_2014', 'some_topic',
+             '2015-10-13 13:14:41', 'description', 'GB', 'en', 'topic_id', '2015-10-31 21:32:17',
+             '8']
+        ]
+        self.check_output(deid.DeidentifyTeamsTask, data, expected)
+
+    def test_teams_membership_deidentification(self):
+        header = ['id', 'user_id', 'team_id', 'date_joined', 'last_activity_at']
+        data = [
+            header,
+            ['1', '123456', '1', '2015-10-13 13:14:41', '2015-10-14 18:41:24']
+        ]
+        expected = [
+            header,
+            ['1', '273678626', '1', '2015-10-13 13:14:41', '2015-10-14 18:41:24']
+        ]
+        self.check_output(deid.DeidentifyTeamsMembershipTask, data, expected)
+
+    def test_verification_status_deidentification(self):
+        header = ['timestamp', 'status', 'course_id',
+                  'checkpoint_location', 'user_id']
+        data = [
+            header,
+            ['2015-09-03 07:19:10', 'submitted', 'course-v1:edX+DemoX+Test_2014',
+             'block-v1:edX+DemoX+Test_2014+type@edx', '123456']
+        ]
+        expected = [
+            header,
+            ['2015-09-03 07:19:10', 'submitted', 'course-v1:edX+DemoX+Test_2014',
+             'block-v1:edX+DemoX+Test_2014+type@edx', '273678626']
+        ]
+        self.check_output(deid.DeidentifyVerificationStatusTask, data, expected)
+
+    def test_wiki_article_deidentification(self):
+        header = ['id', 'current_revision_id', 'created', 'modified', 'owner_id', 'group_id', 'group_read',
+                  'group_write', 'other_read', 'other_write']
+        data = [
+            header,
+            ['1234', '27567', '2013-08-08 22:00:58', '2013-09-30 16:52:21', 'owner_id', 'group_id', '1',
+             '2', '3', '4']
+        ]
+        expected = [
+            header,
+            ['1234', '27567', '2013-08-08 22:00:58', '2013-09-30 16:52:21', 'owner_id', 'group_id', '1',
+             '2', '3', '4']
+        ]
+        self.check_output(deid.DeidentifyWikiArticleTask, data, expected)
+
+    def test_wiki_article_revision_deidentification(self):
+        header = ['id', 'revision_number', 'user_message', 'automatic_log', 'ip_address', 'user_id', 'modified',
+                  'created', 'previous_revision_id', 'deleted', 'locked', 'article_id', 'content', 'title']
+        data = [
+            header,
+            ['23456', '1', 'This is a user message', 'automatic_log', '192.168.1.1', '123456', '2013-08-08 22:00:58',
+             '2013-08-08 22:00:58', '123', '0', '0', '123', 'This is the content', 'TITLE']
+        ]
+        expected = [
+            header,
+            ['23456', '1', '', '', '', '273678626', '2013-08-08 22:00:58',
+             '2013-08-08 22:00:58', '123', '0', '0', '123', 'This is the content', 'TITLE']
+        ]
+        self.check_output(deid.DeidentifyWikiArticleRevisionTask, data, expected)
+
+    def test_mongo_deidentification(self):
+        data = '{"author_id":"123456","author_username":"johndoe","body":"This is a body", ' \
+               '"course_id":"course-v1:edX+DemoX+Test_2014","votes":{"down":["123456"],"up":["12345"]}}'
+        expected = '{"author_id":"273678626","author_username":"username_273678626","body":"This is a body", ' \
+                   '"course_id":"course-v1:edX+DemoX+Test_2014","votes":{"down":["123456"],"up":["12345"]}}'
+        output = self.run_task(task_cls=deid.DeidentifyMongoDumpsTask, source=data)
+        self.assertDictEqual(json.loads(output), json.loads(expected))
+
+
+class TestDeidentifyCourseDumpTask(unittest.TestCase):
+    """Test for DeidentifyCourseDumpTask."""
+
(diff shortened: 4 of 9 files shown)


edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,296 @@
...
+    def run(self):
+        if(len(self.input()) == 0):
+            #throw exception
+            pass
+
+        input_file = self.input()[0].open('r')
+        output_file = self.output().open('w')
+
+        for line in input_file:
+            output_file.write(line)
```
brianhw: Are the files being copied text files?  I'm concerned about relying on newlines to define the buffer size here, if the files are actually binary.  It could end up reading the entire file into line, then writing it out.   And at least one of the files is xml.gz.

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,296 @@
...
+class BaseDeidentifyDump(object):
+
+    course = luigi.Parameter()
+    output_directory = luigi.Parameter()
+    data_directory = luigi.Parameter()
+
+    permutation = Permutations(42,32,32)
+
+    def output(self):
+        output_filename = self.input()[0].path.rsplit('/',1)[-1]
```
brianhw: Any reason not to use something like os.path.basename?

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,296 @@
...
+from edx.analytics.tasks.pathutil import PathSetTask
+from edx.analytics.tasks.url import get_target_from_url, url_path_join
+from edx.analytics.tasks.util.id_codec import Permutations
+import edx.analytics.tasks.util.opaque_key_util as opaque_key_util
+
+import logging
+
+log = logging.getLogger(__name__)
+
+class BaseDeidentifyDump(object):
```
brianhw: I'm somewhat unclear on the naming of classes here.  This "base" class seems to be used more like a "mixin".  But it might be reasonably made into a base "Task" by inheriting from luigi.Task here.  

What could be pulled into a mixin is a RemapIdMixin (or something better named), which would define permutation and remap_id().  I expect that would be reused by the event deid stuff as well.

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,296 @@
...
+    permutation = Permutations(42,32,32)
+
+    def output(self):
+        output_filename = self.input()[0].path.rsplit('/',1)[-1]
+        return get_target_from_url(url_path_join(self.output_directory, output_filename))
+
+    def remap_id(self, id):
+        return self.permutation.permute(int(id))
+
+class DeidentifySqlDumpMixin(BaseDeidentifyDump):
```
brianhw: Similarly, this would probably become a Task instead of a mixin.

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,296 @@
...
+class DeidentifyUserApiUsercoursetagTask(DeidentifySqlDumpMixin, luigi.Task):
+
+    def requires(self):
+        return PathSetTask([self.data_directory], ['*-user_api_usercoursetag-*'])
+
+    def filter_row(self, row):
+        row[1] = self.remap_id(row[1]) #user_id
+        return row
+
+class DeidentifyStudentLanguageproficiencyTask(DeidentifySqlDumpMixin, luigi.Task):
```
brianhw: nit:  consistent camelCase:   LanguageProficiency

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,296 @@
...
+class DeidentifyStudentCourseenrollmentTask(DeidentifySqlDumpMixin, luigi.Task):
+
+    def requires(self):
+        return PathSetTask([self.data_directory], ['*-student_courseenrollment-*'])
+
+    def filter_row(self, row):
+        row[1] = self.remap_id(row[1]) #user_id
+        return row
+
+class DeidentifyUserApiUsercoursetagTask(DeidentifySqlDumpMixin, luigi.Task):
```
brianhw: nit:  consistent camelCase: UserCourseTag

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,296 @@
...
+        row[3] = '' #last_name
+        row[4] = '' #email
+        row[5] = '' #passwords
+
+        for i in xrange(11,len(row)):
+            row[i] = ''
+
+        return row
+
+class DeidentifyStudentCourseenrollmentTask(DeidentifySqlDumpMixin, luigi.Task):
```
brianhw: nit:  consistent camelCase: StudentCourseEnrollment

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,296 @@
...
+
+        input_file = self.input()[0].open('r')
+        output_file = self.output().open('w')
+
+        for line in input_file:
+            output_file.write(line)
+
+        output_file.close()
+
+class DeidentifyAuthUserprofileTask(DeidentifySqlDumpMixin, luigi.Task):
```
brianhw: nit:  consistent camelCase: AuthUserProfile

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,296 @@
...
+class DeidentifyTeamsMembershipTask(DeidentifySqlDumpMixin, luigi.Task):
+
+    def requires(self):
+        return PathSetTask([self.data_directory], ['*-teams_membership-*'])
+
+    def filter_row(self, row):
+        row[1] = self.remap_id(row[1]) #user_id
+        return row
+
+class DeidentifyVerifyStudentVerificationstatus(DeidentifySqlDumpMixin, luigi.Task):
```
brianhw: nit:  consistent camelCase: StudentVerificationStatus.   Also add "Task".  If it's too long, remove the "Status" and add "Task" instead.

brianhw: Actually, no, if length is an issue, I would leave "VerificationStatusTask", and remove the "VerifyStudent" first.

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,296 @@
...
+
+class DeidentifyWikiArticleTask(DeidentifySqlDumpMixin, luigi.Task):
+
+    def requires(self):
+        return PathSetTask([self.data_directory], ['*-wiki_article-*'])
+
+    def filter_row(self, row):
+        return row
+
+class DeidentifyWikiArticlerevisionTask(DeidentifySqlDumpMixin, luigi.Task):
```
brianhw: nit:  consistent camelCase: ArticleRevision

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,296 @@
...
+            # throw exception
+            pass
+
+        input_file = self.input()[0].open('r')
+        output_file = self.output().open('w')
+
+        for line in input_file:
+            row = json.loads(line)
+            filtered_row = self.filter_row(row)
+            output_file.write("{}\n".format(json.dumps(filtered_row, ensure_ascii=False).encode('utf-8')))
```
brianhw: or you could have a second output_file.write('\n') call....

brianhw: When you get around to writing docstrings and comments, explaining the need for ensure_ascii and adding the encoding would be helpful.

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,296 @@
...
+    course = luigi.Parameter()
+    dump_root = luigi.Parameter()
+    output_root = luigi.Parameter()
+
+    def __init__(self, *args, **kwargs):
+        super(DeidentifyCourseDumpTask, self).__init__(*args, **kwargs)
+
+        filename_safe_course_id = opaque_key_util.get_filename_safe_course_id(self.course)
+        auth_userprofile_targets = PathSetTask([url_path_join(self.dump_root, filename_safe_course_id, 'state')], ['*auth_userprofile*']).output()
+        auth_userprofile_paths = map((lambda target: target.path), auth_userprofile_targets)
```
brianhw: This is the same as a list comprehension.  I'm more used to the latter as being 'pythonic', and map as more functional -- does this have advantages?

edx/analytics/tasks/deidentify_dumps.py
HassanJaveed84: Course content is the xml.tar.gz. I tried to open the files with 'b' mode, but it is not supported. However this code still produces correct output files as I it passed a comparison. Do you have a suggestion on a more reliable approach ?

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,296 @@
...
+        latest_auth_userprofile = sorted(auth_userprofile_paths)[-1]
+        self.data_directory = latest_auth_userprofile[0:latest_auth_userprofile.rfind('/')]
```
brianhw: Any reason not to use os.path functions here too?

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,296 @@
...
+            CourseContentTask(**kwargs), # a gzip compressed file, need a way to do simple source/dest copy
+            CourseStructureTask(**kwargs),
+            DeidentifyMongoDumpsTask(**kwargs),
+        )
+
+class DeidentifyDumpsTask(luigi.WrapperTask):
+
+    course = luigi.Parameter(is_list=True)
+    dump_root = luigi.Parameter()
+    output_root = luigi.Parameter(default='hdfs://localhost:9000/edx-analytics-pipeline/output/')
```
brianhw: Eventually such default values should come from config files.  And this looks like it would go in the devstack one.

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,296 @@
...
+            row = json.loads(line)
+            filtered_row = self.filter_row(row)
+            output_file.write("{}\n".format(json.dumps(filtered_row, ensure_ascii=False).encode('utf-8')))
+
+        output_file.close()
+
+    def filter_row(self, row):
+        row['author_id'] = self.remap_id(row['author_id'])
+        row['author_username'] = "author_username_{id}".format(id=row['author_id'])
+        #scrub & encrypt ?
```
brianhw: encrypt?

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,296 @@
...
+        for line in input_file:
+            row = json.loads(line)
+            filtered_row = self.filter_row(row)
+            output_file.write("{}\n".format(json.dumps(filtered_row, ensure_ascii=False).encode('utf-8')))
+
+        output_file.close()
+
+    def filter_row(self, row):
+        row['author_id'] = self.remap_id(row['author_id'])
+        row['author_username'] = "author_username_{id}".format(id=row['author_id'])
```
brianhw: I think we should use the same username across the entire data.  I had suggested "username_{id}".  So I would leave off the "author_".

edx/analytics/tasks/deidentify_dumps.py
brianhw: What did you think of the copy code in encrypt.py?  It uses a fixed buffer.  It would also allow the injection of a counter if the copying took a long time, to keep hadoop from cancelling the task due to inactivity.

edx/analytics/tasks/deidentify_dumps.py
brianhw: Actually I would still prefer to find a way for this instance of the Permutation to be a singleton, rather than recalculated for each class that includes it.  That's different than making the Permutations class a singleton.  But maybe there should be a UserIdRemapper class that is a singleton, that provides Permutations with the initialization parameters, and provides a class-level method to do the remapping.

edx/analytics/tasks/util/id_codec.py
```diff
@@ -12,3 +14,46 @@ def decode_id(encoded_id):
     """Decode an ID string back to the original scope-type-value tuple."""
     scope, id_type, id_value = base64.b32decode(encoded_id).split('|')
     return scope, id_type, id_value
+
+class Permutations:
```
brianhw: This is a pretty general name for the class.  Can it be a bit more descriptive?

brianhw:
That's a pass for now -- sorry I ran out of time to dig further into your specific questions.

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,296 @@
...
+import datetime
+import csv
+import json
+import os
+
+from edx.analytics.tasks.pathutil import PathSetTask
+from edx.analytics.tasks.url import get_target_from_url, url_path_join
+from edx.analytics.tasks.util.id_codec import UserIdRemapperMixin
+import edx.analytics.tasks.util.opaque_key_util as opaque_key_util
+from edx.analytics.tasks.encrypt import _copy_file_to_open_file
```
brianhw: if this is no longer just for encryption, we should move it to a different module.    And take off the leading underscore, since it's no longer internal.

brianhw: Is there a good candidate in utils, or do we need to create a separate one?

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,296 @@
...
+
+import logging
+
+log = logging.getLogger(__name__)
+
+class BaseDeidentifyDumpTask(UserIdRemapperMixin, luigi.Task):
+
```
brianhw: nit: docstrings...

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,296 @@
...
+    course = luigi.Parameter()
+    output_directory = luigi.Parameter()
+    data_directory = luigi.Parameter()
+
+    def output(self):
+        output_filename = os.path.basename(self.input()[0].path)
+        return get_target_from_url(url_path_join(self.output_directory, output_filename))
+
```
brianhw: nit: two blank lines...

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,296 @@
...
+class DeidentifySqlDumpTask(BaseDeidentifyDumpTask):
+
+    def run(self):
+        if(len(self.input()) == 0):
```
brianhw: nit: space after if, and no parentheses around expression.

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,296 @@
...
+            #throw exception
+            pass
```
brianhw: wait, when does this happen?  I'm not understanding when input() would be empty.

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,296 @@
...
+                filtered_row = self.filter_row(row)
+                writer.writerow(filtered_row)
+
+        output_file.close()
+
+    def filter_row(self, row):
+        raise NotImplementedError
+
+class FileCopyTask(object):
+
```
brianhw: Maybe this whole task should go to the utility.  

And is there something reasonable here to share between this and encrypt?

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,296 @@
...
+        finally:
+            output_file.close()
+
+class DeidentifyAuthUserProfileTask(DeidentifySqlDumpTask):
+
+    def requires(self):
+        return PathSetTask([self.data_directory], ['*-auth_userprofile-*'])
+
+    def filter_row(self, row):
+        row[1] = self.remap_id(row[1]) #user_id
```
brianhw: nit: space after '#'

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,296 @@
...
+        output_file = self.output().open('w')
+        try:
+            _copy_file_to_open_file(self.input()[0].path, output_file, progress=report_progress)
+        finally:
+            output_file.close()
+
+class DeidentifyAuthUserProfileTask(DeidentifySqlDumpTask):
+
+    def requires(self):
+        return PathSetTask([self.data_directory], ['*-auth_userprofile-*'])
```
brianhw: seems like the pattern could be pulled out, and the requires() defined in the base class.

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,296 @@
...
+        row[3] = '' # automatic_log
+        row[4] = '' # ip_address
+        row[5] = self.remap_id(row[5]) # user_id
+
+class CourseStructureTask(FileCopyTask, BaseDeidentifyDumpTask):
+
+    def requires(self):
+        return PathSetTask([self.data_directory], ['*-course_structure-*'])
+
+class CourseContentTask(FileCopyTask, BaseDeidentifyDumpTask, luigi.Task):
```
brianhw: don't need luigi.Task?

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,296 @@
...
+class FileCopyTask(object):
+
+    def run(self):
+        if(len(self.input()) == 0):
+            #throw exception
+            pass
+
+        def report_progress(num_bytes):
+            """Update hadoop counters as the file is written"""
+            self.incr_counter('Deidentify State', 'Bytes Written to Output', num_bytes)
```
brianhw: If we do make this more generic, then the counter name would need to be generalized too.

edx/analytics/tasks/util/id_codec.py
```diff
@@ -12,3 +14,54 @@ def decode_id(encoded_id):
...
+        return self.binvec_to_int(permuted)
+
+    def unpermute(self, x):
+        vec = self.int_to_binvec(x)
+        permuted = vec.dot(self.permutation_matrix.T)
+        return self.binvec_to_int(permuted)
+
+class UserIdRemapperMixin(object):
+
+    permutation_generator = PermutationGenerator(42,32,32)
```
brianhw: If this is to provide any measure of security, we need to keep something secret, and I assume that would be the seed.  Anyone could read this code and just unpermute all the ids.

edx/analytics/tasks/util/id_codec.py
```diff
@@ -1,6 +1,8 @@
 """Defines a reversible way to label an id with a scope and type to make it unique."""
```
brianhw: I guess we should change the docstring for the file to something more general.

brianhw:
This is looking pretty solid -- good progress!

edx/analytics/tasks/deidentify_dumps.py
HassanJaveed84: I don't see any obvious candidate without changing the filename. I'll create a new one for now.

edx/analytics/tasks/deidentify_dumps.py
HassanJaveed84: If the file does not exist, this would be empty(list), no ?

HassanJaveed84: Actually, I moved it to the base class's output method.

edx/analytics/tasks/deidentify_dumps.py
HassanJaveed84: I thought of doing something like 

```
class FileCopyTask(object):


    def run(self):
        def report_progress(num_bytes):
            """Update hadoop counters as the file is written"""
            self.incr_counter('FilyCopyTask', 'Bytes Written to Output', num_bytes)

        output_file = self.output().open('w')
        try:
            copy_file_to_open_file(self.input().path, output_file, progress=report_progress)
        finally:
            output_file.close()
```

to make it generic. But PathSetTask returns multiple targets.
So if I were to make use of it like the following, it won't work. Thoughts ?

```
class SomeTask(FileCopyTask):
    def requires(self):
        return PathSetTask([self.data_directory], [self.file_pattern])
```

edx/analytics/tasks/util/id_codec.py
HassanJaveed84: Would the seed value be read from secure config ?

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,304 @@
...
+
+
+class DeidentifyDumpsTask(luigi.WrapperTask):
+
+    course = luigi.Parameter(is_list=True)
+    dump_root = luigi.Parameter()
+    output_root = luigi.Parameter(
+        config_path={'section': 'rdx', 'name': 'output_root'}
+    )
+    output_root = luigi.Parameter(default='hdfs://localhost:9000/edx-analytics-pipeline/output/')
```
brianhw: You don't need both "output_root" definitions. : )

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,304 @@
...
+            DeidentifyMongoDumpsTask(**kwargs),
+        )
+
+
+class DeidentifyDumpsTask(luigi.WrapperTask):
+
+    course = luigi.Parameter(is_list=True)
+    dump_root = luigi.Parameter()
+    output_root = luigi.Parameter(
+        config_path={'section': 'rdx', 'name': 'output_root'}
```
brianhw: I would change the section name to "deidentify-dumps" or similar.  I think 'rdx' will be referring to a much larger process, of which this is only a part.

edx/analytics/tasks/util/id_codec.py
brianhw: Yup.

brianhw:
Just configuring the seed, and a few small tweaks!

HassanJaveed84:
@brianhw can you make another pass at it? Ignoring the docstrings and comments, I still need to add/modify those. Also, how or what part do you see as reusable/shareable between encrypt.py and file_util.p ?

HassanJaveed84:
@brianhw I added unit tests. Wanted like you to take a look at my approach. I'm still thinking about a neater way to define input/output data. Previously I had made separate classes for each kind of test, I've pulled everything into one main class with individual tests. Also, I would need to discuss an issue with reading the seed value from the test.cfg during testing, hopefully over a 1-1 later Today.

edx/analytics/tasks/tests/test_data_deidentification.py
```diff
@@ -0,0 +1,95 @@
+from edx.analytics.tasks.tests import unittest
+from edx.analytics.tasks.tests.target import FakeTarget
+from edx.analytics.tasks.deidentify_dumps import *
```
HassanJaveed84: Should I be specifying only classes I need in the tests or is \* reasonable ?

config/test.cfg
```diff
@@ -37,6 +37,9 @@ required_path_text = FakeServerGroup
 [event-export-course]
 output_root = s3://fake/
```
brianhw, line 39: Does this also need an entry for [deidentify-dumps]output_root?

edx/analytics/tasks/util/file_util.py
```diff
@@ -0,0 +1,48 @@
...
+TRANSFER_BUFFER_SIZE = 1024 * 1024  # 1 MB
+log = logging.getLogger(__name__)
+
+
+class FileCopyTask(object):
+
+    def run(self):
+        def report_progress(num_bytes):
+            """Update hadoop counters as the file is written"""
+            self.incr_counter('FilyCopyTask', 'Bytes Written to Output', num_bytes)
```
brianhw: nit:  s/Fily/File/

edx/analytics/tasks/util/file_util.py
```diff
@@ -0,0 +1,48 @@
...
+
+        output_file = self.output().open('w')
+        if isinstance(self.input(), list):
+            if len(self.input()) == 1:
+                filepath = self.input()[0].path
+            else:
+                # What should we do here, throw exception ?
```
brianhw: I would say 'yes'.  Until there's some reason to do something else.  As far as I know, the only use for this is with single files.

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,292 @@
...
+    file_pattern = '*-wiki_articlerevision-*'
+
+    def filter_row(self, row):
+        row[2] = ''  # user_message
+        row[3] = ''  # automatic_log
+        row[4] = ''  # ip_address
+        row[5] = self.remap_id(row[5])  # user_id
+
+
+class CourseStructureTask(BaseDeidentifyDumpTask, FileCopyTask):
```
brianhw: Is there a reason to order the dependencies with the Base first and the File second?  I would reverse them.  Then again, I would label the FileCopyTask as FileCopyMixin, because it's the BaseDeidentifyDumpTask that inherits from luigi.Task, and the FileCopy doesn't.

edx/analytics/tasks/util/file_util.py
```diff
@@ -0,0 +1,48 @@
+"""
+Utility methods interact with files.
+"""
+import logging
+
+TRANSFER_BUFFER_SIZE = 1024 * 1024  # 1 MB
+log = logging.getLogger(__name__)
+
+
+class FileCopyTask(object):
```
brianhw: As I said earlier, I would make this FileCopyMixin, as it is not a Task but rather needs to be "mixed" with a task to get the .input() and .output() that it depends on.

config/devstack.cfg
```diff
@@ -33,6 +33,12 @@ required_path_text = FakeServerGroup
 [event-export-course]
 output_root = hdfs://localhost:9000/edx-analytics-pipeline/event-export-by-course/output/
 
+[deidentify-dumps]
+output_root = hdfs://localhost:9000/edx-analytics-pipeline/deidentify-dumps/output/
+
+[encrypt]
```
brianhw: The file that is using this is not encrypt.py, so this label is potentially confusing.  Maybe just calling it 'id-codec' would be fine.  A better name?

edx/analytics/tasks/tests/test_data_deidentification.py
brianhw: Another alternative is to "import edx.analytics.tasks.deidentify_dumps as deid", and then you can prefix each class with "deid.".

edx/analytics/tasks/tests/test_data_deidentification.py
```diff
@@ -0,0 +1,95 @@
...
+        task.run()
+        return output_target.buffer.read()
+
+    def reformat(self, data):
+        """Reformat data to make it like a TSV."""
+        return "\n".join(["\t".join(row) for row in data])
+
+    def check_output(self, cls, input, expected):
+        output = self.run_task(task_cls=cls, source=self.reformat(input))
+        self.assertEquals(output, self.reformat(expected) + '\n')
```
brianhw: nit:  You could add the \n here, but it might be a little cleaner to add it in reformat().

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,292 @@
...
+
+                for row in reader:
+                    # Found an empty line at the bottom of test file
+                    if row:
+                        filtered_row = self.filter_row(row)
+                        writer.writerow(filtered_row)
+
+    def filter_row(self, row):
+        raise NotImplementedError
+
```
brianhw: The new Record class just got merged to master as part of the ModuleEngagement PR.  It might make sense to rework this to make use of those classes.  It's solving a different problem, by providing type information that allows database schema to be generated.  But it does function as a namedtuple, which is sort of what a row is, here.  One could also just use a namedtuple instead of a record, as is done in edx/analytics/tasks/reports/reconcile.py.

Either way, the idea would be to have a property used by the base class that defines the record/namedtuple class, and have this defined in each derived class.  

---

@property 
def record_class(self):

##     return AuthUserProfileRecord

Hmm.  Actually the only reason it would be needed would be to check the headers -- to load the auth_userprofile headers into AuthUserProfileRecord.  That's probably not that important to do.  So instead it could be local to the filter_row method.

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,292 @@
...
+
+class DeidentifyAuthUserProfileTask(DeidentifySqlDumpTask):
+
+    file_pattern = '*-auth_userprofile-*'
+
+    def filter_row(self, row):
+        row[1] = self.remap_id(row[1])  # user_id
```
brianhw: So if one had a record/namedtuple, then record = record_class()(row), and modifications would be record.user_id = self.remap_id(record.user_id), record.name = '', and so forth.

edx/analytics/tasks/deidentify_dumps.py
```diff
@@ -0,0 +1,292 @@
...
+        row[2] = ''  # name
+        row[3] = ''  # language
+        row[4] = ''  # location
+        row[5] = ''  # meta
+        row[6] = ''  # courseware
+        row[8] = ''  # mailing_address
+        row[12] = '1'  # allow_certificate
+        row[14] = ''  # city
+        row[15] = ''  # bio
+
```
brianhw: I'm not sure what writerow() requires, but if it requires a list, one can certainly just pass the namedtuple to list().  I don't remember what would be needed to do to a Record to convert it.

edx/analytics/tasks/tests/test_data_deidentification.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def test_auth_user_deidentification(self):
+        data = [
+            ['id', 'username', 'first_name', 'last_name', 'email', 'password', 'is_staff', 'is_active', 'is_superuser', 'last_login', 'date_joined', 'status', 'email_key', 'avatar_type', 'country', 'show_country', 'date_of_birth', 'interesting_tags', 'ignored_tags', 'email_tag_filter_strategy', 'display_tag_filter_strategy', 'consecutive_days_visit_count'],
```
brianhw: I might pull out the header, defining it before data, and including it in data and in expected.

edx/analytics/tasks/util/id_codec.py
```diff
@@ -12,3 +14,58 @@ def decode_id(encoded_id):
...
+        vec = self.int_to_binvec(x)
+        permuted = vec.dot(self.permutation_matrix.T)
+        return self.binvec_to_int(permuted)
+
+class UserIdRemapperMixin(object):
+
+    # seed_value = luigi.Parameter(
+    #     config_path={'section': 'encrypt', 'name': 'seed_value'}
+    # )
+    seed_value = 42 # TODO: need to remove this from here, added to make unit tests work until we figure out a fix.
```
brianhw: Okay, I assume we'll talk about this.... : )

edx/analytics/tasks/util/id_codec.py
```diff
@@ -1,7 +1,9 @@
-"""Defines a reversible way to label an id with a scope and type to make it unique."""
+"""Various helper utilities to help do a 1-1 mapping of sensitive ids"""
```
brianhw: ... utilities to calculate reversible one-to-one mappings of ...

edx/analytics/tasks/tests/test_data_deidentification.py
brianhw: But the general approach seems sound.

brianhw:
I think the unit tests look fine.  The data has to be defined somehow, and that doesn't seem too arduous a way to do it. 

I was originally commenting on using namedtuples or Record classes with the idea that they would help with unit tests, but I don't think that they would add much.  Text in, text out.

config/test.cfg
HassanJaveed84, line 39: I think it would be needed for acceptance test.

edx/analytics/tasks/tests/test_data_deidentification.py
HassanJaveed84: Done

config/devstack.cfg
HassanJaveed84: Couldn't think of any, changed to id-codec.

edx/analytics/tasks/util/file_util.py
HassanJaveed84: Done

edx/analytics/tasks/deidentify_dumps.py
HassanJaveed84: Done.

edx/analytics/tasks/util/id_codec.py
```diff
@@ -12,3 +14,58 @@ def decode_id(encoded_id):
...
+        permuted = vec.dot(self.permutation_matrix.T)
+        return self.binvec_to_int(permuted)
+
+class UserIdRemapperMixin(object):
+
+    seed_value = luigi.Parameter(
+        config_path={'section': 'id-codec', 'name': 'seed_value'}
+    )
+    seed_value = 42 # TODO: need to remove this from here, added to make unit tests work until we figure out a fix
+    permutation_generator = PermutationGenerator(seed_value,32,32)
```
brianhw: I found that removing line 66 and using int(seed_value.value) worked.  Give it a try!

brianhw: (It took a little while to realize that the seed_value.value was a string.  When writing it out, it looked the same....)

HassanJaveed84: Brian, it works in the tests but still throws the missing parameter exception while running the task(actually I had not ran the task again after adding the code to read the seed from config thinking it would just work). 

Does something like this makes sense ?

```
class UserIdRemapperMixin(object):

    seed_value = luigi.Parameter(
        config_path={'section': 'id-codec', 'name': 'seed_value'}
    )
    __generator_instance = None

        super(UserIdRemapperMixin, self).__init__(*args, **kwargs)
        if UserIdRemapperMixin.__generator_instance is None:
            UserIdRemapperMixin.__generator_instance = PermutationGenerator(int(self.seed_value),32,32)
        self.permutation_generator = UserIdRemapperMixin.__generator_instance

    def remap_id(self, id):
        return self.permutation_generator.permute(int(id))
```

HassanJaveed84:
@brianhw seems to be a conflict, should I rebase and fix ?
Also need your review.

edx/analytics/tasks/tests/test_data_deidentification.py
```diff
@@ -0,0 +1,209 @@
+from edx.analytics.tasks.tests import unittest
```
brianhw, line 5: The name of this file doesn't match the name of the file it's testing.  Does it make sense to rename the other to reflect that it's only handling the data deidentification, and not the event processing?

edx/analytics/tasks/tests/test_data_deidentification.py
```diff
@@ -0,0 +1,209 @@
...
+    def reformat(self, data):
+        """Reformat data to make it like a TSV."""
+        return "\n".join(["\t".join(row) for row in data]) + '\n'
+
+    def check_output(self, cls, input_value, expected_value):
+        output = self.run_task(task_cls=cls, source=self.reformat(input_value))
+        self.assertEquals(output, self.reformat(expected_value))
+
+    def test_auth_user_deidentification(self):
+        header = ['id', 'username', 'first_name', 'last_name', 'email', 'password', 'is_staff', 'is_active', 'is_superuser', 'last_login', 'date_joined', 'status', 'email_key', 'avatar_type', 'country', 'show_country', 'date_of_birth', 'interesting_tags', 'ignored_tags', 'email_tag_filter_strategy', 'display_tag_filter_strategy', 'consecutive_days_visit_count']
```
brianhw: were you going to split these lines so they would fit better on a line?

edx/analytics/tasks/tests/test_data_deidentification.py
```diff
@@ -0,0 +1,209 @@
...
+            ['23456', '1', 'This is a user message', 'automatic_log', '192.168.1.1', '123456', '2013-08-08 22:00:58', '2013-08-08 22:00:58', '123', '0', '0', '123', 'This is the content', 'TITLE']
+        ]
+        expected = [
+            header,
+            ['23456', '1', '', '', '', '273678626', '2013-08-08 22:00:58', '2013-08-08 22:00:58', '123', '0', '0', '123', 'This is the content', 'TITLE']
+        ]
+        self.check_output(deid.DeidentifyWikiArticleRevisionTask, data, expected)
+
+    def test_mongo_deidentification(self):
+        data = '{"author_id":"123456","author_username":"johndoe","body":"This is a boady", "course_id":"course-v1:edX+DemoX+Test_2014","votes":{"down":["123456"],"up":["12345"]}}'
```
brianhw: what is a "boady"? : )

edx/analytics/tasks/tests/test_data_deidentification.py
```diff
@@ -0,0 +1,209 @@
...
+            os.makedirs(os.path.dirname(file))
+            open(file, 'a').close()
+
+    def tearDown(self):
+        "Remove temp dir. after running the test."
+        if os.path.exists(self.temp_rootdir):
+            shutil.rmtree(self.temp_rootdir)
+
+    def test_data_directory(self):
+        """Test fo check whether the data_directory for a course is being setup correctly."""
```
brianhw: nit:  s/fo/to/ ?  Also s/setup/set up/

brianhw:
Certainly a rebase will be needed before merging. I also asked about formatting test data, and the naming of the files, nothing major.  Also, has this been run as a Jenkins job yet?

edx/analytics/tasks/tests/test_data_deidentification.py
HassanJaveed84: I did try, but it didn't look any better, kinda worse actually . Can you advise with a sample ?

HassanJaveed84:
I tested it on jenkins using debug-pipeline-task, I'll create a separate job and test again.

edx/analytics/tasks/tests/test_data_deidentification.py
HassanJaveed84, line 5: Done.

edx/analytics/tasks/tests/test_data_deidentification.py
HassanJaveed84: Fixed.

edx/analytics/tasks/tests/test_data_deidentification.py
HassanJaveed84: Done.

edx/analytics/tasks/tests/test_data_deidentification.py
brianhw: I submitted a commit with some reformatting of the data so that it fit within the 120 character limit.  Idea was to break lines according to the worst case in the test, and then break the other lines to match (so that all are breaking at the same columns).  Does that look okay?

brianhw:
Good progress!  Just needs some pep8/pylint attention.  We'll deal with ORA2 stuff much later, as well as the remaining TODOs.

edx/analytics/tasks/tests/test_data_deidentification.py
HassanJaveed84: Looks great, also fits perfectly within my editor window!  Thanks

HassanJaveed84:
@brianhw there is an issue with the FileCopy tasks, came up while running the Task on jenkins. 

FileCopy Task uses 'copy_file_to_open_file' function to do the copying in which the source file is opened using `with open(filepath, 'r') as src_file:`
That would not work for anything that is not a local FS path. 

I can do something like:

```
if isinstance(filepath, luigi.File):
    src_file = filepath.open('r') 
else:
    src_file = open(filepath, 'r')

with srf_file:
    #rest of the original function
```

Does that seem okay ?

brianhw:
Will it do the right thing?  That is, the files are already open, so will the "with" clause close them when exiting the scope?  Other alternative is to move the "with" clause out of the function, and pass the open input file into the function.  That would work for encrypt, but still leaves the question what FileCopyMixin should do.

So in what circumstances did this work?  It worked in devstack, using hdfs for the path?

HassanJaveed84:
I checked it and it closes the file with the scope. I also thought of changing the implementation so both input and output are open files. 
It originally worked with the devstack on local filesystem paths for input files. I was not using hdfs as Pathutil still has that bug where it's behavior for hdfs is not consistent with s3 and local fs. 

So I'm guessing if we move out the 'with', one could pass on the open files from the FileCopyMixin, that is:

```
with self.input().open('r') as input_file:
    copy_file_to_file(input_file, output_file)
```

If we go the other way, we would need to pass in target:
`copy_file_to_open_file(self.input(), output_file)`

brianhw:
Oops, I missed your response.  I think Gabe fixed the Pathutil bug (by adding recurse=True) in his big PR last week, so that should be in master now.  

It looks simpler passing in the open files, no?

HassanJaveed84:
Yup, I'll go with it. 

I came across the 'recursive' bit,  I think it was part of the EventLogSelectionTask. Pathutil probably still has that issue.

brianhw:
Yep, you're right.  Gabe added it to EventLogSelectionTask, but PathTaskSet does not have the fix.  Do you want to try adding it?

HassanJaveed84:
Sure! Will do.
@brianhw I tried fixing PathSetTask by adding the `recursive=True` to `luigi.hdfs.listdir(src)`.
It solves that particular issue with PathSetTask. 

I tried running DataDeidentificationTask with HDFS as the source for the state files, but it does not work. 
The issue is that in my code i'm calling `target.path`(to set the data directory for each course by finding the latest dated folder) and for HDFS targets it returns path without URL scheme,  so instead of getting the full url hdfs://localhost:9000/edx-analytics-pipeline/course_exports/raw... it returns /edx-analytics-pipeline/course_exports/raw...The next time it calls to PathSetTask with /edx-analytics-pipeline/course_exports/raw it tries to look up in the local filesystem and crashes. 

Should I create a PR for the fix for PathSetTask ?

For the issue in my code, this is what it currently looks like:

```
auth_userprofile_targets = PathSetTask([url_path_join(self.dump_root, filename_safe_course_id, 'state')],
                                               ['*auth_userprofile*']).output()
auth_userprofile_paths = [target.path for target in auth_userprofile_targets]
latest_auth_userprofile = sorted(auth_userprofile_paths)[-1]
self.data_directory = os.path.dirname(latest_auth_userprofile)
```

I could try to pick out the 'date' from `/edx-analytics-pipeline/course_exports/raw/course_id/state/2015-11-25/course_id-auth_userprofile-prod-analytics.sql` and then manually set the data_directory to
`url_path_join(self.dump_root, course_id, 'state', date)`. What do you think ?

HassanJaveed84:
I made the change, does it look reasonable ?

brianhw:
Yes, that looks reasonable.  Hoping it works too....

edx/analytics/tasks/util/file_util.py
```diff
@@ -0,0 +1,48 @@
...
+            if len(self.input()) == 1:
+                input_target = self.input()[0]
+            else:
+                raise ValueError("Number of input files should be exactly 1")
+        else:
+            input_target = self.input()
+
+        with input_target.open('r') as input_file:
+            copy_file_to_file(input_file, output_file, progress=report_progress)
+        output_file.close()
```
HassanJaveed84: @brianhw I had to close the file here. Would this be better?

```
with self.output().open('w') as output_file:
    with input_target.open('r') as input_file:
        copy_file_to_file(input_file, output_file, progress=report_progress)
```

and then I can get rid of line 20

brianhw: Yes, that looks much better!

brianhw:
Sorry I missed the comment on PathTaskSet, and only saw the one on FileCopyMixin.  Anyway, you had suggested you could manually set the data_directory to url_path_join(self.dump_root, course_id, 'state', date), and that makes sense.  In fact, you could pull the date off at the time of the list comprehension, so that it's finding all the dates, and it will be clearer that you're getting the latest date.  

At some point, it would be good to factor this kind of logic out, as we probably have a number of places where we would want to get the "latest" existing file.  (Like in the Pentaho stuff Umer is doing now -- for now, we assert that we expect a date, but it would be better to be more robust at some point in the future.)

But we don't have to factor it out right now.....

brianhw:
Refactor looks good.

edx/analytics/tasks/tests/test_data_deidentification.py
```diff
@@ -0,0 +1,265 @@
...
+            ['23456', '1', '', '', '', '273678626', '2013-08-08 22:00:58',
+             '2013-08-08 22:00:58', '123', '0', '0', '123', 'This is the content', 'TITLE']
+        ]
+        self.check_output(deid.DeidentifyWikiArticleRevisionTask, data, expected)
+
+    def test_mongo_deidentification(self):
+        data = '{"author_id":"123456","author_username":"johndoe","body":"This is a body", ' \
+               '"course_id":"course-v1:edX+DemoX+Test_2014","votes":{"down":["123456"],"up":["12345"]}}'
+        expected = '{"author_id":"273678626","author_username":"username_273678626","body":"This is a body", ' \
+               '"course_id":"course-v1:edX+DemoX+Test_2014","votes":{"down":["123456"],"up":["12345"]}}'
```
HassanJaveed84: @brianhw pep8 is complaining here 'E127 continuation line over-indented for visual indent'. Is it ignorable ?

edx/analytics/tasks/util/id_codec.py
```diff
@@ -12,3 +14,66 @@ def decode_id(encoded_id):
...
+    )
+    __generator_instance = None
+
+    def __init__(self, *args, **kwargs):
+        super(UserIdRemapperMixin, self).__init__(*args, **kwargs)
+        if UserIdRemapperMixin.__generator_instance is None:
+            UserIdRemapperMixin.__generator_instance = PermutationGenerator(self.seed_value, 32, 32)
+        self.permutation_generator = UserIdRemapperMixin.__generator_instance
+
+    def remap_id(self, id):
```
HassanJaveed84: @brianhw pylint does not like the name 'id'. Can you suggest a more appropriate one ?

HassanJaveed84:
@brianhw one last review hopefully. Lot's of pylint issues in PermutationGenerator class. Mostly related to the name 'x'.

edx/analytics/tasks/tests/test_data_deidentification.py
brianhw: Not sure how hard to fix.   Just indenting by 4?  That should be easy enough.   Otherwise ignore.

edx/analytics/tasks/util/id_codec.py
brianhw: id_value

brianhw:
I would make that code more readable by replacing 'x' with 'int_value' where appropriate.  Likewise rename 'k' as something like 'matrix_dim'.

HassanJaveed84:
Updated + squashed.

edx/analytics/tasks/data_deidentification.py
```diff
@@ -0,0 +1,311 @@
...
+import json
+import os
+import re
+
+from edx.analytics.tasks.pathutil import PathSetTask
+from edx.analytics.tasks.url import get_target_from_url, url_path_join
+from edx.analytics.tasks.util.id_codec import UserIdRemapperMixin
+import edx.analytics.tasks.util.opaque_key_util as opaque_key_util
+from edx.analytics.tasks.util.file_util import FileCopyMixin
+import edx.analytics.tasks.util.csv_util
```
brianhw, line 15: This is apparently not used.

edx/analytics/tasks/util/id_codec.py
```diff
@@ -12,3 +14,66 @@ def decode_id(encoded_id):
     """Decode an ID string back to the original scope-type-value tuple."""
     scope, id_type, id_value = base64.b32decode(encoded_id).split('|')
     return scope, id_type, id_value
+
+
+class PermutationGenerator(object):
+
+    def __init__(self, seed, k, bits):
```
brianhw: Just curious why you didn't use matrix_dim here for 'k'...

edx/analytics/tasks/util/id_codec.py
```diff
@@ -1,7 +1,9 @@
-"""Defines a reversible way to label an id with a scope and type to make it unique."""
+"""Various helper utilities to calculate reversible one-to-one mappings of sensitive ids"""
 
 import base64
-
+import numpy as np
+import random
+import luigi
```
brianhw, line 7: nit: should be an extra line here (pep8).

edx/analytics/tasks/util/id_codec.py
```diff
@@ -12,3 +14,66 @@ def decode_id(encoded_id):
     """Decode an ID string back to the original scope-type-value tuple."""
     scope, id_type, id_value = base64.b32decode(encoded_id).split('|')
     return scope, id_type, id_value
+
+
+class PermutationGenerator(object):
+
```
brianhw, line 22: docstring?

edx/analytics/tasks/util/id_codec.py
```diff
@@ -12,3 +14,66 @@ def decode_id(encoded_id):
...
+            permutation[i, mapping[i]] = 1
+        return permutation
+
+    def permute(self, int_value):
+        """Given int int_value with bits bits, permute it using the specified bits-by-bits permutation"""
+        vec = self.int_to_binvec(int_value)
+        permuted = vec.dot(self.permutation_matrix)
+        return self.binvec_to_int(permuted)
+
+    def unpermute(self, int_value):
```
brianhw, line 58: docstring?

edx/analytics/tasks/util/id_codec.py
```diff
@@ -12,3 +14,66 @@ def decode_id(encoded_id):
...
+    def __init__(self, seed, k, bits):
+        self.bits = bits
+        self.permutation_matrix = self.random_permutation_matrix(seed, k)
+
+    def int_to_binvec(self, int_value):
+        """Convert int_value, which must be less than 2**bits, to an np vector of bits 0/1 bits"""
+        if int_value < 0 or int_value >= 2 ** self.bits:
+            raise ValueError("{} out of range [0, 2**{}]".format(int_value, self.bits))
+
+        # yay strings
```
brianhw: This is cute.  But I think it can go....

edx/analytics/tasks/util/id_codec.py
```diff
@@ -12,3 +14,66 @@ def decode_id(encoded_id):
...
+    def permute(self, int_value):
+        """Given int int_value with bits bits, permute it using the specified bits-by-bits permutation"""
+        vec = self.int_to_binvec(int_value)
+        permuted = vec.dot(self.permutation_matrix)
+        return self.binvec_to_int(permuted)
+
+    def unpermute(self, int_value):
+        vec = self.int_to_binvec(int_value)
+        permuted = vec.dot(self.permutation_matrix.T)
+        return self.binvec_to_int(permuted)
```
brianhw, line 62: Also, is it possible to add a simple test to the unit tests for this file, that would exercise the unpermute as well as the permute.  Probably enough to show that applying one to the output of the other returns the original value.

edx/analytics/tasks/data_deidentification.py
```diff
@@ -0,0 +1,311 @@
...
+
+    def filter_row(self, row):
+        """Function to deitentify a particular row."""
+        raise NotImplementedError
+
+
+class DeidentifyAuthUserProfileTask(DeidentifySqlDumpTask):
+    """Task to deidentify auth_userprofile dump file."""
+
+    file_pattern = '*-auth_userprofile-*'
```
brianhw: I'm reconsidering this.  Either it should be a constant, and therefore capitalized (FILE_PATTERN), or preferably, made into a @property with a NotImplemented in the base class so that it's clearer that it needs to be defined.  (You've effectively done that for filter_row -- same idea.)

brianhw:
Looks good.  I expect this is the last round of comments.   No, really!  This time for sure!

edx/analytics/tasks/util/id_codec.py
HassanJaveed84: Done.

edx/analytics/tasks/util/id_codec.py
HassanJaveed84, line 58: Done.

edx/analytics/tasks/util/id_codec.py
HassanJaveed84, line 22: Done.

edx/analytics/tasks/util/id_codec.py
HassanJaveed84, line 7: Done.

edx/analytics/tasks/util/id_codec.py
HassanJaveed84: Oops, done now.

edx/analytics/tasks/data_deidentification.py
HassanJaveed84, line 15: I'm using one of the dialects defined by the csv_util, thought I need to import it for it to work. 
Removed it now.

HassanJaveed84, line 15: Actually, unittests fail without it. Thoughts ?

edx/analytics/tasks/util/id_codec.py
HassanJaveed84, line 62: Done.

edx/analytics/tasks/data_deidentification.py
HassanJaveed84: Done.

HassanJaveed84:
The only thing left is the unused 'import edx.analytics.tasks.util.csv_util'. 
I'm using one of the dialects(mysqlpipe) as defined by csv_util.

PR #183, opened 2015-12-02
Fix class name of manifest input format

mulby:
@brianhw this has been breaking some single-node deploys and devstacks


brianhw:
LGTM. :+1:

PR #184, opened 2015-12-09
Compute metric ranges and user segments

mulby:
@brianhw this is ready for review

Open to suggestions about how to best document this workflow, as it is starting to get complicated.

### DIFF
--- a/edx/analytics/tasks/module_engagement.py
+++ b/edx/analytics/tasks/module_engagement.py
@@ -149,17 +149,12 @@ class ModuleEngagementDataTask(EventLogSelectionMixin, OverwriteOutputMixin, Map
                 entity_id = event_data.get('problem_id')
         elif event_type == 'play_video':
             entity_type = 'video'
-            user_actions.append('played')
+            user_actions.append('viewed')
             entity_id = event_data.get('id')
         elif event_type.startswith('edx.forum.'):
-            entity_type = 'forum'
+            entity_type = 'discussion'
             if event_type.endswith('.created'):
-                if event_type == 'edx.forum.comment.created':
-                    user_actions.append('commented')
-                elif event_type == 'edx.forum.response.created':
-                    user_actions.append('responded')
-                elif event_type == 'edx.forum.thread.created':
-                    user_actions.append('created')
+                user_actions.append('contributed')
 
             entity_id = event_data.get('commentable_id')
 
@@ -482,9 +477,9 @@ class ModuleEngagementSummaryDataTask(
                 elif record.event == 'completed':
                     output_record.problems_completed.add(record.entity_id)
             elif record.entity_type == 'video':
-                if record.event == 'played':
+                if record.event == 'viewed':
                     output_record.videos_viewed.add(record.entity_id)
-            elif record.entity_type == 'forum':
+            elif record.entity_type == 'discussion':
                 output_record.discussions_contributed += count
             else:
                 log.warn('Unrecognized entity type: %s', record.entity_type)
--- a/edx/analytics/tasks/module_engagement.py
+++ b/edx/analytics/tasks/module_engagement.py
@@ -649,6 +649,7 @@ class ModuleEngagementSummaryMetricRangesDataTask(
                     metric_values[metric].append(value)
 
         for metric, values in metric_values.items():
+            values = metric_values[metric]
             range_values = numpy.percentile(  # pylint: disable=no-member
                 values, [0.0, self.low_percentile, self.high_percentile, 100.0]
             )
@@ -867,7 +868,7 @@ class ModuleEngagementUserSegmentDataTask(
                     else:
                         segments[SEGMENT_HIGHLY_ENGAGED].add(metric)
 
-        for segment in segments:
+        for segment in sorted(segments):
             yield ModuleEngagementUserSegmentRecord(
                 course_id=course_id,
                 username=username,
--- a/edx/analytics/tasks/module_engagement.py
+++ b/edx/analytics/tasks/module_engagement.py
@@ -1,8 +1,8 @@
 """Measure student engagement with individual modules in the course"""
-
+from collections import defaultdict
+import datetime
 import logging
-
-log = logging.getLogger(__name__)
+from operator import attrgetter
 
 import luigi
 import luigi.task
@@ -15,12 +15,20 @@ from edx.analytics.tasks.url import get_target_from_url
 from edx.analytics.tasks.util import eventlog
 from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
 from edx.analytics.tasks.vertica_load import VerticaCopyTask
-from edx.analytics.tasks.mysql_load import IncrementalMysqlInsertTask
+from edx.analytics.tasks.mysql_load import IncrementalMysqlInsertTask, MysqlInsertTask
 
 from edx.analytics.tasks.util.hive import (
     WarehouseMixin, BareHiveTableTask, HivePartitionTask
 )
-from edx.analytics.tasks.util.record import Record, StringField, IntegerField, DateField
+from edx.analytics.tasks.util.record import Record, StringField, IntegerField, DateField, FloatField
+
+log = logging.getLogger(__name__)
+
+try:
+    import numpy
+except ImportError:
+    log.warn('Unable to import numpy')
+    numpy = None  # pylint: disable=invalid-name
 
 
 class ModuleEngagementRecord(Record):
@@ -200,13 +208,8 @@ class ModuleEngagementPartitionTask(ModuleEngagementDownstreamMixin, HivePartiti
             overwrite=self.overwrite,
         )
 
-    def requires(self):
-        yield self.data_task
-        yield self.hive_table_task
-
     @property
     def data_task(self):
-        """The task that generates the raw data inside this hive partition."""
         return ModuleEngagementDataTask(
             date=self.date,
             n_reduce_tasks=self.n_reduce_tasks,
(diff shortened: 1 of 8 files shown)


edx/analytics/tasks/module_engagement.py
```diff
@@ -358,3 +356,572 @@ def requires(self):
...
+
+        If a student has performed many attempts but not completed *any* problems successfully, a value of float('inf')
+        is returned.
+        """
+        if num_problems_completed > 0:
+            attempts_per_completion = float(num_problem_attempts) / num_problems_completed
+        elif num_problem_attempts > 0:
+            attempts_per_completion = float('inf')
+        else:  # both are 0
+            attempts_per_completion = float(0)
```
brianhw, line 519: Is '0' what you want to return here?  As opposed to 'nan' or similar?

edx/analytics/tasks/module_engagement.py
```diff
@@ -358,3 +356,572 @@ def requires(self):
...
+
+    def mapper(self, line):
+        record = ModuleEngagementSummaryRecord.from_tsv(line)
+        yield record.course_id, line.rstrip('\n')
+
+    def reducer(self, course_id, lines):
+        """
+        Analyze all summary records for particular course.
+
+        This will include all students performed any activity of interest in the past week.
```
brianhw, line 634: nit:  students _who_ performed....

edx/analytics/tasks/module_engagement.py
```diff
@@ -358,3 +356,572 @@ def requires(self):
...
+    This task currently computes "low" and "high" ranges. The "low" range is defined as the [min(), 15th percentile)
+    interval. The "high" range is defined as the [85th percentile, max()). There is an implied "middle" range that is
+    not persisted since it is fully defined by the low and the high. Note that all intervals are left-closed.
+    """
+
+    EXPAND_RANGE_BY = 0.1
+
+    output_root = luigi.Parameter()
+    low_percentile = luigi.FloatParameter(default=15.0)
+    high_percentile = luigi.FloatParameter(default=85.0)
```
brianhw, line 615: Hard-coded defaults, or pulled from config?

edx/analytics/tasks/module_engagement.py
```diff
@@ -358,3 +356,572 @@ def requires(self):
...
+                first_record = record
+            for metric, value in record.get_metrics():
+                if value != 0:
+                    # NOTE: a lot of people don't participate, so don't include them in the analysis. Otherwise it would
+                    # look like doing *anything* in the course meant you were doing really well.
+                    metric_values[metric].append(value)
+
+        for metric, values in metric_values.items():
+            values = metric_values[metric]
+            range_values = numpy.percentile(  # pylint: disable=no-member
```
brianhw, line 653: Add a comment as to what numpy.percentile returns.  Counts in each of the percentile ranges?  No.  Value closest to that percentile?

edx/analytics/tasks/module_engagement.py
```diff
@@ -358,3 +356,572 @@ def requires(self):
...
+):
+    """
+    Summarize the metrics of interest and persist the interesting ranges.
+
+    This task currently computes "low" and "high" ranges. The "low" range is defined as the [min(), 15th percentile)
+    interval. The "high" range is defined as the [85th percentile, max()). There is an implied "middle" range that is
+    not persisted since it is fully defined by the low and the high. Note that all intervals are left-closed.
+    """
+
+    EXPAND_RANGE_BY = 0.1
```
brianhw, line 611: Don't understand this.   Does this just need to be a non-integer?

edx/analytics/tasks/util/tests/test_record.py
```diff
@@ -481,3 +511,42 @@ def test_hive_type(self):
...
+
+
+@ddt
+class FloatFieldTest(unittest.TestCase):
+    """Tests for FloatField"""
+
+    @data(
+        1,
+        10.0,
+        float('inf'),
```
brianhw, line 523: also 'nan'?

edx/analytics/tasks/tests/test_module_engagement.py
```diff
@@ -193,3 +198,682 @@ def test_replacement_of_count(self):
...
+        self.initialize_task([
+            self.range_record.replace(
+                metric='problem_attempts_per_completion',
+                low_value=float('inf')
+            )
+        ])
+        output = self._get_reducer_output(
+            [
+                self.input_record.replace(
+                    problem_attempts_per_completion=100000.0
```
brianhw, line 875: How does 'inf' end up at 100000?

edx/analytics/tasks/tests/test_module_engagement.py
```diff
@@ -193,3 +198,682 @@ def test_replacement_of_count(self):
...
+
+@ddt
+class ModuleEngagementSummaryDataTaskReducerTest(ReducerTestMixin, unittest.TestCase):
+    """Base class for test analysis of student engagement summaries"""
+
+    task_class = ModuleEngagementSummaryDataTask
+    output_record_type = ModuleEngagementSummaryRecord
+
+    input_record = ModuleEngagementRecord(
+        course_id='foo/bar/baz',
```
brianhw, line 247: self.reduce_key uses self.COURSE_ID.  Should this?

edx/analytics/tasks/module_engagement.py
```diff
@@ -358,3 +356,572 @@ def requires(self):
...
+        self.days_active = set()
+
+
+class ModuleEngagementSummaryDataTask(
+    ModuleEngagementDownstreamMixin, OptionalVerticaMixin, OverwriteOutputMixin, MapReduceJobTask
+):
+    """
+    Store a summary of student engagement with their courses on particular dates aggregated using a sliding window.
+
+    Only emits a record if the user did something in the course on that particular day, this dramatically reduces the
```
brianhw, line 437: nit:   "day, this" => "day.  This"

edx/analytics/tasks/module_engagement.py
```diff
@@ -358,3 +356,572 @@ def requires(self):
...
+    volume of data in this table and keeps it manageable.
+
+    Currently this analyzes a sliding one week window of data to generate the summary.
+    """
+
+    # NOTE: We could have done this in Hive, but this MR job is much more simple than the necessary Hive query.
```
brianhw, line 443: nit:  "Hive query." => "Hive query,"

edx/analytics/tasks/module_engagement.py
```diff
@@ -358,3 +356,572 @@ def requires(self):
...
+
+    def mapper(self, line):
+        record = ModuleEngagementRecord.from_tsv(line)
+        yield ((record.course_id, record.username), line.rstrip('\r\n'))
+
+    def reducer(self, key, lines):
+        """Calculate counts for events corresponding to user and course in a given time period."""
+        course_id, username = key
+
+        output_record = ModuleEngagementSummaryRecordBuilder()
```
brianhw, line 466: nit:  this really isn't a record.  "record_builder" seems most reasonable.  Or "record_stats"?  or "output_builder"?

edx/analytics/tasks/module_engagement.py
```diff
@@ -358,3 +356,572 @@ def requires(self):
...
+        for line in lines:
+            record = ModuleEngagementSummaryRecord.from_tsv(line)
+            if first_record is None:
+                # There is some information we need to copy out of the summary records, so just grab one of them. There
+                # will be at least one, or else the reduce function would have never been called.
+                first_record = record
+            for metric, value in record.get_metrics():
+                if value != 0:
+                    # NOTE: a lot of people don't participate, so don't include them in the analysis. Otherwise it would
+                    # look like doing *anything* in the course meant you were doing really well.
```
brianhw, line 648: So when we calculate the bottom of those who answered problems, we're ignoring anyone who didn't answer.  Are they handled separately?  Seems like they're as disengaged as those who did fewer but still non-zero amounts than their peers.

edx/analytics/tasks/module_engagement.py
brianhw, line 653: Yes, I think so.

edx/analytics/tasks/module_engagement.py
```diff
@@ -358,3 +356,572 @@ def requires(self):
...
+                    metric_values[metric].append(value)
+
+        for metric, values in metric_values.items():
+            values = metric_values[metric]
+            range_values = numpy.percentile(  # pylint: disable=no-member
+                values, [0.0, self.low_percentile, self.high_percentile, 100.0]
+            )
+            min_value = range_values[0] - self.EXPAND_RANGE_BY
+            max_value = range_values[3] + self.EXPAND_RANGE_BY
```
brianhw, line 657: Again, I don't understand the need to expand the range.

brianhw, line 657: And if it's to make sure ranges are inclusive, shouldn't it be done to decrease range_values[2](below), rather than max_value, since the intervals are left-closed.  Otherwise, perhaps the docstring needs to be updated?

edx/analytics/tasks/module_engagement.py
mulby, line 519: I've since changed this to case to actually return float('inf') which is consistent with python's division logic which doesn't distinguish between 0/0 and any other number divided by 0.

This also "fixes" the sort ordering somewhat, as we want to group these cases with the learners who have gotten nothing correct yet.

```
~$ python
Python 2.7.6 (default, Jun 22 2015, 17:58:13) 
[GCC 4.8.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> 0 / 0
  File "<stdin>", line 1, in <module>
ZeroDivisionError: integer division or modulo by zero
>>> 10 / 0
  File "<stdin>", line 1, in <module>
ZeroDivisionError: integer division or modulo by zero
>>>
```

edx/analytics/tasks/module_engagement.py
mulby, line 611: The intent is to avoid floating point equality issues. The interval is supposed to be left-closed, but if the bottom of the interval is say 13.0 and the value we are comparing with is 12.99999999999 due to floating point messiness, we want that value to be "included" in the interval.

Open to ideas about a better way to handle this, we could round to a few digits perhaps?

mulby:
Closing, this is now part of #215

PR #188, opened 2015-12-17
WIP: EventsDidentificationTask

HassanJaveed84:
@mulby @brianhw


config/devstack.cfg
```diff
@@ -18,7 +18,7 @@ engine = hadoop
 marker = hdfs://localhost:9000/edx-analytics-pipeline/marker/
 
 [event-logs]
-pattern = .*tracking.log.*
+pattern = .*
```
HassanJaveed84: This is temporary, plan to revert it.

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,151 @@
...
+    course = luigi.Parameter(default=None)
+    explicit_events_filepath = luigi.Parameter(default=None)
+    dump_root = luigi.Parameter(default=None)
+
+    def requires_local(self):
+        filename_safe_course_id = opaque_key_util.get_filename_safe_course_id(self.course)  # do we need to do this ?
+        results = {
+            'auth_user': PathSetTask([url_path_join(self.dump_root, filename_safe_course_id, 'state')], ['*-auth_user-*'])
+        }
+        if self.explicit_events_filepath is not None:
```
HassanJaveed84: AN-6468 indicates that there must be a parameter incase we want to read from a file on s3 or elsewhere, if the parameter is not set, the default behavior would read from the constants fie.

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,151 @@
...
+            results['explicit_events'] = ExternalURL(url=self.explicit_events_filepath)
+        return results
+
+    def init_local(self):
+        super(DeidentifyCourseEventsTask, self).init_local()
+
+        self.explicit_events = []
+        if self.input_local().get('explicit_events') is not None:
```
HassanJaveed84, line 55: Should I check the explicit_events_filepath parameter here instead ?

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def filter_implicit_event(self, event):
+        event_type = event.get('event_type')
+
+        match = opaque_key_util.COURSE_REGEX.match(event_type)
+        if match:
+            course_id_string = match.group('course_id')
+            event_type = event_type.replace(course_id_string, '(course_id)')
+
+        for included_event_type in constants.IMPLICIT_EVENT_TYPE_PATTERNS:
```
HassanJaveed84: Using the patterns from constants file.

edx/analytics/tasks/events_deidentification.py
mulby, line 55: I actually like the way you have it right now. This condition is more clear about what it needs (the key to exist in the inputs).

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+class DeidentifyCourseEventsTask(EventLogSelectionMixin, MultiOutputMapReduceJobTask):
+
+    course = luigi.Parameter(default=None)
+    explicit_events_filepath = luigi.Parameter(default=None)
+    dump_root = luigi.Parameter(default=None)
+
+    def requires_local(self):
+        filename_safe_course_id = opaque_key_util.get_filename_safe_course_id(self.course)  # do we need to do this ?
```
mulby: I'm pretty sure you do. If I remember correctly, the exporter uses this same method to create its directories. That said, I haven't been following the changes you guys have been making recently so that may have changed.

https://github.com/edx/edx-analytics-exporter/blob/8efd4b5ece7e499e621a02b1f1860c59fe5200c3/exporter/course_export.py#L130

edx/analytics/tasks/util/constants.py
```diff
@@ -0,0 +1,167 @@
...
+    "\/courses\/\(course_id\)\/wiki((?!\/_).)*$",
+    "\/courses\/\(course_id\)\/discussion\/(threads|comments)",
+    "\/courses\/\(course_id\)\/discussion\/(upload|users|forum\/?)$",
+    "/\courses\/\(course_id\)\/discussion\/[\w\-.]+\/threads\/create$",
+    "\/courses\/\(course_id\)\/discussion\/forum\/[\w\-.]+\/(inline|search|threads)$",
+    "\/courses\/\(course_id\)\/discussion\/forum\/[\w\-.]+\/threads\/\w+$",
+]
+
+EXPLICIT_EVENTS = [
+    ('admin', 'browser', 'edx.instructor.report.downloaded'),
```
mulby: Suggestion: it might be more clear to define a very simple namedtuple that tells us what the parts of this tuple are. I'm not sure what the first element is (for example)

``` python
from collections import namedtuple

ExplicitEventFilter = namedtupe("ExplicitEventFilter", ["category", "event_source", "event_type"])

EXPLICIT_EVENTS = [
    ExplicitEventFilter('admin', 'browser', 'edx.instructor.report.downloaded'),
    # ...
]
```

edx/analytics/tasks/util/constants.py
```diff
@@ -0,0 +1,167 @@
+IMPLICIT_EVENT_TYPE_PATTERNS = [
+    "\/courses\/\(course_id\)\/jump_to_id\/",
```
mulby: consider using the "r" prefix on these strings.

``` python
r"\/courses\/\(course_id\)\/jump_to_id\/"
```

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,151 @@
...
+        event_type = event.get('event_type')
+
+        event_source = event.get('event_source')
+
+        if event_type is None or event_source is None:
+            return None
+
+        if event_source == 'server' and event_type.startswith('/'):
+            return self.filter_implicit_event(event)
+
```
mulby: nit: remove the newline between the "if" and "elif" blocks, at first glance I thought they were two separate "if" blocks. If we group them more closely together it helps the reader see that the second is an "elif".

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,151 @@
...
+        elif (event_source, event_type) not in [explicit_event[1:3] for explicit_event in self.explicit_events]:
```
mulby: Do we use the first element of these tuples anywhere?

mulby:
Took a pass, this is coming along nicely. Great work so far.

edx/analytics/tasks/events_deidentification.py
HassanJaveed84: Nopes, the explicit events data is from a file which had
'category', 'event_source' and 'type. I initially placed only the source and type, but later added category to the constant as the sample file Brian sent(quite some time ago for event type distribution task) had category, event_source, event_type.

edx/analytics/tasks/util/constants.py
HassanJaveed84: Actually I don't think I need to escape forward slashes. So this would become

``` python
r"/courses/\(course_id\)/jump_to_id/"
```

edx/analytics/tasks/util/constants.py
HassanJaveed84: Done. 
Also does this makes sense ?

``` python
if (event_source, event_type) not in [(explicit_event.event_source, explicit_event.event_type) for explicit_event in self.explicit_events]:
```

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,177 @@
...
+        event_type = event.get('event_type')
+
+        event_source = event.get('event_source')
+
+        if event_type is None or event_source is None:
+            return None
+
+        if event_source == 'server' and event_type.startswith('/'):
+            return self.filter_implicit_event(event)
+        elif (event_source, event_type) in [(explicit_event.event_source, explicit_event.event_type) for explicit_event in self.explicit_events]:
```
mulby: Looking at this more closely I'm thinking that we should just discard the "category" information, or perhaps just include it in a comment above a block of event filters that share a category.

Then we can make self.explicit_events a native python set containing `ExplicitEventFilter` named tuples. When we read them from a file, we can just construct ExplicitEventFilter objects from the content in the file.

That will make the lookups in this set _very_ fast and avoid having to recreate this list so often. I'd like to be able to say:

``` python
elif ExplicitEventFilter(event_source, event_type) in self.explicit_events:
    # do stuff
```

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,177 @@
...
+
+        for included_event_type in constants.IMPLICIT_EVENT_TYPE_PATTERNS:
+            match = re.match(included_event_type, event_type)
+            if match:
+                return event
+
+        return None
+
+    def deidentify_event(self, event):
+        event['context']['user_id'] = self.remap_id(event['context']['user_id'])
```
mulby: Not all events will have a "context", and even if they do, they may not contain the "user_id" field, we should program this more defensively:

``` python
if 'context' in event and 'user_id' in event['context']:
    event['context']['user_id'] = self.remap_id(event['context']['user_id'])
```

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,177 @@
...
+        event.pop('host', None)  # delete host
+        event.pop('ip', None)  # delete ip
+        event['username'] = self.remap_username(event['username'])
+
+        if 'user_id' in event['event']:  # 'event' dict is always present ?
```
mulby: Not necessarily... see this function:

https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/util/eventlog.py#L147

It should always exist, but sometimes it might be a "string" instead of a dict and has to be parsed as JSON first.

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,177 @@
...
+        # TODO: add answer.file_upload_key & saved_response.file_upload_key, is it nested like event['event']['answer']['file_upload_key'] etc. ?
+        for key in ['certificate_id', 'certificate_url', 'source_url', 'fileName', 'GET', 'POST']:
+            event['event'].pop(key, None)
+
+        # TODO: clean ?
+
+        return event
+
+    def remap_username(self, username):
+        # what should we do when username is not in the username_map ?
```
mulby: Open to ideas here. My vote would be to log an error and discard the event.

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,177 @@
...
+        return "username_{id}".format(id=self.remap_id(self.username_map[username]))
```
mulby: We may want to share most of this function with the database deid code, something like this:

``` python
def generate_username_from_user_id(self, user_id):
    return "username_{0}".format(self.remap_id(user_id))
```

The code in this file can call this like so:

``` python
self.generate_username_from_user_id(self.username_map[username])
```

And in the database deid, it can just call it directly with the user_id.

edx/analytics/tasks/util/constants.py
```diff
@@ -0,0 +1,171 @@
...
+    r"/courses/\(course_id\)/discussion/(upload|users|forum/?)$",
+    r"/courses/\(course_id\)/discussion/[\w\-.]+/threads/create$",
+    r"/courses/\(course_id\)/discussion/forum/[\w\-.]+/(inline|search|threads)$",
+    r"/courses/\(course_id\)/discussion/forum/[\w\-.]+/threads/\w+$",
+]
+
+ExplicitEventFilter = namedtuple("ExplicitEventFilter", ["category", "event_source", "event_type"])
+
+EXPLICIT_EVENTS = [
+    ExplicitEventFilter('admin', 'browser', 'edx.instructor.report.downloaded'),
```
mulby: As described above, consider changing this to:

``` python
EXPLICIT_EVENTS = set([
    # Administrative Events
    ExplicitEventFilter('browser', 'edx.instructor.report.downloaded'),
    # ...
])
```

mulby:
@HassanJaveed84 great progress. I made another review pass.

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,177 @@
...
+            with self.input_local()['explicit_events'].open('r') as explicit_events_file:
+                for line in explicit_events_file:
+                    if not line.strip():
+                        continue
+                    if line.startswith('#'):
+                        continue
+                    category, event_source, event_type = line.rstrip('\n').split("\t")
+                    self.explicit_events.append((category, event_source, event_type))
+        else:
+            self.explicit_events = constants.EXPLICIT_EVENTS
```
brianhw: I guess I was imagining that the format of the external file and the checked-in defaults would be the same.  That way, someone could take the checked-in version, make a small modification to it, and put it elsewhere in their system.  We can come back to that later, I guess.   Also, I was imagining it to be a separate file, not one with all the possible constants of the pipeline, which is what utils/constants.py implies to me.

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,177 @@
...
+        self.explicit_events = []
+        if self.input_local().get('explicit_events') is not None:
+            with self.input_local()['explicit_events'].open('r') as explicit_events_file:
+                for line in explicit_events_file:
+                    if not line.strip():
+                        continue
+                    if line.startswith('#'):
+                        continue
+                    category, event_source, event_type = line.rstrip('\n').split("\t")
+                    self.explicit_events.append((category, event_source, event_type))
```
brianhw: You can toss the category at this point, since this code doesn't need it.  It's just that the file is (was?) shared with the event_type_dist code.

edx/analytics/tasks/events_deidentification.py
brianhw: And I was hoping that whatever mechanism worked for this, would also be able to be used for event_type_dist, to provide a default file that included category info, that would be used by event_type_dist and ignored by this.

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,177 @@
...
+        event.pop('host', None)  # delete host
+        event.pop('ip', None)  # delete ip
+        event['username'] = self.remap_username(event['username'])
+
+        if 'user_id' in event['event']:  # 'event' dict is always present ?
+            event['event']['user_id'] = self.remap_id(event['event']['user_id'])
+        if 'username' in event['event']:
+            event['event']['username'] = self.remap_username(event['event']['username'])
+
+        # TODO: add answer.file_upload_key & saved_response.file_upload_key, is it nested like event['event']['answer']['file_upload_key'] etc. ?
```
brianhw: yes, that's what I meant by the dots.

edx/analytics/tasks/events_deidentification.py
brianhw: For now.  But we should run it on real data and actually look at the error logs, and look at why it would be happening and how often and all.

edx/analytics/tasks/events_deidentification.py
mulby: Ah, I misunderstood the intent here.

To ship a file with the pipeline, and read it later from the application, I've used data files before, see:

https://github.com/edx/edx-analytics-pipeline/blob/master/setup.cfg#L15

To actually find the file once it's shipped with the app, I used this:

https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/launchers/remote.py#L13

Note that the files are copied to sys.prefix. I'm not sure this works in all cases.

edx/analytics/tasks/util/id_codec.py
```diff
@@ -79,3 +83,7 @@ def __init__(self, *args, **kwargs):
     def remap_id(self, id_value):
         "Returns a reversible mapping of input id."
         return self.permutation_generator.permute(int(id_value))
+
+    # TODO: use this in data_deidentification.py
+    def generate_username_from_user_id(self, user_id):
```
HassanJaveed84: should the name be made more obvious of what this method does ?

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,220 @@
...
+import edx.analytics.tasks.util.csv_util
+import edx.analytics.tasks.util.constants as constants
+from edx.analytics.tasks.util import eventlog
+
+log = logging.getLogger(__name__)
+
+ExplicitEventFilter = namedtuple("ExplicitEventFilter", ["event_source", "event_type"])
+
+
+class DeidentifyCourseEventsTask(UserIdRemapperMixin, EventLogSelectionMixin, MultiOutputMapReduceJobTask):
```
mulby, line 29: Before I forget, shouldn't this be processing the output from EventExportByCourseTask instead of the raw tracking log?

brianhw, line 29: It should.  

I was thinking it should override the definition of "pattern" defined for [event-log] in configuration, so that it's effectively *.log (or whatever), and provide an interval.  But really, we just want it to read all tracking logs in the particular directory, regardless of date.  Is there a simpler way to do this than to use EventLogSelectionMixin?  Just PathSetTask, perhaps.

mulby, line 29: There are really 2 parts to this:
- I want EventExportByCourseTask to be an upstream dependency of this, so that when I run this, that gets run if necessary.
- The hadoop job should only be ingesting the events from the files output from the above task.

We may need to do this by adding EventExportByCourseTask to "requires_hadoop(self)" to indicate the upstream dependency and using either PathSetTask or ExternalURL in "input_hadoop(self)" to actually point to the exact files we want to ingest.

brianhw, line 29: I'm not sure I would want EventExportByCourseTask to be an upstream dependency.  Mainly because I would expect courses to be constrainable to a narrower window than could be know by the dependency system.  I would prefer for it to fail.  Note that it's not an incremental system.  We should be doing a one-time event dump.   Also, EventExportByCourseTask is presumably able to identify the directory that contains the output log files, but it's not going to know what the subset of dates are that actually contain events.    If we run with the interval 2012-2015, we may only get a few months of events, and there may be gaps.

HassanJaveed84, line 29: @gabe Right now it is indeed processing the output from the EventExportByCourseTask. 
I set the source to the output of EventExportByCourseTask here.
https://github.com/edx/edx-analytics-pipeline/blob/hassan/event-deidentification/edx/analytics/tasks/events_deidentification.py#L210

mulby, line 29: Thanks for the explanations, I'm OK with the approach here.

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,220 @@
...
+            # we already have course events dumped separately, so each DeidentifyCourseEventsTask for a course would have a different source. Better way ?
+            source = [url_path_join(self.dump_root, course, 'events')]
+            kwargs = {
+                'course': course,
+                'dump_root': self.dump_root,
+                'output_root': self.output_root,
+                'source': source,
+                'explicit_events_filepath': self.explicit_events_filepath,
+                'n_reduce_tasks': self.n_reduce_tasks,
+                'interval': self.interval,
```
mulby, line 236: you should be able to override the pattern here instead of in the config file

HassanJaveed84, line 236: Done

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,221 @@
...
+
+    course = luigi.Parameter(is_list=True)
+    dump_root = luigi.Parameter() # course-exports/raw/
+    output_root = luigi.Parameter() # deidentified-course-exports/
+    explicit_events_filepath = luigi.Parameter(default=None)
+
+    def requires(self):
+        for course in self.course:
+            # we already have course events dumped separately, so each DeidentifyCourseEventsTask for a course would have a different source. Better way ?
+            source = [url_path_join(self.dump_root, course, 'events')]
```
mulby: this needs to use filename_safe_course_id in the path I think

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,221 @@
...
+            finally:
+                outfile.close()
+
+    def output_path_for_key(self, key):
+        filename_safe_course_id = opaque_key_util.get_filename_safe_course_id(self.course)
+
+        return url_path_join(
+            self.output_root,
+            filename_safe_course_id,
+            'deidentified-events'
```
mulby: There is a missing comma at the end of this line

mulby:
General comment - we also should make the deidentified data folder structure match the raw structure.

Perhaps:

```
{output_root}/{filename_safe_course_id}/state/{iso date}/{data deid file}  (one per export artifact)
{output_root}/{filename_safe_course_id}/events/{event deid file}  (one per day)
```

@brianhw does this match your expectations?

brianhw:
Yes, that folder structure looks right.

edx/analytics/tasks/tests/acceptance/test_deidentification.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def setUp(self):
+        super(DeidentificationAcceptanceTest, self).setUp()
+        self.create_temporary_directories()
+
+    def create_temporary_directories(self):
+        self.temporary_dir = tempfile.mkdtemp()
+        #self.addCleanup(shutil.rmtree, self.temporary_dir)
+
+        self.dump_root = os.path.join(self.temporary_dir, 'dump_root')
```
mulby: I think this may be your problem.

When this runs on jenkins, the machine that runs the test code (jenkins-ci.analytics.edx.org) is _different_ than the master node of the cluster which is running the luigi workflow. However, on devstack they are the same machine.

You need the parameter passed into the task to be the path that these files are uploaded to (which will either be in S3 or HDFS).

edx/analytics/tasks/tests/acceptance/test_deidentification.py
```diff
@@ -0,0 +1,68 @@
...
+    def run_data_deidentification_task(self, course_id):
+        self.task.launch([
+            'DataDeidentificationTask',
+            '--course', self.filename_safe_course_id,
+            '--output-root', self.test_out,
+            '--dump-root', self.dump_root
+        ])
+
+    def validate_deidentification(self):
+        pass
```
mulby: I think it might make the most sense to use PathSetTask here to select the files in self.test_out for comparison purposes.

That way it should be able to list the directory properly regardless of whether it's HDFS or S3 based.

edx/analytics/tasks/tests/acceptance/test_deidentification.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+class DeidentificationAcceptanceTest(AcceptanceTestCase):
+
+    COURSE_ID = 'course-v1:edX+DemoX+Demo_Course'
+    INPUT_FILE = 'event_export_tracking.log'
+
+    def setUp(self):
+        super(DeidentificationAcceptanceTest, self).setUp()
+        self.temporary_dir = tempfile.mkdtemp()
```
mulby, line 31: This should be cleaned up.

``` python
self.addCleanup(shutil.rmtree, self.temporary_dir)
```

edx/analytics/tasks/tests/acceptance/test_deidentification.py
```diff
@@ -0,0 +1,88 @@
...
+        self.dump_root = os.path.join(self.test_src, 'course_exports', 'raw')
+        self.filename_safe_course_id = get_filename_safe_course_id(self.COURSE_ID)
+
+    def setup_state_files(self):
+        state_files_dir = os.path.join(self.data_dir, 'input', 'deidentification', 'state')
+        for filename in os.listdir(state_files_dir):
+            local_filepath = os.path.join(state_files_dir, filename)
+            dst = url_path_join(self.dump_root, self.filename_safe_course_id, 'state', '2015-11-13', filename)
```
mulby: We should make '2015-11-13' a constant defined at the class scope.

mulby:
@HassanJaveed84 - just took a pass, this is looking great so far. I assume you are going to to add validation for the event output.

Once that is done, we should tag Brian on this PR so that he can take a pass and address his comments, then merge.

HassanJaveed84:
@brianhw @mulby 
For the events deid part, should I include more input events to cover everything or should that be done as a unit test ?

mulby:
The short answer is to use unit tests.

We focus the acceptance tests on the data hand-offs between tasks and systems. It's really about testing the happy path with a minimal functional dataset. The unit tests should explore all of the odd corner cases.

edx/analytics/tasks/events_deidentification.py
brianhw, line 29: Maybe some of this should make it into comments or docstrings, so the approach is captured for ourselves and others who are bound to ask the same questions.

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,224 @@
...
+            with open(default_filepath, 'r') as explicit_events_file:
+                self.explicit_events = self.parse_explicit_events_file(explicit_events_file)
+
+        # read the latest auth_user for this course in memory, needed to remap usernames
+        latest_auth_user_target = sorted(self.input_local()['auth_user'], key=lambda target: target.path)[-1]
+
+        self.username_map = {}
+
+        with latest_auth_user_target.open('r') as auth_user_file:
+            reader = csv.reader(auth_user_file, dialect='mysqlpipe')
```
brianhw: Should this be mysqlpipe or mysqlexport?  Or does it just happen that for this file it didn't matter?

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,224 @@
...
+                self.username_map[username] = user_id
+
+    def parse_explicit_events_file(self, explicit_events_file):
+        events = set()
+        for line in explicit_events_file:
+            if not line.strip():
+                continue
+            if line.startswith('#'):
+                continue
+            _, event_source, event_type = line.rstrip('\n').split("\t")  # is it okay to use _ for unused ?
```
brianhw: Yes, _ is okay for unused variables.  Preferred, I believe.

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,224 @@
...
+
+        if event['username']:
+            try:
+                event['username'] = self.generate_username_from_user_id(self.username_map[event['username']])
+            except KeyError:
+                log.error("Unable to find username: %s in the username_map", event['username'])
+                return None
+
+        # should we use get_event_data or check event field for dict/string and transform accordingly ?
+        # lots of events with unparseable event fields(strings/lists).
```
brianhw: I'm not sure my answer on this.  There are a number of kinds of event_data that probably get thrown away by get_event_data(), because it tries to parse any string as JSON, and there are some values that are not JSON.   The POST values that are truncated are not important here, because we are going to toss POSTs anyway.  However, there are browser events that contain key-value pairs, that are neither JSON nor dicts.  It's just a string.  
  The bigger question is whether to return the event to its original form as much as possible.  If we were to do that, then we would re-encode the event_data dict as a JSON string if it were originally a JSON-encoded string.  I'm concerned that somebody out there might have logic that says if it is a browser event, then expect a string and parse it.  We can also say that a little cleaning here is an okay thing to do, so if they get a dict instead of a string for the event payload, they should be happier.

brianhw: I'm not sure my answer on this.  There are a number of kinds of event_data that probably get thrown away by get_event_data(), because it tries to parse any string as JSON, and there are some values that are not JSON.   The POST values that are truncated are not important here, because we are going to toss POSTs anyway.  However, there are browser events that contain key-value pairs, that are neither JSON nor dicts.  It's just a string.  
  The bigger question is whether to return the event to its original form as much as possible.  If we were to do that, then we would re-encode the event_data dict as a JSON string if it were originally a JSON-encoded string.  I'm concerned that somebody out there might have logic that says if it is a browser event, then expect a string and parse it.  We can also say that a little cleaning here is an okay thing to do, so if they get a dict instead of a string for the event payload, they should be happier.

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,224 @@
...
+
+            for key in ['certificate_id', 'certificate_url', 'source_url', 'fileName', 'GET', 'POST']:
+                event_data.pop(key, None)
+
+            for key in ['answer', 'saved_response']:
+                if key in event_data and 'file_upload_key' in event_data[key]:
+                    event_data[key].pop('file_upload_key')
+
+            # TODO: clean ?
+            event['event'] = event_data # do we transform it back to a string if it was a string originally ?
```
brianhw: Okay, you asked the same question I tried to raise above.  And I'm inclined to say 'yes', unfortunately.

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,224 @@
...
+
+log = logging.getLogger(__name__)
+
+ExplicitEventFilter = namedtuple("ExplicitEventFilter", ["event_source", "event_type"])
+
+
+class DeidentifyCourseEventsTask(UserIdRemapperMixin, EventLogSelectionMixin, MultiOutputMapReduceJobTask):
+
+    course = luigi.Parameter(default=None)
+    explicit_events_filepath = luigi.Parameter(
```
brianhw: This name is really starting to bug me.  Lots of things are filepaths, so it seems less informative here.  How about 'explicit_event_whitelist'?

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,224 @@
...
+    course = luigi.Parameter(is_list=True)
+    dump_root = luigi.Parameter() # course-exports/raw/
+    output_root = luigi.Parameter() # deidentified-course-exports/
+    explicit_events_filepath = luigi.Parameter(
+        config_path={'section': 'events-deidentification', 'name': 'explicit_events_filename'}
+    )
+
+    def requires(self):
+        for course in self.course:
+            # we already have course events dumped separately, so each DeidentifyCourseEventsTask for a course would have a different source. Better way ?
```
brianhw: nit: wrap the comment

edx/analytics/tasks/tests/acceptance/fixtures/input/deidentification/state/edX-DemoX-Demo_Course-courseware_studentmodule-acceptance.sql
```diff
@@ -0,0 +1,5 @@
+id	module_type	module_id	student_id	state	grade	state	created	modified	max_grade	done	course_id
```
brianhw: why does 'state' appear in the header twice?

edx/analytics/tasks/tests/acceptance/test_deidentification.py
```diff
@@ -0,0 +1,116 @@
...
+        self.filename_safe_course_id = get_filename_safe_course_id(self.COURSE_ID)
+
+    def setup_state_files(self):
+        state_files_dir = os.path.join(self.data_dir, 'input', 'deidentification', 'state')
+        for filename in os.listdir(state_files_dir):
+            local_filepath = os.path.join(state_files_dir, filename)
+            dst = url_path_join(self.dump_root, self.filename_safe_course_id, 'state', self.EXPORT_DATE, filename)
+            self.upload_file(local_filepath, dst)
+
+    def test_deidentificaiton(self):
```
brianhw: nit:  s/ficaiton/fication/

edx/analytics/tasks/tests/acceptance/test_deidentification.py
```diff
@@ -0,0 +1,116 @@
...
+            expected_output_filepath = os.path.join(self.data_dir, 'output', 'deidentification', 'state', output_filename)
+
+            if output_filename.endswith('mongo'):
+                with open(output_filepath) as mongo_output_file:
+                    output_json = [json.loads(line) for line in mongo_output_file]
+                with open(expected_output_filepath) as expected_output_file:
+                    expected_output_json = [json.loads(line) for line in expected_output_file]
+                self.assertEqual(output_json, expected_output_json)
+            else:
+                shell.run(['diff', output_filepath, expected_output_filepath])
```
brianhw, line 92: We may find that with other files where we parse json, we may need to do a more complicated comparison like that for the forum data above.  That's at least courseware-studentmodule.  (You might want data with more than one element in state.)  On the other hand, if we always turn on sorting when outputting cleaned json, then 'diff' should work.

edx/analytics/tasks/tests/acceptance/test_deidentification.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def validate_events_deidentification(self):
+        for output_filepath in self.download_output_files(url_path_join(self.test_out, self.filename_safe_course_id, 'events')):
+            output_filename = os.path.basename(output_filepath)
+            decompressed_file_name = output_filepath[:-len('.gz')]
+            log.error("output_filepath: %s", output_filepath)
```
brianhw: Is this an error because that's what it takes to output it?  Seems like we should 'fix' that, because reading this made me think that outputting was an error....

edx/analytics/tasks/util/constants.py
```diff
@@ -0,0 +1,18 @@
+IMPLICIT_EVENT_TYPE_PATTERNS = [
```
brianhw: I still do not like the naming of this file.  And with deidentification patterns being put into something like deid_util, I figure this pattern list can go into the same file.

edx/analytics/tasks/util/csv_util.py
```diff
@@ -33,9 +33,17 @@ class MySQLPipeDialect(csv.Dialect):
     strict = True
 
 
+class MySQLExportDialect(MySQLPipeDialect):
+    """
+    CSV Dialect for files created by edx-analytics-exporter.
+    """
+    escapechar = None
+
```
brianhw, line 42: What dialect is used by Courseware_studentmodule, which is dumped by the pipeline?  
Also, don't the other SQL files use backslash as an escape character?

brianhw:
Looks good!  Just a few comments....

edx/analytics/tasks/util/csv_util.py
mulby, line 42: RE: the backslash

We do use "\" as an escape char, however, the CSV parser treats them as a literal escape, so the string "\n" is interpreted to be simply "n". This is not what we want, since we need the entire character sequence "\n" to be included in the output.

RE: courseware_studentmodule - good question, we should double check that.

We definitely should be using the mysqlexport dialect for all tables exported by edx-analytics-exporter.

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,225 @@
...
+        return [numpy]
+
+
+class EventDeidentificationTask(EventLogSelectionDownstreamMixin, MapReduceJobTaskMixin, luigi.WrapperTask):
+
+    course = luigi.Parameter(is_list=True)
+    dump_root = luigi.Parameter()
+    output_root = luigi.Parameter()
+    explicit_event_whitelist = luigi.Parameter(
+        config_path={'section': 'events-deidentification', 'name': 'explicit_event_whitelist'}
```
HassanJaveed84, line 221: Should we have a mixin for paramters?

edx/analytics/tasks/tests/map_reduce_mixins.py
```diff
@@ -67,7 +67,7 @@ def create_event_log_line(self, **kwargs):
     def create_event_dict(self, **kwargs):
         """Create an event log with test values, as a dict."""
         # Define default values for event log entry.
-        template_name = kwargs.get('template_name', self.default_event_template)
+        template_name = kwargs.pop('template_name', self.default_event_template)
```
HassanJaveed84, line 70: @mulby @brianhw does it make sense to use 'pop' for 'template_name' here ?

edx/analytics/tasks/util/deid_util.py
```diff
@@ -0,0 +1,18 @@
...
+    r"/courses/\(course_id\)/jump_to_id/",
+    r"/courses/\(course_id\)/courseware/",
+    r"/courses/\(course_id\)/info/?$",
+    r"/courses/\(course_id\)/progress/?$",
+    r"/courses/\(course_id\)/course_wiki/?$",
+    r"/courses/\(course_id\)/about/?$",
+    r"/courses/\(course_id\)/teams/?$",
+    r"/courses/\(course_id\)/[a-fA-F\d]{32}/?$",
+    r"/courses/\(course_id\)/?$",
+    r"/courses/\(course_id\)/pdfbook/\d+(/chapter/\d+(/\d+)?)?$",  # found an event with /pdfbook/0/, should we allow trailing slash
```
HassanJaveed84: should we allow trailing slash here?

brianhw: Yes, I think that's a reasonable addition.

edx/analytics/tasks/events_deidentification.py
mulby, line 221: I added this in a subsequent PR, fine to leave the way is for now.

edx/analytics/tasks/tests/map_reduce_mixins.py
mulby, line 70: Yup, this is a good fix.

edx/analytics/tasks/util/id_codec.py
mulby: definitely open to suggestions

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,225 @@
...
+from edx.analytics.tasks.util.id_codec import UserIdRemapperMixin
+
+import edx.analytics.tasks.util.opaque_key_util as opaque_key_util
+import edx.analytics.tasks.util.csv_util
+import edx.analytics.tasks.util.deid_util as deid_util
+from edx.analytics.tasks.util import eventlog
+
+log = logging.getLogger(__name__)
+
+EXPLICIT_EVENT_FILTER = namedtuple("EXPLICIT_EVENT_FILTER", ["event_source", "event_type"])
```
mulby: This should be camel cased:

ExplicitEventFilter

You can disable the pep8/pylint warning about this, the "namedtuple" function is actually defining a class, so the thing it returns should be a class.

edx/analytics/tasks/events_deidentification.py
mulby, line 55: We should unit test this fallback logic.

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,225 @@
...
+                for value in values:
+                    outfile.write(value.strip())
+                    outfile.write('\n')
+                    # WARNING: This line ensures that Hadoop knows that our process is not sitting in an infinite loop.
+                    # Do not remove it.
+                    self.incr_counter('Deidentified Event Exports', 'Raw Bytes Written', len(value) + 1)
+            finally:
+                outfile.close()
+
+    def output_path_for_key(self, key):
```
mulby, line 116: Unit test this path construction as well

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,225 @@
...
+            event_type = event_type.replace(course_id_string, '(course_id)')
+
+        for included_event_type in deid_util.IMPLICIT_EVENT_TYPE_PATTERNS:
+            match = re.match(included_event_type, event_type)
+            if match:
+                return event
+
+        return None
+
+    def deidentify_event(self, event):
```
mulby, line 163: I think this method needs more unit test coverage.

We should have examples of events with "host" and "ip" fields that need to be removed, for example.

edx/analytics/tasks/tests/acceptance/test_deidentification.py
mulby, line 92: We don't sort currently... do we? This could fail depending on the machine that is running the test due to dictionary key ordering weirdness, no?

mulby:
This is pretty much ready as far as I'm concerned, my only blocking concern is about the tests failing on different machines due to dictionary ordering semantics.

edx/analytics/tasks/events_deidentification.py
HassanJaveed84, line 163: Would https://github.com/edx/edx-analytics-pipeline/blob/hassan/event-deidentification/edx/analytics/tasks/tests/test_events_deidentification.py#L168 cover it ?
Or should I write a separate test for this function ?

mulby, line 163: Ah yes, sorry, I missed that one.

edx/analytics/tasks/tests/acceptance/test_deidentification.py
HassanJaveed84, line 92: We don't use sort.
The data deid task reads and writes line by line for all sql files, a file copy for course_structure and course content. Only json deided yet is the mongo forum data which I validate differently. 
I think this would fail after we include scrubbing/deid on json colums.

Should I change the deid task so it sorts output json ?

brianhw, line 92: When I was doing deid on the Mongo-dumped forum data, I found that I could not get a round trip to preserve content after I parsed the json and output it again.  The original had no embedded spaces, and entries were in alphabetic order.  This is addressed by modifying the separators and setting sort_keys when calling json.dumps().  But then I found there are character encodings that are also different, as to when \u notation is used for a character as opposed to a utf8 encoding of the character between Mongo and python's json.  So I gave up getting it to match.  

Those are things that affect the comparison of before-after for forums.  For CWSM, it's messier because the order output is not sorted by the LMS when it updates state.  (Grepping for sort_keys=True in LMS finds it used in export and dumping course structure, mainly.)  So any round-trip won't match.

But if one is willing to give up on matching the order of the input (except when cleaning), and willing to take a little more time in the deid-ing, then sorting the output would result in stable comparisons of output against expected.  

The other consideration is that if we choose to use cjson for some of these steps, I don't know if it has the same option.  My deid code right now is an odd mix of json and cjson, unfortunately.

edx/analytics/tasks/util/csv_util.py
brianhw, line 42: It would be good to capture that in a comment in the code, explaining why the escapechar is None.

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,225 @@
...
+        event.pop('ip', None)  # delete ip
+
+        if event['username']:
+            try:
+                event['username'] = self.generate_username_from_user_id(self.username_map[event['username']])
+            except KeyError:
+                log.error("Unable to find username: %s in the username_map", event['username'])
+                return None
+
+        event_data = eventlog.get_event_data(event)
```
brianhw, line 179: apparently, event_data is not always stored as a dict, or as a JSON-encoded dict.  It can also be just a string, a key-value pair.  Right now, get_event_data doesn't handle that, and presumably just returns None.  But if it were changed, then the logic below can no longer assume event_data is a dict as it currently does.

brianhw:
Right now, the only output that is not stably sorted is mongo, because CWSM state isn't being parsed yet.  So we don't have an issue right now with the sorting.  So I"m willing to postpone getting the comparison to work until the CWSM is being dealt with.

edx/analytics/tasks/events_deidentification.py
HassanJaveed84, line 179: Right now, get_event_data returns None if the 'event' is not parseable as json(so in case of a string, it does not modify/deid anything). Should I add a 
isinstance(event_data, dict) check ?

edx/analytics/tasks/events_deidentification.py
HassanJaveed84, line 55: The fallback logic uses sys.prefix which I think would be set differently during unittest execution(i saw it pointing to the venv folder during test execution).  

How do you think I should test this part ?

edx/analytics/tasks/events_deidentification.py
HassanJaveed84, line 116: done.

edx/analytics/tasks/events_deidentification.py
mulby, line 179: we don't have that isinstance check anywhere else in the code...

I think it is safe to assume it is a dict at this point and _not_ perform any de-id on non-dict payloads. Basically just leave this code the way it is.

edx/analytics/tasks/events_deidentification.py
mulby, line 29: Concur, we should add a comment here with this explanation in it.

edx/analytics/tasks/events_deidentification.py
mulby, line 55: Ah, I forgot about this issue. Don't worry about it for now.

edx/analytics/tasks/tests/acceptance/test_deidentification.py
mulby, line 92: I agree. We should defer solving this problem until we are actually scrubbing data out of CWSM.

I like what you did with the mongo data and am hoping we can do something similar for CWSM. That is - our algorithm will not guarantee stable key ordering, and I think that's OK.

edx/analytics/tasks/util/csv_util.py
mulby, line 42: +1

mulby:
I'm good with merging this as is, but would like a couple more comments added for some of the tricky bits (input to the job, mysqlexport csv dialect).

:+1:

HassanJaveed84:
Added in the comments, please see if they make sense.

mulby:
They look good. Once Brian approves this is can be squashed and merged.

HassanJaveed84:
@mulby @brianhw

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,239 @@
...
+            filename_safe_course_id,
+            'events',
+            '{course}-events-{date}.log.gz'.format(
+                course=filename_safe_course_id,
+                date=key,
+            )
+        )
+
+    def filter_event(self, event):
+        """Filter event using different event filtering critera."""
```
brianhw: nit:  criteria

edx/analytics/tasks/events_deidentification.py
```diff
@@ -0,0 +1,239 @@
...
+from edx.analytics.tasks.util.id_codec import UserIdRemapperMixin
+
+import edx.analytics.tasks.util.opaque_key_util as opaque_key_util
+import edx.analytics.tasks.util.csv_util
+import edx.analytics.tasks.util.deid_util as deid_util
+from edx.analytics.tasks.util import eventlog
+
+log = logging.getLogger(__name__)
+
+ExplicitEventFilter = namedtuple("ExplicitEventFilter", ["event_source", "event_type"])
```
brianhw: I would suggest naming this "ExplicitEventType".  This is not a filter object.  (And to split hairs further, it's actually used for whitelisting, rather than for filtering.)

edx/analytics/tasks/tests/acceptance/test_deidentification.py
```diff
@@ -0,0 +1,123 @@
...
+        """Downloads the files on the path and copies them to a temporary directory."""
+        output_filepaths = []
+        for output_target in PathSetTask([path]).output():
+            output_filename = os.path.basename(output_target.path)
+            output_filepath = os.path.join(self.temporary_dir, output_filename)
+            output_filepaths.append(output_filepath)
+
+            # TODO: Fix this hack solution properly.
+            # We need the target to be of type S3Target instead of S3HdfsTarget on Jenkins.
+            output_target = get_target_from_url(output_target.path.replace('s3://', 's3+https://'))
```
brianhw, line 117: Is this something that needs to be addressed now?  How do we know if we're on Jenkins?  Is there something else that this can check to determine when this replacement needs to be performed?

edx/analytics/tasks/util/csv_util.py
```diff
@@ -33,9 +33,17 @@ class MySQLPipeDialect(csv.Dialect):
     strict = True
 
 
+class MySQLExportDialect(MySQLPipeDialect):
+    """
+    CSV Dialect for files created by edx-analytics-exporter.
+    """
+    escapechar = None  # Needed to preserve the character sequence, otherwise csv parser would interpret '\n' as 'n'
```
brianhw: nit: I would prefer to have such lengthy comments on the preceding line, rather than inline.  (And I generally prefer the shorter ones on the preceding line as well...)

edx/analytics/tasks/util/id_codec.py
brianhw: generate_deid_username_from_user_id?

brianhw:
Looks good - nothing substantive at this point.  A few nits, and a question about a hack.

edx/analytics/tasks/tests/acceptance/test_deidentification.py
HassanJaveed84, line 117: I'm not really sure how to address this.
@mulby should have better insight.

edx/analytics/tasks/events_deidentification.py
HassanJaveed84: done

edx/analytics/tasks/events_deidentification.py
HassanJaveed84: done

edx/analytics/tasks/util/csv_util.py
HassanJaveed84: Done.

mulby:
:+1: ready to merge

edx/analytics/tasks/tests/acceptance/test_deidentification.py
mulby, line 117: The more I think about this, it may be a valid transform whether or not the system is running on the Jenkins box or not.

That said, a real fix for this issue could be to refactor the tests such that they are actually executed on the master node of the cluster. I think that would simplify things quite a bit.

PR #189, opened 2015-12-24
Change MySQL requirement location

mulby:
@brianhw @HassanJaveed84 

FYI - this was breaking a bunch of jobs, I already hotfixed a patch into release


brianhw:
Thanks, Gabe!

PR #190, opened 2016-01-04
Change MySQL requirement location

mulby:
@nedbat - as discussed in hipchat

Context for others:

http://stackoverflow.com/questions/34489271/i-cannot-install-mysql-connector-python-using-pip/34489738
https://bitbucket.org/pypa/pypi/issues/365/many-packages-have-empty-list-of-links


nedbat:
:+1:

PR #191, opened 2016-01-08
Write enrollment job output directly to destination

mulby:
@brianhw - here is the second part of the patch


brianhw:
LGTM.  :+1:

PR #192, opened 2016-01-12
Deid Integration

HassanJaveed84:
@mulby please review.

### DIFF
--- a/edx/analytics/tasks/data_deidentification.py
+++ b/edx/analytics/tasks/data_deidentification.py
@@ -1,7 +1,6 @@
 """Deidentify course state files by removing/stubbing user information."""
 
 import luigi
-import luigi.s3
 import csv
 import json
 import os
@@ -12,9 +11,10 @@ from edx.analytics.tasks.url import get_target_from_url, url_path_join
 from edx.analytics.tasks.util.id_codec import UserIdRemapperMixin
 import edx.analytics.tasks.util.opaque_key_util as opaque_key_util
 from edx.analytics.tasks.util.file_util import FileCopyMixin
-import edx.analytics.tasks.util.csv_util
 
 import logging
+
+import edx.analytics.tasks.util.csv_util
 
 log = logging.getLogger(__name__)
 
@@ -34,6 +34,7 @@ class BaseDeidentifyDumpTask(UserIdRemapperMixin, luigi.Task):
                 filename=self.file_pattern, course=self.course
             ))
 
+        # TODO: should we change the filename to indicate that it has been de-identified?
         output_filename = os.path.basename(self.input()[0].path)
         return get_target_from_url(url_path_join(self.output_directory, output_filename))
 
@@ -283,7 +284,7 @@ class DeidentifyMongoDumpsTask(BaseDeidentifyDumpTask):
         return row
 
 
-class DeidentifyCourseDumpTask(luigi.WrapperTask):
+class DeidentifiedCourseDumpTask(luigi.WrapperTask):
     """Wrapper task to deidentify data for a particular course."""
 
     course = luigi.Parameter()
@@ -291,13 +292,14 @@ class DeidentifyCourseDumpTask(luigi.WrapperTask):
     output_root = luigi.Parameter()
 
     def __init__(self, *args, **kwargs):
-        super(DeidentifyCourseDumpTask, self).__init__(*args, **kwargs)
+        super(DeidentifiedCourseDumpTask, self).__init__(*args, **kwargs)
 
         filename_safe_course_id = opaque_key_util.get_filename_safe_course_id(self.course)
         auth_userprofile_targets = PathSetTask([url_path_join(self.dump_root, filename_safe_course_id, 'state')],
                                                ['*auth_userprofile*']).output()
         # TODO: Refactor out this logic of getting latest file. Right now we expect a date, so we use that
         dates = [re.search(r"\d{4}-\d{2}-\d{2}", target.path).group() for target in auth_userprofile_targets]
+        # TODO: Make the date a parameter that defaults to the most recent, but allows the user to override?
         latest_date = sorted(dates)[-1]
         self.data_directory = url_path_join(self.dump_root, filename_safe_course_id, 'state', latest_date)
         self.output_directory = url_path_join(self.output_root, filename_safe_course_id, 'state', latest_date)
@@ -325,22 +327,3 @@ class DeidentifyCourseDumpTask(luigi.WrapperTask):
             CourseStructureTask(**kwargs),
             DeidentifyMongoDumpsTask(**kwargs),
         )
-
-
-class DataDeidentificationTask(luigi.WrapperTask):
-    """Wrapper task for data deidentification."""
-
-    course = luigi.Parameter(is_list=True)
-    dump_root = luigi.Parameter()
-    output_root = luigi.Parameter(
-        config_path={'section': 'data-deidentification', 'name': 'output_root'}
-    )
-
-    def requires(self):
-        for course in self.course:
-            kwargs = {
-                'dump_root': self.dump_root,
-                'course': course,
-                'output_root': self.output_root,
-            }
-            yield DeidentifyCourseDumpTask(**kwargs)
--- a/edx/analytics/tasks/deidentification.py
+++ b/edx/analytics/tasks/deidentification.py
@@ -1,0 +1,182 @@
+"""Tasks to deidentify course data for RDX."""
+
+import json
+import os
+import errno
+import logging
+import tarfile
+import urlparse
+
+import luigi
+
+from edx.analytics.tasks.encrypt import make_encrypted_file
+from edx.analytics.tasks.util.file_util import copy_file_to_file
+
+from edx.analytics.tasks.data_deidentification import DeidentifiedCourseDumpTask
+from edx.analytics.tasks.events_deidentification import DeidentifyCourseEventsTask
+from edx.analytics.tasks.pathutil import PathSetTask
+from edx.analytics.tasks.url import url_path_join, get_target_from_url
+from edx.analytics.tasks.url import ExternalURL
+from edx.analytics.tasks.util import opaque_key_util
+from edx.analytics.tasks.util.tempdir import make_temp_directory
+
+
+log = logging.getLogger(__name__)
+
+
+class DeidentifiedCourseTaskMixin(object):
+    """Parameters used by DeidentifiedCourseTask."""
+
+    deidentified_output_root = luigi.Parameter(
+        config_path={'section': 'deidentification', 'name': 'deidentified_output_root'}
+    )
+    dump_root = luigi.Parameter(
+        config_path={'section': 'deidentification', 'name': 'dump_root'}
+    )
+    format_version = luigi.Parameter()
+    pipeline_version = luigi.Parameter()
+
+
+class DeidentifiedPackageTaskMixin(object):
+    """Parameters used by DeidentifiedPackageTask."""
+
+    deidentified_output_root = luigi.Parameter(
+        config_path={'section': 'deidentification', 'name': 'deidentified_output_root'}
+    )
+    gpg_key_dir = luigi.Parameter(
+        config_path={'section': 'deidentification', 'name': 'gpg_key_dir'}
+    )
+    gpg_master_key = luigi.Parameter(
+        config_path={'section': 'deidentification', 'name': 'gpg_master_key'}
+    )
+    output_root = luigi.Parameter(
+        config_path={'section': 'deidentification', 'name': 'output_root'}
+    )
+    recipient = luigi.Parameter(is_list=True)
+    format_version = luigi.Parameter()
+
+
+class DeidentifiedCourseTask(DeidentifiedCourseTaskMixin, luigi.Task):
+    """Depends on DeidentifiedCourseDumpTask & DeidentifyCourseEventsTask, writes metadata file to the output."""
+
+    course = luigi.Parameter()
+
+    def requires(self):
+        output_root_with_version = url_path_join(self.deidentified_output_root, self.format_version)
+        yield DeidentifiedCourseDumpTask(
+            dump_root=self.dump_root,
+            course=self.course,
+            output_root=output_root_with_version,
+        )
+        yield DeidentifyCourseEventsTask(
+            dump_root=self.dump_root,
+            course=self.course,
+            output_root=output_root_with_version,
+        )
+
+    def run(self):
+        with self.output().open('w') as metadata_file:
+            json.dump({
+                'format_version': self.format_version,
+                'pipeline_version': self.pipeline_version
+            }, metadata_file)
+
+    def output(self):
+        filename_safe_course_id = opaque_key_util.get_filename_safe_course_id(self.course)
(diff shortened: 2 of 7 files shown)


edx/analytics/tasks/deidentification.py
```diff
@@ -0,0 +1,129 @@
...
+                ['*.*']
+        )
+
+        with make_temp_directory(prefix='deid-archive.') as tmp_directory:
+            for target in path_task.output():
+                log.info(target.path)
+                with target.open('r') as input_file:
+                    relative_url_path = urlparse.urlparse(course_files_url).path
+                    r_index = target.path.find(relative_url_path) + len(relative_url_path)
+                    relative_path = target.path[r_index:].lstrip('/')
```
HassanJaveed84, line 126: target.path for HdfsTarget returns path without the url scheme(no hdfs://..). However, it returns the complete url for S3Target(s3://bucket/...). Is there a better way of doing this ?

edx/analytics/tasks/deidentification.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+log = logging.getLogger(__name__)
+
+
+class DeidentificationTaskMixin(object):
+    dump_root = luigi.Parameter(
+        config_path={'section': 'deidentification', 'name': 'dump_root'}
+    )
+    intermediate_output_root = luigi.Parameter(
```
mulby: After more thought, maybe we should change this parameter name to "deidentified_output_root" since it will contain the de-identified data before it is packaged up.

edx/analytics/tasks/deidentification.py
mulby, line 126: I think the way you do it here is good. It was a little difficult to understand at first. I would put in a few comments explaining the procedure and why it's necessary.

edx/analytics/tasks/deidentification.py
```diff
@@ -0,0 +1,129 @@
...
+                    try:
+                        os.makedirs(os.path.dirname(local_file_path))
+                    except OSError as exc:
+                        if exc.errno != errno.EEXIST:
+                            raise
+                    with open(local_file_path, 'w') as temp_file:
+                        copy_file_to_file(input_file, temp_file)
+
+            def report_encrypt_progress(num_bytes):
+                """Update hadoop counters as the file is encrypted"""
```
mulby: nit: I forgot to update this comment to remove the bit about hadoop counters...

mulby:
Generally looking good.

I would like to roll the version file changes into this PR as well.

See this diff: https://github.com/edx/edx-analytics-pipeline/compare/hassan/deid-integration...gabe/deid-version

edx/analytics/tasks/deidentification.py
```diff
@@ -0,0 +1,142 @@
...
+            for target in path_task.output():
+                log.info(target.path)
+                with target.open('r') as input_file:
+                    # Get path without urlscheme.
+                    course_files_path = urlparse.urlparse(course_files_url).path
+                    # Calculates target's relative path to course_files_path by getting the substring that
+                    # occurs after course_files_path substring in target's path.
+                    # Needed as target.path returns path with urlscheme for s3target & without for hdfstarget.
+                    r_index = target.path.find(course_files_path) + len(course_files_path)
+                    relative_path = target.path[r_index:].lstrip('/')
```
HassanJaveed84, line 126: @mulby added comments, not sure if the language is clear enough. What do you think ?

HassanJaveed84:
Added versioning info, updated the acceptance test.

edx/analytics/tasks/deidentification.py
mulby, line 126: Maybe include a couple of examples (with made up data in them)?

edx/analytics/tasks/deidentification.py
```diff
@@ -0,0 +1,148 @@
...
+                    # Get path without urlscheme.
+                    course_files_path = urlparse.urlparse(course_files_url).path
+                    # Calculates target's relative path to course_files_path by getting the substring that
+                    # occurs after course_files_path substring in target's path.
+                    # Needed as target.path returns path with urlscheme for s3target & without for hdfstarget.
+                    # Examples:
+                    # target.path: /edx-analytics-pipeline/output/edX_Demo_Course/events/edX_Demo_Course-events-2015-08-30.log.gz
+                    # relative_path: events/edX_Demo_Course-events-2015-08-30.log.gz
+                    # target.path: s3://some_bucket/output/edX_Demo_Course/state/2015-11-25/edX-Demo-Course-auth_user-prod-analytics.sql
+                    # relative_path: state/2015-11-25/edX-Demo-Course-auth_user-prod-analytics.sql
```
HassanJaveed84, line 124: @mulby is this close to what you had suggested ?

HassanJaveed84:
@brianhw please review.

edx/analytics/tasks/deidentification.py
mulby, line 124: Yes, perfect! thanks.

HassanJaveed84:
@brianhw @mulby please review. I've updated the acceptance test, what parts should have more coverage ?

edx/analytics/tasks/deidentification.py
```diff
@@ -0,0 +1,184 @@
...
+        return get_target_from_url(url_path_join(self.deidentified_output_root, self.format_version, filename_safe_course_id, 'version.json'))
+
+
+class DeidentifiedPackageTask(DeidentifiedPackageTaskMixin, luigi.Task):
+
+    course = luigi.Parameter()
+
+    def requires(self):
+        return {
+            'version_file': ExternalURL(url_path_join(self.course_files_url, 'version.json')),
```
HassanJaveed84: If the path to ExternalURL does not exist, the task does not run. However, the process still exists with 0. How do we tackle this ?

mulby: That's a pervasive issue that we can address some other time. The task won't actually run, which is the important part.

edx/analytics/tasks/deidentification.py
```diff
@@ -0,0 +1,184 @@
...
+        config_path={'section': 'deidentification', 'name': 'gpg_master_key'}
+    )
+    output_root = luigi.Parameter(
+        config_path={'section': 'deidentification', 'name': 'output_root'}
+    )
+    recipient = luigi.Parameter(is_list=True)
+    format_version = luigi.Parameter()
+
+
+class DeidentificationTask(DeidentificationTaskMixin, luigi.Task):
```
brianhw: I'd probably just name this DeidentifyCourseTask.

edx/analytics/tasks/deidentification.py
```diff
@@ -0,0 +1,184 @@
...
+            dump_root=self.dump_root,
+            course=self.course,
+            output_root=output_root_with_version,
+            source=source,
+            interval=self.event_log_interval,
+            pattern='.*.gz',
+        )
+
+    def run(self):
+        with self.output().open('w') as version_file:
```
brianhw: A small nit:  I would like to see us more generally generate metadata files that provide information about the output.  This is an excellent start in that direction.  Towards that end, I suggest calling this metadata_file instead of just version_file, since it will hopefully hold more information....

edx/analytics/tasks/deidentification.py
brianhw: Same suggestion here:   "version" => "metadata".

edx/analytics/tasks/deidentification.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+class DeidentifiedPackageTask(DeidentifiedPackageTaskMixin, luigi.Task):
+
+    course = luigi.Parameter()
+
+    def requires(self):
+        return {
+            'version_file': ExternalURL(url_path_join(self.course_files_url, 'version.json')),
+            'path_task': PathSetTask([self.course_files_url], ['*.*'])
```
brianhw: Is *.* going to include the version/metadata file?

brianhw: Oh, right, I guess we want it to...

edx/analytics/tasks/deidentification.py
```diff
@@ -0,0 +1,184 @@
...
+        if self.gpg_master_key is not None:
+            recipients.add(self.gpg_master_key)
+        key_file_targets = [
+            get_target_from_url(url_path_join(self.gpg_key_dir, recipient))
+            for recipient in recipients
+        ]
+
+        with make_temp_directory(prefix='deid-archive.') as tmp_directory:
+            for target in self.input()['path_task']:
+                log.info(target.path)
```
brianhw: Is this just here for debugging?  Not a lot of information about what is being logged...

edx/analytics/tasks/deidentification.py
brianhw: I don't understand the purpose of 'path_task' being in the requirements.  It's not a requirement, in that it's passively matching whatever is already complete.  I think it should be moved down into the run().

brianhw:
Generally looks good.  Just a few questions.

edx/analytics/tasks/data_deidentification.py
```diff
@@ -298,6 +300,7 @@ def __init__(self, *args, **kwargs):
                                                ['*auth_userprofile*']).output()
         # TODO: Refactor out this logic of getting latest file. Right now we expect a date, so we use that
         dates = [re.search(r"\d{4}-\d{2}-\d{2}", target.path).group() for target in auth_userprofile_targets]
+        # TODO: Make the date a parameter that defaults to the most recent, but allows the user to override?
```
brianhw, line 302: It should also be clear what to do if there are no files present, if this is being refactored.  At the moment, we get an index-out-of-range error below when trying to access dates[-1].  I added a check for len(dates) in my branch, but if this gets refactored, the new code should handle this.

edx/analytics/tasks/data_deidentification.py
```diff
@@ -325,22 +328,3 @@ def requires(self):
             CourseStructureTask(**kwargs),
             DeidentifyMongoDumpsTask(**kwargs),
         )
-
-
-class DataDeidentificationTask(luigi.WrapperTask):
```
brianhw, line 330: Hmm.  I'm finding it useful to have a task to do this over multiple courses, as part of development now and (I expect) validation later.  We could make clear in the docstring (when that gets added) that this is for development, and not for production.  (And rename to DeidentifiedMultipleCourseDumpTask?)  (And rename DeidentifyCourseDumpTask to DeidentifiedCourseDumpTask? ... see below.)

edx/analytics/tasks/deidentification.py
brianhw: Hmm.  In reviewing our task naming, it seems like we're a little inconsistent.  Most cases identify the result:  <OutputArtifact>Task.  Some identify the process:  <DoingSomething>Task.   To fit with the more common usage, I suggest this be named "DeidentifiedCourseTask", since that is the result.

edx/analytics/tasks/data_deidentification.py
HassanJaveed84, line 302: Should I add the length check for now?

edx/analytics/tasks/data_deidentification.py
HassanJaveed84, line 330: Renamed DeidentifyCourseDumpTask-> DeidentifiedCourseDumpTask.
Should I add a MultiCourseDeidentifiedCourseDumpTask for development ?

edx/analytics/tasks/deidentification.py
HassanJaveed84: moved to run()

edx/analytics/tasks/deidentification.py
HassanJaveed84: removed.

edx/analytics/tasks/deidentification.py
HassanJaveed84: done.

edx/analytics/tasks/deidentification.py
HassanJaveed84: done. Should I also change DeidentifyCourseEventsTask to DeidentifiedCourseEventsTask(events_deidentification.py) ?

edx/analytics/tasks/data_deidentification.py
mulby, line 330: I would say leave it off this branch and @brianhw can add it back in on his since he appears to be actively using it.

mulby:
LGTM
:+1:

edx/analytics/tasks/data_deidentification.py
brianhw, line 302: I can add it.  But I'm not planning on refactoring....

edx/analytics/tasks/data_deidentification.py
brianhw, line 330: That's fine.

edx/analytics/tasks/deidentification.py
```diff
@@ -0,0 +1,181 @@
...
+
+class DeidentifiedCourseTaskMixin(object):
+    deidentified_output_root = luigi.Parameter(
+        config_path={'section': 'deidentification', 'name': 'deidentified_output_root'}
+    )
+    dump_root = luigi.Parameter(
+        config_path={'section': 'deidentification', 'name': 'dump_root'}
+    )
+    event_log_interval = luigi.DateIntervalParameter(
+        config_path={'section': 'deidentification', 'name': 'event_log_interval'}
```
brianhw: Is this something that we expect to store in a config file?  I'm expecting that the interval depends on the course being processed.  In a first pass, where we are running old courses, we would specify the interval to be all-time.  But it seems overkill to require that we scan all the event files ever for courses that start in 2015 (let alone those coming on line in 2016).   But this points to an architectural inefficiency of this current approach -- that we end up doing a separate pass over all event files for each and every course -- again and again.  Pretty wasteful.

brianhw: Sigh.  So the viability of processing events for multiple courses at the same time depends on what kind of "state" is loaded.  If we load per-course data, then we have to process the events that way.  If we load global data (e.g. all of auth_userprofile), then we can process more courses at a time. 

I think that determination is out of scope for this PR, and is a refactoring for another PR if we want to do that.  I'll just flag this interval as something I don't expect to be in an config file.  Thoughts?

brianhw:
LGTM.  Just a question about event intervals being read from config.

edx/analytics/tasks/deidentification.py
HassanJaveed84: Events-deid task only uses the output produced by EventExportByCourseTask, I think that is where we should address this issue.

brianhw: Sigh.  Just to recap, the selection I was thinking of takes place when we get the events as part of course export, as you pointed out.  It's not relevant here, where we just want to process all events for a course.

HassanJaveed84:
@mulby @brianhw changed events-id to use PathSetTask instead of EventLogSelectionMixin.

brianhw:
LGTM.  I assume it's not going to pass pep8.  Can you add missing docstrings?

HassanJaveed84:
Passes pep8. 
Lots of 'Line too long' for pylint.
Added docstrings. 

I'll merge it after a final thumbsup.

brianhw:
LGTM.  :+1:

mulby:
:+1:

PR #193, opened 2016-01-20
Fix missing explicit_event_whitelist parameter.

HassanJaveed84:


mulby:
I think I actually preferred it the way you had it before. I don't forsee us wanting to override this parameter from the commandline, so having it only be read from the config file seems very reasonable to me.

HassanJaveed84:
Alright, I also renamed the config section from events-deidentification to deidentification. Should I do that in a separate PR, or update this one to keep only that stuff ?

mulby:
I like that idea, doing it in this PR is fine. Not sure if there are any other settings that are only used by the deidentification workflow, but if there are we should also pull them into this section. I like the idea of having one instead of separate "data-deidentification", "events-deidentification" and "deidentification" sections.

brianhw:
Ah, well.  I tend to disagree with Gabe on whether to provide command-line support for parameters or require that they be updated in a config file.  He doesn't like the clutter of long arglists being passed through, and I don't like having to check in changes to the secure branch for any experiment I want to run.

Because we're talking about a whitelist, I actually expect it to need to change relatively frequently.  However, if the location being pointed to is a location in the pipeline codebase, we should be able to deal with such "experiments" by branching pipeline rather than secure.  So I'm okay with it in this case.  Note that such changes to the whitelist should lead to a change in the RDX version number.

HassanJaveed84:
The RDX version number is the latest tag in the pipeline repo that starts with rdx-, so unless a new tag is created, the version stays same. Actually, now that I think there are two verions, 
format_version & pipeline_version where fomat_version is the one that looks for the tag and pipeline_version is the hash of the HEAD of the branch.

mulby:
@brianhw - I don't want to hijack this PR too much, but I wonder if there isn't a compromise solution that meets both our needs? Something like a generic parameter mechanism that allows us to override arbitrary config file values from the command line?

`launch-task SomeFunTask --local-scheduler --config-var "id-codec.seed_value=42"`

This is what tools like ansible, hadoop, hive and others do. Obviously there are limitations, and some thought needs to be put into the way the paramters are passed in and parsed etc, but the overall intent is that we don't have to fill up the code with plumbing while still maintaining the flexibility to perform ad-hoc overrides during development.

@HassanJaveed84 - That is the desired behavior. I think @brianhw is just pointing out that when we change that file we should create a new tag.

brianhw:
@mulby Good point.  I think something like that behavior is available in a later version of Luigi as well, except that instead of specifying the config section, you specify the task name.  Not quite the same thing, but it may still simplify our param-passing needs (though not eliminate it, I would argue).

Otherwise :+1:  on this.

mulby:
:+1:

mulby:
Probably need to update the acceptance tests too? Don't they pass in a request id?

HassanJaveed84:
The acceptance test runs DeidentifiedCourseTask & DeidentifiedPackageTask.
request_id was part of the MultiCourseDeidentifiedPackageTask and was not used in the acceptance test.

mulby:
Got it, I think this is ready to merge after a squash.

brianhw:
:+1:

PR #194, opened 2016-01-21
WIP: cache user information during deidentification

mulby:
@brianhw here is a POC

Stuff that could use work:
- Loading the tables takes so much memory that we cannot run it in the map phase since the large number of map tasks require too many copies of the tables. Doing the join in the reducer allows us to tightly control the number of copies of the tables in memory on a particular node.
- I only wanted to store a single copy of the tables per python process, thus the module scope storage. I also wanted to fail fast if we ever tried to load multiple different copies.
- Passing in the whole path to the tables isn't great... we may want to figure out a better way.
- The way the requirements are handled is not great either.


brianhw:
Cherry-picked from this, so this is not needed.

PR #196, opened 2016-01-25
Cleaning text for obfuscation

brianhw:
This PR adds cleaning of text fields for deidentification of events and database dumps.  It includes the definition of regexp patterns, and the lookup against Sqoop versions of auth_user and auth_userprofile.  This user info needs to be generated externally.  It also includes a standalone script that permits running the regular expressions against local files.  

This lacks unit test coverage and docstrings, and acceptance tests have not been run.  But a review would be helpful nonetheless.  @mulby @HassanJaveed84


edx/analytics/tasks/data_deidentification.py
```diff
@@ -29,30 +30,39 @@ class BaseDeidentifyDumpTask(UserIdRemapperMixin, luigi.Task):
...
+        return base_reqs
 
     @property
     def file_pattern(self):
         """Provides the file pattern for input file."""
         raise NotImplementedError
 
+    @property
+    def file_input_target(self):
+        return self.input()['data'][0]
```
mulby: It occurs to me that we should search and replace `self.input()['data'][0]` with `self.file_input_target`

edx/analytics/tasks/data_deidentification.py
```diff
@@ -236,10 +280,31 @@ def file_pattern(self):
         return '*-wiki_articlerevision-*'
 
     def filter_row(self, row):
+        user_id = row[5]
+        user_info = {}
+        if user_id != 'NULL':
+            user_id = int(user_id)
+            user_info['user_id'] = [user_id,]
+            try:
```
mulby: nit: we may want to make this a function given that it is also done on line 174

edx/analytics/tasks/util/eventlog.py
```diff
@@ -131,6 +137,17 @@ def get_event_time(event):
         return None
 
 
+def get_event_username(event):
+    """Returns a username from an event object, if present."""
+    username = event.get('username')
+    # Some usernames have trailing newlines, so remove that.
```
mulby, line 143: they do? If so, this is almost certainly breaking our parsing of events in hadoop jobs since hadoop splits its input files on newline characters, so it's probably splitting these events into two separate lines, neither of which can be parsed by the json parser.

edx/analytics/tasks/util/deid_util.py
```diff
@@ -1,18 +1,619 @@
...
+    auth_user_path = luigi.Parameter()
+    auth_userprofile_path = luigi.Parameter()
+
+
+class UserInfoMixin(UserInfoDownstreamMixin):
+
+    def user_info_requirements(self):
+        return {
+            # 'auth_user': ExternalURL(self.auth_user_path.rstrip('/') + '/'),
+            # 'auth_userprofile': ExternalURL(self.auth_userprofile_path.rstrip('/') + '/')
```
mulby: remove this commented out code

edx/analytics/tasks/util/deid_util.py
```diff
@@ -1,18 +1,619 @@
...
+
+
+class Deidentifier(object):
+
+    log_context = DEFAULT_LOG_CONTEXT
+    entities = DEFAULT_ENTITIES
+
+    def __init__(self, **kwargs):
+        if 'log_context' in kwargs:
+            self.log_context = kwargs['log_context']
```
mulby: This pattern feels odd since it can result in "self.log_context" not being defined after the constructor has been executed.

mulby: Ohhhh... I didn't see the class level definition of the defaults.

I think I would prefer something like this:

``` python
def __init__(self, log_context=DEFAULT_LOG_CONTEXT, entities=DEFAULT_ENTITIES):
    self.log_context = log_context
    self.entities = entities
```

edx/analytics/tasks/data_deidentification.py
```diff
@@ -308,7 +424,11 @@ def requires(self):
         kwargs = {
             'course': self.course,
             'output_directory': self.output_directory,
-            'data_directory': self.data_directory
+            'data_directory': self.data_directory,
+            'entities': self.entities,
+            'log_context': self.log_context,
```
mulby: As I typically do, I would advocate for removing these parameters from the interface for these tasks and pulling them from the config file. I just don't see us tweaking them very frequently and am OK with the added friction associated with having to create branch on the secure repo during development.

edx/analytics/tasks/util/deid_util.py
```diff
@@ -1,18 +1,619 @@
...
+                    new_list.append(value)
+            if changed:
+                return new_list
+            else:
+                return None
+        elif isinstance(obj, unicode):
+            # First perform backslash decoding on string, if needed.
+            if needs_backslash_decoding(obj):
+                decoded_obj = backslash_decode_value(obj)
+                new_label = u"{}*d".format(label)
```
mulby: What is the significance of "*d"?

edx/analytics/tasks/util/deid_util.py
```diff
@@ -1,18 +1,619 @@
...
+                    new_list.append(updated_value)
+                else:
+                    new_list.append(value)
+            if changed:
+                return new_list
+            else:
+                return None
+        elif isinstance(obj, unicode):
+            # First perform backslash decoding on string, if needed.
+            if needs_backslash_decoding(obj):
```
mulby: Why do we only deidentify if it needs backslash decoding? That seems orthogonal to me...

mulby: ah, nevermind I see the block below... this was somewhat unexpected.

I see now that it has to recursively perform backslash decoding... it might be clearer to just do it in a tight loop:

``` python
to_deidentify = obj
num_decodes = 0
while needs_backslash_decoding(to_deidentify):
    to_deidentify = backslash_decode_value(to_deidentify)
    num_decodes += 1

new_label = u"{0}{1}".format(label, '*d'*num_decodes)
updated_value = self.deidentify_text(obj, user_info, log_context, entities)
```

mulby:
meta note: we should get a successful run before addressing my comments, I don't see any show stoppers in here (yet)

edx/analytics/tasks/events_deidentification.py
```diff
@@ -101,16 +84,16 @@ def mapper(self, line):
...
-            return
-
-        yield date_string.encode('utf-8'), cjson.encode(deidentified_event)
+        yield date_string.encode('utf-8'), line.rstrip('\r\n')
 
     def multi_output_reducer(self, _key, values, output_file):
         with gzip.GzipFile(mode='wb', fileobj=output_file) as outfile:
             try:
                 for value in values:
+                    filtered_event = eventlog.parse_json_event(value)
```
mulby: I should have added a comment here about the reasoning behind the move to the reducer

edx/analytics/tasks/events_deidentification.py
```diff
@@ -161,43 +146,183 @@ def filter_implicit_event(self, event):
...
+            username = username.decode('utf8')
+            remapped_username = self.remap_username(username, user_info)
+            if remapped_username is not None:
+                event['username'] = remapped_username
+            else:
+                # TODO: what to do if the username isn't found.  Do we delete it?  Leave it?  Stub it?
+                pass
+
+        # Get the user_id from context, either as an int or None, and remap.
+        userid_entry = None
```
mulby: Is this used?

edx/analytics/tasks/events_deidentification.py
```diff
@@ -161,43 +146,183 @@ def filter_implicit_event(self, event):
...
+        # Note that eventlog.get_event_username() does a strip on the username and checks for zero-len,
+        # so we don't have to do so here.
+        username = eventlog.get_event_username(event)
+        if username is not None:
+            username = username.decode('utf8')
+            remapped_username = self.remap_username(username, user_info)
+            if remapped_username is not None:
+                event['username'] = remapped_username
+            else:
+                # TODO: what to do if the username isn't found.  Do we delete it?  Leave it?  Stub it?
```
mulby: Maybe replace with a constant string like "[REDACTED]"?

brianhw:
Mostly complete now in terms of unit test coverage.  I also went ahead with changing "deidentify" to the more modest (and less inaccurate) "obfuscate", including renaming files.  Remaining work consists of adding docstrings and other pylint fixes, updating secure repo with new config section name, and updating (and running) jenkins jobs.   

Remaining wart:  I would also like to verify that the "original" email pattern is no longer used, and remove it.   (I switched the order so it's run after the more complete pattern now, so it should in theory never find anything anymore.)

FYI @mulby and @HassanJaveed84 .

edx/analytics/tasks/util/obfuscate_util.py
```diff
@@ -0,0 +1,635 @@
...
+    matches = pattern.finditer(string)
+    for match in matches:
+        start, end = match.span()
+        if log_context >= 0:
+            left_edge = start - log_context if start > log_context else 0
+            left = backslash_encode_value(string[left_edge:start])
+            right = backslash_encode_value(string[end:end + log_context])
+            value1 = match.group(0)
+            value = unicode(value1)
+            log.info(u"Found %s:  %s<<%s>>%s", label, left, value, right)
```
mulby, line 231: Change to log.debug

edx/analytics/tasks/util/obfuscate_util.py
```diff
@@ -0,0 +1,635 @@
...
+# Some failures:
+#
+# Regular is too limited, so it truncates:
+# Found EMAIL:  ed to my account in <<blah@yahoomail.com>>.au
+
+# But extended doesn't deal well with case!
+# Found POSSIBLE_EMAIL:  il me at First.L<<ast@example.co.uk>>.
+
+# Leaving off name that precedes, as in email headers:
+# Found POSSIBLE_EMAIL:  rom: "First Last" <<<firstlast@example.com.au>>>
```
mulby: Are these captured by unit tests?

edx/analytics/tasks/util/obfuscate_util.py
```diff
@@ -0,0 +1,635 @@
...
+    patterns.append(u" ".join(names))
+    for name in names:
+        if len(name) > 2 and name not in STOPWORDS and not name.endswith('.'):
+            patterns.append(name)
+
+    # Because we're operating with unicode instead of raw strings, make sure that
+    # the slashes are escaped.
+    fullname_pattern = re.compile(
+        u'\\b({})\\b'.format(u"|".join(patterns)),
+        re.IGNORECASE,
```
mulby: for "\b" to do the right thing, don't we need to us re.UNICODE here?

edx/analytics/tasks/util/tests/test_obfuscate_util.py
```diff
@@ -0,0 +1,611 @@
...
+    )
+    def test_needs_no_backslash_decoding(self, text):
+        self.assertFalse(obfuscate_util.needs_backslash_decoding(text))
+
+    @data(
+        'This is a test.\\nThis is another.',
+        'This is a test.\\\\\\nThis is another.',
+    )
+    def test_decoding_round_trip(self, text):
+        self.assertEquals(text, obfuscate_util.backslash_encode_value(obfuscate_util.backslash_decode_value(text)))
```
mulby, line 39: In addition to the round trip, we should have at least one test of the one-way. Otherwise both functions could be nops!

edx/analytics/tasks/util/obfuscate_util.py
mulby: nevermind, I see them below!

edx/analytics/tasks/util/tests/test_obfuscate_util.py
```diff
@@ -0,0 +1,611 @@
...
+
+    #####################
+    # email
+    #####################
+
+    @data(
+        'testuser@gmail.com',
+        'test.user@gmail.com',
+        'test_user@gmail.com',
+        'Ttestuser@gmail.com',
```
mulby, line 149: test.user+foo@gmail.com

mulby:
A couple of nits, but overall looks great.

edx/analytics/tasks/util/obfuscate_util.py
brianhw, line 231: This is deliberately set at info, as I don't view it as "debug" logic.  It is development logic, yes, but it is controlled by log_context, not by log level.  Same thing in the code that calls this, that provides info about the context of the matches.

edx/analytics/tasks/util/tests/test_obfuscate_util.py
brianhw, line 149: Done.

edx/analytics/tasks/util/obfuscate_util.py
brianhw: Removed dead comments.

edx/analytics/tasks/util/obfuscate_util.py
brianhw: Yes, good catch!  I added a test case for it as well.

edx/analytics/tasks/util/tests/test_obfuscate_util.py
brianhw, line 39: Done.

mulby:
:+1: ready for squash and merge IMHO

PR #197, opened 2016-01-28
Simplified user_activity loading into vertica

HassanJaveed84:
@mulby @brianhw 
test_internal_reporting_user_activity acceptance test breaks because of the introduction of ExternalURL as the dependency. How can I set that up in the test to make it work ?


edx/analytics/tasks/load_internal_reporting_user_activity.py
```diff
@@ -224,3 +220,24 @@ def requires(self):
                 history_schema=self.history_schema
             )
         ]
+
+
+class UserActivityWorkflow(luigi.WrapperTask):
```
HassanJaveed84, line 225: Should this be moved to user_activity.py ?

HassanJaveed84:
@mulby updated acceptance test.

edx/analytics/tasks/load_internal_reporting_user_activity.py
brianhw, line 225: Good question.  I don't think it hugely matters, but I'm inclined to say 'no'.  This is the work that needs to be run every week, so putting it with the other weekly code might make sense.  But this file is dependent on the other file anyway, and having the other file be dependent on this one as well seems unnecessarily circular.  I would also put the workflow that does the daily writing to Vertica in this file, not in user_activity.

brianhw:
LGTM. :+1:

edx/analytics/tasks/load_internal_reporting_user_activity.py
mulby, line 225: I'm a little confused, when will these be called? Is it the new entry point?

If so, what is happening with InternalReportingUserActivityWorkflow?

brianhw, line 225: I understand this to be the new entry point for a weekly job.  Then there is a separate daily job to write this output of the weekly job to Vertica.

mulby, line 225: Ah, that makes sense.

mulby:
:+1:

PR #198, opened 2016-02-04
Fixed deidentification acceptance test.

HassanJaveed84:
@brianhw @mulby 
This test actually won't run until it gets rebased of master with the updated test.cfg. 
I've already tested it in a separate branch.


mulby:
:+1: we can probably just merge this into the branch

We'll review it again before merging into master

PR #201, opened 2016-02-16
WIP: remove potentially sensitive xblock fields

mulby:
Still needs tests etc, but here's a sketch of the procedure.


share/xblock_obfuscation_config.yml
```diff
@@ -0,0 +1,406 @@
+fields:
```
mulby, line 1: @ormsbee - do any of you see any fields or blocks in here that might contain secrets?

The blocks and/or fields I've commented out here are either not used often enough to justify further investigation or seem like they might contain secrets.

Any field or element in the course export that does not appear in these lists will be removed. As always, I'm trying to balance the utility of the artifact for research purposes with data security concerns. If you see any field in here that you think _might_ contain secrets, we should just remove it.

Methodology:
- I wrote a short management command that dumped all blocks the LMS knew about and any field in the "settings" scope. It also dumped the fields in the `settings.XBLOCK_MIXINS` classes in the same scope. I ran this on my devstack.
- I ran a query to identify the blocks used by courses in production and count usages to ensure we don't exclude any of the most common blocks, I can share this data with you if you want.
- I commented out blocks that were used very infrequently.
- I commented fields that sounded like they might contain secrets.

mulby:
@brianhw this is ready for a preliminary review

share/xblock_obfuscation_config.yml
```diff
@@ -0,0 +1,406 @@
...
+- graded
+# - annotation_token_secret
```
brianhw: I like the idea of including the items that have been deliberately excluded as a comment like this.  I would go further and add some kind of indication, either on the same line or in a preceding comment, explaining why it is commented out.  In this case, I might judge from the name that it's because the value might be a secret.  However, there are items below that are commented out for different reasons, and it would be good to spell them out as well.

brianhw: Actually, I have a different thought.  Instead of having a single "fields" list, it might be more maintainable to have a separate list of fields that are to be removed or ignored.  That way, the code can check whether a decision has been made about a given field (or a given xblock, for that matter), and could log a warning (or error) for those things that show up in the data that are not in this enumeration.  We would at least know what we were missing, and it would help with answering such questions and also with updating this file as necessary.

share/xblock_obfuscation_config.yml
```diff
@@ -0,0 +1,406 @@
...
+#    fields:
+#    - settings
+#  acid_parent:
+#    fields:
+#    - settings
+#    has_children: true
+  annotatable:
+    fields: []
+#  book:
+#    fields: []
```
brianhw: For example, is this commented out because books contain secrets?  Or because it's not really part of normal courses?  Or some other reason?

edx/analytics/tasks/data_obfuscation.py
```diff
@@ -306,21 +310,146 @@ def filter_row(self, row):
...
+            for attribute in element.attrib.keys():
+                if attribute != 'url_name':
+                    element.attrib.pop(attribute)
+            for child in element:
+                element.remove(child)
+
+            element.set("redacted", "true")
+        else:
+            fields = set(block_config.get('fields', []))
+            fields.update(self.xblock_whitelist_config['fields'])
```
brianhw: Wondering if this would be clearer to be called "global_fields", since it's a whitelist that's added to every xblock.

edx/analytics/tasks/data_obfuscation.py
```diff
@@ -306,21 +310,146 @@ def filter_row(self, row):
...
+                    category = block_structure.get('category')
+                    if category == 'course':
+                        metadata = block_structure.get('metadata', {})
+                        to_remove = frozenset(metadata.keys()) - self.course_field_whitelist
+                        for attribute in to_remove:
+                            metadata.pop(attribute)
+                            log.info('Removed course attribute "{0}"'.format(attribute))
+                    elif category not in self.xblock_whitelist_config['xblocks']:
+                        block_structure['metadata'] = {}
+                        block_structure['redacted'] = True
```
brianhw: Maybe "metadata_redacted"?

brianhw:
Done with my pass.

share/xblock_obfuscation_config.yml
mulby: done

share/xblock_obfuscation_config.yml
mulby: Done

edx/analytics/tasks/data_obfuscation.py
mulby: I modified this so that it adds a field to the block called "redacted_metadata" that contains a list of the fields that were removed.

share/xblock_obfuscation_config.yml
```diff
@@ -0,0 +1,478 @@
+fields:
+- graded
+- giturl
```
ormsbee: A `giturl` can have a username and password encoded in it.

share/xblock_obfuscation_config.yml
```diff
@@ -0,0 +1,478 @@
...
+    - is_entrance_exam
+    - is_time_limited
+    - exam_review_rules
+    - is_proctored_enabled
+    - default_time_limit_minutes
+    - is_practice_exam
+    has_children: true
+  course:
+    fields:
+    - remote_gradebook
```
ormsbee: I think the data in `remote_gradebook` might be dangerous. It's unclear now because the code that uses it has been ripped out, so we ignore the value.

share/xblock_obfuscation_config.yml
```diff
@@ -0,0 +1,478 @@
...
+    - certificates_display_behavior
+    - allow_anonymous
+    - is_proctored_enabled
+    - html_textbooks
+    - end
+    - invitation_only
+    - no_grade
+    - catalog_visibility
+    - entrance_exam_enabled
+    - certificates
```
ormsbee: I think the `certificates` dict will allow folks to find the URLs of various professors' signature images. Is that ok?

share/xblock_obfuscation_config.yml
```diff
@@ -0,0 +1,478 @@
...
+    - show_calculator
+    - info_sidebar_name
+    - display_coursenumber
+    - allow_public_wiki_access
+    - enrollment_end
+    - is_time_limited
+    - hide_progress_tab
+    - exam_review_rules
+    - cosmetic_display_price
+    - ccx_connector
```
ormsbee: While it's not really a secret, we probably shouldn't give out `ccx_connector`, since it's pointing to someone's server.

share/xblock_obfuscation_config.yml
```diff
@@ -0,0 +1,478 @@
...
+    - editor
+    - filename
+  image:
+    fields: []
+  imageannotation:
+    fields:
+    - data
+    - instructor_tags
+    - annotation_mode
+    - instructor_email
```
ormsbee: Is it okay to have instructor emails?

share/xblock_obfuscation_config.yml
```diff
@@ -0,0 +1,478 @@
...
+    - custom_parameters
+    - weight
+    - ask_to_send_email
+    - accept_grades_past_due
+    - button_text
+    - has_score
+    - hide_launch
+    - open_in_a_new_page
+    - launch_url
+    - lti_id
```
ormsbee: We should definitely remove `lti_id` since it's a client_id/secret pairing.

share/xblock_obfuscation_config.yml
```diff
@@ -0,0 +1,478 @@
...
+    - ask_to_send_email
+    - accept_grades_past_due
+    - button_text
+    - has_score
+    - hide_launch
+    - launch_target
+    - modal_width
+    - launch_url
+    - modal_height
+    - lti_id
```
ormsbee: Same here, `lti_id` needs to be removed.

share/xblock_obfuscation_config.yml
ormsbee, line 1: @mulby: Made my first pass at the field list. Given the sensitive nature of this, you might want to get a second pass done by someone else who's been around the code base for a while.

mulby, line 1: @ormsbee thanks for the review, very helpful.

@cpennington can you also take a pass over this list?

cpennington, line 1: Did you decide not to dump the content scope? That's going to have a lot of the meat (especially of CAPA problems).

mulby, line 1: I decided to whitelist the content scoped fields. Note that I manually merged them into the existing whitelist... hopefully I didn't miss any.

This list should include both Scope.content and Scope.settings fields.

mulby:
@brianhw this is ready for another review pass, still working on more unit tests, but it's stabilizing quite a bit

share/xblock_obfuscation_config.yml
brianhw, line 1: It's sort of hard to scan some of these lists that are getting pretty long -- it would help to have the items in alphabetical order.

share/xblock_obfuscation_config.yml
```diff
@@ -0,0 +1,478 @@
...
+    - max_count
+    - capa_type
+    - source_library_version
+    - has_score
+    - mode
+    has_children: true
+  lti:
+    fields:
+    - description
+    - custom_parameters
```
brianhw: Is this also risky?

share/xblock_obfuscation_config.yml
```diff
@@ -0,0 +1,478 @@
...
+    - weight
+    - ask_to_send_email
+    - accept_grades_past_due
+    - button_text
+    - has_score
+    - hide_launch
+    - open_in_a_new_page
+    - launch_url
```
brianhw: Are credentials included in this?

share/xblock_obfuscation_config.yml
```diff
@@ -0,0 +1,478 @@
...
+    - mod_queue
+    - debug
+    - hint_pk
+    - moderate
+    - data
+    - hints
+    has_children: true
+  discuss:  # not frequently used
+    fields: []
+  google-calendar:  # unsure if safe
```
brianhw: Just trying to understand:  the concern here is that the _content_ of the xblock might be problematic, rather than just the metadata?  If it were just the metadata, then they could be excluded rather than the xblock.

edx/analytics/tasks/data_obfuscation.py
```diff
@@ -1,21 +1,25 @@
 """Obfuscate course state files by removing/stubbing user information."""
+import shutil
+import tarfile
```
brianhw, line 3: FYI:  with pep8/pylint upgrade, this funky ordering here will likely result in pep8/pylint issues.

edx/analytics/tasks/data_obfuscation.py
```diff
@@ -307,21 +311,183 @@ def filter_row(self, row):
...
+
+    def get_field_blacklist(self, block_type):
+        return self.get_field_list(block_type, 'blacklist')
+
+    def remove_fields_from_dict(self, block_type, dict_to_modify):
+        to_remove = frozenset(dict_to_modify.keys()) - self.get_field_whitelist(block_type)
+        for attribute in to_remove:
+            dict_to_modify.pop(attribute)
+            if attribute in self.get_field_blacklist(block_type):
+                log.info('Removed blacklisted attribute "{0}" from "{1}"'.format(attribute, block_type))
```
brianhw: info or debug?

edx/analytics/tasks/data_obfuscation.py
```diff
@@ -307,21 +311,183 @@ def filter_row(self, row):
...
+
+        return self._field_lists
+
+    def get_field_list(self, block_type, list_type):
+        return self.field_lists.get(block_type, {}).get(list_type, frozenset())
+
+    def get_field_whitelist(self, block_type):
+        return self.get_field_list(block_type, 'whitelist')
+
+    def get_field_blacklist(self, block_type):
```
brianhw, line 355: prepend a bunch of these with underscore.  The mixin really contributes only requires(), remove_fields_from_dict() and the two properties.  The three above are internal.

edx/analytics/tasks/data_obfuscation.py
```diff
@@ -307,21 +311,183 @@ def filter_row(self, row):
...
+                    elif category not in self.xblock_config['xblocks']:
+                        removed_fields = set(metadata.keys())
+                        for key in removed_fields:
+                            del metadata[key]
+                        if category in self.xblock_config.get('exclude_xblocks', set()):
+                            log.info('Removed metadata from blacklisted block "{0}"'.format(category))
+                        else:
+                            log.info('Removed metadata from unknown block "{0}"'.format(category))
+                    else:
+                        removed_fields = set()
```
brianhw: Why does this not call remove_fields_from_dict as well?  Do non-course xblocks not have fields to be removed?

edx/analytics/tasks/data_obfuscation.py
```diff
@@ -307,21 +311,183 @@ def filter_row(self, row):
...
+        removed_fields = set()
+        removed_children = set()
+        if not block_config:
+            if element.tag in self.xblock_config.get('exclude_xblocks', set()):
+                log.info('Blacklisted element "{0}", removing all attributes and children'.format(element.tag))
+            else:
+                log.warning('Unrecognized element "{0}", removing all attributes and children'.format(element.tag))
+
+            for attribute in element.attrib.keys():
+                if attribute != 'url_name':
```
brianhw: Not clear why this line is here:  needs a comment.

share/xblock_obfuscation_config.yml
```diff
@@ -0,0 +1,478 @@
...
+- static_asset_path
+- name
+- show_reset_button
+- hide_from_toc
+- group_access
+- rerandomize
+- user_partitions
+- graceperiod
+- visible_to_staff_only
+- url_name
```
brianhw, line 30: This is the attribute that everything has?  So some of these things are attributes of the modules/xblocks, but some are instead artifacts of the XML representation?  Is it just this?

edx/analytics/tasks/data_obfuscation.py
brianhw: Is this the only attribute that we need to keep?  Do we also need to keep parent_url and name too?  Just wondering what the set of attributes that are always kept, even for blacklisted and unrecognized blocks.

edx/analytics/tasks/data_obfuscation.py
```diff
@@ -307,21 +311,183 @@ def filter_row(self, row):
...
+                    element.attrib.pop(attribute)
+                    removed_fields.add(attribute)
+            for child in list(element):
+                removed_children.add(child.tag)
+                element.remove(child)
+        else:
```
brianhw: nit: I would reverse if and else, since I expect the found case before the not-found case.  Real small nit.

edx/analytics/tasks/data_obfuscation.py
```diff
@@ -307,21 +311,183 @@ def filter_row(self, row):
...
+            removed_fields = self.remove_fields_from_dict(element.tag, element.attrib)
+
+            if block_config.get('has_children', True):
+                for child_element in element:
```
brianhw: Above you call "for child in list(element)" and here "for child_element in element".  Both should be the same.  Is list() needed?  Would also be good to standardize on using child_element.

share/xblock_obfuscation_config.yml
mulby: The difference is exactly what you specified. These blocks aren't used frequently to justify any investigation, I can just update the comment to reflect this fact. I assume they don't have children, but I just excluded all rarely used blocks without really even looking into them to see if children were a possibility or not.

edx/analytics/tasks/data_obfuscation.py
mulby: Good call, I meant to make these debug before merging.

edx/analytics/tasks/data_obfuscation.py
brianhw: Oh, I'm guessing that the list() is needed to make a copy before doing the remove.  (Add a comment...)

edx/analytics/tasks/data_obfuscation.py
mulby, line 355: Not sure I follow, what do you mean by "internal"?

edx/analytics/tasks/data_obfuscation.py
```diff
@@ -307,21 +311,183 @@ def filter_row(self, row):
...
+                    if child_element.tag not in block_config.get('fields', []):
+                        self.clean_element(child_element)
```
brianhw: I'm not clear how we decide when to clean something and when to remove it.  If an element being cleaned is not on the xblock list, then its children are removed.  If it is on the xblock list and is not marked as having children, then any children it has are preserved without cleaning (nor are any of its further descendants).  If it is on the xblock list and is marked as having children and a child's tag appears in the xblock's field list, then the child is not cleaned (nor are any of its further descendants).  Otherwise children only get cleaned if the parent is labelled as having children and the child is not listed in fields.  Wow, that would be great to get into a comment.  Wondering if it makes sense to ensure there are no children if has_children is not set.

edx/analytics/tasks/data_obfuscation.py
mulby: Great catch, this is somewhat old logic that I hadn't updated.

edx/analytics/tasks/data_obfuscation.py
mulby: parent_url is only used for drafts which we exclude. By name do you mean "display_name"?

edx/analytics/tasks/data_obfuscation.py
```diff
@@ -307,21 +311,183 @@ def filter_row(self, row):
...
+
+        if len(removed_fields) > 0:
+            element.set("redacted_attributes", ",".join(removed_fields))
```
brianhw: Hopefully nothing secret is included as a key name....

edx/analytics/tasks/data_obfuscation.py
mulby: Added a comment.

edx/analytics/tasks/data_obfuscation.py
mulby: Concur! I usually try to do this...

edx/analytics/tasks/data_obfuscation.py
brianhw, line 355: internal = only used within this class.  I.e. part of its implementation and not its interface.  -> _get_field_blacklist.   Not critical, more of a nit, but it helps to define the contract of the class with those using it when reading the code.

edx/analytics/tasks/data_obfuscation.py
mulby: Added a comment.

edx/analytics/tasks/data_obfuscation.py
mulby: I put in a comment to help clarify this... hopefully it is effective... here is the text:

``` python
            # The logic here is a bit tricky, a sub-element can either be a field of the current block or a child block.
            # We just assume everything is a child block unless we are told otherwise by the configuration. Fields that
            # can appear as sub-elements need to be declared in the field whitelist. Note that some wonkiness is
            # possible if a field name happens to conflict with a block name. This could result in us thinking a child
            # block is actually a field and ignoring it.
            #
            # We only "clean" a sub-element if and only if:
            # 1) The current block does not have the property has_children=False. This is used to prevent traversal of
            #    sub-elements for blocks that can have arbitrary children (like raw HTML). Note that by default we
            #    traverse the sub-elements, we only disable this behavior if we have explicitly set this field to false.
            # 2) The sub-element name does not appear in the field list for the current block.
```

PR #204, opened 2016-02-19
Improve workflow for running acceptance tests

itsjeyd:
This PR aims to improve the workflow for running acceptance tests in instances of the Analytics Devstack. It introduces a wrapper script that can run the whole test suite or just a specific test module/class/method. It also adds code that makes sure tests depending on services that are unavailable in a local test environment are skipped correctly.

**Test instructions**

To prepare for testing, complete the following steps:
- Test https://github.com/edx/configuration/pull/2790.
- Starting from the `analyticstack` directory, update the version of edx-analytics-pipeline that you checked out while testing https://github.com/edx/configuration/pull/2790 and run a pipeline task (make sure to adjust `--vagrant-path` before):
  
  ``` sh
  cd ../edx-analytics-pipeline
  source venv/bin/activate
  git remote add fork https://github.com/open-craft/edx-analytics-pipeline.git
  git fetch fork
  git checkout tim/improve-pipeline-workflow
  make bootstrap
  export WHEEL_URL=https://edx-wheelhouse.s3-website-us-east-1.amazonaws.com/Ubuntu/precise
  remote-task --vagrant-path /path/to/analytics/analyticstack --remote-name devstack --override-config ${PWD}/config/devstack.cfg --wheel-url $WHEEL_URL --wait ImportEnrollmentsIntoMysql --local-scheduler --interval-end $(date +%Y-%m-%d -d "tomorrow") --n-reduce-tasks 1 
  ```
  
  This will cause edx-analytics-pipeline to "install itself" on your Analytics Devstack (which is necessary because edx-analytics-pipeline is currently _not_ installed like any other repo when creating an instance of Analytics Devstack).

To test the changes in this PR, follow these steps:
- Starting from the `analyticstack` directory, shut down Analytics Devstack and check out another copy of edx-analytics-pipeline:
  
  ``` sh
  cd ../analyticstack
  vagrant halt
  git clone https://github.com/open-craft/edx-analytics-pipeline.git
  cd edx-analytics-pipeline
  git checkout tim/improve-pipeline-workflow
  ```
- Modify Vagrantfile to mount the repository at `/edx/app/analytics_pipeline/analytics_pipeline`:
  
  ``` ruby
    :analytics_pipeline => {:repo => "edx-analytics-pipeline", :local => "/edx/app/analytics_pipeline/analytics_pipeline", :owner => "hadoop"},
  ```
  
  **Note**: As of https://github.com/edx/configuration/commit/3f6082eedf862a2729679ba51a542b0d5d5aafe5, this step is no longer necessary.
- Bring Analytics Devstack back up and use new script to run acceptance tests:
  
  ``` sh
  vagrant up
  cd /var/lib/analytics-tasks/devstack/repo/scripts  # This is where edx-analytics-pipeline installed itself
  ./run_acceptance.sh --test test_video --update
  ```
  
  As mentioned above, you can also specify that you want to limit execution to a certain test class (e.g., `test_video:VideoAcceptanceTest`) or method (e.g., `test_video:VideoAcceptanceTest.test_video_timeline`) when running tests. The `--update` option ensures that the venv in which the tests will be executed is up-to-date; you can leave it off when running another test.
- Verify that the tests that have been annotated to require specific services (such as S3) are skipped when you try to run them. For example, the following command
  
  ``` sh
  ./run_acceptance.sh --test test_course_catalog
  ```
  
  should output:
  
  ```
  Tests the workflow for the course subjects, end to end. ... SKIP: S3 is not available
  
  ----------------------------------------------------------------------
  XML: nosetests.xml
  ----------------------------------------------------------------------
  Ran 1 test in 0.001s
  
  OK (SKIP=1)
  ```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,6 +49,9 @@ test: test-requirements develop test-local
 test-acceptance: test-requirements
 	LUIGI_CONFIG_PATH='config/test.cfg' python -m coverage run --rcfile=./.coveragerc -m nose --nocapture --with-xunit -A acceptance $(ONLY_TESTS)
 
+test-acceptance-local:
+	REMOTE_TASK=$(shell which remote-task) LUIGI_CONFIG_PATH='config/test.cfg' ACCEPTANCE_TEST_CONFIG="/var/tmp/acceptance.json" python -m coverage run --rcfile=./.coveragerc -m nose --nocapture --with-xunit -A acceptance --stop -v $(ONLY_TESTS)
+
 coverage-local: test-local
 	python -m coverage html
 	python -m coverage xml -o coverage.xml
@@ -67,4 +70,3 @@ coverage: test coverage-local
 
 todo:
 	pylint --disable=all --enable=W0511 edx
-
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ The Hadoop-based data pipeline.
 Installation instructions
 --------------------
 
-We have some draft docs: 
+We have some draft docs:
 
 * https://openedx.atlassian.net/wiki/display/OpenOPS/edX+Analytics+Installation describes how to install a small scale analytics stack on a single machine. It'll get merged into the next doc below at some point. Help welcome!
 * http://edx.readthedocs.org/projects/edx-installing-configuring-and-running/en/latest/analytics/install_analytics.html has an overview of a larger scale AWS install.
@@ -56,6 +56,9 @@ Values for `<OPERATING SYSTEM>/<OS VARIANT>`:
 
 Running the Tests
 -----------------
+
+### Unit tests
+
 Run `make test` to install the Python requirements and run the unit tests.
 
 Some of the tests rely on AWS. If you encounter errors such as `NoAuthHandlerFound: No handler was ready to authenticate. 1 handlers were checked. ['HmacAuthV1Handler'] Check your credentials`,
@@ -65,6 +68,111 @@ valid credentials for the tests to pass, so the commands below should fix the fa
     export AWS_ACCESS_KEY_ID='AK123'
     export AWS_SECRET_ACCESS_KEY='abc123'
 
+### Acceptance tests
+
+**Preconditions**
+
+To be able to run acceptance tests on an instance of the Analytics
+Devstack, the following conditions must hold:
+
+- [`/var/tmp/acceptance.json`](https://github.com/edx/configuration/blob/master/playbooks/roles/analytics_pipeline/files/acceptance.json)
+  is present on your instance of Analytics Devstack. This file
+  contains the default configuration for running acceptance tests.
+
+- The `pipeline001` user has access to tables storing acceptance test
+  data. You can verify this by doing the following (as `vagrant` user):
+
+  ```sh
+  mysql -u root
+  SHOW GRANTS FOR 'pipeline001'@'localhost';
+  ```
+
+  If your instance is configured correctly, you should get the
+  following output:
+
+  ```
+  +----------------------------------------------------------------------+
+  | Grants for pipeline001@localhost                                     |
+  +----------------------------------------------------------------------+
+  | ...                                                                  |
+  | GRANT ALL PRIVILEGES ON `acceptance%`.* TO 'pipeline001'@'localhost' |
+  +----------------------------------------------------------------------+
+  ```
+
+- The pipeline has installed itself on your instance of Analytics
+  Devstack. This happens when you run a pipeline task on the instance
+  as described in [this section](https://edx.readthedocs.org/projects/edx-installing-configuring-and-running/en/latest/devstack/analytics_devstack.html#run-the-open-edx-analytics-pipeline)
+  of the [official installation instructions](https://edx.readthedocs.org/projects/edx-installing-configuring-and-running/en/latest/devstack/analytics_devstack.html).
+
+If you are running a recent version of the Analytics Devstack (and
+followed the official installation instructions to set it up), the
+conditions above should already be satisfied.
+
+**Steps**
+
+1. If you haven't already, check out a local copy of this repository
+   into the local directory of your Analytics Devstack installation.
+   Restart the instance and make sure the repository is mounted at
+   `/edx/app/analytics_pipeline/analytics_pipeline`.
+
+2. If the code you would like to test lives on a different branch than
+   `master`, navigate to `/edx/app/analytics_pipeline/analytics_pipeline`
+   and check out that branch.
+
+3. Navigate to `/var/lib/analytics-tasks/devstack/repo/scripts`
+   (this is where the pipeline installs itself by default) and check
+   out the branch that contains the test code you would like to use.
+   *This step is only necessary if you modified existing tests or
+   created new tests that you'd like to run. You can skip it if you
+   want to run acceptance tests to verify changes to existing pipeline
+   tasks*.
+
+4. From `/var/lib/analytics-tasks/devstack/repo/scripts`, run the
+   following command to launch the entire test suite:
+
+   ```sh
+   ./run_acceptance.sh --update
+   ```
+
+   The `--update`/`-u` option ensures that the venv in which the tests
+   will be executed is up-to-date; you can leave it off when running
+   another test.
+
+   You can also limit test execution to a certain module, class, or
+   method. To do this, use the `--test`/`-t` option:
+
+   ```sh
+   ./run_acceptance.sh --test test_course_catalog
+   ./run_acceptance.sh --test test_course_catalog:CourseSubjectsAcceptanceTest
+   ./run_acceptance.sh --test test_course_catalog:CourseSubjectsAcceptanceTest.test_course_subjects
+   ```
+
+   The `--test`/`t` option allows you to specify multiple test
+   modules/classes/methods at the same time:
+
+   ```sh
+   ./run_acceptance.sh -t test_course_catalog -t test_video
+   ```
+
+   Note that when running multiple tests, execution will stop
+   after the first error or failure. (This is helpful because
+   acceptance tests can take a long time to finish.)
+
+   Note also that some tests depend on the availability of specific
+   services to run successfully. When running these tests on an
+   instance of the Analytics Devstack, they will be skipped with
+   output similar to this:
+
+   ```
+   Tests the workflow for the course subjects, end to end. ... SKIP: S3 is not available
+
+   ----------------------------------------------------------------------
+   XML: nosetests.xml
+   ----------------------------------------------------------------------
+   Ran 1 test in 0.001s
+
+   OK (SKIP=1)
+   ```
 
 How to Contribute
 -----------------
--- a/edx/analytics/tasks/tests/acceptance/__init__.py
+++ b/edx/analytics/tasks/tests/acceptance/__init__.py
@@ -1,17 +1,80 @@
+import boto
+import hashlib
 import json
 import logging
-import hashlib
 from luigi.s3 import S3Client
 import os
-import sys
 import shutil
 import unittest
 
-from edx.analytics.tasks.url import url_path_join, get_target_from_url
 from edx.analytics.tasks.tests.acceptance.services import fs, db, task, hive, vertica
+from edx.analytics.tasks.url import url_path_join, get_target_from_url
 
 
 log = logging.getLogger(__name__)
+
+# Decorators for tagging tests
+
+def when_s3_available(function):
+    access_keys_defined = getattr(when_s3_available, 'access_keys_defined', None)
+    if access_keys_defined is None:
+        aws_access_key_id = os.getenv('AWS_ACCESS_KEY_ID')
+        aws_secret_access_key = os.getenv('AWS_SECRET_ACCESS_KEY')
+        access_keys_defined = aws_access_key_id is not None and aws_secret_access_key is not None
+        when_s3_available.access_keys_defined = access_keys_defined  # Cache result to avoid having to compute it again
+    if access_keys_defined:
+        s3_available = getattr(when_s3_available, 's3_available', None)
+        if s3_available is None:
+            connection = boto.connect_s3()  # This will not error out if
+                                            # AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY are set,
+                                            # so it can't be used to check if we have a valid connection to S3
+            try:
+                connection.get_all_buckets()
+            except boto.exception.S3ResponseError:
+                s3_available = False
+            else:
+                s3_available = True
+            finally:
+                when_s3_available.s3_available = s3_available  # Cache result to avoid having to compute it again
+        return unittest.skipIf(
+            not s3_available, 'S3 is not available'
+        )(function)
+    else:
+        return unittest.skip('AWS access keys not defined')(function)
+
+def when_exporter_available(function):
+    return unittest.skipIf(
+        os.getenv('EXPORTER') is None, 'Private Exporter code is not available'
+    )(function)
+
+def when_geolocation_data_available(function):
+    config = get_test_config()
+    geolocation_data = config.get('geolocation_data')
+    geolocation_data_available = bool(geolocation_data) and os.path.isfile(geolocation_data)
+    return unittest.skipIf(
+        not geolocation_data_available, 'Geolocation data is not available'
+    )(function)
+
+def when_vertica_available(function):
+    config = get_test_config()
+    vertica_available = bool(config.get('vertica_creds_url'))
+    return unittest.skipIf(
+        not vertica_available, 'Vertica service is not available'
+    )(function)
+
+# Utility functions
+
+def get_test_config():
+    config_json = os.getenv('ACCEPTANCE_TEST_CONFIG')
+    try:
+        with open(config_json, 'r') as config_json_file:
+            config = json.load(config_json_file)
+    except (IOError, TypeError):
+        try:
+            config = json.loads(config_json)
+        except TypeError:
+            config = {}
+    return config
 
 
 class AcceptanceTestCase(unittest.TestCase):
@@ -26,15 +89,7 @@ class AcceptanceTestCase(unittest.TestCase):
         except Exception:
             self.s3_client = None
 
-        config_json = os.getenv('ACCEPTANCE_TEST_CONFIG')
-        try:
-            with open(config_json, 'r') as config_json_file:
-                self.config = json.load(config_json_file)
-        except (IOError, TypeError):
-            try:
-                self.config = json.loads(config_json)
-            except TypeError:
-                self.config = {}
+        self.config = get_test_config()
 
         # The name of an existing job flow to run the test on
         assert('job_flow_name' in self.config or 'host' in self.config)
--- a/edx/analytics/tasks/tests/acceptance/test_enrollment_validation.py
+++ b/edx/analytics/tasks/tests/acceptance/test_enrollment_validation.py
@@ -9,8 +9,8 @@ import StringIO
 
 from luigi.s3 import S3Target
 
+from edx.analytics.tasks.tests.acceptance import AcceptanceTestCase, when_s3_available
 from edx.analytics.tasks.url import url_path_join
-from edx.analytics.tasks.tests.acceptance import AcceptanceTestCase
 
 
 log = logging.getLogger(__name__)
@@ -28,6 +28,7 @@ class EnrollmentValidationAcceptanceTest(AcceptanceTestCase):
     WIDER_DATE_INTERVAL = "{}-{}".format(START_DATE, END_DATE + datetime.timedelta(days=1))
     SQL_FIXTURE = 'load_student_courseenrollment_for_enrollment_validation.sql'
 
+    @when_s3_available
     def test_enrollment_validation(self):
         # Initial setup.
         self.upload_tracking_log(self.INPUT_FILE, self.START_DATE)
--- a/scripts/run_acceptance.sh
+++ b/scripts/run_acceptance.sh
@@ -1,0 +1,75 @@
+#!/usr/bin/env bash
+#
+# Use this script to run the entire suite of acceptance tests
+# or one or more specific test modules/classes/methods
+#
+
+function usage
+{
+    echo "Usage: ./run_acceptance.sh [[[-t test] [-u]] | [-h]]"
+}
+
+TESTS=()
+UPDATE_VENV=
+
+while [ "$1" != "" ]; do
+    case "$1" in
+        -t | --test )
+            shift
+            TEST="$1"
+            TESTS+=("edx.analytics.tasks.tests.acceptance."$TEST)
+            ;;
+        -u | --update )
+            UPDATE_VENV=1
+            ;;
+        -h | --help )
+            usage
+            exit
+            ;;
+        * )
+            usage
+            exit 1
+    esac
+    shift
+done
+
+# Environment
+
+WORKING_DIRECTORY=/var/lib/analytics-tasks/devstack/repo
+
+TASKS_REPO=/edx/app/analytics_pipeline/analytics_pipeline
+
+# Check if repo has been cloned and mounted at default location
+if [ ! -d "$TASKS_REPO" ]
+then
+    echo "Please make sure edx-analytics-pipeline is mounted at ${TASKS_REPO}. See README.md for more info."
+    exit 1
+fi
+
+# Prepare environment as "vagrant" user
+if [ -n "$UPDATE_VENV" ]
+then
+    sudo -Hu vagrant /bin/bash<<-PREPARE
+cd $WORKING_DIRECTORY
+
+# Activate virtualenv
+source ../venv/bin/activate
+
+# Prepare virtualenv for testing
+make develop-local
+make test-requirements
+PREPARE
+fi
+
+# Tests
+
+# Run tests as "hadoop" user
+sudo -Hu hadoop /bin/bash<<-RUN
+cd $WORKING_DIRECTORY
+
+# Activate virtualenv
+source ../venv/bin/activate
+
+# Run tests
+make test-acceptance-local ONLY_TESTS="${TESTS[@]}"
+RUN
--- a/share/task.yml
+++ b/share/task.yml
@@ -82,7 +82,7 @@
         - "{{ log_dir }}"
 
     - name: analytics tasks repository checked out
-      git: repo={{ repo }} dest={{ working_repo_dir }} version=release
+      git: repo={{ repo }} dest={{ working_repo_dir }} version=master
 
     - name: branch fetched
       command: git fetch --all chdir={{ working_repo_dir }}
--- a/edx/analytics/tasks/tests/acceptance/services/vertica.py
+++ b/edx/analytics/tasks/tests/acceptance/services/vertica.py
@@ -63,8 +63,6 @@ class VerticaService(object):
         """
         Connect to the Vertica server.
         """
-        if self.disabled:
-            raise unittest.SkipTest('The vertica service is disabled')
         return vertica_python.connect(user=self.credentials.get('user'), password=self.credentials.get('password'),
                                       database='', host=self.credentials.get('host'))
 
--- a/edx/analytics/tasks/tests/acceptance/test_answer_dist.py
+++ b/edx/analytics/tasks/tests/acceptance/test_answer_dist.py
@@ -7,7 +7,7 @@ import logging
 
 from luigi.s3 import S3Target
 
-from edx.analytics.tasks.tests.acceptance import AcceptanceTestCase
+from edx.analytics.tasks.tests.acceptance import AcceptanceTestCase, when_s3_available
 from edx.analytics.tasks.url import url_path_join
 
 
@@ -41,6 +41,7 @@ class BaseAnswerDistributionAcceptanceTest(AcceptanceTestCase):
 class AnswerDistributionAcceptanceTest(BaseAnswerDistributionAcceptanceTest):
     """Acceptance test for the CSV-generating Answer Distribution Task"""
 
+    @when_s3_available
     def test_answer_distribution(self):
         self.task.launch([
             'AnswerDistributionOneFilePerCourseTask',
@@ -78,6 +79,7 @@ class AnswerDistributionAcceptanceTest(BaseAnswerDistributionAcceptanceTest):
 class AnswerDistributionMysqlAcceptanceTests(BaseAnswerDistributionAcceptanceTest):
     """Acceptance tests for Answer Distribution Tasks -> MySQL"""
 
+    @when_s3_available
     def test_answer_distribution_mysql(self):
         self.task.launch([
             'AnswerDistributionToMySQLTaskWorkflow',
--- a/edx/analytics/tasks/tests/acceptance/test_course_catalog.py
+++ b/edx/analytics/tasks/tests/acceptance/test_course_catalog.py
@@ -7,7 +7,7 @@ import logging
 
 import pandas
 
-from edx.analytics.tasks.tests.acceptance import AcceptanceTestCase
+from edx.analytics.tasks.tests.acceptance import AcceptanceTestCase, when_s3_available, when_vertica_available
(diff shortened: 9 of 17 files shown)


openedx-webhooks:
Thanks for the pull request, @itsjeyd! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-analytics-pipeline&number=204

edx/analytics/tasks/tests/acceptance/__init__.py
```diff
@@ -16,6 +16,9 @@
 
 log = logging.getLogger(__name__)
 
+SKIP_S3_TESTS = os.getenv('SKIP_S3_TESTS')
+SKIP_S3_TESTS_REASON = 'S3 is not available'
+
```
smarnach, line 78: How about exporting a decorator here instead of these constants?  Like

```
def when_s3_available(function):
    return unittest.skipIf(SKIP_S3_TESTS, 'S3 is not available')(function)
```

The tagged tests would read a bit more succinctly and to the point:

```
@when_s3_available
def test_…
```

instead of

```
@unittest.skipIf(SKIP_S3_TESTS, SKIP_S3_TESTS_REASON)
def test_
```

edx/analytics/tasks/tests/acceptance/test_enrollment_validation.py
```diff
@@ -4,13 +4,15 @@
 import gzip
 import json
 import logging
+import os
```
smarnach: Why this import?

scripts/run_acceptance.sh
```diff
@@ -0,0 +1,93 @@
...
+function usage
+{
+    echo "Usage: ./run_acceptance.sh [[[-t tests] [-u]] | [-h]]"
+}
+
+TESTS=
+UPDATE_VENV=
+
+while [ "$1" != "" ]; do
+    case $1 in
```
smarnach: You should use double quotes around `$1` here (like in the line above).

scripts/run_acceptance.sh
```diff
@@ -0,0 +1,93 @@
...
+        -t | --tests )
+            shift
+            TESTS=$1
```
smarnach: Same here.

scripts/run_acceptance.sh
```diff
@@ -0,0 +1,93 @@
...
+if [ ! -d "$TASKS_REPO" ]
+then
+    echo "Please clone edx-analytics-pipeline to your local machine and modify your Vagrantfile to mount it at ${TASKS_REPO}."
+    exit 1
+fi
+
+# Prepare environment as "vagrant" user
+if [ -n "$UPDATE_VENV" ]
+then
+    sudo -Hu vagrant /bin/bash<<-PREPARE
```
smarnach, line 52: `sudo` has a `-s` option meaning "run the current shell given in $SHELL (basically, the current shell) as that user", which could be used here.  (No need to change anything.  The option is handy for interactive use since it saves some typing, but you did the typing here already.)

smarnach:
@itsjeyd Thanks for the nice instructions!  I managed to follow all steps, and setting up the devstack went much smoother than last year. :)

Running the acceptance tests also worked fine, though the repo in `/var/lib/analytics-tasks/devstack/repo` was on the `origin/release` branch after following the instructions, so the new script wasn't there yet.  Manually adding the `open-craft` fork right there and switiching to your branch fixed this problem, and running the acceptance tests went fine, so :+1: after at least reading the few minor comments I had.

The whole setup is still a bit confusing.  Among other thing, there is definitely too big a number of checkouts of `edx-analytics-pipeline` involved, but that's kind of unrelated to this ticket.

itsjeyd:
@mulby @brianhw Please review.

mulby:
A few general thoughts:
- The intent of the JSON blob config file was to enable lots of parameters to be passed into the acceptance test framework. Is there any way we could just rely on that instead of a combination of environment variables and the JSON configuration? Perhaps we just hardcode a location for the geolocation file and in our logic, check if the file exists?
- As for the S3 disabling logic, could we detect this automatically? Not sure if boto has a way to basically say "check if I have a valid connection to S3", I suppose we could do something like list buckets?
- I would like to unify your bash script with the logic already in the Makefile. At a minimum define a new target in the Makefile and clarify the difference between the two.
- I love the decorators! Could you create one for Vertica and the exporter as well and roll that out to the tests that require them? I want to have a consistent pattern for skipping tests.

Great work! Thanks for helping us make this better for everyone.

itsjeyd:
Thanks @mulby for the comments!

> The intent of the JSON blob config file was to enable lots of parameters to be passed into the acceptance test framework. Is there any way we could just rely on that instead of a combination of environment variables and the JSON configuration? Perhaps we just hardcode a location for the geolocation file and in our logic, check if the file exists?

Providing a sensible default for the location of the geolocation file is a good idea I think. Where do you normally store it?

Beyond that the mechanism for specifying the path to a geolocation file on the command line might be worth keeping around as a convenience. (I thought implementing something like this was what you had in mind when you suggested to "provide a mechanism for users to download the file themselves and provide it to the test", but I might have misinterpreted :)) But let me know, I'll remove it if you think it's not useful.

> I would like to unify your bash script with the logic already in the Makefile. At a minimum define a new target in the Makefile and _clarify the difference between the two_.

Do you mean clarify the difference between a new make target for acceptance tests and the existing make target for unit tests ("test-local")? If you could elaborate a bit on what you have in mind here, that would be great.

> I love the decorators! Could you create one for Vertica and the exporter as well and roll that out to the tests that require them? I want to have a consistent pattern for skipping tests.

Thanks! Done :)

mulby:
> Where do you normally store it?

I haven't actually ever run them on devstack before... maybe somewhere in /var/tmp?

> Beyond that the mechanism for specifying the path to a geolocation file on the command line might be worth keeping around as a convenience. (I thought implementing something like this was what you had in mind when you suggested to "provide a mechanism for users to download the file themselves and provide it to the test", but I might have misinterpreted :)) But let me know, I'll remove it if you think it's not useful.

I'm not convinced it's worth the effort if we just document where they should put the file. I don't feel strongly about this as long as the code only has to look one place for the path (in the JSON blob), if the wrapper re-wrote the JSON before passing it into python, that would work for me too.

> Do you mean clarify the difference between a new make target for acceptance tests and the existing make target for unit tests ("test-local")? If you could elaborate a bit on what you have in mind here, that would be great.

I actually meant the "test-acceptance" target.

itsjeyd:
@mulby 

> I don't feel strongly about this as long as the code only has to look one place for the path (in the JSON blob), ...

OK, I removed the option for specifying the path to the geolocation file on the command line, and added a default path to the JSON blob over on https://github.com/edx/configuration/pull/2790.

> As for the S3 disabling logic, could we detect this automatically? Not sure if boto has a way to basically say "check if I have a valid connection to S3", I suppose we could do something like list buckets?

Doing something like this in `@when_s3_available` might work:

``` python
def when_s3_available(function):
    s3_available = True
    import boto
    connection = boto.connect_s3()  # This will not error out if
                                    # AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY are set,
                                    # so it can't be used to check if we have a valid connection to S3
    try:
        connection.get_all_buckets()
    except boto.exception.S3ResponseError:
        s3_available = False
    return unittest.skipIf(
        not s3_available, 'S3 is not available'
    )(function)
```

With this in place, annotated tests are skipped on Analytics Devstack. Assuming that the Jenkins environment uses valid/non-dummy values for `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` (and any other relevant settings), `get_all_buckets` should not error out in that environment (allowing the tests to run). Is there any chance you could verify this on your end?

smarnach:
@itsjeyd Two minor comments about your proposed decorator:
1. The way it is written, it will try to connect to S3 and list all buckets for every single usage of the decorator.  It would make sense to do the test only once, and cache the result in a global variable (e.g. in `when_s3_available._have_s3`).
2. I suspect that `boto.connect_s3()` can fail, e.g. when the access key and id are not set at all (not sure).  If it can fail, it should also be enclosed in a try/except block.

itsjeyd:
@smarnach Thanks @smarnach for the feedback on the changes I proposed for `@when_s3_available`!

@mulby I added a new make target to the Makefile (cf. most recent commit) As for clarifying the difference between it and the existing target (test-acceptance): Were you using the existing target to run tests in a Jenkins environment? Or, more generally, what use cases does it cover that go beyond running acceptance tests locally?

mulby:
@itsjeyd - correct, that target is used by jenkins. However, the Jenkins job also does a bunch of stuff before it calls that target (setting up the exporter etc). My gut says we should have at least two targets, one for devstack/local and one that is more explicitly for jenkins.

itsjeyd:
@mulby OK thanks.

I added information about running acceptance tests to the README.

The only item from your original feedback that hasn't been resolved completely is the question of whether or not we can detect S3 availability automatically. Could you please confirm whether the check I proposed in [this comment](https://github.com/edx/edx-analytics-pipeline/pull/204#issuecomment-188212477) would pass in a Jenkins environment?

mulby:
@itsjeyd - verified. I did the following:
- created an empty jenkins job
- added a virtualenv build step that just runs "pip install boto==2.22.1"
- added a python step that runs

``` python
s3_available = True
import boto
connection = boto.connect_s3()
try:
    connection.get_all_buckets()
except boto.exception.S3ResponseError:
    s3_available = False
print s3_available
```
- Ran the job. It printed "True" to the console.

itsjeyd:
@mulby Thanks! I updated `when_s3_available` to use the check (with caching).

All your original comments have now been addressed, and I also rebased both PRs for this feature (https://github.com/edx/configuration/pull/2790 and this one). Let me know if there's anything else.

scripts/run_acceptance.sh
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env bash
```
mulby, line 1: I'm nervous this file actually confuses things more than it helps. I'm trying to move away from the situation where you need two copies of the pipeline (one to manage the other) in order to run stuff.

I think I would prefer something like this:
- Assume that there is a working pipeline that has been setup with `remote-task` somewhere
- cd into that directory and activate the virtualenv
- run something like "make test-acceptance" which would, in turn, depend on `develop` and `test-requirements`
- alternately you could run "make test-acceptance-local" which would only depend on `develop-local`

Perhaps we could have a small wrapper script for the devstack that does all of the above:

``` bash
cd /var/lib/analytics-tasks-devstack/repo
. ../venv/bin/activate
make test-acceptance
```

Then the rest of the logic could live in the Makefile. I'm envisioning something like this:

```
REMOTE_TASK=$(shell which remote_task) LUIGI_CONFIG_PATH='config/test.cfg' ACCEPTANCE_TEST_CONFIG="/var/tmp/acceptance.json" python -m coverage run --rcfile=./.coveragerc -m nose --nocapture --with-xunit -A acceptance $(ONLY_TESTS)
```

mulby:
Generally looking great. Just one concern about the Makefile/shell script.

scripts/run_acceptance.sh
itsjeyd, line 1: @mulby I updated the script and Makefile; the roles are now reversed. Instead of the Makefile calling the script, the script now calls `make` to run tests (and the actual test logic lives in the Makefile). To run the entire test suite, you can do `./run_acceptance.sh` from the `scripts` folder, and to run specific tests you can do `./run_acceptance.sh --tests test_course_catalog`.

I did not add `develop-local` as a dependency to `test-acceptance-local` because I think it's more convenient to be able to skip updating the virtualenv (at least for the purpose of running acceptance tests locally). This is the default behavior; if you want to force an update you can call the script with `--update`.

scripts/run_acceptance.sh
```diff
@@ -0,0 +1,84 @@
...
+# Environment
+
+BASE_DIRECTORY=/var/lib/analytics-tasks/devstack
+VENV_DIRECTORY=$BASE_DIRECTORY/venv
+WORKING_DIRECTORY=$BASE_DIRECTORY/repo
+
+TASKS_REPO=/edx/app/analytics_pipeline/analytics_pipeline
+
+# Check if repo has been cloned and mounted at default location
+if [ ! -d "$TASKS_REPO" ]
```
mulby, line 43: Is this check necessary? (also the very existence of the TASK_REPO variable)

scripts/run_acceptance.sh
```diff
@@ -0,0 +1,84 @@
...
+            exit 1
+    esac
+    shift
+done
+
+# Environment
+
+BASE_DIRECTORY=/var/lib/analytics-tasks/devstack
+VENV_DIRECTORY=$BASE_DIRECTORY/venv
+WORKING_DIRECTORY=$BASE_DIRECTORY/repo
```
mulby: Are all of these variables still used?

scripts/run_acceptance.sh
```diff
@@ -0,0 +1,84 @@
...
+
+# Prepare virtualenv for testing
+make develop-local
+make test-requirements
+PREPARE
+fi
+
+# Tests
+
+if [ -n "$TESTS" ]
```
mulby: This works differently than ONLY_TESTS which I suspect could be confusing. The name also implies that you can specify multiple tests but the logic seems to imply that you should only be specifying one.

scripts/run_acceptance.sh
```diff
@@ -0,0 +1,84 @@
...
+# Run tests as "hadoop" user
+sudo -Hu hadoop /bin/bash<<-RUN
+cd $WORKING_DIRECTORY
+
+# Activate virtualenv
+source ../venv/bin/activate
+
+# Export AWS keys
+export AWS_ACCESS_KEY_ID='AK123'
+export AWS_SECRET_ACCESS_KEY='abc123'
```
mulby: Why AWS keys here?

scripts/run_acceptance.sh
mulby: I should clarify, since ONLY_TESTS is passed directly in to nose you can actually specify multiple tests and there are a bunch of different formats you can use to "choose" which tests get run.

scripts/run_acceptance.sh
mulby, line 1: I generally like the change. I am still wary of putting logic outside of the Makefile. It continues to be unclear to me where this would be run from.

It seems like you would run it from "/edx/app/analytics_pipeline/analytics_pipeline", but it "targets" the other pipeline installation on the box in /var/lib/. This is confusing to me and in the long term I hope to only have 1 version of the pipeline installed on the box and have everything run from it. Does that make sense?

itsjeyd, line 1: @mulby The script would be run from `/var/lib/analytics-tasks-devstack/repo/scripts`. For now -- once we address [OLIVE-22](https://openedx.atlassian.net/browse/OLIVE-22) ("Make it so pipeline is installed like any other component"), it should be possible to run it from `/edx/app/analytics_pipeline/analytics_pipeline/scripts`, assuming that the pipeline will get its own venv that is equivalent to the venv for `/var/lib/analytics-tasks-devstack/repo` during provisioning.

The idea behind the current setup is the following: `/edx/app/analytics_pipeline/analytics_pipeline/` is the default `TASKS_REPO` that code is checked out from when running acceptance tests (cf. [acceptance.json](https://github.com/edx/configuration/pull/2790/files#diff-ef286cd07b3ec9b07b2dec4c950a0951)). This makes it possible to work on pipeline code locally and test it without having to push it to a remote repository first. Since `TASKS_BRANCH` is set to `HEAD` by default, the code to test can also live on any branch. The only overhead is that you might need to check out a different branch to `/var/lib/analytics-tasks-devstack/repo/` if the tests you want to run don't live on the currently checked out branch. But this is again only temporary. Once `/edx/app/analytics_pipeline/analytics_pipeline` becomes the default (and only) location for the pipeline, the test code and the code to test will automatically come from the same branch.

---

I updated the documentation to reflect the changes from https://github.com/open-craft/edx-analytics-pipeline/commit/d2f487c9ff27d4adea1b95a77e8ba8041e2be6df.

scripts/run_acceptance.sh
itsjeyd, line 43: @mulby For the setup described [above](https://github.com/edx/edx-analytics-pipeline/pull/204#discussion_r55066739), yes.

mulby, line 43: Oh, I understand now. Thanks for the clarification.

Makefile
```diff
@@ -49,6 +49,9 @@ test: test-requirements develop test-local
 test-acceptance: test-requirements
 	LUIGI_CONFIG_PATH='config/test.cfg' python -m coverage run --rcfile=./.coveragerc -m nose --nocapture --with-xunit -A acceptance $(ONLY_TESTS)
 
+test-acceptance-local:
+	AWS_ACCESS_KEY_ID='AK123' AWS_SECRET_ACCESS_KEY='abc123' REMOTE_TASK=$(shell which remote-task) LUIGI_CONFIG_PATH='config/test.cfg' ACCEPTANCE_TEST_CONFIG="/var/tmp/acceptance.json" python -m coverage run --rcfile=./.coveragerc -m nose --nocapture --with-xunit -A acceptance --stop -v $(ONLY_TESTS)
```
mulby: heh - sorry to keep asking about this... I still don't understand why we define AWS access keys here?

scripts/run_acceptance.sh
mulby, line 1: OK, I think I understand now.

The only part of this that is still counter-intuitive to me is that the virtualenv for /var/lib/analytics-tasks/devstack will be used to run the code in /edx/app/analytics_pipeline/analytics_pipeline.

I think this is a good step forward, though. We can always iterate later.

mulby:
@brianhw can you give this a quick review as well?

Makefile
itsjeyd: @mulby When I started working on this, a few tests (that turned out not to depend on S3) were failing with something similar to this:

```
  File "/var/lib/analytics-tasks/devstack/venv/local/lib/python2.7/site-packages/boto/utils.py", line 219, in retry_url
    r = opener.open(req)
  File "/usr/lib/python2.7/urllib2.py", line 431, in open
    response = self._open(req, data)
  File "/usr/lib/python2.7/urllib2.py", line 449, in _open
    '_open', req)
  File "/usr/lib/python2.7/urllib2.py", line 409, in _call_chain
    result = func(*args)
  File "/usr/lib/python2.7/urllib2.py", line 1227, in http_open
    return self.do_open(httplib.HTTPConnection, req)
  File "/usr/lib/python2.7/urllib2.py", line 1197, in do_open
    raise URLError(err)
URLError: <urlopen error timed out>
boto: ERROR: Unable to read instance data, giving up
```

Setting `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` before running these tests fixed the problem. However, as of today I'm unable to reproduce the problem, so I removed the definitions from the Makefile (cf. https://github.com/open-craft/edx-analytics-pipeline/commit/5857e648ed516638bb42ea6c9eaff1347b93df43).

share/task.yml
```diff
@@ -82,7 +83,7 @@
         - "{{ log_dir }}"
 
     - name: analytics tasks repository checked out
-      git: repo={{ repo }} dest={{ working_repo_dir }} version=release
+      git: repo={{ repo }} dest={{ working_repo_dir }} version={{ remote }}/{{ branch }}
```
mulby: Is this change necessary? it would require quite a bit of rework to our Jenkins jobs...

itsjeyd: @mulby Yes, please see the commit message of https://github.com/open-craft/edx-analytics-pipeline/commit/f6a2ca5d38ec72cc2129a652db356a572c43c3d1 for the reasoning.

mulby: What if we just checked out a bare copy of the repo? I think that would solve both of our issues... it would remove the local copy of the release branch and still allow me to pass in "origin/foo" for the branch parameter.

README.md
```diff
@@ -65,6 +68,111 @@ valid credentials for the tests to pass, so the commands below should fix the fa
...
+### Acceptance tests
+
+**Preconditions**
+
+To be able to run acceptance tests on an instance of the Analytics
+Devstack, the following conditions must hold:
+
+- `/var/lib/acceptance.json` is present on your instance of Analytics
+  Devstack. This file contains the default configuration for running
+  acceptance tests.
```
brianhw: Having this documentation is great!  Is it possible to offer some guidance here as to where a user might find a template or example for this file?

brianhw:
One optional doc comment, otherwise this LGTM. :+1:

mulby:
:+1: from me as well. Ready to squash and merge :)

README.md
itsjeyd: @brianhw I added a link to the default version of acceptance.json to the README. The file is being introduced in the companion PR (https://github.com/edx/configuration/pull/2790) for this, so the link will work once that PR is merged.

itsjeyd:
@mulby @brianhw Thanks for reviewing! Could you please also have a look at/add thumbs-up to the companion PR (https://github.com/edx/configuration/pull/2790) for this?

If possible, I'd like to merge the PRs at the same time.

README.md
```diff
@@ -65,6 +68,111 @@ valid credentials for the tests to pass, so the commands below should fix the fa
...
 
+### Acceptance tests
+
+**Preconditions**
+
+To be able to run acceptance tests on an instance of the Analytics
+Devstack, the following conditions must hold:
+
+- [`/var/lib/acceptance.json`](https://github.com/edx/configuration/blob/master/playbooks/roles/analytics_pipeline/files/acceptance.json)
+  is present on your instance of Analytics Devstack. This file
```
brianhw, line 79: Thanks for adding the pointer.  I discovered that when reviewing the other PR.  

But I just noticied that this recommends putting the file in /var/lib.  However, the other PR puts it in /var/tmp, and the makefile expects it in /var/tmp as well.

brianhw:
Just the one correction in the documentation, then :+1:

share/task.yml
itsjeyd: @mulby I made a mistake yesterday when testing the approach you suggested (forgot to update the venv, so modifications from https://github.com/open-craft/edx-analytics-pipeline/commit/33537a51a29ec6d4f90cb89d1630d80087a7bbdb weren't being used). It turns out that when replacing `version=release` with `bare=yes`, the playbook fails because the `branch checked out` task needs a work tree to succeed:

```
TASK: [branch checked out] **************************************************** 
<localhost> ESTABLISH CONNECTION FOR USER: hadoop
<localhost> EXEC ['ssh', '-tt', '-vvv', '-o', 'ForwardAgent=yes', '-o', 'UserKnownHostsFile=/dev/null', '-o', 'StrictHostKeyChecking=no', '-o', 'Port=22', '-o', 'KbdInteractiveAuthentication=no', '-o', 'PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey', '-o', 'PasswordAuthentication=no', '-o', 'ConnectTimeout=10', 'localhost', "/bin/sh -c 'mkdir -p $HOME/.ansible/tmp/ansible-1457430542.22-182101216527929 && chmod a+rx $HOME/.ansible/tmp/ansible-1457430542.22-182101216527929 && echo $HOME/.ansible/tmp/ansible-1457430542.22-182101216527929'"]
<localhost> REMOTE_MODULE command git checkout HEAD chdir=/var/lib/analytics-tasks/2c0a5e8921aa235b148da415269e91c9/repo
<localhost> PUT /tmp/tmpHN8Fdg TO /edx/app/hadoop/.ansible/tmp/ansible-1457430542.22-182101216527929/command
<localhost> EXEC ['ssh', '-tt', '-vvv', '-o', 'ForwardAgent=yes', '-o', 'UserKnownHostsFile=/dev/null', '-o', 'StrictHostKeyChecking=no', '-o', 'Port=22', '-o', 'KbdInteractiveAuthentication=no', '-o', 'PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey', '-o', 'PasswordAuthentication=no', '-o', 'ConnectTimeout=10', 'localhost', "/bin/sh -c '/usr/bin/python /edx/app/hadoop/.ansible/tmp/ansible-1457430542.22-182101216527929/command; rm -rf /edx/app/hadoop/.ansible/tmp/ansible-1457430542.22-182101216527929/ >/dev/null 2>&1'"]
failed: [localhost] => {"changed": true, "cmd": ["git", "checkout", "HEAD"], "delta": "0:00:00.001818", "end": "2016-03-08 09:49:03.546037", "rc": 128, "start": "2016-03-08 09:49:03.544219"}
stderr: fatal: This operation must be run in a work tree

FATAL: all hosts have already failed -- aborting

PLAY RECAP ******************************************************************** 
           to retry, use: --limit @/edx/app/hadoop/task.retry

localhost                  : ok=8    changed=4    unreachable=0    failed=1   

Exiting with status = 2
```

I reverted the changes. If you have any other suggestions that would help avoid making changes to Jenkins jobs, please let me know. Otherwise this should be good to go (addressed Brian's pointers about the documentation and squashed the commits).

README.md
itsjeyd, line 79: @brianhw Thanks for catching that! Fixed.

share/task.yml
mulby: In that case, can we just revert these changes entirely?

I don't see any problem with having a ghost release branch sitting on the machine, and it will cause a huge amount of work on our end to go through all of our jenkins jobs and fix this.

Once we have the jobs checked in as DSL files, this type of change will be much easier to make :smile:

itsjeyd: @mulby 

> In that case, can we just revert these changes entirely?

That would be possible, at the cost of adding another manual setup step: `/edx/app/analytics_pipeline/analytics_pipeline` needs to have a local tracking branch for `origin/release`, otherwise the `analytics tasks repository checked out` task will fail:

```
TASK: [analytics tasks repository checked out] ******************************** 
<localhost> ESTABLISH CONNECTION FOR USER: hadoop
<localhost> EXEC ['ssh', '-tt', '-vvv', '-o', 'ForwardAgent=yes', '-o', 'UserKnownHostsFile=/dev/null', '-o', 'StrictHostKeyChecking=no', '-o', 'Port=22', '-o', 'KbdInteractiveAuthentication=no', '-o', 'PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey', '-o', 'PasswordAuthentication=no', '-o', 'ConnectTimeout=10', 'localhost', "/bin/sh -c 'mkdir -p $HOME/.ansible/tmp/ansible-1457461780.75-16575766786959 && chmod a+rx $HOME/.ansible/tmp/ansible-1457461780.75-16575766786959 && echo $HOME/.ansible/tmp/ansible-1457461780.75-16575766786959'"]
<localhost> REMOTE_MODULE git repo=/edx/app/analytics_pipeline/analytics_pipeline dest=/var/lib/analytics-tasks/2c0a5e8921aa235b148da415269e91c9/repo version=release
<localhost> PUT /tmp/tmpWy3sm0 TO /edx/app/hadoop/.ansible/tmp/ansible-1457461780.75-16575766786959/git
<localhost> EXEC ['ssh', '-tt', '-vvv', '-o', 'ForwardAgent=yes', '-o', 'UserKnownHostsFile=/dev/null', '-o', 'StrictHostKeyChecking=no', '-o', 'Port=22', '-o', 'KbdInteractiveAuthentication=no', '-o', 'PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey', '-o', 'PasswordAuthentication=no', '-o', 'ConnectTimeout=10', 'localhost', "/bin/sh -c '/usr/bin/python /edx/app/hadoop/.ansible/tmp/ansible-1457461780.75-16575766786959/git; rm -rf /edx/app/hadoop/.ansible/tmp/ansible-1457461780.75-16575766786959/ >/dev/null 2>&1'"]
failed: [localhost] => {"failed": true}
msg: Failed to checkout release

FATAL: all hosts have already failed -- aborting

PLAY RECAP ******************************************************************** 
           to retry, use: --limit @/edx/app/hadoop/task.retry

localhost                  : ok=6    changed=2    unreachable=0    failed=1   

Exiting with status = 2
```

Also, the default for `tasks_branch` in `acceptance.json` needs to be adjusted for this to work (`"HEAD"` ⟶ `"origin/HEAD"`).

Let me know if that's an acceptable trade-off for you; I'll make the necessary adjustments to this PR and https://github.com/edx/configuration/pull/2790 in that case.

(Automating the step of creating a local tracking branch for `origin/release` could be addressed in the context of [OLIVE-22](https://openedx.atlassian.net/browse/OLIVE-22), so the trade-off would likely be temporary.)

mulby: What if we just change the default branch to "origin/release"? And then procede with the checkout normally?

Not sure I understand why we have to change HEAD to origin/HEAD but I think I'd rather do that then change this code.

itsjeyd: @mulby 

> What if we just change the default branch to "origin/release"?

That doesn't make a difference (i.e., it produces the same error if `/edx/app/analytics_pipeline/analytics_pipeline` does not have a local `release` branch). This is because the copy of the pipeline that is checked out for the purpose of running tests doesn't "see" the remote tracking branches that `/edx/app/analytics_pipeline/analytics_pipeline` has. It only sees the local branches. To give a concrete example, a branch `foo` that is local to `/edx/app/analytics_pipeline/analytics_pipeline` is available to the working copy as `origin/foo`. (For the `git` module, `version=origin/release` is the same as `version=release` because the `remote` parameter -- which we're not customizing in the `analytics tasks repository checked out` task -- defaults to `origin`.)

> Not sure I understand why we have to change HEAD to origin/HEAD ...

If the version that the `analytics tasks repository checked out` task checks out (`release`) is different from the version that we want to run our tests on (the development branch that `/edx/app/analytics_pipeline/analytics_pipeline` is on), checking out `HEAD` in the `branch checked out` task that follows won't give the desired result (i.e., the current branch will stay on `release`).

---

To sum up, the following changes would make it possible for you to avoid having to modify Jenkins jobs:
1. Revert https://github.com/open-craft/edx-analytics-pipeline/commit/f6a2ca5d38ec72cc2129a652db356a572c43c3d1.
2. Update README to make it clear that `/edx/app/analytics_pipeline/analytics_pipeline` needs a local `release` branch.
3. Change value of `tasks_branch` to `origin/HEAD` in acceptance.json.

Can I go ahead with these (and merge afterwards)?

mulby: I think I now understand. I was thinking of "origin" as github, but it's actually the repo stored at `/edx/app/analytics_pipeline` apologies for the confusion.

What if we changed the default branch to `master`? Since that's the default branch I think it's likely that it will exist locally in origin.

We should definitely change `tasks_branch` to `origin/HEAD`.

OK with merging once we have this resolved.

itsjeyd: @mulby Changing the default branch to `master` works :) I did that instead of reverting https://github.com/open-craft/edx-analytics-pipeline/commit/f6a2ca5d38ec72cc2129a652db356a572c43c3d1 and updating the README. Adjusted `tasks_branch` in acceptance.json as well and squashed the commits for https://github.com/edx/configuration/pull/2790. The build for that PR is green. Good to go?

mulby:
@itsjeyd ready to merge once I run the acceptance tests on this branch on our jenkins box, which I'm kicking off now

itsjeyd:
@mulby OK thanks, let me know.

mulby:
@itsjeyd: The test run passed, this is ready to merge whenever you're ready.

Great work!

bradenmacdonald:
@itsjeyd has either just left for vacation or is about to, so I'm going to merge this for him.

Thanks for the great work on this @itsjeyd, and thanks for the review @mulby @brianhw @smarnach.

:tada:

itsjeyd:
@mulby @bradenmacdonald Thanks guys!

PR #205, opened 2016-02-19
ensure static data files are installed when developing

mulby:
@brianhw this should fix the jenkins build failure


brianhw:
LGTM. :+1:

PR #206, opened 2016-02-20
Fix acceptance tests

mulby:
The acceptance tests found a bug in my PR that would have prevented it from working on S3 files, I also needed to update the dummy tarball and fix the course structure comparison logic. The test appears to be passing now.

Reviewers: @brianhw @HassanJaveed84


brianhw:
Hooray for acceptance tests!   LGTM. :+1:

PR #207, opened 2016-02-24
Generate edx.analytics.task and parameter documentation with Sphinx

pomegranited:
This PR adds a custom `sphinx.ext.autodoc` plugin which appends `luigi.Parameter` attribute names and description values to the documentation generated for each `luigi.Task` class.

**JIRA tickets**: Implements [OC-1391](https://tasks.opencraft.com/browse/OC-1391), [OLIVE-16](https://openedx.atlassian.net/browse/OLIVE-16), [OLIVE-15](https://openedx.atlassian.net/browse/OLIVE-15)

**Discussions**: Implementation decisions came out of a discovery task document, created by @e-kolpakov, and reviewed by @mulby and @bradenmacdonald.

**Dependencies**: Sphinx 1.3.5

**Screenshots**:
[Generated docs.zip](https://github.com/edx/edx-analytics-pipeline/files/151665/docs.zip) **Notes** The attached docs.zip are the PDFs generated by merging the code from this PR with the description changes for PR #208.  They're also subject to print CSS, which removes the lovely tables of contents sidebar.

edx.analytics.pipeline home
workflow_entry_points
supporting_tasks
task index
module index

**Sandbox URL**: TBD 

**Testing instructions**: 
1. go to edx-analytics-pipeline 
2. make install 
3. make docs
4. open docs/index.html

Or check the attached PDF.

**Author notes and concerns**:
1. The Analytics Tasks section in `sphinx_source/index.rst` is automatically generated from the edx.analytics.tasks listed in `setup.cfg`, so any tasks missing from that list will not be documented.

**Reviewers**
- [] Code: (@e-kolpakov)
- [] Product: (TBD)
- [] Platform: (TBD)
- [] UX: (TBD)
- [] TNL: (TBD)
- [] a11y: (TBD)
- [] doc/strings: (TBD)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=207&repo=edx%2Fedx-analytics-pipeline

requirements/default.txt
```diff
@@ -11,7 +11,7 @@ isoweek==1.3.0		# BSD
 mechanize==0.2.5	# BSD
 http://cdn.mysql.com/Downloads/Connector-Python/mysql-connector-python-1.2.2.zip  	# GPL v2 with FOSS License Exception
 ndg-httpsclient==0.4.0  # BSD
-numpy==1.8.0 		# BSD
+git+https://github.com/numpy/numpy.git@maintenance/1.8.x#egg=numpy-1.8.x
```
pomegranited: I had to make this change because the pypi numpy package refused to build on Ubuntu.
Unsure whether it be included in this PR?

e-kolpakov: @pomegranited I believe I had the same issue - it was solved by reinstalling `libblas3` and `libblas-dev` packages. Give it a try.

pomegranited: @e-kolpakov Did the trick, thank you!  I've reverted the change.

setup.cfg
```diff
@@ -31,6 +31,7 @@ edx.analytics.tasks =
     course-enroll = edx.analytics.tasks.course_enroll:CourseEnrollmentChangesPerDay
     enrollments_and_registrations_workflow-manifest = edx.analytics.tasks.reports.enrollments_and_registrations_workflow_manifest:EnrollmentsandRegistrationsWorkflow
     answer-dist = edx.analytics.tasks.answer_dist:AnswerDistributionPerCourse
+    answer-dist-workflow = edx.analytics.tasks.answer_dist:AnswerDistributionWorkflow
```
pomegranited: Added these two classes because [OLIVE-16]() wanted them documented, however I'm unsure of the implications.
@mulby Is it ok to be changing this file, and to be using it to generate the list of tasks to document?

Also, the tasks are currently documented in the order they're presented in this file.  Would you rather they were alphabetized?

sphinx_source/luigi_sphinx.py
```diff
@@ -0,0 +1,52 @@
...
+                    param['default'] = membervalue.default
+                    param['type'] = u'{type}, optional'.format(**param)
+
+                    # Show default value, if not already in the description.
+                    # NB: This test is useful to avoid redundant descriptions,
+                    # and for dynamically determined defaults like date.today()
+                    if not default_re.search(param['description']):
+                        param['description'] = u'{description}\nDefault is {default}.'.format(**param)
+
+                # Mark insignificant parameters
```
pomegranited, line 55: @mulby Is it useful for the documentation to mark insignificant parameters? See attached docs.pdf `edx.analytics.tasks.reports.enrollments.EnrollmentsByWeek` for example.

edx/analytics/tasks/reports/total_enrollments.py
```diff
@@ -145,19 +145,7 @@ class WeeklyAllUsersAndEnrollments(luigi.Task, AllCourseEnrollmentCountMixin):
     """
     Calculates total users and enrollments across all (known) courses per week.
 
-    Parameters:
-        enrollments: Location of daily enrollments per date. The format is a
```
pomegranited, line 149: Note: `enrollments` doesn't appear to be a valid parameter anymore, so I've removed this comment.

edx/analytics/tasks/reports/incremental_enrollments.py
```diff
@@ -18,18 +18,7 @@ class WeeklyIncrementalUsersAndEnrollments(luigi.Task, AllCourseEnrollmentCountM
     """
     Calculates weekly incremental changes in users and enrollments across courses.
 
-    Parameters:
-        registrations: Location of daily registrations per date. The format is a
```
pomegranited, line 22: Note: `enrollments` and `registrations` doesn't appear to be a valid parameter anymore, so I've removed this comments.

edx/analytics/tasks/reports/incremental_enrollments.py
```diff
@@ -197,17 +186,7 @@ class DailyRegistrationsEnrollmentsAndCourses(luigi.Task,
     """
     Calculates users registration and total enrollments across courses.
 
-    Parameters:
-        registrations: Location of daily registrations per date. The format is a
```
pomegranited, line 201: Note: `enrollments` and `registrations` doesn't appear to be a valid parameter anymore, so I've removed this comments.

sphinx_source/luigi_sphinx.py
```diff
@@ -0,0 +1,52 @@
...
+                # Mark optional parameters
+                if hasattr(membervalue, 'default'):
+                    param['default'] = membervalue.default
+                    param['type'] = u'{type}, optional'.format(**param)
+
+                    # Show default value, if not already in the description.
+                    # NB: This test is useful to avoid redundant descriptions,
+                    # and for dynamically determined defaults like date.today()
+                    if not default_re.search(param['description']):
+                        param['description'] = u'{description}\nDefault is {default}.'.format(**param)
```
pomegranited: @mulby I'm including the parameter `default` value in the description, which is fine in most cases.  However, some params have default values set via `config_path`, which might result in confusion if people have their configuration defaults set to something different than what the documentation was generated with.

Is that a concern?  Is there a better way to denote this?

See `edx.analytics.tasks.answer_dist.AnswerDistributionWorkflow` `database` parameter for an example.  This parameter comes from `edx.analytics.tasks.mysql_load.MysqlInsertTaskMixin`.  The generated default value is `reports`, and is being pulled from `client.cfg`.

edx/analytics/tasks/reports/enrollments.py
```diff
@@ -156,17 +183,7 @@ def select_weekly_values(self, daily_values, start, weeks):
...
-            tsv file, with fields course_id, date and count.
-        destination: Location of the resulting report. The output format is a
-            excel csv file with course_id and one column per requested week.
-        offsets: Location of seed values for each course. The format is a
-            hadoop tsv file, with fields course_id, date and offset.
-        date: End date of the last week requested.
-        weeks: Number of weeks from the end date to request.
-
-    Output:
+    Returns:
```
pomegranited: Changed heading to Returns, because Output isn't a Sphinx keyword, so it doesn't get highlighted.

edx/analytics/tasks/reports/incremental_enrollments.py
```diff
@@ -18,18 +18,7 @@ class WeeklyIncrementalUsersAndEnrollments(luigi.Task, AllCourseEnrollmentCountM
...
-            TSV file, with fields date and count.
-        enrollments: Location of daily enrollments per date. The format is a
-            TSV file, with fields course_id, date and count.
-        destination: Location of the resulting report. The output format is an
-            excel-compatible CSV file.
-        date: End date of the last week requested.
-        weeks: Number of weeks from the end date to request.
-
-    Output:
```
pomegranited, line 31: Changed heading to Returns, because Output isn't a Sphinx keyword, so it doesn't get highlighted.

edx/analytics/tasks/answer_dist.py
```diff
@@ -867,8 +869,12 @@ class AnswerDistributionWorkflow(
     """Calculate answer distribution and output to files and to database."""
 
     # Add additional args for MultiOutputMapReduceJobTask.
-    output_root = luigi.Parameter()
-    marker = luigi.Parameter()
+    output_root = luigi.Parameter(
```
pomegranited: No comments were available for the `output_root` and `marker` descriptions, so I made them up using other similar parameters.

pomegranited: No comments were available for the `output_root` and `marker` descriptions, so I made them up using other similar parameters.

edx/analytics/tasks/mapreduce.py
```diff
@@ -28,19 +28,31 @@ class MapReduceJobTaskMixin(object):
 
     mapreduce_engine = luigi.Parameter(
         config_path={'section': 'map-reduce', 'name': 'engine'},
-        significant=False
+        significant=False,
+        description='Name of the map reduce job engine to use.  Use `hadoop` (the default) or `local`.',
```
pomegranited: Made up the `mapreduce_engine` description from [UsersPerCountryReportWorkflow, L339](https://github.com/edx/edx-analytics-pipeline/pull/207/files#diff-0438beb1d074451c5190ecc525872b54L339)

edx/analytics/tasks/mapreduce.py
```diff
@@ -28,19 +28,31 @@ class MapReduceJobTaskMixin(object):
...
-    lib_jar = luigi.Parameter(is_list=True, default=[], significant=False)
-
-    # Override the parent class definition of this parameter. This typically wants to scale with the cluster size so the
-    # user should be able to tweak it depending on their particular configuration.
-    n_reduce_tasks = luigi.Parameter(default=25, significant=False)
-
+    input_format = luigi.Parameter(
+        default=None,
+        significant=False,
+        description='The input_format for Hadoop job to use. For example, when '
```
pomegranited: Made up the `input_format` description from [UsersPerCountryReportWorkflow, L340](https://github.com/edx/edx-analytics-pipeline/pull/207/files#diff-0438beb1d074451c5190ecc525872b54L340)

edx/analytics/tasks/mapreduce.py
```diff
@@ -28,19 +28,31 @@ class MapReduceJobTaskMixin(object):
...
+        'running with manifest file, specify "oddjob.ManifestTextInputFormat" for input_format.',
+    )
+    lib_jar = luigi.Parameter(
+        is_list=True,
+        default=[],
+        significant=False,
+        description='An list of library jars that the Hadoop job can make use of.',
+    )
```
pomegranited: Made up the `lib_jar` description from [MapReduceJobRunner L130](https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/mapreduce.py#L130)

edx/analytics/tasks/util/hive.py
```diff
@@ -25,7 +25,8 @@ class WarehouseMixin(object):
     """Task that is aware of the data warehouse."""
 
     warehouse_path = luigi.Parameter(
-        config_path={'section': 'hive', 'name': 'warehouse_path'}
+        config_path={'section': 'hive', 'name': 'warehouse_path'},
+        description='A URL location of the data warehouse.',
```
pomegranited: Made `warehouse_path` description up completely, no idea if it's correct.

edx/analytics/tasks/util/hive.py
```diff
@@ -375,8 +376,14 @@ class HiveTableFromParameterQueryTask(HiveTableFromQueryTask):  # pylint: disabl
...
 
-    overwrite = luigi.BooleanParameter(default=True)  # Overwrite the MySQL data?
-    hive_overwrite = luigi.BooleanParameter(default=False)
+    overwrite = luigi.BooleanParameter(
+        default=True,
+        description='If True, overwrite the MySQL data.',
+    )
+    hive_overwrite = luigi.BooleanParameter(
+        default=False,
+        description='If True, overwrite the hive data.',
```
pomegranited: Made `hive_overwrite` description up completely, no idea if it's correct.

mulby:
@pomegranited - this is a huge step forward, and shaping up nicely. I need to look at it more closely.

One thing that jumped out at me, however, is that the classes listed in setup.cfg are not all of the tasks. We simply need to include one task from each module in order for luigi to detect all of the tasks in that module, so the list is far from comprehensive.

You can use stevedore to enumerate all of the loaded classes, see this snippet:

https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/launchers/local.py#L45

Note that we do distinguish, to some extent, between tasks that we expect users to run directly and tasks that are somewhat internal and only relied upon by other tasks in that workflow. For example, I would be surprised if someone ran the Sqoop import task directly - although this is possible.

Perhaps there is somewhat to annotate the tasks we expect people to actually run? The docs should either only include those tasks, or separate them so that it's really clear they should be the focus of the user.

Does that make sense?

Overall this looks great! Hopefully I can dig into this more comprehensively today or tomorrow.

mulby:
@lamagnifica - getting this on to your radar. We are planning on auto-generating RST based documentation for data pipeline tasks. I would like to eventually run this automatically from readthedocs and link to it from docs.edx.org (somewhere).

pomegranited:
@mulby That makes sense why `setup.cfg`'s list of classes isn't complete!  And I hear that we don't just want to show all the Tasks, as this will confuse users.  So we either need to differentiate the useful Tasks from the less-useful ones, or only show the useful Tasks.

@lamagnifica Where do you stand on the above question?  Should we risk confusing users with by presenting documentation of Tasks they may not need, or is it better to give them a full picture, with the most useful Tasks split out above?

As far as the implementation goes, I think a `@decorator` annotation is the best approach.  Any time people have to depart from the source code to add documentation, it becomes less likely that the documentation will be maintained.  And that's the whole point of adding Sphinx support anyway!

So I'll add in a couple of `@decorators` to this PR, pick out the Task classes in OLIVE-16 as the top-level Important Tasks, and leave the rest as Supporting Tasks.  If you could provide a full list of Important Tasks, then I'll make sure they get annotated.

Thank you for the clarifications!  Looking forward to working with you both.

mulby:
Sounds like a good plan.

pomegranited:
@mulby The updated `make docs-local` command generates documentation for two lists of classes from `edx.analytics.tasks`:
- (Main) Analytics Tasks – classes decorated with `@edx.analytics.tasks.decorators.sphinx_main_task`.
- Supporting Tasks – classes decorated with `@edx.analytics.tasks.decorators.sphinx_helper_task`.

Another option is to filter the Supporting Tasks list using `luigi.task.Task`, which would include every task.  That way, you wouldn't have to mark all the `@sphinx_helper_task`s, just the `@sphinx_main_task`s. 

I've only marked Main Tasks from OLIVE-16, and `SqoopImportTask` as a Supporting Task, so you can see how it looks. I've updated `gen_tasks.py` to take a list of `--output` arguments `*.inc` file names, and a corresponding list of `--filter` arguments, which are optional subclass filters for the generated task lists.  So if we want more than 2 lists, or to just go back to 1 list, it's just a config change.

If you could please provide the following information, I'll get the PR updated and ready to go:
- List of Main Task classes
- List of Helper Task classes (or `luigi.task.Task` for all tasks)
- Any text changes or additions you'd like in the generated document.

It's my Friday night now, so I won't be working much until [Monday morning](http://www.timeanddate.com/worldclock/fixedtime.html?msg=Monday+morning&iso=20160229T09&p1=5), but if there's anything urgent, please let me know, and I'll get online and reply.

P.S. _Unfortunately I wasn't able to use `stevedore` to iterate through the available tasks, because it's is limited to the installed classes listed in the `setup.cfg` entry_points list.  So I do some python inspection magic to find the list of classes available in the `edx.analytics.tasks` module, and to filter out the relevant subclasses.  So the code isn't as pretty as it was, but at least it does the job :smile:._

mulby:
@pomegranited

As far as the dynamic loading of the classes, I think I found the way that luigi does it internally.

Try something like this:

``` python
from luigi.task import Register
import stevedore

# This just ensures that the dynamically loaded modules are loaded into memory
stevedore.ExtensionManager('edx.analytics.tasks')

# Once those modules are parsed, luigi's internal Register now knows about all of the tasks and can be used to retrieve them
print Register.get_reg().items()
```

pomegranited:
Omg that's so much cleaner.. Will put that in ASAP. Thank you!

mulby:
As for the pydoc.locate issue. I would recommend just annotating the classes with a hidden attribute instead of extending the set of base classes. I've seen that pattern used fairly frequently. Maybe something like this:

``` python
def workflow_entry_point(cls):
    cls._category = 'workflow_entry_point'
    return cls
```

And then you can filter like this:

``` python
class_categories = defaultdict(list)
for cls in all_task_classes:
    category = getattr(cls, '_category')
    if category:
        class_categories[category].append(cls)
```

A side note - I like the idea of these decorators so much I'm thinking that we could extend them to be used in other ways. For example, when I run "launch-task --help" it would be really great if I could _just_ see the "main" classes. Perhaps if I ran something else like "launch-task --help-all" it could show the other classes.

For the above reason I would advocate for naming them more generically. Something like "workflow_entry_point"... open to other ideas.

I'm not sure there is value in defining three categories at this point ("workflow entry point", "helper", everything else). I'd be fine with just "workflow entry point" and "everything else" initially. We can always add more categories later if we wanted to. If it's simple to do so, I would recommend writing the code to support N categories even if we only start with 1.

Great work on this so far!

e-kolpakov:
@pomegranited One suggestion - can we split commits that add descriptions into a dedicated PR - it is very hard to review this, as those "added descriptions" changes clutter the overall result.

e-kolpakov:
@pomegranited nit: lots of SEVERE errors in console:

```
edX/edx-analytics-pipeline/sphinx_source/index.rst:7: SEVERE: Duplicate ID: "module-edx.analytics.tasks.answer_dist".
edX/edx-analytics-pipeline/sphinx_source/index.rst:7: SEVERE: Duplicate ID: "module-edx.analytics.tasks.answer_dist".
edX/edx-analytics-pipeline/sphinx_source/index.rst:7: SEVERE: Duplicate ID: "module-edx.analytics.tasks.answer_dist".
edX/edx-analytics-pipeline/sphinx_source/index.rst:7: SEVERE: Duplicate ID: "module-edx.analytics.tasks.answer_dist".
```

Might make sense to group tasks by modules to avoid this. Alternatively, use full-qualified name insteam of `..module:` directive, i.e. `edx.analytics.tasks.answer_dist.AnswerDistributionPerCourse`

e-kolpakov:
@pomegranited nit: resulting doc is a bit hard to navigate - it basically dumps all the classes on the same page - it might make sense to split it into multiple pages (with hyperlinks between them), i.e. 
- Doc root
  - Main tasks
  - Support tasks
    - Support task category (or module)
    - Support task category (or module)

pomegranited:
@e-kolpakov The easiest fix to the "SEVERE: Duplicate ID:" warnings is to fully-qualify the task name and don't use ..module:, but then no Module Index page gets generated.  There was a link to that in your example index.rst.. Do you know if it is required?

e-kolpakov:
@pomegranited I'm not sure if it's required, but I believe it you could just group them together:

```
.. module:: edx.analytics.tasks.answer_dist
.. autoclass:: AnswerDistributionOneFilePerCourseTask
.. autoclass:: AnswerDistributionPerCourse
.. autoclass:: AnswerDistributionToMySQLTaskWorkflow
.. autoclass:: BaseAnswerDistributionTask
```

It will still likely throw warnings if there are multiple `..module::` directives in different "included" files, but this could be solved by splitting them into actual pages.

e-kolpakov:
@pomegranited includes some luigi tasks - from module index:

```
luigi   
    luigi.hadoop    
    luigi.hive  
    luigi.interface 
    luigi.task  
    luigi.util
```

I don't think they are needed here, as they are (or at least, should) be documented in luigi docs.

e-kolpakov:
@pomegranited nit: module index might need a custom style - `edx.analytics.tasks.reports.enrollments_and_registrations_workflow_manifest` does not fit into one line, get wrapped and as a result it looks like it's a new top-level module for the tasks down below.

image

Also, maybe we should change the entrypoint so that moduleindex is built from `edx.analytics.tasks` right away? I believe it would significantly improve it's usability:

image

Hint:  

```
# in conf.py
sys.path.insert(0, local_path('../edx/analytics/tasks')) 
# in gen_tasks.py
out['fp'].write("\n.. module:: {module}".format(module=cls.__module__.replace('edx.analytics.tasks.', '')))
```

pomegranited:
@e-kolpakov I've applied your suggestions, and it's looking much better, thank you for your suggestions! 
I'm working on making this extendable to any number of categories they might want to create.

e-kolpakov:
@pomegranited a couple of final suggestions:
1. Might make sense to output module docstring; not sure if it is easy with current schema (i.e. module + autoclass), but it is totally possible with `automodule` + `members` (i.e. each `autoclass` record becomes `member` of `automodule`)
2. Even if module docstings are not needed, it might make sense to visually separate modules from each other, i.e. something like

```
Module course_catalog
---------------------
.. module:: course_catalog
.. autoclass:: CourseCatalogWorkflow
.. autoclass:: DailyLoadSubjectsToVerticaTask
.. autoclass:: DailyProcessFromCatalogSubjectTask
.. autoclass:: DailyPullCatalogTask

Module course_enroll
--------------------

.. module:: course_enroll
.. autoclass:: CourseEnrollmentChangesPerDay
.. autoclass:: CourseEnrollmentEventsPerDay
```

pomegranited:
@e-kolpakov I think I've done everything you mentioned..
- Duplicate ID errors were fixed by separating out each category into separate files, and grouping the classes by module.
- Removed the edx.analytics.tasks prefix, which makes the indexes awesome.
- Filtered out the luigi tasks when I removed the edx.analytics.tasks prefix
- Added custom template + stylesheet to force wrap long module names
- Using automodule to output module docstrings
- Added module headings, which makes the page sidebar TOC really nice too.

Note: I was getting these warnings, but they disappeared when I merged in `jill/pipeline-docs-descriptions` from PR #208.

```
reading sources... [100%] workflow_entry_point

/edx/src/edx-analytics-pipeline/edx/analytics/tasks/answer_dist.py:docstring of answer_dist.AnswerDistributionOneFilePerCourseTask:9: WARNING: Definition list ends without a blank line; unexpected unindent.

/edx/src/edx-analytics-pipeline/edx/analytics/tasks/user_location.py:docstring of user_location.UsersPerCountryReportWorkflow:8: WARNING: Definition list ends without a blank line; unexpected unindent.

/edx/src/edx-analytics-pipeline/edx/analytics/tasks/user_location.py:docstring of user_location.UsersPerCountryReportWorkflow:10: ERROR: Unexpected indentation.

/edx/src/edx-analytics-pipeline/edx/analytics/tasks/user_location.py:docstring of user_location.UsersPerCountryReportWorkflow:11: WARNING: Block quote ends without a blank line; unexpected unindent.

/edx/src/edx-analytics-pipeline/edx/analytics/tasks/user_location.py:docstring of user_location.UsersPerCountryReportWorkflow:19: ERROR: Unexpected indentation.

/edx/src/edx-analytics-pipeline/edx/analytics/tasks/user_location.py:docstring of user_location.UsersPerCountryReportWorkflow:20: WARNING: Block quote ends without a blank line; unexpected unindent.
```

e-kolpakov:
@pomegranited now it is totally awesome :+1:

pomegranited:
@mulby The docs are much better now :)  Could you have a look and let me know what else you think needs fixing?  We've moved the actual docstring/description changes to PR #208 for clarity.

Also, if I could get a list of tasks that should be included in the Workflow Entry Points list, I'll decorate them.

sphinx_source/luigi_sphinx.py
```diff
@@ -0,0 +1,55 @@
...
+                # Mark optional parameters
+                if hasattr(membervalue, 'default'):
+                    param['default'] = membervalue.default
+                    param['type'] = u'{type}, optional'.format(**param)
+
+                    # Show default value, if not already in the description.
+                    # NB: This test is useful to avoid redundant descriptions,
+                    # and for dynamically determined defaults like date.today()
+                    if not default_re.search(param['description']):
+                        param['description'] = u'{description}. Default is {default}.'.format(**param)
```
mulby: If the description already ends in a "." we probably want to dedup it so that this doesn't look like:

"This is some description.. Default is None."

mulby:
This is looking great, I called out one minor issue, not a blocker, however.

I'm OK starting with just the entry points listed on this page: https://github.com/edx/edx-analytics-pipeline/wiki/Tasks-to-Run-to-Update-Insights

I think you have tagged them all already, but I didn't look very closely.

:+1:

mulby:
oh, another non-blocker: the entry-point name is coincidentally the same as the root package name.

In the future we may have entry points defined in other packages... we can fix that then, though, if it's difficult to do now.

pomegranited:
@mulby Cool, I'll get the full-stop de-duped, and look at the entry-point issue to see how hard it'll be.

I had a couple of questions from outdated diffs:
-  Is it useful for the documentation to mark insignificant parameters? See attached docs.pdf edx.analytics.tasks.reports.enrollments.EnrollmentsByWeek for example.  It's easy to remove if that's confusing to users.
-  I'm including the parameter default value in the description, which is fine in most cases. However, some params have default values set via config_path, which might result in confusion if people have their configuration defaults set to something different than what the documentation was generated with.

Is that a concern?  I'm really not sure how to distinguish the situation, so am unsure how to resolve it.

See edx.analytics.tasks.answer_dist.AnswerDistributionWorkflow database parameter for an example. This parameter comes from edx.analytics.tasks.mysql_load.MysqlInsertTaskMixin. The generated default value is reports, and is being pulled from client.cfg.

mulby:
> Is it useful for the documentation to mark insignificant parameters? See attached docs.pdf edx.analytics.tasks.reports.enrollments.EnrollmentsByWeek for example. It's easy to remove if that's confusing to users.

I think so, yes.

> I'm including the parameter default value in the description, which is fine in most cases. However, some params have default values set via config_path, which might result in confusion if people have their configuration defaults set to something different than what the documentation was generated with.

Hrmm, yeah, I could see how that could be confusing. Would it be possible to just say in the docs that the default is pulled from "database_exports.database" (or whatever the config path happens to be) in the config file?

pomegranited:
@mulby Unfortunately, luigi.Parameter stores the `config_path` as `self.__config`, but it's not something I should be accessing..  And there's no flag to indicate that a config value is being used :frowning:  

@e-kolpakov Do you have any ideas?  Context in previous two comments above.

pomegranited:
@mulby Confirmed that the tasks in https://github.com/edx/edx-analytics-pipeline/wiki/Tasks-to-Run-to-Update-Insights are listed on the Workflow Entry Points page.

mulby:
@pomegranited given the nature of this tool (it's effectively performing static analysis), I'm OK with reaching down in to the private interface to extract the config path. Not sure which class it's defined in, but you should be able to directly access the mangled attribute name. Something like `some_parameter._Parameter__config`.

pomegranited:
@mulby Cool, I've updated the docs to include the configured section and name, if there is one.  There's a screenshot above, and the docs.zip shows them all.

Still ok for this to be merged?  (I'll rebase first.)

edx/analytics/tasks/answer_dist.py
```diff
@@ -868,7 +868,9 @@ class AnswerDistributionWorkflow(
 
     # Add additional args for MultiOutputMapReduceJobTask.
     output_root = luigi.Parameter()
+    """ Test docstring for output_root attribute """
```
mulby: nit: remove these test docstrings

setup.cfg
```diff
@@ -30,7 +30,7 @@ edx.analytics.tasks =
     inc-enrollments-report = edx.analytics.tasks.reports.incremental_enrollments:WeeklyIncrementalUsersAndEnrollments
     course-enroll = edx.analytics.tasks.course_enroll:CourseEnrollmentChangesPerDay
     enrollments_and_registrations_workflow-manifest = edx.analytics.tasks.reports.enrollments_and_registrations_workflow_manifest:EnrollmentsandRegistrationsWorkflow
-    answer-dist = edx.analytics.tasks.answer_dist:AnswerDistributionPerCourse
+    answer-dist-workflow = edx.analytics.tasks.answer_dist:AnswerDistributionWorkflow
```
mulby: nit: I think we can revert this?

sphinx_source/luigi_sphinx.py
```diff
@@ -0,0 +1,66 @@
...
+                    param['description'] = membervalue.description
+
+                    # Append a full stop, for consistency.
+                    if not param['description'].endswith('.'):
+                        param['description'] = '{description}.'.format(description=param['description'])
+
+                # Mark configured parameters (requires protected-access)
+                # pylint: disable=W0212
+                if hasattr(membervalue, '_Parameter__config') and membervalue._Parameter__config is not None:
+                    param['default'] = ' pulled from ``{section}.{name}``'.format(**membervalue._Parameter__config)
```
mulby, line 40: nit: I would prefer " read from configuration " instead of "pulled from"

mulby:
Yeah this is ready to merge, I called out a few last nits, but they are all optional.

If you rebase I'll merge.

mulby:
Great work! I'm very excited to see this land.

pomegranited:
@mulby nits picked, and merged :smile:  

Be aware you should merge PR #207 then PR #208; it's got all the docstrings moved into parameter descriptions.

It's been lovely working with you!

PR #209, opened 2016-03-01
Update ansible to 1.9.4.

mtyaka:
Version 1.4.5 does not work correctly on Ubuntu precise.

JIRA ticket: https://openedx.atlassian.net/browse/OLIVE-25


edx-webhook:
Thanks for the pull request, @mtyaka! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=209&repo=edx%2Fedx-analytics-pipeline

edx-webhook:
Thanks for the pull request, @mtyaka! I've created [OSPR-1171](https://openedx.atlassian.net/browse/OSPR-1171) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

omarkhan:
:+1:

nedbat:
@mulby ?

mulby:
@mtyaka - can you be more specific about what isn't working?

also, if we upgrade this hear, we should probably also upgrade edx-analytics-configuration

mtyaka:
@mulby 1.4.x fails with nonsensical errors such as `ERROR: file is not a legal parameter in an Ansible task or handler` or `ERROR: template is not a legal parameter in an Ansible task or handler`.

We were seeing the same problem with edx-analytics-configuration and already have a PR that updates ansible open: https://github.com/edx/edx-analytics-configuration/pull/19

mulby:
I'm OK upgrading ansible if this "just works".

As an aside, however, we have seen this error before when the wrong version of pip is used to install ansible. Newer versions of pip use wheels, which this version of ansible does not support. The effect of the issue is that ansible can't find any of its core modules, which causes the error message you describe.

mtyaka:
@mulby Yes, I believe it "just works".  We have been running analytics pipeline with upgraded ansible in production without any issues for almost a year now, although we're using `1.8.4` rather than `1.9.4` in that deployment. We haven't run into any issues while testing `1.9.4`.

openedx-webhooks:
Thanks for the pull request, @mtyaka! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-analytics-pipeline&number=209

mulby:
@mtyaka IIRC we pin the pipeline to pip 1.5.6 since installation of ansible required an old version of pip to prevent it from trying install ansible from wheels (which causes the error you saw above). We should consider removing this constraint in the requirements if we are upgrading ansible:

https://github.com/edx/edx-analytics-pipeline/blob/master/requirements/default.txt#L35

mtyaka:
@mulby I updated the patch to remove `pip` from `requirements/default.txt` and use the most recent version (`8.1.1`) in `requirements/pre.txt`. I don't entirely understand the implications of AN-4322, but I tested this patch and it seems to work fine.

bradenmacdonald:
@mtyaka @mulby Given that https://github.com/edx/edx-analytics-configuration/pull/27 had to be reverted, should we close this PR as well, and revisit once the new Jenkins DSL magic is in place?

mtyaka:
Yeah, I guess closing this ticket until Jenkins DLS magic is in place makes sense.

PR #213, opened 2016-03-04
WIP: Task to load d_country in warehouse

HassanJaveed84:
@mulby @brianhw up for initial review.

### DIFF
--- a/edx/analytics/tasks/load_internal_reporting_country.py
+++ b/edx/analytics/tasks/load_internal_reporting_country.py
@@ -1,0 +1,123 @@
+"""
+Loads the country table into the warehouse through the pipeline via Hive.
+"""
+
+import luigi
+import logging
+
+from edx.analytics.tasks.vertica_load import VerticaCopyTask
+from edx.analytics.tasks.util.hive import HiveTableFromQueryTask, WarehouseMixin, HivePartition
+from edx.analytics.tasks.url import url_path_join, ExternalURL
+from edx.analytics.tasks.location_per_course import LastCountryOfUserMixin, QueryLastCountryPerCourseMixin, ImportLastCountryOfUserToHiveTask
+
+log = logging.getLogger(__name__)
+
+
+class LoadInternalReportingCountryTableHive(LastCountryOfUserMixin, HiveTableFromQueryTask):
+    """Loads internal_reporting_d_country Hive table from last_country_of_user Hive table."""
+
+    def requires(self):
+        return ImportLastCountryOfUserToHiveTask(
+            mapreduce_engine=self.mapreduce_engine,
+            n_reduce_tasks=self.n_reduce_tasks,
+            source=self.source,
+            interval=self.interval,
+            pattern=self.pattern,
+            geolocation_data=self.geolocation_data,
+            overwrite=self.overwrite,
+            user_country_output=self.user_country_output,
+        )
+
+    @property
+    def partition(self):
+        return HivePartition('dt', self.interval.date_b.isoformat())  # pylint: disable=no-member
+
+    @property
+    def table(self):
+        return 'internal_reporting_d_country'
+
+    @property
+    def columns(self):
+        return [
+            ('country_name', 'STRING'),
+            ('user_last_location_country_code', 'STRING')
+        ]
+
+    @property
+    def insert_query(self):
+        return """
+            SELECT
+              collect_set(lcu.country_name)[0]
+            , lcu.country_code as user_last_location_country_code
+            FROM last_country_of_user lcu
+            GROUP BY lcu.country_code
+            """
+
+
+class ImportCountryWorkflow(QueryLastCountryPerCourseMixin, LastCountryOfUserMixin, luigi.WrapperTask):
+
+    def requires(self):
+        yield (
+            LoadInternalReportingCountryTableHive(
+                mapreduce_engine=self.mapreduce_engine,
+                n_reduce_tasks=self.n_reduce_tasks,
+                source=self.source,
+                interval=self.interval,
+                pattern=self.pattern,
+                geolocation_data=self.geolocation_data,
+                overwrite=self.overwrite,
+                user_country_output=self.user_country_output,
+            ),
+            InsertToMysqlCourseEnrollByCountryWorkflow(
+                mapreduce_engine=self.mapreduce_engine,
+                n_reduce_tasks=self.n_reduce_tasks,
+                source=self.source,
+                interval=self.interval,
+                pattern=self.pattern,
+                geolocation_data=self.geolocation_data,
+                overwrite=self.overwrite,
+                user_country_output=self.user_country_output,
+                course_country_output=self.course_country_output,
+            ),
+        )
+
+
+class LoadInternalReportingCountryToWarehouse(WarehouseMixin, VerticaCopyTask):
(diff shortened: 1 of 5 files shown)


edx/analytics/tasks/load_internal_reporting_country.py
```diff
@@ -0,0 +1,93 @@
...
+
+from edx.analytics.tasks.vertica_load import VerticaCopyTask
+from edx.analytics.tasks.util.hive import HiveTableFromQueryTask, WarehouseMixin, HivePartition
+from edx.analytics.tasks.url import url_path_join, ExternalURL
+from edx.analytics.tasks.location_per_course import LastCountryOfUserMixin, ImportLastCountryOfUserToHiveTask
+
+log = logging.getLogger(__name__)
+
+
+class LoadInternalReportingCountryTableHive(LastCountryOfUserMixin, HiveTableFromQueryTask):
```
mulby, line 16: I wonder if there isn't a more straightforward way to accomplish this. Can we get maxmind to just give us a dump of country code to country names?

HassanJaveed84, line 16: I think we can maybe parse GeoCountry data for this, which would have the country codes and names.

brianhw, line 16: I don't understand.  Do you mean parse the actual raw data, or just run IP addresses through the API to generate possible country codes.  I'm pretty reluctant to introduce a dependency on our ability to parse third-party data on our own like that.

mulby, line 16: no. I was simply wondering if the maxmind API exposed this information directly to us, something like:

``` python
for country_code, country_name in geoip.all_country_list():
    tsv_file.write('\t'.join([country_code, country_name]))
    tsv_file.write('\n')
```

HassanJaveed84, line 16: Right now, the country/code are fetched based on the IP using maxmind's geoip country data(if I'm not wrong). If all we need is country_name/code list, we can just generate them from the geoip country data(using their CSV).

HassanJaveed84, line 16: My comment came in too late, I get the point now :) I'll refer to the API.

HassanJaveed84, line 16: @mulby I don't see anything in pygeoip that can provide us with such info.

mulby, line 16: OK cool, thanks for looking!

HassanJaveed84:
I've added an acceptance test for LoadInternalReportingCountryToWarehouse. 

Does it make sense to add one for LoadInternalReportingCountryTableHive or the wrapper task or both ?

edx/analytics/tasks/load_internal_reporting_country.py
```diff
@@ -0,0 +1,114 @@
...
+            ('user_last_location_country_code', 'STRING')
+        ]
+
+    @property
+    def insert_query(self):
+        return """
+            SELECT
+              collect_set(lcu.country_name)[0]
+            , lcu.country_code as user_last_location_country_code
+            FROM last_country_of_user lcu group by lcu.country_code
```
brianhw: nit:  I would capitalize GROUP BY.  Maybe a separate line.

edx/analytics/tasks/load_internal_reporting_country.py
```diff
@@ -0,0 +1,114 @@
...
+    @property
+    def auto_primary_key(self):
+        """No automatic primary key here; user's id is enough."""
+        return None
+
+    @property
+    def columns(self):
+        return [
+            ('country_name', 'VARCHAR(45)'),
+            ('user_last_location_country_code', 'VARCHAR(45) NOT NULL')
```
brianhw, line 111: I think this should just be country_code.  The goal of this table is to provide a name for a given code.  It shouldn't matter what the code is actually used for.  (That is, if there were another location that were computed, e.g. user_first_location, it would use this same mapping.)

HassanJaveed84, line 111: I used user_last_location_country_code to match production schema(production.d_country has user_last_location_country_code & country_name columns). 
Thought I should relay this before I make the change.

brianhw, line 111: Fair enough.  I think it should eventually change, but we don't have to do that right now in this PR.  (But if we were going to do it sometime, this seems like a pretty good time.)

brianhw:
LGTM, generally.  I think I'm fine with that level of acceptance testing.  Probably just needs docstrings and the like.

edx/analytics/tasks/tests/acceptance/test_internal_reporting_country.py
```diff
@@ -0,0 +1,47 @@
...
+
+class InternalReportingCountryLoadAcceptanceTest(AcceptanceTestCase):
+
+    DATE = '2014-07-01'
+
+    def setUp(self):
+        super(InternalReportingCountryLoadAcceptanceTest, self).setUp()
+        self.upload_file(os.path.join(self.data_dir, 'input', 'internal_reporting_d_country'), url_path_join(self.warehouse_path, 'internal_reporting_d_country', 'dt=2014-07-01', 'internal_reporting_d_country'))
+
+    def test_internal_reporting_country(self):
```
mulby, line 29: This should now be decorated with `when_vertica_available`

mulby:
:+1: once the decorator is added

edx/analytics/tasks/load_internal_reporting_country.py
```diff
@@ -0,0 +1,123 @@
...
+    @property
+    def auto_primary_key(self):
+        """No automatic primary key here; user's id is enough."""
+        return None
+
+    @property
+    def columns(self):
+        return [
+            ('country_name', 'VARCHAR(45)'),
+            ('country_code', 'VARCHAR(45) NOT NULL')
```
HassanJaveed84: @brianhw  I've changed it to just 'country_code' as per your suggestion. Does this change need to be communicated/documented somehow ?

edx/analytics/tasks/load_internal_reporting_country.py
```diff
@@ -0,0 +1,123 @@
...
+        return """
+            SELECT
+              collect_set(lcu.country_name)[0]
+            , lcu.country_code as user_last_location_country_code
+            FROM last_country_of_user lcu
+            GROUP BY lcu.country_code
+            """
+
+
+class ImportCountryWorkflow(QueryLastCountryPerCourseMixin, LastCountryOfUserMixin, luigi.WrapperTask):
```
HassanJaveed84, line 57: Does the class name convey the functionality here ?

HassanJaveed84:
Should I merge this in ?

edx/analytics/tasks/load_internal_reporting_country.py
mulby: my opinion is that we shouldn't make any schema changes like this during the port to reduce risk and simplify comparison between old and new

mulby:
I think this is basically ready to merge, maybe we should chat briefly after standup today about the column name. I want to make sure you, Brian and I agree.

PR #214, opened 2016-03-04
Support RHEL-type OS

brianhw:
This is the work done to get a cluster to run on AWS EMR 3.x and 4.x.  

FYI:  @mulby @HassanJaveed84

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,12 +6,6 @@ else
 	PIP_INSTALL = pip install --allow-external mysql-connector-python
 endif
 
-ifeq (,$(wildcard /usr/bin/yum))
-    PKG_INSTALL = apt-get
-else
-    PKG_INSTALL = yum
-endif
-
 .PHONY:	requirements test test-requirements .tox
 
 uninstall:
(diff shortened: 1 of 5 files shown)


share/task.yml
```diff
@@ -29,11 +36,18 @@
...
-        # EMR ships with python 2.7
+        # Match the version of Python define in virtualenv_python above.
         WHEEL_PYVER: 2.7
+    - common_debian_variants:
+      - Ubuntu
+      - Debian
+    - common_redhat_variants:
+      - CentOS
+      - Red Hat Enterprise Linux
+      - Amazon
```
brianhw, line 50: This is repeated twice.  Is there a better way to define these just once somewhere, and just refer to them in both places?

Makefile
```diff
@@ -27,9 +27,14 @@ develop-local: uninstall
 	python setup.py install_data
 
 system-requirements:
+ifeq (,$(wildcard /usr/bin/yum))
 	sudo apt-get update -q
 	# This is not great, we can't use these libraries on slave nodes using this method.
 	sudo apt-get install -y -q libmysqlclient-dev libatlas3gf-base libpq-dev python-dev libffi-dev libssl-dev libxml2-dev libxslt1-dev
+else
+	sudo yum update -q -y
+	sudo yum install -y -q postgresql-devel libffi-devel
```
brianhw, line 36: This is very likely an incomplete list.  I don't believe that either of these have been actually tested, let alone the external server requests (e.g. paypal) that led to many of the additional libraries being included.

mulby:
@brianhw my understanding is that this was merged as part of #231

PR #215, opened 2016-03-11
Generate a searchable student roster

mulby:
@brianhw - in case you have time to start review when I'm gone, I figured I would at least post this.

Open to suggestions about how to make this easier to review.

TODO:
- [x] finish tests cases for elasticsearch_load
- [x] lots of docstrings and comments!
- [x] document the sort ordering scheme
- [x] pep8, pylint
- [x] make sure the acceptance test isn't broken (I haven't run it in a while)

Other optional tasks:
- [x] Handle the case when there is no tracking logs for a particular date. This currently causes a failure, which makes testing on devstack difficult.

Deferred:
~~The big hive query runs and overwrites any existing data if the underlying data exists but the hive partition does not. This can happen fairly frequently during testing if you use a new cluster with existing data.~~

### DIFF
--- a/config/devstack.cfg
+++ b/config/devstack.cfg
@@ -69,3 +69,10 @@ interval = 2012-01-01-2020-01-01
 [videos]
 dropoff_threshold = 0.05
 
+[elasticsearch]
+# Point to the vagrant host's port 9201 where we assume elasticsearch is running
+host = http://172.17.0.1:9201/
+
+[module-engagement]
+alias = roster
+number_of_shards = 5
--- a/edx/analytics/tasks/elasticsearch_load.py
+++ b/edx/analytics/tasks/elasticsearch_load.py
@@ -1,0 +1,420 @@
+"""Load records into elasticsearch clusters."""
+
+from itertools import islice
+import logging
+import random
+import time
+
+try:
+    import elasticsearch
+    import elasticsearch.helpers
+    from elasticsearch.exceptions import TransportError
+except ImportError:
+    elasticsearch = None
+import luigi
+
+from edx.analytics.tasks.mapreduce import MapReduceJobTask
+try:
+    from edx.analytics.tasks.util.aws_elasticsearch_connection import AwsHttpConnection
+except ImportError:
+    AwsHttpConnection = None
+from edx.analytics.tasks.util.elasticsearch_target import ElasticsearchTarget
+from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
+
+
+log = logging.getLogger(__name__)
+
+
+# These are standard HTTP status codes used by elasticsearch to represent various error conditions
+HTTP_CONNECT_TIMEOUT_STATUS_CODE = 408
+REJECTED_REQUEST_STATUS = 429
+HTTP_GATEWAY_TIMEOUT_STATUS_CODE = 504
+
+
+class ElasticsearchIndexTaskMixin(OverwriteOutputMixin):
+    """A task that either loads data into elasticsearch or depends on a task that does."""
+
+    host = luigi.Parameter(
+        is_list=True,
+        config_path={'section': 'elasticsearch', 'name': 'host'},
+        description='Hostnames for the elasticsearch cluster nodes. They can be specified in any of the formats'
+                    ' accepted by the elasticsearch-py library. This includes complete URLs such as http://foo.com/, or'
+                    ' host port pairs such as foo:8000. Note that if you wish to use SSL you should specify a full URL'
+                    ' and the "https" scheme.'
+    )
+    timeout = luigi.FloatParameter(
+        config_path={'section': 'elasticsearch', 'name': 'timeout'},
+        significant=False,
+        default=60,
+        description='Maximum number of seconds to wait when attempting to make connections to the elasticsearch cluster'
+                    ' before assuming the cluster is not responding and giving up with a timeout error.'
+    )
+    connection_type = luigi.Parameter(
+        config_path={'section': 'elasticsearch', 'name': 'connection_type'},
+        significant=False,
+        default='urllib',
+        description='If not specified, default to using urllib3 to make HTTP requests to elasticsearch. The other valid'
+                    ' value is "aws" which can be used to connect to clusters that are managed by AWS. See'
+                    ' `AWS elasticsearch service <https://aws.amazon.com/elasticsearch-service/>`_'
+    )
+    alias = luigi.Parameter(
+        description='Name of the alias in elasticsearch that will point to the complete index when loaded. This value '
+                    ' should match the settings of edx-analytics-data-api.'
+    )
+    number_of_shards = luigi.Parameter(
+        default=None,
+        description='Number of `shards <https://www.elastic.co/guide/en/elasticsearch/reference/current/glossary.html'
+                    '#glossary-shard>`_ to use in the elasticsearch index.'
+    )
+    throttle = luigi.FloatParameter(
+        default=0.1,
+        significant=False,
+        description='Wait this many seconds between batches of records submitted to the cluster to be indexed. This can'
+                    ' be used to tune the indexing process, allowing the cluster to successfully "keep up" with the'
+                    ' loader. Note that often the hadoop cluster can load records much more quickly than the cluster'
+                    ' can index them, which eventually causes queues to overflow within the elasticsearch cluster.'
+    )
+    batch_size = luigi.IntParameter(
+        default=1000,
+        significant=False,
+        description='Number of records to submit to the cluster to be indexed in a single request. A small value here'
+                    ' will result in more, smaller, requests and a larger value will result in fewer, bigger requests.'
+    )
+    indexing_tasks = luigi.IntParameter(
+        default=None,
+        significant=False,
+        description='Number of parallel processes to use to submit records to be indexed from. The stream of records'
+                    ' will be divided up evenly among these processes during the indexing procedure.'
+    )
+    max_attempts = luigi.IntParameter(
+        default=10,
+        significant=False,
+        description='If the elasticsearch cluster rejects a batch of records (usually because it is too busy) the'
+                    ' indexing process will retry up to this many times before giving up. It uses an exponential back-'
+                    'off strategy, so a high value here can result in very significant wait times before retrying.'
+    )
+
+
+class ElasticsearchIndexTask(ElasticsearchIndexTaskMixin, MapReduceJobTask):
+    """
+    Index a stream of documents in an elasticsearch index.
+
+    This task is intended to do the following:
+    * Create a new index that is unique to this task run (all significant parameters).
+    * Load all of the documents into this unique index.
+    * If the alias is already pointing at one or more indexes, switch it so that it only points at this newly loaded
+      index.
+    * Delete any indexes that were previously pointed at by the alias, leaving only the newly loaded index.
+
+    """
+
+    # These attributes should be overridden, but don't need to be.
+    settings = {}
+    properties = {}
+
+    def __init__(self, *args, **kwargs):
+        super(ElasticsearchIndexTask, self).__init__(*args, **kwargs)
+
+        self.other_reduce_tasks = self.n_reduce_tasks
+        if self.indexing_tasks is not None:
+            self.n_reduce_tasks = self.indexing_tasks
+
+        self.batch_index = 0
+        self.index = self.alias + '_' + str(hash(self.update_id()))
+        self.indexes_for_alias = set()
+
+    def init_local(self):
+        super(ElasticsearchIndexTask, self).init_local()
+
+        elasticsearch_client = self.create_elasticsearch_client()
+
+        # Find all indexes that are referred to by this alias (currently). These will be deleted after a successful
+        # load of the new index.
+        self.indexes_for_alias.update(
+            elasticsearch_client.indices.get_aliases(name=self.alias).keys()
+        )
+
+        if self.index in self.indexes_for_alias:
+            if not self.overwrite:
+                raise RuntimeError('Index {0} is currently in use by alias {1}'.format(self.index, self.alias))
+            else:
+                # These indexes will be deleted, after the alias swap, make sure we don't delete the index we just
+                # populated.
+                self.indexes_for_alias.remove(self.index)
+
+        if not self.overwrite and len(self.indexes_for_alias) > 1:
+            raise RuntimeError(
+                'Invalid state, multiple existing indexes ({0}) found for alias {1}'.format(
+                    ', '.join(self.indexes_for_alias),
+                    self.alias
+                )
+            )
+
+        if elasticsearch_client.indices.exists(index=self.index):
+            elasticsearch_client.indices.delete(index=self.index)
+
+        settings = {
+            'refresh_interval': -1,
+        }
+        if self.number_of_shards is not None:
+            settings['number_of_shards'] = self.number_of_shards
+
+        if self.settings:
+            settings.update(self.settings)
+
+        elasticsearch_client.indices.create(index=self.index, body={
+            'settings': settings,
+            'mappings': {
+                self.doc_type: {
+                    'properties': self.properties
+                }
+            }
+        })
+
+    def create_elasticsearch_client(self):
+        """Build an elasticsearch client using the various parameters passed into this task."""
+        kwargs = {}
+        if self.connection_type == 'aws':
+            kwargs['connection_class'] = AwsHttpConnection
+        return elasticsearch.Elasticsearch(
+            hosts=self.host,
+            timeout=self.timeout,
+            retry_on_status=(HTTP_CONNECT_TIMEOUT_STATUS_CODE, HTTP_GATEWAY_TIMEOUT_STATUS_CODE),
+            retry_on_timeout=True,
+            **kwargs
+        )
+
+    def mapper(self, line):
+        yield (random.randrange(int(self.n_reduce_tasks)), line.rstrip('\r\n'))
+
+    def reducer(self, _key, lines):
+        """
+        Given a batch of records, transmit them to the elasticsearch cluster to be indexed.
+
+        There should be one reducer per parallel indexing thread. Controlling the number of reducers is the way to
+        control the level of parallelism in the load process.
+        """
+        elasticsearch_client = self.create_elasticsearch_client()
+
+        document_iterator = self.document_generator(lines)
+        first_batch = True
+        while True:
+            bulk_action_batch = self.next_bulk_action_batch(document_iterator)
+
+            if not bulk_action_batch:
+                break
+
+            if not first_batch and self.throttle:
+                time.sleep(self.throttle)
+            first_batch = False
+
+            if self.send_bulk_action_batch(elasticsearch_client, bulk_action_batch):
+                self.incr_counter('Elasticsearch', 'Committed Batches', 1)
+
+                # Note that each document produces two entries in the bulk_action_batch list.
+                num_records = len(bulk_action_batch) / 2
+                self.incr_counter('Elasticsearch', 'Records Indexed', num_records)
+            else:
+                raise IndexingError('Batch of records rejected too many times. Aborting.')
+
+        # Luigi requires the reducer to actually return something, so we just return empty strings that are written
+        # to a temp file in HDFS that is immediately cleaned up after the job finishes.
+        yield ('', '')
+
+    def next_bulk_action_batch(self, document_iterator):
+        """
+        Read a batch of documents from the iterator and convert them into bulk index actions.
+
+        Elasticsearch expects each document to actually be transmitted on two lines the first of which details the
+        action to take, and the second contains the actual document.
+
+        See the `Cheaper in Bulk <https://www.elastic.co/guide/en/elasticsearch/guide/1.x/bulk.html>`_ guide.
+
+        Arguments:
+            document_iterator (iterator of dicts):
+
+        Returns: A list of dicts that can be transmitted to elasticsearch using the "bulk" request.
+        """
+        bulk_action_batch = []
+        for raw_data in islice(document_iterator, self.batch_size):
+            action, data = elasticsearch.helpers.expand_action(raw_data)
+            bulk_action_batch.append(action)
+            if data is not None:
+                bulk_action_batch.append(data)
+        return bulk_action_batch
+
+    def send_bulk_action_batch(self, elasticsearch_client, bulk_action_batch):
+        """
+        Given a batch of actions, transmit them in bulk to the elasticsearch cluster.
+
+        This method handles back-pressure from the elasticsearch cluster which queues up writes. When the queue is full
+        the cluster will start rejecting additional bulk indexing requests. This method implements an exponential
+        back-off, allowing the cluster to catch-up with the client.
+
+        Arguments:
+            elasticsearch_client (elasticsearch.Elasticsearch): A reference to an elasticsearch client.
+            bulk_action_batch (list of dicts): A list of bulk actions followed by their respective documents.
+
+        Raises:
+            IndexingError: If a record cannot be indexed by elasticsearch this method assumes that is a fatal error and
+                it immediately raises this exception. If we try to transmit a batch repeatedly and it is continually
+                rejected by the cluster, this method will give up after `max_attempts` and raise this error.
+
+        Returns: True iff the batch of actions was successfully transmitted to and acknowledged by the elasticsearch
+            cluster.
+        """
+        attempts = 0
+        batch_written_successfully = False
+        while True:
+            try:
+                resp = elasticsearch_client.bulk(bulk_action_batch, index=self.index, doc_type=self.doc_type)
+            except TransportError as transport_error:
+                if transport_error.status_code != REJECTED_REQUEST_STATUS:
+                    raise transport_error
+            else:
+                num_errors = 0
+                for raw_data in resp['items']:
+                    _op_type, item = raw_data.popitem()
+                    successful = 200 <= item.get('status', 500) < 300
+                    if not successful:
+                        log.error('Failed to index: %s', str(item))
+                        num_errors += 1
+
+                if num_errors == 0:
+                    batch_written_successfully = True
+                    break
+                else:
+                    raise IndexingError('Failed to index {0} records. Aborting.'.format(num_errors))
+
+            attempts += 1
+            if attempts < self.max_attempts:
+                sleep_duration = 2 ** attempts
+                self.incr_counter('Elasticsearch', 'Rejected Batches', 1)
+                log.warn(
+                    'Batch of records rejected. Sleeping for %d seconds before retrying.',
+                    sleep_duration
+                )
+                time.sleep(sleep_duration)
+            else:
+                batch_written_successfully = False
+                break
+
+        return batch_written_successfully
+
+    def document_generator(self, lines):
+        """
+        Given lines of raw text, generates structured documents that will be indexed by elasticsearch.
+
+        The returned document should have roughly the following structure:
+
+            {
+                "_id": "(optional) your custom identifier for the document",
+                "_source": {
+                    "prop0": "you should have one key-value pair for each property and its value"
+                }
+            }
+
+        Note that you can also specify other "special" fields other than "_id":
+
+        - _index
+        - _parent
+        - _percolate
+        - _routing
+        - _timestamp
+        - _ttl
+        - _type
+        - _version
+        - _version_type
+        - _retry_on_conflict
+
+        The "_source" field is required.
+
+        Arguments:
+            lines (iterable of unicode strings): This is the raw data to be indexed.
+
+        Yields:
+            dict: The document to index in the format expected by the elasticsearch bulk loading process.
+        """
+        raise NotImplementedError
+
+    @property
+    def doc_type(self):
+        """
+        Elasticsearch `document type <https://www.elastic.co/guide/en/elasticsearch/guide/current/mapping.html>`_.
+        """
+        raise NotImplementedError
+
+    def extra_modules(self):
+        import urllib3
+
+        packages = [elasticsearch, urllib3]
+
+        return packages
+
+    def jobconfs(self):
+        jcs = super(ElasticsearchIndexTask, self).jobconfs()
+        jcs.append('mapred.reduce.tasks.speculative.execution=false')
+        return jcs
+
+    def update_id(self):
+        """A unique identifier for this task instance that is used to determine if it should be run again."""
+        return self.task_id
+
+    def output(self):
+        return ElasticsearchTarget(
+            client=self.create_elasticsearch_client(),
+            index=self.alias,
+            doc_type=self.doc_type,
+            update_id=self.update_id()
+        )
+
+    def commit(self):
+        """
+        If all documents have been loaded successfully, make the changes visible to users.
+        """
+        # The ordering of operations here is sensitive.
+
+        elasticsearch_client = self.create_elasticsearch_client()
+
+        # First "refresh" the newly loaded index. We disable refreshes during the load to keep throughput high. This
+        # step is necessary to ensure all of the documents are properly indexed and user-visible.
+        elasticsearch_client.indices.refresh(index=self.index)
+
+        # Perform an atomic swap of the alias.
+        actions = []
+        for old_index in self.indexes_for_alias:
+            actions.append({"remove": {"index": old_index, "alias": self.alias}})
+        actions.append({"add": {"index": self.index, "alias": self.alias}})
+        elasticsearch_client.indices.update_aliases({"actions": actions})
+
+        # Update the luigi metadata to indicate that the task ran successfully.
+        self.output().touch()
+
+        # Attempt to remove any old indexes that are now no longer user-visible.
+        for old_index in self.indexes_for_alias:
+            elasticsearch_client.indices.delete(index=old_index)
+
+    def rollback(self):
+        """
+        If something goes wrong during the load, attempt to clean up the partially loaded index.
+        """
+        elasticsearch_client = self.create_elasticsearch_client()
+        try:
+            if elasticsearch_client.indices.exists(index=self.index):
+                elasticsearch_client.indices.delete(index=self.index)
+        except Exception:  # pylint: disable=broad-except
+            log.exception("Unable to rollback the elasticsearch load.")
+
+    def run(self):
+        try:
+            super(ElasticsearchIndexTask, self).run()
+        except Exception:  # pylint: disable=broad-except
+            self.rollback()
+            raise
+        else:
+            self.commit()
+
+
+class IndexingError(RuntimeError):
+    """Something went wrong during the indexing operation."""
+    pass
--- a/edx/analytics/tasks/module_engagement.py
+++ b/edx/analytics/tasks/module_engagement.py
@@ -1,38 +1,60 @@
-"""Measure student engagement with individual modules in the course"""
+"""
+Measure student engagement with individual modules in the course.
 
+See ModuleEngagementWorkflowTask for more extensive documentation.
+"""
+
+from collections import defaultdict
+import datetime
 import logging
+import random
 
+log = logging.getLogger(__name__)
+
+# pylint: disable=wrong-import-position
 import luigi
 import luigi.task
 from luigi import date_interval
-from luigi.configuration import get_config
+try:
+    import numpy
+except ImportError:
+    log.warn('Unable to import numpy')
+    numpy = None  # pylint: disable=invalid-name
+
+from edx.analytics.tasks.database_imports import ImportAuthUserTask, ImportCourseUserGroupUsersTask, \
+    ImportAuthUserProfileTask
+from edx.analytics.tasks.database_imports import ImportCourseUserGroupTask
+from edx.analytics.tasks.decorators import workflow_entry_point
+from edx.analytics.tasks.elasticsearch_load import ElasticsearchIndexTask
+from edx.analytics.tasks.enrollments import CourseEnrollmentTableTask
 
 from edx.analytics.tasks.mapreduce import MapReduceJobTask, MapReduceJobTaskMixin
 from edx.analytics.tasks.pathutil import EventLogSelectionMixin, EventLogSelectionDownstreamMixin
-from edx.analytics.tasks.url import get_target_from_url
+from edx.analytics.tasks.url import get_target_from_url, url_path_join
 from edx.analytics.tasks.util import eventlog
 from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
-from edx.analytics.tasks.vertica_load import VerticaCopyTask
-from edx.analytics.tasks.mysql_load import IncrementalMysqlInsertTask
+from edx.analytics.tasks.mysql_load import IncrementalMysqlInsertTask, MysqlInsertTask
 
 from edx.analytics.tasks.util.hive import (
-    WarehouseMixin, BareHiveTableTask, HivePartitionTask
-)
-from edx.analytics.tasks.util.record import Record, StringField, IntegerField, DateField
-
-log = logging.getLogger(__name__)
+    WarehouseMixin, BareHiveTableTask, HivePartitionTask,
+    hive_database_name)
+from edx.analytics.tasks.util.record import Record, StringField, IntegerField, DateField, FloatField
 
 
 class ModuleEngagementRecord(Record):
     """Represents a count of interactions performed by a user on a particular entity (usually a module in a course)."""
 
-    course_id = StringField(length=255, nullable=False)
-    username = StringField(length=30, nullable=False)
-    date = DateField(nullable=False)
-    entity_type = StringField(length=10, nullable=False)
-    entity_id = StringField(length=255, nullable=False)
-    event = StringField(length=30, nullable=False)
-    count = IntegerField(nullable=False)
+    course_id = StringField(length=255, nullable=False, description='Course the learner interacted with.')
+    username = StringField(length=30, nullable=False, description='Learner\'s username.')
+    date = DateField(nullable=False, description='The learner interacted with the entity on this date.')
+    entity_type = StringField(length=10, nullable=False, description='Category of entity that the learner interacted'
+                                                                     ' with. Example: "video".')
+    entity_id = StringField(length=255, nullable=False, description='A unique identifier for the entity within the'
+                                                                    ' course that the learner interacted with.')
+    event = StringField(length=30, nullable=False, description='The interaction the learner had with the entity.'
+                                                               ' Example: "viewed".')
+    count = IntegerField(nullable=False, description='Number of interactions the learner had with this entity on this'
+                                                     ' date.')
 
 
 class ModuleEngagementDownstreamMixin(
@@ -41,7 +63,11 @@ class ModuleEngagementDownstreamMixin(
     """Common parameters and base classes used to pass parameters through the engagement workflow."""
 
     # Required parameter
-    date = luigi.DateParameter()
+    date = luigi.DateParameter(
+        description='Upper bound date for the end of the interval to analyze. Data produced before 00:00 on this'
+                    ' date will be analyzed. This workflow is intended to run nightly and this parameter is intended'
+                    ' to be set to "today\'s" date, so that all of yesterday\'s data is included and none of today\'s.'
+    )
 
     # Override superclass to disable this parameter
     interval = None
@@ -69,6 +95,10 @@ class ModuleEngagementDataTask(EventLogSelectionMixin, OverwriteOutputMixin, Map
 
     # Override superclass to disable this parameter
     interval = None
+
+    # Write the output directly to the final destination and rely on the _SUCCESS file to indicate whether or not it
+    # is complete. Note that this is a custom extension to luigi.
+    enable_direct_output = True
 
     def __init__(self, *args, **kwargs):
         super(ModuleEngagementDataTask, self).__init__(*args, **kwargs)
@@ -141,17 +171,12 @@ class ModuleEngagementDataTask(EventLogSelectionMixin, OverwriteOutputMixin, Map
                 entity_id = event_data.get('problem_id')
         elif event_type == 'play_video':
             entity_type = 'video'
-            user_actions.append('played')
+            user_actions.append('viewed')
             entity_id = event_data.get('id')
         elif event_type.startswith('edx.forum.'):
-            entity_type = 'forum'
+            entity_type = 'discussion'
             if event_type.endswith('.created'):
-                if event_type == 'edx.forum.comment.created':
-                    user_actions.append('commented')
-                elif event_type == 'edx.forum.response.created':
-                    user_actions.append('responded')
-                elif event_type == 'edx.forum.thread.created':
-                    user_actions.append('created')
+                user_actions.append('contributed')
 
             entity_id = event_data.get('commentable_id')
 
@@ -164,8 +189,17 @@ class ModuleEngagementDataTask(EventLogSelectionMixin, OverwriteOutputMixin, Map
     def output(self):
         return get_target_from_url(self.output_root)
 
+    def complete(self):
+        if self.overwrite and not self.attempted_removal:
+            return False
+        else:
+            return get_target_from_url(url_path_join(self.output_root, '_SUCCESS')).exists()
+
     def run(self):
         self.remove_output_on_overwrite()
+        output_target = self.output()
+        if not self.complete() and output_target.exists():
+            output_target.remove()
         return super(ModuleEngagementDataTask, self).run()
 
 
@@ -200,13 +234,8 @@ class ModuleEngagementPartitionTask(ModuleEngagementDownstreamMixin, HivePartiti
             overwrite=self.overwrite,
         )
 
-    def requires(self):
-        yield self.data_task
-        yield self.hive_table_task
-
     @property
     def data_task(self):
-        """The task that generates the raw data inside this hive partition."""
         return ModuleEngagementDataTask(
             date=self.date,
             n_reduce_tasks=self.n_reduce_tasks,
@@ -218,11 +247,19 @@ class ModuleEngagementPartitionTask(ModuleEngagementDownstreamMixin, HivePartiti
 class ModuleEngagementMysqlTask(ModuleEngagementDownstreamMixin, IncrementalMysqlInsertTask):
     """
     This table is appended to every time this workflow is run, so it is expected to grow to be *very* large. For this
-    reason, the records are stored in a clustered index allowing for very fast point queries and date range queries
-    for individual users in particular courses.
+    reason, the records are indexed to allow for very fast point queries and date range queries for individual users in
+    particular courses.
 
     This allows us to rapidly generate activity charts over time for small numbers of users.
+
+    Note that it would have been better to use a clustered index and stored the records in sorted order, however, the
+    Django ORM does not support composite primary key indexes, so we have to use a secondary index.
     """
+
+    allow_empty_insert = luigi.BooleanParameter(
+        default=False,
+        config_path={'section': 'module-engagement', 'name': 'allow_empty_insert'},
+    )
 
     @property
     def table(self):
@@ -233,37 +270,13 @@ class ModuleEngagementMysqlTask(ModuleEngagementDownstreamMixin, IncrementalMysq
         return "date='{date}'".format(date=self.date.isoformat())  # pylint: disable=no-member
 
     @property
-    def auto_primary_key(self):
-        # Instead of using an auto incrementing primary key, we define a custom compound primary key. See keys() defined
-        # below. This vastly improves the performance of our most common query pattern.
-        return None
-
-    @property
     def columns(self):
         return ModuleEngagementRecord.get_sql_schema()
 
     @property
-    def keys(self):
-        # Use a primary key since we will always be pulling these records by course_id, username ordered by date
-        # This dramatically speeds up access times at the cost of write speed.
-
-        # From: http://dev.mysql.com/doc/refman/5.6/en/innodb-restrictions.html
-
-        # The InnoDB internal maximum key length is 3500 bytes, but MySQL itself restricts this to 3072 bytes. This
-        # limit applies to the length of the combined index key in a multi-column index.
-
-        # The total for this key is:
-        #   course_id(255 characters * 3 bytes per utf8 char)
-        #   username(30 characters * 3 bytes per utf8 char)
-        #   date(3 bytes per DATE)
-        #   entity_type(10 characters * 3 bytes per utf8 char)
-        #   entity_id(255 characters * 3 bytes per utf8 char)
-        #   event(30 characters * 3 bytes per utf8 char)
-        #   count(4 bytes per INTEGER)
-
-        # Total = 1747
+    def indexes(self):
         return [
-            ('PRIMARY KEY', ['course_id', 'username', 'date', 'entity_type', 'entity_id', 'event'])
+            ('course_id', 'username', 'date')
         ]
 
     @property
@@ -276,64 +289,12 @@ class ModuleEngagementMysqlTask(ModuleEngagementDownstreamMixin, IncrementalMysq
         return partition_task.data_task
 
 
-class ModuleEngagementVerticaTask(ModuleEngagementDownstreamMixin, VerticaCopyTask):
-    """
-    Replicate the hive table in Vertica.
-
-    Note that it is updated incrementally.
-    """
-
-    @property
-    def insert_source_task(self):
-        partition_task = ModuleEngagementPartitionTask(
-            date=self.date,
-            n_reduce_tasks=self.n_reduce_tasks,
-            overwrite=self.overwrite,
-        )
-        return partition_task.data_task
-
-    @property
-    def table(self):
-        return 'f_module_engagement'
-
-    @property
-    def default_columns(self):
-        """List of tuples defining name and definition of automatically-filled columns."""
-        return None
-
-    @property
-    def columns(self):
-        return ModuleEngagementRecord.get_sql_schema()
-
-
-class OptionalVerticaMixin(object):
-    """
-    If a vertica connection is present, replicate the data there. Otherwise, don't require those insertion tasks.
-    """
-
-    vertica_schema = luigi.Parameter(default=None)
-    vertica_credentials = luigi.Parameter(default=None)
-
-    def __init__(self, *args, **kwargs):
-        super(OptionalVerticaMixin, self).__init__(*args, **kwargs)
-
-        if not self.vertica_credentials:
-            self.vertica_credentials = get_config().get('vertica-export', 'credentials', None)
-
-        if not self.vertica_schema:
-            self.vertica_schema = get_config().get('vertica-export', 'schema', None)
-
-        self.vertica_enabled = self.vertica_credentials and self.vertica_schema
-
-
-class ModuleEngagementIntervalTask(
-    MapReduceJobTaskMixin, EventLogSelectionDownstreamMixin, WarehouseMixin, OptionalVerticaMixin,
-    OverwriteOutputMixin, luigi.WrapperTask
-):
-    """Compute engagement information over a range of dates and insert the results into Hive, Vertica and MySQL"""
+class ModuleEngagementIntervalTask(MapReduceJobTaskMixin, EventLogSelectionDownstreamMixin, WarehouseMixin,
+                                   OverwriteOutputMixin, luigi.WrapperTask):
+    """Compute engagement information over a range of dates and insert the results into Hive and MySQL"""
 
     def requires(self):
-        for date in self.interval:
+        for date in reversed([d for d in self.interval]):  # pylint: disable=not-an-iterable
             yield ModuleEngagementPartitionTask(
                 date=date,
                 n_reduce_tasks=self.n_reduce_tasks,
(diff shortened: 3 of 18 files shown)


mulby:
Oh also:
- [x] Make a single entry point that depends on both the metric range MySQL task and the roster indexing task

mulby:
@HassanJaveed84 - I would love your feedback on this PR as well. As an engineer without as much context, you will be able to identify confusing bits and missing documentation.

edx/analytics/tasks/elasticsearch_load.py
```diff
@@ -0,0 +1,292 @@
...
+                    ' indexing process will retry up to this many times before giving up. It uses an exponential back-'
+                    'off strategy, so a high value here can result in very significant wait times before retrying.'
+    )
+
+
+class ElasticsearchIndexTask(ElasticsearchIndexTaskMixin, MapReduceJobTask):
+    """
+    Index a stream of documents in an elasticsearch index.
+
+    The is intended to do the following:
```
HassanJaveed84: Should 'The' be 'This' here ?

config/devstack.cfg
```diff
@@ -69,3 +69,9 @@ interval = 2012-01-01-2020-01-01
 [videos]
 dropoff_threshold = 0.05
 
+[elasticsearch]
+host = http://172.17.0.1:9201/
```
brianhw, line 74: Is this temporary, for local testing?  If permanent, it should have some kind of comment saying what it is.

edx/analytics/tasks/elasticsearch_load.py
```diff
@@ -0,0 +1,292 @@
...
+except ImportError:
+    BotoHttpConnection = None
+
+
+log = logging.getLogger(__name__)
+
+
+REJECTED_REQUEST_STATUS = 429
+HTTP_CONNECT_TIMEOUT_STATUS_CODE = 408
+HTTP_GATEWAY_TIMEOUT_STATUS_CODE = 504
```
brianhw, line 31: Are these codes returned by Elastic Search?  Needs a comment to say so.  
nit: list in sorted order.  Probably by value.

edx/analytics/tasks/elasticsearch_load.py
```diff
@@ -0,0 +1,292 @@
...
+        description='Wait this many seconds between batches of records submitted to the cluster to be indexed. This can'
+                    ' be used to tune the indexing process, allowing the cluster to successfully "keep up" with the'
+                    ' loader. Note that often the hadoop cluster can load records much more quickly than the cluster'
+                    ' can index them, which eventually causes queues to overflow within the elasticsearch cluster.'
+    )
+    batch_size = luigi.IntParameter(
+        default=1000,
+        significant=False,
+        description='Number of records to submit to the cluster to be indexed in a single request. A small value here'
+                    'will result in more, smaller, requests and a larger value will result in fewer, bigger requests.'
```
brianhw: nit: leading space in continuation line.  (i.e. ' will...")

edx/analytics/tasks/elasticsearch_load.py
```diff
@@ -0,0 +1,292 @@
...
+        config_path={'section': 'elasticsearch', 'name': 'connection_type'},
+        significant=False,
+        default='urllib',
+        description='If not specified, default to using urllib3 to make HTTP requests to elasticsearch. The other valid'
+                    ' value is "boto" which can be used to connect to clusters that are managed by AWS. See'
+                    ' `AWS elasticsearch service <https://aws.amazon.com/elasticsearch-service/>`_'
+    )
+    alias = luigi.Parameter(
+        description='Name of the alias in elasticsearch that will point to the complete index when loaded. This value '
+                    'should match the settings of edx-analytics-data-api.'
```
brianhw: nit: leading space:  ' should....'

edx/analytics/tasks/elasticsearch_load.py
```diff
@@ -0,0 +1,292 @@
...
+                try:
+                    resp = es.bulk(bulk_action_batch, index=self.index, doc_type=self.doc_type)
+                except TransportError as e:
+                    if e.status_code != REJECTED_REQUEST_STATUS:
+                        raise e
+                else:
+                    num_errors = 0
+                    for raw_data in resp['items']:
+                        op_type, item = raw_data.popitem()
+                        ok = 200 <= item.get('status', 500) < 300
```
HassanJaveed84: Do we treat any status code in 200-299 as success ?

edx/analytics/tasks/elasticsearch_load.py
```diff
@@ -0,0 +1,292 @@
...
+                    'will result in more, smaller, requests and a larger value will result in fewer, bigger requests.'
+    )
+    indexing_tasks = luigi.IntParameter(
+        default=None,
+        significant=False,
+        description='Number of parallel processes to use to submit records to be indexed from. The stream of records'
+                    ' will be divided up evenly among these processes during the indexing procedure.'
+    )
+    max_attempts = luigi.IntParameter(
+        default=10,
```
brianhw, line 90: should this be significant=False?

edx/analytics/tasks/elasticsearch_load.py
```diff
@@ -0,0 +1,292 @@
...
+
+    def init_local(self):
+        super(ElasticsearchIndexTask, self).init_local()
+
+        es = self.create_elasticsearch_client()
+
+        # Find all indexes that are referred to by this alias (currently). These will be deleted after a successful
+        # load of the new index.
+        for index_name, aliases in es.indices.get_aliases(name=self.alias).iteritems():
+            if self.alias in aliases.get('aliases', {}):
```
brianhw: naive question:  seems that get_aliases(name=self.alias) ought to return only indices that contain self.alias as one of its aliases.  So is this if-statement just confirming an expected state?  Is it sometimes not found?  If so, should it warn, or raise?

edx/analytics/tasks/elasticsearch_load.py
```diff
@@ -0,0 +1,292 @@
...
+
+        if self.index in self.indexes_for_alias:
+            if not self.overwrite:
+                raise RuntimeError('Index {0} is currently in use by alias {1}'.format(self.index, self.alias))
+            else:
+                # These indexes will be deleted, after the alias swap, make sure we don't delete the index we just
+                # populated.
+                self.indexes_for_alias.remove(self.index)
+
+        if not self.overwrite and len(self.indexes_for_alias) > 1:
```
brianhw, line 145: Why is this an error when overwrite is not set, and not an error when it is set?  Is it that the overwrite mode is what is used to fix the error, were it to occur on a non-overwrite run?    (Briefly comment...)

edx/analytics/tasks/elasticsearch_load.py
```diff
@@ -0,0 +1,292 @@
...
+        if es.indices.exists(index=self.index):
+            es.indices.delete(index=self.index)
+
+        settings = {
+            'refresh_interval': -1,
+        }
+        if self.number_of_shards is not None:
+            settings['number_of_shards'] = self.number_of_shards
+
+        if self.settings:
```
brianhw, line 162: Where does this get set?  I see it being initialized to {} along with self.properties, but I don't see where these are populated.

edx/analytics/tasks/elasticsearch_load.py
```diff
@@ -0,0 +1,292 @@
...
+                except TransportError as e:
+                    if e.status_code != REJECTED_REQUEST_STATUS:
+                        raise e
+                else:
+                    num_errors = 0
+                    for raw_data in resp['items']:
+                        op_type, item = raw_data.popitem()
+                        ok = 200 <= item.get('status', 500) < 300
+                        if not ok:
+                            log.error('Failed to index: {0}'.format(str(raw_data)))
```
brianhw: Does this message actually log the item in the raw_data that wasn't okay?  Seems like it's been popped?

edx/analytics/tasks/elasticsearch_load.py
```diff
@@ -0,0 +1,292 @@
...
+                else:
+                    break
+
+            if succeeded:
+                self.incr_counter('Elasticsearch', 'Committed Batches', 1)
+                self.incr_counter('Elasticsearch', 'Records Indexed', num_records)
+            else:
+                raise RuntimeError('Batch of records rejected too many times. Aborting.')
+
+        yield ('', '')
```
brianhw, line 222: What does this do?  Is it necessary for the reducer to produce something for the job to be able to succeed?

edx/analytics/tasks/module_engagement.py
```diff
@@ -70,6 +89,8 @@ class ModuleEngagementDataTask(EventLogSelectionMixin, OverwriteOutputMixin, Map
     # Override superclass to disable this parameter
     interval = None
 
+    enable_direct_output = True
```
brianhw, line 101: Deserves a one-line comment, since this is our Luigi extension.

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+
+class ModuleEngagementSummaryDataTask(
+    ModuleEngagementDownstreamMixin, OverwriteOutputMixin, MapReduceJobTask
+):
     """
-    Replicate the hive table in Vertica.
+    Store a summary of student engagement with their courses on particular dates aggregated using a sliding window.
 
-    Note that it is updated incrementally.
+    Only emits a record if the user did something in the course on that particular day, this dramatically reduces the
```
brianhw: nit:   'day, this' => 'day. This'

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+    volume of data in this table and keeps it manageable.
+
+    Currently this analyzes a sliding one week window of data to generate the summary.
     """
 
+    # NOTE: We could have done this in Hive, but this MR job is much more simple than the necessary Hive query.
```
brianhw, line 467: nit:  more simple => simpler

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+            self.interval.date_a,
+            self.interval.date_b,
+            output_record.problem_attempts,
+            len(output_record.problems_attempted),
+            len(output_record.problems_completed),
+            attempts_per_completion,
+            len(output_record.videos_viewed),
+            output_record.discussion_contributions,
+            len(output_record.days_active)
+        ).to_string_tuple()
```
brianhw: Not sure where the best place to draw the line between this reducer() method and the ModuleEngagementSummaryRecordBuilder().  As we see here, the latter doesn't actually construct the ModuleEngagementSummaryRecord object.  It only maintains some state, and the logic of what it keeps is intimately tied to the login in the "for line in lines" loop.   Maybe I'm just having trouble with the suffix "Builder", and it would feel better if it were instead "State" or "Aggregate" or something, so it doesn't promise as much.

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
         )
         return partition_task.data_task
 
+    def mapper(self, line):
+        record = ModuleEngagementSummaryRecord.from_tsv(line)
+        yield record.course_id, line.rstrip('\n')
+
+    def reducer(self, course_id, lines):
+        """
+        Analyze all summary records for particular course.
```
brianhw: nit: for a particular course.

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+
+        This will include all students performed any activity of interest in the past week.
```
brianhw: nit: "students who performed" or "students that performed".

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+                if value != 0:
+                    # NOTE: a lot of people don't participate, so don't include them in the analysis. Otherwise it would
+                    # look like doing *anything* in the course meant you were doing really well.
+                    metric_values[metric].append(value)
+
+        for metric in sorted(metric_values):
+            values = metric_values[metric]
+            range_values = numpy.percentile(  # pylint: disable=no-member
+                values, [0.0, self.low_percentile, self.high_percentile, 100.0]
+            )
```
brianhw, line 622: It would help me to be reminded what the values of range_values are.  Is it practical to not use range_values, and just use explicit values?  Like range_min, range_low, range_high, range_max = numpy.percentile(...)

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+            date=self.date,
+            n_reduce_tasks=self.n_reduce_tasks,
+            output_root=self.partition_location,
+        )
+
+
+class ModuleEngagementSummaryMetricRangesMysqlTask(ModuleEngagementDownstreamMixin, MysqlInsertTask):
+    """Result store storage for the metric ranges."""
+
+    overwrite = luigi.BooleanParameter(default=True)
```
brianhw: as opposed to using OverwriteMixin?

edx/analytics/tasks/tests/map_reduce_mixins.py
```diff
@@ -49,7 +52,7 @@ def create_task(self, **kwargs):
         merged_arguments.update(kwargs)
 
         for attr in merged_arguments:
-            if not hasattr(self.task_class, attr):
+            if getattr(self.task_class, attr, None) is None:
```
brianhw, line 55: Hmm.  I thought there might be a case where we wanted the ability to set an attribute value to None, to replace a default value of non-None.   I can't name it now, but that's my sense.  Is this fixing an undesirable behavior in the code?

edx/analytics/tasks/tests/map_reduce_mixins.py
```diff
@@ -162,16 +169,18 @@ def setUp(self):
...
         self.task = self.task_class(**new_kwargs)  # pylint: disable=not-callable
-
         self.task.init_local()
-        self.reduce_key = tuple()
 
     def assert_no_output(self, input_value):
         """Asserts that the given input produces no output."""
         output = self._get_reducer_output(input_value)
-        self.assertEquals(len(output), 0)
+        self.assertEquals(len(output), 0, output)
```
brianhw: Huh!  Who knew?  It turns out that as of 2.7, assertEquals() has been deprecated in favor of assertEqual().  https://docs.python.org/2/library/unittest.html#deprecated-aliases

edx/analytics/tasks/util/boto_connection.py
```diff
@@ -0,0 +1,74 @@
...
+    def __init__(self, host='localhost', port=443, aws_access_key_id=None, aws_secret_access_key=None,
+                 region=None, **kwargs):
+        super(BotoHttpConnection, self).__init__(host=host, port=port, **kwargs)
+        connection_params = {'host': host, 'port': port}
+
+        # If not provided, boto will attempt to use default environment variables to fill
+        # the access credentials.
+        connection_params['aws_access_key_id'] = aws_access_key_id
+        connection_params['aws_secret_access_key'] = aws_secret_access_key
+        connection_params['region'] = region
```
brianhw: Why are the first two args in one line, and these three are on separate lines?  Grouped because of the comment?

edx/analytics/tasks/util/boto_connection.py
```diff
@@ -0,0 +1,74 @@
...
+        # Remove 'None' values so that we don't overwrite defaults
+        connection_params = {key: val for key, val in connection_params.items() if val is not None}
+        self.connection = ESConnection(**connection_params)
+
+    # pylint: disable=unused-argument
+    def perform_request(self, method, url, params=None, body=None, timeout=None, ignore=()):
+        """
+        Called when making requests elasticsearch.  Requests are signed and
```
brianhw: nit: requests _to_ elasticsearch

edx/analytics/tasks/util/boto_connection.py
```diff
@@ -0,0 +1,74 @@
...
+        http status, headers, and response is returned.
```
brianhw: nit: s/is/are

edx/analytics/tasks/util/boto_connection.py
```diff
@@ -0,0 +1,74 @@
+import json
```
brianhw: Naming issue:  this functionality is elasticsearch-specific, rather than generic boto.  Should that be indicated in the name somehow?  Even es_boto_connection.py?    Or elasticsearch_connection.py?

edx/analytics/tasks/util/boto_connection.py
```diff
@@ -0,0 +1,74 @@
...
+        See: https://github.com/boto/boto/blob/develop/boto/connection.py#L533
+        """
+        if not isinstance(body, basestring):
+            body = json.dumps(body)
+        start = time.time()
+        response = self.connection.make_request(method, url, params=params, data=body)
+        duration = time.time() - start
+        raw_data = response.read()
+
+        # raise errors based on http status codes and let the client handle them
```
brianhw: nit:  I thought our style was full sentences.  Init cap at beginning, period at end.  That way one can add additional sentences later, without having to change the existing comments.

edx/analytics/tasks/util/elasticsearch_target.py
```diff
@@ -0,0 +1,70 @@
...
+            self.es.get(index=self.marker_index, doc_type=self.marker_doc_type, id=self.marker_index_document_id())
+            return True
+        except elasticsearch.NotFoundError:
+            log.debug('Marker document not found.')
+        except elasticsearch.ElasticsearchException as err:
+            log.warn(err)
+        return False
+
+    def create_marker_index(self):
+        if not self.es.indices.exists(index=self.marker_index):
```
brianhw: I assume this is internal, so the name can begin with underscore.  This is only called the once, and is two lines, so it could even be inlined.

But comparing this with exists()' use of self.es.get instead of self.es.indices.exists(), is there a reason to use one in one place and the other in the other?

edx/analytics/tasks/util/record.py
```diff
@@ -250,6 +259,14 @@ def to_ordered_dict(self):
 
         return field_values
 
+    def to_separated_values(self, sep='\t', null_value=DEFAULT_NULL_VALUE):
+        """
+        Convert this record to a string with fields delimited by `sep`.
+
+        Returns: a UTF8 string representation of the record.
```
brianhw, line 270: Does the utf8-ness of this depend at all on the sep that is passed in?  Or is it enough that to_string_tuple() returns utf8 strings?

edx/analytics/tasks/util/record.py
```diff
@@ -381,15 +427,21 @@ def hive_type(self):
...
     """Represents a field that contains a relatively short string."""
 
     hive_type = 'STRING'
+    elasticsearch_type = 'string'
 
     def __init__(self, **kwargs):
         super(StringField, self).__init__(**kwargs)
-        self.length = kwargs.pop('length', None)
+        if not hasattr(self, 'length'):
+            self.length = None
```
brianhw, line 478: Not clear on how this changes things.  Isn't it still coming from kwargs?

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+            values.
+        highly_engaged: The user is in the top 15% for any of the metrics except problem_attempts
+            and problem_attempts_per_completed.
+    """
+
+    output_root = luigi.Parameter()
+
+    def requires_local(self):
+        # NOTE: Most of this table is pickled on the master node and distributed to the slaves via the distributed
+        # cache.
```
brianhw: I'd like to get some background on how this pickling works.  Don't know that it needs to go in here...

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+        This can be used by downstream map reduce jobs to read all of the raw data.
+        """
+        for task in self.requires():
+            if isinstance(task, ModuleEngagementPartitionTask):
+                yield task.data_task
+
+
+class ModuleEngagementSummaryRecord(Record):
+    """
+    Summarizes a user's engagement with a particular course on a particular day with simple counts of activity.
```
brianhw: Is this for a particular day?  Seems to be a time interval, defined by start_date and end_date, no?

edx/analytics/tasks/module_engagement.py
```diff
@@ -41,7 +56,11 @@ class ModuleEngagementDownstreamMixin(
     """Common parameters and base classes used to pass parameters through the engagement workflow."""
 
     # Required parameter
-    date = luigi.DateParameter()
+    date = luigi.DateParameter(
+        description='Upper bound date for the end of the interval to analyze. Data produced before midnight on this'
+                    ' date will be analyzed. This workflow is intended to run nightly and this parameter is intended'
+                    ' to be set to "today\'s" date.'
+    )
```
brianhw, line 70: So if this is set to today's date, it will include all data for yesterday.  When I first read this, I interpreted "midnight on this date" as being the end of the day, not the very beginning.  Maybe adding a clarifying clause at the end:  "date, so that all of yesterday's data is included and none of today's."

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+        yield (record.course_id, record.username), line.rstrip('\n')
+
+    def reducer(self, key, lines):
+        course_id, username = key
+
+        records = [ModuleEngagementSummaryRecord.from_tsv(line) for line in lines]
+
+        # Maps segment names to a set of "reasons" that tell why the user was placed in that segment
+        segments = defaultdict(set)
+        most_recent_summary = records[-1]
```
brianhw: I don't understand the name of this.  Do you assume that the records presented to the reducer are in sorted order, from least recent to most recent?

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+        self.high_metric_ranges = defaultdict(dict)
+
+        # Grab all of the "high" ranges since those are the only ones we are going to use. This will load
+        # O(num_courses * num_metrics) records into memory, and then into the pickled job instance. This should be a
+        # relatively small number (thousands).
+        with self.input_local()['range_data'].open('r') as metric_ranges_target:
+            for line in metric_ranges_target:
+                range_record = ModuleEngagementSummaryMetricRangeRecord.from_tsv(line)
+                if range_record.range_type == METRIC_RANGE_HIGH:
+                    self.high_metric_ranges[range_record.course_id][range_record.metric] = range_record
```
brianhw, line 788: There should only be one record for a given course_id/metric pair, right?  So this shouldn't be overwriting anything...

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+                    segments[SEGMENT_STRUGGLING].add(metric)
+                else:
+                    segments[SEGMENT_HIGHLY_ENGAGED].add(metric)
+
+        for segment in sorted(segments):
+            yield ModuleEngagementUserSegmentRecord(
+                course_id=course_id,
+                username=username,
+                start_date=records[0].start_date,
+                end_date=records[0].end_date,
```
brianhw, line 830: Do we want this to work for multiple time intervals in future?  I guess that would just entail putting the start_date and end_date into the key.

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
     """
 
-    vertica_schema = luigi.Parameter(default=None)
-    vertica_credentials = luigi.Parameter(default=None)
+    course_id = StringField()
+    username = StringField()
+    start_date = DateField()
+    end_date = DateField()
+    segment = StringField()
+    reason = StringField()
```
brianhw: Add descriptions for these, like was done for the roster?

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+                                                        ' made in the course.')
+    segments = StringField(analyzed=True, description='Classifiers that help group learners by analyzing their activity'
+                                                      ' patterns. Example: "inactive" indicates the user has not'
+                                                      ' engaged with the course recently. This field is analyzed by'
+                                                      ' elasticsearch so that searches can be made for learners that'
+                                                      ' either have or don\'t have particular classifiers.')
+    attempt_ratio_order = IntegerField(description='Value can be -1, 0, or 1 and allows the learners to be sorted by'
+                                                   ' their problem_attempts_per_completed ratio in a way that'
+                                                   ' highlights highly engaged learners and struggling learners,'
+                                                   ' depending on the sort order.')
```
brianhw: I can't tell from this description what -1, 0 or 1 mean here.

edx/analytics/tasks/module_engagement.py
```diff
@@ -224,6 +244,8 @@ class ModuleEngagementMysqlTask(ModuleEngagementDownstreamMixin, IncrementalMysq
     This allows us to rapidly generate activity charts over time for small numbers of users.
     """
 
+    allow_empty_insert = True
```
HassanJaveed84: When would be have an empty insert and why ?

edx/analytics/tasks/module_engagement.py
HassanJaveed84: I agree, had the same notion while reading the definition of ModuleEngagementSummaryRecordBuilder().

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+            n_reduce_tasks=self.n_reduce_tasks,
+            output_root=self.partition_location,
+        )
+
+
+class ModuleEngagementRosterRecord(Record):
+    course_id = StringField(description='Course the learner is enrolled in.')
+    username = StringField(description='Learner\'s username.')
+    start_date = DateField(description='Analysis includes all data from midnight on this day.')
+    end_date = DateField(description='Analysis includes all data up to but not including midnight of this date.')
```
HassanJaveed84: I'm slightly confused by this. By midnight of this date, do you mean midnight at the beginning of the day or at eod ?

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+
+    date = luigi.DateParameter()
+    interval = None
+    partition_value = None
+
+    def __init__(self, *args, **kwargs):
+        super(ModuleEngagementRosterPartitionTask, self).__init__(*args, **kwargs)
+
+        start_date = self.date - datetime.timedelta(weeks=1)
+        self.interval = date_interval.Custom(start_date, self.date)
```
HassanJaveed84: Does it make sense to pull this logic into some kind of helper method as it appears in ModuleEngagementSummaryDataTask as well ?

edx/analytics/tasks/elasticsearch_load.py
```diff
@@ -0,0 +1,292 @@
...
+
+    def commit(self):
+        es = self.create_elasticsearch_client()
+        es.indices.refresh(index=self.index)
+        actions = []
+        for old_index in self.indexes_for_alias:
+            actions.append({"remove": {"index": old_index, "alias": self.alias}})
+        actions.append({"add": {"index": self.index, "alias": self.alias}})
+        es.indices.update_aliases({"actions": actions})
+        self.output().touch()
```
brianhw, line 391: Wondering if this should be named more explicitly:  i.e. touch() => mark_as_succeeded().   To me, touch() merely suggests updating the last-modified property on something, and this does more than that.

edx/analytics/tasks/module_engagement.py
brianhw: I presume when no users have had any activity on a given day.  Less likely in production, but often the norm for developers.

edx/analytics/tasks/module_engagement.py
brianhw: Actually, it looks to me like MysqlInsertTask already includes OverwriteOutputMixin.  So I now see that the difference here is in the default value.  I would add a comment explaining that's the reason this is defined here.   Actually that may arise when you add the description to this parameter.

edx/analytics/tasks/module_engagement.py
brianhw: Yes, I had the same issue at the top of this PR.  Midnight is the beginning of the day, i.e. 00:00.  One could avoid the "midnight" ambiguity by just dropping it:  "up to but not including this date".

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+    problem_attempts_per_completed = FloatField(description='Ratio of the number of attempts the learner has made on'
+                                                            ' any problem to the number of unique problems they have'
+                                                            ' completed correctly in the course.')
+    videos_viewed = IntegerField(description='Number of unique videos the learner has watched any part of in the'
+                                             ' course.')
+    discussion_contributions = IntegerField(description='Total number of posts, responses and comments the learner has'
+                                                        ' made in the course.')
+    segments = StringField(analyzed=True, description='Classifiers that help group learners by analyzing their activity'
+                                                      ' patterns. Example: "inactive" indicates the user has not'
+                                                      ' engaged with the course recently. This field is analyzed by'
```
brianhw, line 909: "analyzed"?  Or just "indexed"?

edx/analytics/tasks/module_engagement.py
brianhw: Something like WeekIntervalMixin?  Maybe set self.end_date = self.date first.

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+        query = """
+        USE {database_name};
+        INSERT OVERWRITE TABLE {table} PARTITION ({partition.query_spec}) {if_not_exists}
+        SELECT
+            ce.course_id,
+            au.username,
+            '{start}',
+            '{end}',
+            au.email,
+            regexp_replace(regexp_replace(aup.name, '\\\\t|\\\\n|\\\\r', ' '), '\\\\\\\\', ''),
```
brianhw, line 971: Can we (or should we) do this normalization in auth_userprofile export somehow, so we don't have to perform this in every user of aup...

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+                ",",
+                IF(ce.at_end = 0, "unenrolled", NULL),
+                IF(COALESCE(old_eng.days_active, 0) = 0 AND COALESCE(eng.days_active, 0) = 0, "inactive", NULL),
+                IF(COALESCE(old_eng.days_active, 0) > 0 AND COALESCE(eng.days_active, 0) = 0, "disengaging", NULL),
+                seg.segments
+            ),
+            IF(
+                eng.problem_attempts_per_completed IS NULL,
+                -COALESCE(eng.problem_attempts, 0),
+                COALESCE(eng.problem_attempts, 0)
```
brianhw, line 1004: I think we discovered that we couldn't put inline comments into these queries.  But there are several calculations here that would benefit from having some comments nearby.  Like explaining why we want a negative value for this (including negative zero?).

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+        super(ModuleEngagementRosterPartitionTask, self).__init__(*args, **kwargs)
+
+        start_date = self.date - datetime.timedelta(weeks=1)
+        self.interval = date_interval.Custom(start_date, self.date)
+        self.partition_value = self.date.isoformat()
+
+    def query(self):
+        # Join with calendar data only if calculating weekly engagement.
+        last_complete_date = self.interval.date_b - datetime.timedelta(days=1)  # pylint: disable=no-member
+        iso_weekday = last_complete_date.isoweekday()
```
brianhw: Not clear why we need isoweekday() here.  Isn't it enough to just have the query below use "where ce.date = last_complete_date.isoformat()" in the where clause, rather than "ce.date < end and ce.date >= start and cal.isoweekday = iso_weekday"...

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+            ImportCourseUserGroupUsersTask(**kwargs_for_db_import),
+            ImportAuthUserProfileTask(**kwargs_for_db_import),
+        )
 
-    def output(self):
-        return [task.output() for task in self.requires()]
+
+@workflow_entry_point
+class ModuleEngagementRosterIndexTask(ModuleEngagementDownstreamMixin, ElasticsearchIndexTask):
+    __doc__ = """
```
brianhw: Is the way we do docstrings that this **doc** only gets defined for @workflow_entry_point tasks?

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+    A rapidly searchable learner roster for each course with aggregate statistics about that learner's performance.
+
+    Each record written to the elasticsearch index represents a single learner's performance in the course in the last
+    week. Note that the index will contain records for users that have since unenrolled in the course, so there will
+    be one record for every user who has ever been enrolled in that course.
+
+    Given that each week the learner's statistics may have changed, the entire index is re-written every time this task
+    is run. Elasticsearch does not support transactional updates. For this reason, we chose to write to a separate index
+    on each update, using an alias to point to the "live" index while we build the next version. Once the next version
+    is complete, we switch over the index to point to the newly built index and drop the old one. This allows us to
```
HassanJaveed84: Do you mean 'we switch over the 'alias' to point to the newly built index' ?

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+    continue to expose a consistent view of the roster while we are writing out a new version. Without an atomic toggle
+    like this, it is possible an instructor could run a query and see a mix of data from the old and new versions of
+    the index.
+
+    For more information about this strategy see `Index Aliases and Zero Downtime`_.
+
+    We also chose to organized the data in a single index that contains the data for all courses. We rely on the default
```
HassanJaveed84: nit: organized => organize

edx/analytics/tasks/module_engagement.py
brianhw, line 971: Also, do we only have to worry about slashes?  There is quite a variety of cruft in user names, in my experience.  I expect that some of it could actually be cleaned up before display in the UI, rather than in the pipeline, but if this is being used to drive search terms, then some greater form of normalization may be useful to do.  E.g. removing quotes, asterisks, and other random (or not-so-random) punctuation.

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+            eng.problem_attempts_per_completed,
+            COALESCE(eng.videos_viewed, 0),
+            COALESCE(eng.discussion_contributions, 0),
+            CONCAT_WS(
+                ",",
+                IF(ce.at_end = 0, "unenrolled", NULL),
+                IF(COALESCE(old_eng.days_active, 0) = 0 AND COALESCE(eng.days_active, 0) = 0, "inactive", NULL),
+                IF(COALESCE(old_eng.days_active, 0) > 0 AND COALESCE(eng.days_active, 0) = 0, "disengaging", NULL),
+                seg.segments
+            ),
```
brianhw, line 989: I'm hoping that CONCAT_WS will skip over arguments that are equal to NULL.  And if not, that someone else will.

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+            ON (au.id = cohort.user_id AND ce.course_id = cohort.course_id)
+        LEFT OUTER JOIN module_engagement_summary eng
+            ON (ce.course_id = eng.course_id AND au.username = eng.username AND eng.end_date = '{end}')
+        LEFT OUTER JOIN module_engagement_summary old_eng
+            ON (ce.course_id = old_eng.course_id AND au.username = old_eng.username AND old_eng.end_date = DATE_SUB('{end}', 7))
+        LEFT OUTER JOIN (
+            SELECT
+                course_id,
+                user_id,
+                MAX(date) AS last_enrollment_date
```
brianhw: The record documentation talked about first_enrollment_date.  Should we be storing first or last?

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+class ModuleEngagementRosterRecord(Record):
+    course_id = StringField(description='Course the learner is enrolled in.')
+    username = StringField(description='Learner\'s username.')
+    start_date = DateField(description='Analysis includes all data from midnight on this day.')
+    end_date = DateField(description='Analysis includes all data up to but not including midnight of this date.')
+    email = StringField(description='Learner\'s email address.')
+    name = StringField(analyzed=True, description='Learner\'s full name including first, middle and last names. '
+                                                  'This field can be searched by instructors.')
+    enrollment_mode = StringField(description='Learner is enrolled in the course with this mode. Example: verified.')
+    enrollment_date = DateField(description='First date the learner enrolled in the course.')
```
brianhw, line 893: In the calculation below, I'm seeing last_enrollment_date being found, not first.  Which do we want?

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+        WHERE
+            ce.date >= '{start}'
+            AND ce.date < '{end}'
+            AND cal.iso_weekday = {iso_weekday}
+        """.format(
+            start=self.interval.date_a.isoformat(),  # pylint: disable=no-member
+            end=self.interval.date_b.isoformat(),  # pylint: disable=no-member
+            iso_weekday=iso_weekday,
+            partition=self.partition,
+            if_not_exists='' if self.overwrite else 'IF NOT EXISTS',
```
brianhw, line 1053: Just so I understand:  we could always include 'IF NOT EXISTS'.  But we don't include it on overwrite because we expect the table to be deleted, and so should not exist already?

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+                IF(COALESCE(old_eng.days_active, 0) > 0 AND COALESCE(eng.days_active, 0) = 0, "disengaging", NULL),
+                seg.segments
+            ),
+            IF(
+                eng.problem_attempts_per_completed IS NULL,
+                -COALESCE(eng.problem_attempts, 0),
+                COALESCE(eng.problem_attempts, 0)
+            )
+        FROM course_enrollment ce
+        INNER JOIN calendar cal ON (ce.date = cal.date)
```
brianhw: As mentioned above, I don't understand why we include calendar here.  It is only used in the where clause, and don't see what value it provides there.

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+                date=(self.date - datetime.timedelta(weeks=1)),
+                n_reduce_tasks=self.n_reduce_tasks,
+                overwrite=self.overwrite,
+            ),
+            CourseEnrollmentTableTask(
+                interval_end=self.date,
+                n_reduce_tasks=self.n_reduce_tasks,
+                overwrite=self.overwrite,
+            ),
+            CalendarTableTask(),
```
brianhw: If we don't actually need the calendar, then this can be removed.

edx/analytics/tasks/module_engagement.py
brianhw: Hmm.  Maybe I should have read this first.  Perhaps we should put a pointer at the top of the document, saying that further information about the workflow implemented here can be be found in the docstring for the ModuleEngagementRosterIndexTask below.

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+class ModuleEngagementRosterIndexTask(ModuleEngagementDownstreamMixin, ElasticsearchIndexTask):
+    __doc__ = """
+    A rapidly searchable learner roster for each course with aggregate statistics about that learner's performance.
+
+    Each record written to the elasticsearch index represents a single learner's performance in the course in the last
+    week. Note that the index will contain records for users that have since unenrolled in the course, so there will
+    be one record for every user who has ever been enrolled in that course.
+
+    Given that each week the learner's statistics may have changed, the entire index is re-written every time this task
+    is run. Elasticsearch does not support transactional updates. For this reason, we chose to write to a separate index
```
brianhw: nit: I would remove the "choice" here.    "For this reason, we write to a separate index..."

edx/analytics/tasks/module_engagement.py
brianhw: nit: I suggest:  "We organize the data"...  Forget about highlighting the "choice" involved.

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+    is complete, we switch over the index to point to the newly built index and drop the old one. This allows us to
+    continue to expose a consistent view of the roster while we are writing out a new version. Without an atomic toggle
+    like this, it is possible an instructor could run a query and see a mix of data from the old and new versions of
+    the index.
+
+    For more information about this strategy see `Index Aliases and Zero Downtime`_.
+
+    We also chose to organized the data in a single index that contains the data for all courses. We rely on the default
+    elasticsearch sharding strategy instead of manually attempting to shard the data by course (or some other
+    dimension). This choice was made largely because it is more simple to implement and manage.
```
brianhw: /was/is/

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+
+    The index is optimized for the following access patterns within a single course:
+
+    - Find learners by some part of their name. For example: "John" or "John Doe".
```
brianhw: For my edification :  Username has to be exact,  I understand.  But names do not.  Are there implicit boundaries, however?  In particular, if I search for "ohn", will it find "John Doe"?

edx/analytics/tasks/module_engagement.py
```diff
@@ -276,85 +298,885 @@ def insert_source_task(self):
...
+                    if maybe_null_field == 'segments':
+                        maybe_null_value = maybe_null_value.split(',')
+                    document['_source'][maybe_null_field] = maybe_null_value
+
+            original_id = document['_id']
+            for i in range(self.scale_factor):
+                document = document.copy()
+                if i > 0:
+                    document['_id'] = original_id + '|' + str(i)
+                yield document
```
brianhw, line 1180: Maybe it's not a big deal, but we always make the copy every time, even if  i == 0.  Could move the document.copy() inside the if I > 0: clause.

edx/analytics/tasks/tests/map_reduce_mixins.py
brianhw: Maybe the message should be something like "Expected no output, found {}".format(output).

edx/analytics/tasks/tests/test_elasticsearch_load.py
```diff
@@ -0,0 +1,179 @@
...
+    def doc_type(self):
+        return 'raw_text'
+
+    def document_generator(self, lines):
+        pass
+
+
+class ElasticsearchIndexTaskReduceTest(BaseIndexTest, ReducerTestMixin, unittest.TestCase):
+
+    def test_foo(self):
```
brianhw: Is this needed?  What does it test?

edx/analytics/tasks/tests/test_elasticsearch_load.py
```diff
@@ -0,0 +1,179 @@
...
+    def test_other_settings(self):
+        self.task.settings = {
+            'foo': 'bar'
+        }
+        self.task.init_local()
+        _args, kwargs = self.mock_es.indices.create.call_args
+        self.assertEqual(kwargs['body']['settings'], {'refresh_interval': -1, 'foo': 'bar'})
+
+    def test_boto_connection_type(self):
+        self.create_task(connection_type='boto')
```
brianhw: Do we need to test the other connection type as well?

edx/analytics/tasks/tests/test_enrollment_validation.py
```diff
@@ -147,8 +147,8 @@ def setUp(self):
         self.reduce_key = ('foo/bar/baz', 0)
         self.mode = 'honor'
 
-    def create_task(self, generate_before=True, tuple_output=True, include_nonstate_changes=True,
-                    earliest_timestamp=None, expected_validation=None):
+    def create_validation_task(self, generate_before=True, tuple_output=True, include_nonstate_changes=True,
```
brianhw, line 150: I don't see where this is added.  Is this new?  Same with create_enrollment_task in the next file.

edx/analytics/tasks/tests/test_module_engagement.py
```diff
@@ -1,14 +1,20 @@
...
 
 import luigi
 from ddt import ddt, data, unpack
+from mock import MagicMock
 
-from edx.analytics.tasks.module_engagement import ModuleEngagementDataTask
+from edx.analytics.tasks.module_engagement import ModuleEngagementDataTask, ModuleEngagementSummaryDataTask, \
+    ModuleEngagementRecord, ModuleEngagementSummaryRecord, ModuleEngagementSummaryMetricRangesDataTask, \
+    ModuleEngagementSummaryMetricRangeRecord, ModuleEngagementUserSegmentDataTask, ModuleEngagementUserSegmentRecord, \
+    ModuleEngagementRosterIndexTask, ModuleEngagementRosterRecord
```
brianhw: nit: my understanding is that we prefer using parentheses over backslashes.

edx/analytics/tasks/tests/test_module_engagement.py
```diff
@@ -193,3 +199,810 @@ def test_replacement_of_count(self):
             inputs,
             (('\t'.join(self.reduce_key), 4),)
         )
+
+
+@ddt
+class ModuleEngagementSummaryDataTaskMapTest(MapperTestMixin, unittest.TestCase):
+    """Base class for test analysis of student engagement summaries"""
```
HassanJaveed84, line 207: Should we have separate docstrings for the mapper and reducer tests ?

edx/analytics/tasks/elasticsearch_load.py
mulby: According to this page: https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

That is actually the correct behavior. Note that this snippet is very similar to a snippet in the elasticsearch-py library that handles a similar case.

edx/analytics/tasks/elasticsearch_load.py
mulby, line 90: Agreed. Done.

edx/analytics/tasks/elasticsearch_load.py
mulby: Good catch, removed. This was, in fact, redundant.

edx/analytics/tasks/elasticsearch_load.py
mulby, line 145: Correct. The idea is that --overwrite could be used to recover from wacky scenarios.

edx/analytics/tasks/elasticsearch_load.py
mulby, line 162: The idea is that it can be overridden by classes that inherit from this base class.

edx/analytics/tasks/elasticsearch_load.py
mulby: Good catch, fixed.

edx/analytics/tasks/util/record.py
mulby, line 478: Yes it is, see the super() constructor it basically just sets all of the kwargs as attributes on the Field object. This code is just validating the length parameter. I'll restructure a bit to make that more clear.

edx/analytics/tasks/module_engagement.py
mulby: This is default luigi behavior, I've changed around these comments to hopefully clarify this.

edx/analytics/tasks/module_engagement.py
mulby: Yeah this is a vestigial name. There should only be a single record here. I've added in an explicit check and renamed this variable to hopefully clarify the situation.

I used to actually pull two summary records here, but that is no longer necessary.

edx/analytics/tasks/module_engagement.py
mulby: @brianhw is correct, however, I've changed this to be a parameter since we likely only want this behavior in our development environment. In production it's still a useful check. Small installations can disable this behavior if they, in fact, have days with no user activity.

edx/analytics/tasks/module_engagement.py
mulby: Good catch, I've used your simplified logic here.

edx/analytics/tasks/module_engagement.py
mulby, line 971: This is really just about removing delimiters from the data. Elasticsearch will ignore characters that it doesn't recognize when it indexes the name.

edx/analytics/tasks/module_engagement.py
mulby: Great catch. It should be first, fixed.

edx/analytics/tasks/module_engagement.py
mulby, line 893: First. Fixed.

edx/analytics/tasks/module_engagement.py
mulby, line 1053: Not exactly, note that the beginning of the statement is "INSERT OVERWRITE TABLE {table} PARTITION ({partition.query_spec}). So, by default, it will overwrite the partition whether or not it exists. If we specify "IF NOT EXISTS" it will prevent hive from overwriting the partition if it already exists.

edx/analytics/tasks/module_engagement.py
mulby: concur, fixed.

edx/analytics/tasks/module_engagement.py
mulby: Fixed.

edx/analytics/tasks/module_engagement.py
mulby: Yes, there are implicit boundaries. We rely on elasticsearch's analyzer to tokenize the string as it sees fit. We could customize the analyzer but we currently use the default, which I believe tokenizes on whitespace, so "ohn" would not match "John", but "john" and "John" would match.

edx/analytics/tasks/tests/test_enrollment_validation.py
mulby, line 150: It's because I now define a default "create_task" method in the base class and this needs specialized behavior, so I renamed it to avoid the naming conflict.

edx/analytics/tasks/tests/test_module_engagement.py
mulby, line 207: Probably, I'll take a pass over all pylint, pep8 and docstrings before merging.

edx/analytics/tasks/elasticsearch_load.py
mulby: done.

config/devstack.cfg
mulby, line 74: done.

edx/analytics/tasks/elasticsearch_load.py
mulby, line 31: done

edx/analytics/tasks/elasticsearch_load.py
mulby: done

edx/analytics/tasks/elasticsearch_load.py
mulby: done

edx/analytics/tasks/elasticsearch_load.py
mulby, line 222: done.

edx/analytics/tasks/module_engagement.py
mulby, line 101: done.

edx/analytics/tasks/module_engagement.py
mulby: done.

edx/analytics/tasks/module_engagement.py
mulby, line 467: done.

edx/analytics/tasks/module_engagement.py
mulby: done.

edx/analytics/tasks/module_engagement.py
mulby: done.

edx/analytics/tasks/module_engagement.py
mulby: done.

edx/analytics/tasks/module_engagement.py
mulby, line 622: done

edx/analytics/tasks/module_engagement.py
mulby: done.

edx/analytics/tasks/tests/map_reduce_mixins.py
mulby, line 55: Yes, in this case, self.task_class overrides the "interval" parameter with the value `None` since it is defined by the parent class but computed by the child class.

I'll change this logic to check if the attribute is a parameter.

edx/analytics/tasks/tests/map_reduce_mixins.py
mulby: done.

edx/analytics/tasks/util/boto_connection.py
mulby: correct

edx/analytics/tasks/util/boto_connection.py
mulby: done

edx/analytics/tasks/util/boto_connection.py
mulby: done

edx/analytics/tasks/util/boto_connection.py
mulby: done

edx/analytics/tasks/util/boto_connection.py
mulby: done

edx/analytics/tasks/util/elasticsearch_target.py
mulby: inlined the method.

es.get() checks if a particular record exists within an index. es.indices.exists() checks if the index itself exists.

edx/analytics/tasks/util/record.py
mulby, line 270: done.

edx/analytics/tasks/module_engagement.py
mulby: done.

edx/analytics/tasks/module_engagement.py
mulby, line 70: done.

edx/analytics/tasks/module_engagement.py
mulby, line 788: True, was it implied that it would be overwriting?

edx/analytics/tasks/module_engagement.py
mulby, line 830: I think we can tackle that if/when that ever becomes a requirement.

edx/analytics/tasks/module_engagement.py
mulby: done.

edx/analytics/tasks/module_engagement.py
mulby: done

edx/analytics/tasks/module_engagement.py
mulby: done

edx/analytics/tasks/elasticsearch_load.py
mulby, line 391: Other parts of the codebase use touch, so I'm inclined to keep it.

edx/analytics/tasks/module_engagement.py
mulby, line 909: All fields are indexed by elasticsearch, but only some are "analyzed" this has special meaning in elasticsearch land.

It basically means that the string is tokenized and actually searchable. Other fields can only exact match.

edx/analytics/tasks/module_engagement.py
mulby, line 1004: I added some comments since I see them in other queries

edx/analytics/tasks/module_engagement.py
mulby: done.

edx/analytics/tasks/module_engagement.py
mulby: I did! good catch...

edx/analytics/tasks/module_engagement.py
mulby: done.

edx/analytics/tasks/module_engagement.py
mulby, line 989: It does!

edx/analytics/tasks/module_engagement.py
mulby: done

edx/analytics/tasks/module_engagement.py
mulby, line 1180: done, great catch...

edx/analytics/tasks/tests/test_elasticsearch_load.py
mulby: I hadn't finished the tests yet :smile:

edx/analytics/tasks/tests/test_elasticsearch_load.py
mulby: It's implicitly tested since the default value is "urllib"

edx/analytics/tasks/tests/test_module_engagement.py
mulby: We do, but pycharm prefers this style...

mulby:
@HassanJaveed84 I refactored the metric ranges code a little bit to make it slightly more clear, this will likely break your branch off of this branch. Hopefully it will be fairly straightforward to rebase on your end?

mulby:
@HassanJaveed84 and @brianhw - this is ready for final review

HassanJaveed84:
Already did :smile:

edx/analytics/tasks/elasticsearch_load.py
```diff
@@ -52,12 +54,12 @@ class ElasticsearchIndexTaskMixin(OverwriteOutputMixin):
         significant=False,
         default='urllib',
```
brianhw, line 55: nit:  this default value is actually never used, best I can tell.  The value either needs to be "aws" or something else, and setting to None seems marginally clearer.

edx/analytics/tasks/module_engagement.py
```diff
@@ -1,13 +1,26 @@
...
+"""
+
 from collections import defaultdict
 import datetime
 import logging
 import random
 
+log = logging.getLogger(__name__)
+
+# pylint: disable=wrong-import-position
```
brianhw, line 14: Maybe you could let me know offline what you discovered the expected import position stuff to be.   If it is just due to defining the log before the numpy import below (that uses it), then maybe we can put this on the log statement with a comment, so we know the reason going forward.

edx/analytics/tasks/module_engagement.py
```diff
@@ -536,15 +566,15 @@ class ModuleEngagementSummaryMetricRangeRecord(Record):
...
     """
     Summarize the metrics of interest and persist the interesting ranges.
 
-    This task currently computes "low" and "high" ranges. The "low" range is defined as the [min(), 15th percentile)
-    interval. The "high" range is defined as the [85th percentile, max()). There is an implied "middle" range that is
-    not persisted since it is fully defined by the low and the high. Note that all intervals are left-closed.
+    This task currently computes "low", "normal" and "high" ranges. The "low" range is defined as the [0, 15th
+    percentile) interval. The "normal" range is defined as [15th percentile, 85th percentile). The "high" range is
+    defined as the [85th percentile, float('inf')). Note that all intervals are left-closed. Also note that it is not
+    strictly necessary to persist all three intervals, however, we do so for clarity.
```
brianhw, line 577: It might be good to add mention somewhere of the assumption that the inputs to the ranges are [0, inf), since that assumption is built into the ranges that are output (at least at the moment).  Of course, if this is going to change soon, then hold off on doing so until the other work is done.

edx/analytics/tasks/module_engagement.py
```diff
@@ -587,13 +617,13 @@ def reducer(self, course_id, lines):
...
+            normal_lower_bound, normal_upper_bound = numpy.percentile(  # pylint: disable=no-member
                 values, [self.low_percentile, self.high_percentile]
             )
             ranges = [
-                (METRIC_RANGE_LOW, 0, range_values[0]),
-                (METRIC_RANGE_NORMAL, range_values[0], range_values[1]),
-                (METRIC_RANGE_HIGH, range_values[1], float('inf')),
+                (METRIC_RANGE_LOW, 0, normal_lower_bound),
+                (METRIC_RANGE_NORMAL, normal_lower_bound, normal_upper_bound),
+                (METRIC_RANGE_HIGH, normal_upper_bound, float('inf')),
```
brianhw, line 626: I mentioned this above, but it might be good to have a comment here as well that [0,inf) is assumed for the inputs, since it's built into the output ranges.

edx/analytics/tasks/module_engagement.py
```diff
@@ -330,20 +324,28 @@ def get_raw_data_tasks(self):
...
-    end_date = DateField()
-    problem_attempts = IntegerField(is_metric=True)
-    problems_attempted = IntegerField(is_metric=True)
-    problems_completed = IntegerField(is_metric=True)
-    problem_attempts_per_completed = FloatField(is_metric=True)
-    videos_viewed = IntegerField(is_metric=True)
-    discussion_contributions = IntegerField(is_metric=True)
-    days_active = IntegerField()
+    course_id = StringField(description='Course the learner interacted with.')
+    username = StringField(description='Learner\'s username.')
```
brianhw, line 331: Just curious:  are you proposing that we consider standardizing on using single quotes for descriptions, and escaping them when necessary?  I've never heard of any standard, and I've used double quotes when strings contain single quotes, because escapes are ugly to read.  And yet it's nice to be uniform, at least within a region of code.

edx/analytics/tasks/module_engagement.py
```diff
@@ -1145,8 +1185,9 @@ def document_generator(self, lines):
 SURNAMES = ['smith', 'johnson', 'williams', 'jones', 'brown', 'davis', 'miller', 'wilson', 'moore', 'taylor']
```
brianhw, line 1185: mulley, stroilova, friedman, jens, javeed, ...

edx/analytics/tasks/util/record.py
```diff
@@ -334,6 +339,17 @@ def get_hive_schema(cls):
 
     @classmethod
     def get_elasticsearch_properties(cls):
+        """
+        An elasticsearch mapping that could store this data.
+
+        This schema type recognizes the "analyzed" kwarg that can be passed into the Field definition. By default
+        Fields are not analyzed, however, if the Field is declared with analyzed=True, then it will be analyzed by
```
brianhw, line 346: nit:  s/analyzed, however/analyzed.  However/

edx/analytics/tasks/util/record.py
```diff
@@ -334,6 +339,17 @@ def get_hive_schema(cls):
...
+        elasticsearch.
+
+            foo = StringField(analyzed=True)
```
brianhw, line 349: Just looking at that, and imagining a random record Field with that, wouldn't be very clear what "analyzed" means.  Or maybe that's just my lack of familiarity with elasticsearch.  But if this is an elasticsearch property, perhaps it would be enough to call the property "elasticsearch".  Hence "foo = StringField(description = 'blah, blah', elasticsearch=True)".

brianhw, line 349: And then below, since we're adding "not_analyzed", we could define here "elasticsearch_analyzed" and "elasticsearch_not_analyzed", if the latter value is actually used somewhere.

brianhw:
Done with reviewing the diff since my last review.  Maybe one or two nits at most.  Biggest question for me is whether "analyze" is a clear name on a Field, but it's really only a suggestion for a better name, not a critical blocker.

PR #216, opened 2016-03-16
Add projections to Vertica tables.

brianhw:
Also defines projections to use for f_user_course.

@mulby @HassanJaveed84

### DIFF
--- a/edx/analytics/tasks/load_internal_reporting_user_course.py
+++ b/edx/analytics/tasks/load_internal_reporting_user_course.py
@@ -4,7 +4,9 @@ Loads the user_course table into the warehouse through the pipeline via Hive.
 import luigi
 import logging
 
-from edx.analytics.tasks.vertica_load import VerticaCopyTask
+from edx.analytics.tasks.vertica_load import (
+    VerticaCopyTask, VerticaProjection, PROJECTION_TYPE_NORMAL, PROJECTION_TYPE_AGGREGATE
+)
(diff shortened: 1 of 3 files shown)


edx/analytics/tasks/load_internal_reporting_user_course.py
```diff
@@ -52,3 +54,131 @@ def auto_primary_key(self):
...
+
+        # Use templates to format actual values.
+        return [
+            VerticaProjection
+            (
+                template.name.format(schema=self.schema, table=self.table),
+                template.type,
+                template.definition.format(schema=self.schema, table=self.table),
+            ) for template in projection_templates
+        ]
```
mulby: I feel like we are going to have to repeat this pattern anytime we override this property...

edx/analytics/tasks/vertica_load.py
```diff
@@ -172,6 +187,29 @@ def create_table(self, connection):
         log.debug(query)
         connection.cursor().execute(query)
 
+    def drop_aggregate_projections(self, connection):
+        """
+        Drop any projections that are aggregates.
+
+        Aggregate projections must be removed from a table before its contents can be deleted.
+        """
+        for projection in self.projections:
```
mulby: We should only do this if `self.overwrite` is truthy since that is the only time that a delete is executed. I want to make sure we don't break incremental loading.

brianhw: Done.  Moved call into init_copy

mulby:
:+1:

PR #218, opened 2016-03-20
Import order discount/coupon data into finance reports

bradenmacdonald:
This PR:
- Incorporates some basic data about coupons/discounts into various finance reports. (WL-364)
- Fixes [SOL-1662](https://openedx.atlassian.net/browse/SOL-1662) (using Otto enrollment codes caused error codes to appear in the audit reports.) - this part is the only change to any logic; the rest of this PR just adds new data columns.

The new columns are:
- discount_amount: The total $ amount by which the order was discounted, or $0 if there was no discount.
- voucher_code: the voucher code (e.g. `SVCAIGI3IVXJQKZ7` or `50_OFF_PROMO`)
- voucher_id: the ID of the voucher in the Otto/ShoppingCart database
- coupon_id: With Otto, a single coupon can have many voucher codes. The coupon_id can be used to tell that the voucher codes are all part of the same coupon. This coupon ID is what appears in the URL when viewing/editing a coupon - e.g. on devstack it's http://localhost:8002/coupons/**4**/
#### Basic requirements
1. Set up an [analytics devstack](http://edx.readthedocs.org/projects/edx-installing-configuring-and-running/en/latest/installation/analytics/index.html).
2. Edit the Vagrantfile to add ecommerce (Otto) to the stack:
   - Replace `-e configuration_version=master` with `-e configuration_version=braden/ecommerce-analytics`
   - Add [two mount dirs and the 8002 port  - same changes as to the `release` Vagrant file in this commit](https://github.com/edx/configuration/pull/2878/files#diff-549ab5b641d2bf3723816a8e3b8c9696).
3. Run `vagrant provision` to install ecommerce. (Note this might discard any uncommited changes in your analytics git repos.)

Now, LMS, ShoppingCart, and Otto will all be installed and configured to work together.
#### Set up ShoppingCart:
1. ShoppingCart should be working out of the box on your devstack.
2. But, if you have been testing the new Otto payment page, you'll need to go to http://localhost:8000/admin/commerce/commerceconfiguration/ and add a new configuration entry with "Checkout on ecommerce service" set to False, since that setting breaks the "Checkout" buttons for ShoppingCart course orders.
#### Generate Shopping cart data:
1. Login to Studio as `staff@example.com` and create a couple (empty) courses for testing. Call one "SC Verified Option", and one "SC Paid".  After creation, edit each course's schedule so its start date is in the past.
2. Go to http://localhost:8000/admin/course_modes/coursemode/ .
3. Add a two new entries for "SC Verified Option" (put in the correct `course-v1:...` course ID): One with mode `honor` and price $0 (free), and one with mode `verified` and price of your choice (>0). Leave the **SKU blank** to ensure the checkout will use ShoppingCart and not Otto.
4. Add a new entry for the "SC Paid" course, with mode `honor` and price >0. Leave the SKU blank.  
5. Go to http://localhost:8000/admin/shoppingcart/coupon/ and create a coupon for each course, granting some discount (e.g. 50% off). Use codes like `SC_CODE_VER` and `SC_CODE_PAID`.<br><br>Do the rest in an incognito window:<br><br>
6. Create a new user account (e.g. `user1@example.com`) and login to the LMS.
7. **Do a normal purchase of enrollment codes:** Click "Find courses", find the "SC Paid" course, and click the big "Add to cart" button (if you see "enroll" instead, the course modes are not configured correctly).
8. Go to your cart, and increase the quantity of the course (to any _n_  > 1), to switch to "enrollment code purchase" mode.
9. Complete the purchase of the enrollment codes. Make a note of the quantity purchased.<br><br>
10. **Do a discounted purchase of enrollment codes:** Click "Find courses", find the "SC Paid" course, and click the big "Add to cart" button.
11. Go to your cart, and increase the quantity of the course, to switch to "enrollment code purchase" mode. Now, enter the `SC_CODE_PAID` discount code.
12. Complete the purchase of the discounted enrollment codes. Make a note of the quantity purchased.<br><br>
13. **Do a normal purchase of the Paid course:** Add to cart but this time leave the quantity at 1. "Pay" and then you should be enrolled. If you are asked to verify your identity, just skip that part ("verify later").<br><br>
14. **Register a new user account, and do a discounted purchase of the Paid course:** Same as the previous step, but this time use the discount code before "paying".<br><br>
15. **Do a normal purchase of the verified upgrade course**: Enroll in the "SC Verified Option" course (should be free to enroll; if not your course modes are not configured correctly). You should be prompted to verify at the end of the enrollment process (if not, there should be an upgrade to verified button on the dashboard). Do that, and complete the "payment" process to upgrade to verified.<br><br>
16. **Register a new user account, and do a discounted purchase of the verified upgrade course**: Same as the previous step, but this time use the discount code before "paying".<br><br>

Note: With ShoppingCart, unlike Otto, redeeming an enrollment code does not create any "order" and so is not relevant to this PR. That's why I have not included any enrollment code redeptions above. 
#### Set up Otto
1. Go to http://localhost:8000/admin/credit/creditprovider/ and create a credit provider. The values of most fields do not matter.
2. Login to Studio as `staff@example.com` and create a few (empty) courses for testing. Call one "Otto Verified", one "Otto Professional", and one "Otto Credit."  After creation, edit each course's schedule so its start date is in the past.
3. Set up Otto: [Follow steps 12.2, 12.3, and 12.4 from the official documentation](http://edx.readthedocs.org/projects/edx-installing-configuring-and-running/en/latest/ecommerce/index.html). (n.b. 12.1 is not necessary on a devstack - just do "sudo su ecommerce" and "make serve" to start Otto)
   - When asked about payment processors, put `paypal` only.
   - When creating products, only create products for the "Otto" courses you created in step 1. Set whatever price you want. Set up the "Verified" course, the "Professional Education" course, and the "Credit" course, choosing the appropriate "Course Type" for each one.
4. Go to http://localhost:8002/coupons/ and create a bunch of coupons and enrollment codes for each of the three Otto courses. Note that the "Mode" dropdown is enabled but doesn't show any values until after you've entered a course ID, which is something I initially found confusing.
5. When testing Otto, you'll need a test PayPal account, so set up PayPal:
   - Go to https://developer.paypal.com , and log in with your personal PayPal account.
   - Click “Dashboard”, then “My Apps & Credentials”. Enter a name like "devstack"
   - Get the sandbox credentials: the ID and the secret.
   - Click “Sandbox” → “Accounts”, find the test buyer account, then click “Profile”, then “Change password”. Set a password, and make a note of the buyer account's email and password.
   - In ecommerce/ecommerce/settings/ copy PAYMENT_PROCESSOR_CONFIG from devstack.py to private.py. Put the newly obtained PayPal `client_id` and `client_secret` into private.py. You will also need to put `from django.conf import settings` and `from ecommerce.settings import get_lms_url` at the top of private.py.
6. Go to http://localhost:8000/admin/commerce/commerceconfiguration/ and add new Commerce Configuration, with Active True, checkout on ecommerce True, checkout URL `/basket/single-item/` (the default)  (Note that enabling this seems to break the ShoppingCart checkout button, so turn it off again if you go back to trying a ShoppingCart purchase)
#### Generate Otto data:
1. Login to the LMS as a new user, e.g. `user1@example.com`. Enroll in each Otto course (choose to upgrade to verified if the initial enrollment is free). **Pay full price** and use PayPal with test credentials during checkout. Skip verification of your identity.
2. Login to the LMS as a new user, e.g. `user2@example.com`. Enroll in each Otto course (choose to upgrade to verified if the initial enrollment is free). **Use a coupon to get a discount on each course**. Use PayPal with test credentials during checkout. Skip verification of your identity.
3. Login to the LMS as a new user, e.g. `user3@example.com`. **Use an enrollment code to get free enrollment in each course** (The URL to use these codes is available when creating them from the Otto coupon console). Skip verification of your identity.
#### Run the OrderTableTask:

On your host computer:
1. `cd edx-analytics-pipeline/`
2. Fetch and checkout this `reports-with-coupons` branch (to get the updated `devstack.cfg`)
3. `source venv/bin/activate`
4. `export WHEEL_URL=http://edx-wheelhouse.s3-website-us-east-1.amazonaws.com/Ubuntu/precise`
5. `remote-task --vagrant-path ../ --remote-name devstack --repo https://github.com/open-craft/edx-analytics-pipeline.git --branch reports-with-coupons --override-config ${PWD}/config/devstack.cfg --wheel-url $WHEEL_URL --wait OrderTableTask --local-scheduler`
#### Check the OrderTableTask:

On the devstack:
1. `sudo su hadoop`
2. `hive`
3. `select user_email, order_id, course_key, coupon_id, discount_amount, voucher_id, voucher_code, status, line_item_price, line_item_quantity, line_item_unit_price, order_processor from order;`

Carefully check that the output matches what you'd expect for all of the orders you placed, especially for the four new fields (coupon_id, discount_amount, voucher_id, voucher_code). Note that for ShoppingCart purchases, `voucher_id` has the coupon ID and `coupon_id` will be NULL since ShoppingCart does not distinguish between those concepts.
## Author concerns:
- I have been unable to determine if the django-oscar `OrderDiscount` model's `amount` field is a total discount amount or a per-unit discount amount. For now it doesn't much matter since Otto only allows purchasing one item at a time.
- The Otto version assumes that there is only one course per order, and up to one coupon applied to the order (i.e. that the single-item basket is enforced). The Otto code currently enforces that, but might that change in the future?
## Other notes:
- The existing acceptance test `FinancialReportsAcceptanceTest.test_end_to_end` does not seem to check the actual data values - just the shape of the data. I wrote an alternate test which only runs when Vertica is not available (i.e. on devstack). It checks the shape and the values of the data. It caught a discrepancy between the existing test fixtures and the expected output csv: a001ff742a59e3259070b53e5fe0a60f69614bd2 (order_refunded_amount and order_refunded_quantity were both '0' for an order that had a refund, and whose refund table fixture clearly specified '100' and '1')
## TODO:
- [x] Add some new acceptance test data, with coupon codes included


e-kolpakov:
@bradenmacdonald :+1:

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -133,13 +137,15 @@ def mapper(self, line):
...
-                fields[17] = '0'
+            # Convert Hive null values ('\\N') in fields like 'product_detail' to string typed values:
+            typed_defaults = {
+                'discount_amount': '0.0',
+                'refunded_amount': '0.0',
+                'refunded_quantity': '0',
+            }
+            for index, value in enumerate(fields):
+                if value == '\\N':
+                    fields[index] = typed_defaults.get(ORDERITEM_FIELDS[index], '')
```
smarnach: I implemented a similar change first, but reverted it since I remembered that performance is quite important in the map/reduce code.  The original code consists of operations that are very fast in Python: local variable access, indexing lists etc.  You added quite a bit to that: Creating a temporary dictionary for each record, comparing all 20 fields with `\\N` instead of only 3, attribute lookups etc.  I'd expect the updated version to be an order of magnitude slower than the original.

smarnach: I'm using your version for now anyway, since otherwise rebasing my changes on top of yours becomes kind of crazy.  It would be almost impossible to keep the indices in sync.

bradenmacdonald:
@e-kolpakov I just pushed some updated acceptance test data and fixed an existing bug that the greatly-expanded acceptance tests found (I realized later the bug is SOL-1662). Can you please take a quick look and make sure your +1 stands?

edx/analytics/tasks/reports/reconcile.py
bradenmacdonald: @smarnach Yeah, I think we can optimize this once our PRs are merged; or feel free to propose an optimized version, as long as it doesn't use manual indexing of the columns.

edx/analytics/tasks/tests/acceptance/test_financial_reports.py
```diff
@@ -75,3 +77,48 @@ def test_end_to_end(self):
...
+        expected = pandas.read_csv(expected_output_csv, parse_dates=True)
+
+        output = StringIO(raw_output.replace('\t\\N', '\t'))
+        data = pandas.read_table(output, header=None, names=columns, parse_dates=True)
+        # Re-order dataframe for consistent comparison:
+        for frame in (data, expected):
+            frame.sort(['payment_ref_id', 'transaction_type'], inplace=True, ascending=[True, False])
+            frame.reset_index(drop=True, inplace=True)
+        try:
+            assert_frame_equal(data, expected)
```
e-kolpakov, line 118: @bradenmacdonald there are too much records in the dataframes compared - consider row-by-row comparison - if something goes wrong, this code will output two giant tables and you'll have to manually find an error.

e-kolpakov:
@bradenmacdonald :+1: conditional new note is addressed

bradenmacdonald:
@e-kolpakov Thanks! Per your suggestion, I added 4d0e8935dcda2c449d62829541a470f80fe4531b which will point out exactly which row is the first to differ. I tested it and confirmed that it works and it also handles the case where the # of rows or columns differs.

bradenmacdonald:
@mjfrey When you have some time, can you please check this out? Specifically I'm hoping you can read through my PR description, testing instructions, and especially "Author concerns", and advise if I've misunderstood any aspects of coupons/vouchers, or if there are any key cases I didn't consider.

@brianhw Heads up that I'm hoping you can review this PR next week.

bradenmacdonald:
@mjfrey @brianhw ^ Friendly reminder to make sure this is on your radar for review.

brianhw:
I've run this against production data (using code that was branched from this, in PR 223), and I'm looking over the results to see if they look reasonable.   This should also include PR220 as well.  (It does not yet include the actual invoice work.  That will be a later run, when it's more ready.)

edx/analytics/tasks/reports/reconcile.py
brianhw: I ran into an issue with the handling of coupon_id and voucher_id.  In much of the order data, these values are '\N'.  The code here maps them to a zero-length string, which then fails when trying to load these values into an INT type.  (Voucher_code is fine because it's a String/Varchar.) I assume that these could be set to None so that they would be mapped to NULL in the output, as is done when there is no orderitem found for a given transaction.

bradenmacdonald: Thanks @brianhw. I guess I assumed that the task had to return a string type given that the existing code was converting numeric values to strings like `'0'`. I have pushed deaae5ded600780e9cae08316c45778fd520e35e which should fix the issue, and should also address Sven's performance concern, at least partially.

brianhw:
Looks good.  I'm running with an equivalent now, and will hopefully get it successfully into Vertica tonight.

smarnach:
@bradenmacdonald Could you please rebase this PR?  I've based #220 on this, and currently both PRs have conflicts.

bradenmacdonald:
@smarnach Sure, rebased deaae5ded600780e9cae08316c45778fd520e35e on top of master.

The only conflict was a really minor one, due to 8b10872f234f37734793c2ebd29fa2840e8d7f6f which was just a drive-by pep8 fix in a file I happened to touch for this PR.

bradenmacdonald:
@mjfrey I think we still need you to take a look at this, and make sure my assumptions are correct.

mjfrey:
@bradenmacdonald can you specifically call out what I should look at?  I really don't know what I am looking at.

bradenmacdonald:
@mjfrey Yep, ' Specifically I'm hoping you can read through my PR description, testing instructions, and especially "Author concerns", and advise if I've misunderstood any aspects of coupons/vouchers, or if there are any key cases I didn't consider.'

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -228,6 +242,13 @@ def _get_audit_code_for_orders_without_transactions(self, orderitems):
             elif orderitem.line_item_unit_price == 0.0:
                 order_audit_code = 'ORDER_BALANCED'
                 orderitem_audit_code = 'NO_COST'
+            elif orderitem.line_item_price == 0.0:
+                # The order would normally have a cost but has been discounted 100% or an enrollment code was used.
+                if orderitem.discount_amount == orderitem.line_item_unit_price:
+                    order_audit_code = 'ORDER_BALANCED'
+                    orderitem_audit_code = 'NO_COST'
+                else:
+                    orderitem_audit_code = 'ERROR_MISMATCHED_DISCOUNT'
```
brianhw: I would just remove all mention of line_item_unit_price.  It turns out to be broken (as far as I'm concerned), in that it's list_unit_price from otto and net_unit_price from shoppingcart.  But that's not really relevant at this point.  The line_item_price is net, and if it's zero, the orderitem should be tagged as NO_COST.   So I would remove the previous elif for line_item_unit_price, and remove the if-else here that checks discount against unit_price.  (It will be a separate effort to fix unit_price information for downstream consumers of this table, once it's decided what it should be.)

brianhw:
From my standpoint, things look good.  I just had one comment about slimming down the fix to SOL-1662.  Otherwise :+1: .

mjfrey:
This looks good to me.  To clarify on @brianhw comments.  An Invoice in Otto for a Coupon order is the total price to be billed to the customer and not the total cost of the Seat.   It could be discounted.  If that helps.
:+1:

brianhw:
In terms of addressing author's concerns:  some of those concerns are also analytics concerns.  First, to comment on current system behavior:  I did note that several orderitems in the audit table had discounts and had quantities > 1, but these were all shoppingcart.  All the order_orderdiscount data I see is tied to an order that has a single order_line, which has a quantity of 1.  On the other hand, I also quickly checked to see if there were _any_ orders that had multiple order_line entries, and found four orders with two order_lines, and one with three order_lines.  

But I'm also concerned about the design of the Order Discount stuff.  It seems to me that these coupons or vouchers or whatever should be applied at the order_line level.  At the moment, they actually are, in that the discounts are reflected in the difference between line_price_incl_tax and line_price_before_discounts_incl_tax on the order_line.  So maybe a partial solution is to use the difference between those values, rather than explicitly referencing the order_orderdiscount.amount directly.

@mjfrey I didn't understand your reference to "invoice" here.  This PR makes no reference to invoices.  Is that your name for the order_order entry, or the order_orderdiscount entry?

edx/analytics/tasks/reports/reconcile.py
brianhw: Actually, having decided what it should be, the change seems  small enouhg that it would be great to get it in here as well.  Please change "oi.unit_cost AS line_item_unit_price," to "oi.list_price AS line_item_unit_price,".   That will bring the definition of line_item_unit_price to agree across the board, and particularly to agree with the otto meaning (as _before_ discount).

This invariant should hold going forward:   "line_item_price = (line_item_unit_price \* quantity) - discount"

bradenmacdonald: @brianhw I have taken your suggestion and implemented it in 1fc611db978584108ba84a5c3e19160d6d716680

Since I discovered that `oi.list_price` is sometimes NULL (which means "list_price = unit_price"; though the converse does not hold - just because unit_price = list_price does not imply that list_price is NULL), the actual change is slightly different that what you quoted, but it's the same idea.

Since `line_item_unit_price` is now consistent, I kept the audit code logic in `_get_audit_code_for_orders_without_transactions` unchanged as I think it's more useful this way.

edx/analytics/tasks/tests/acceptance/fixtures/output/expected_financial_report.csv
```diff
@@ -6,7 +6,7 @@ ERROR_ORDER_NOT_BALANCED,ERROR_NO_TRANSACTION,NO_TRANSACTION,1,1,ONUG64DQNFXGOY3
...
 ERROR_ORDER_NOT_BALANCED,ERROR_NO_TRANSACTION,NO_TRANSACTION,12,12,ONUG64DQNFXGOY3BOJ2HY33SMRSXEX3JMR6DCMQ=,2016-03-22 21:18:46,,,,,,,,,,,,,32,ONUG64DQNFXGOY3BOJ2HY3DJNZSV62LUMVWV62LEPQZTE===,1,512,512,1,,0,,,0,0,11,test_user,test@example.com,seat,verified,course-v1:AccTest+ShoppingCart+Verified,AccTest,shoppingcart
-ORDER_NOT_BALANCED,NO_TRANS_WHITE_LABEL,NO_TRANSACTION,13,13,ONUG64DQNFXGOY3BOJ2HY33SMRSXEX3JMR6DCMY=,2016-03-22 21:27:03,,,,,,,,,,,,,34,ONUG64DQNFXGOY3BOJ2HY3DJNZSV62LUMVWV62LEPQZTI===,3,102.4,25.6,4,,409.6,3,SC_PAID_80_PCT_OFF,0,0,11,test_user,test@example.com,reg-code,honor,course-v1:AccTest+ShoppingCart+Paid,AccTest,shoppingcart
+ORDER_NOT_BALANCED,NO_TRANS_WHITE_LABEL,NO_TRANSACTION,13,13,ONUG64DQNFXGOY3BOJ2HY33SMRSXEX3JMR6DCMY=,2016-03-22 21:27:03,,,,,,,,,,,,,34,ONUG64DQNFXGOY3BOJ2HY3DJNZSV62LUMVWV62LEPQZTI===,3,102.4,128,4,,409.6,3,SC_PAID_80_PCT_OFF,0,0,11,test_user,test@example.com,reg-code,honor,course-v1:AccTest+ShoppingCart+Paid,AccTest,shoppingcart
```
bradenmacdonald, line 9: This one line of output in the acceptance test is updated since we made the definition of `line_item_unit_price` consistent.

Now, in this output CSV:
order_line_item_price: 102.4
order_line_item_unit_price: 128 (was 25.6)
order_line_item_quantity: 4
order_discount_amount: 409.6

This satisfies the invariant "line_item_price = (line_item_unit_price \* quantity) - discount"

The other shopping cart lines are unchanged since they have no discount.

bradenmacdonald:
@brianhw Following up on your comment above: I have taken your idea of using "the difference between line_price_incl_tax and line_price_before_discounts_incl_tax on the order_line" to calculate the discount amount, which I think is a good idea to help future-proof us against orders that may contain both multiple line items and a discount code. 

I also used that same technique to only join the `order_orderdiscount` table to the `order_line` information when a discount is active. Since I believe that both the old "Otto + LMS checkout", and the new "Otto + Otto checkout" enforce only one discount code per order and that is unlikely to ever change, even if multiple line items per order were/are/may-be/will-be supported, this should give us a reliable way to associate the discount information with the specific order line even though it is missing from the DB schema.

Re: "All the order_orderdiscount data I see is tied to an order that has a single order_line, which has a quantity of 1" - I'm very glad to know that; thanks so much for checking!

Re. the fact that you "found four orders with two order_lines, and one with three order_lines.": I assume that these orders must have been placed using Otto with the old LMS-hosted checkout page. The new Otto-hosted checkout page does not support multiple order lines.

Please check out my latest commit and let me know if this is good to merge. (Also, please let me know if you want me to squash it.)

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -244,6 +244,7 @@ def _get_audit_code_for_orders_without_transactions(self, orderitems):
                 orderitem_audit_code = 'NO_COST'
             elif orderitem.line_item_price == 0.0:
                 # The order would normally have a cost but has been discounted 100% or an enrollment code was used.
+                # We expect this invariant: line_item_price = (line_item_unit_price * quantity) - discount_amount
                 if orderitem.discount_amount == orderitem.line_item_unit_price:
```
brianhw: I'm afraid I find this confusing.  If we want to determine if the orderitem had no cost, whatever the reason, then we shouldn't need to check line_item_unit_price == 0.0, and just check that line_item_price == 0.0 here.  If the goal is to also find a place to test the invariant, then this seems an odd place to do so, in this one case.  I would put that somewhere else that's more globally applied, and maybe have it prepend  "ERROR_MISMATCH_DISCOUNT_" to the orderitem's audit code.  But as coded here, it seems to assume that the quantity is one, which is not necessarily true for discounts coming out of shoppingcart orders.

brianhw: I can't remember, but is there a unit test that confirms the ERROR_MISMATCHED_DISCOUNT actually gets returned in the right circumstance?  (If you decide to leave it in, that is.)  I'm just looking at the diff of new code right now.

edx/analytics/tasks/tests/acceptance/fixtures/output/expected_financial_report.csv
brianhw, line 9: Nice!

brianhw:
@bradenmacdonald Once ready, I would like to run this and #220 together on prod data one last time to confirm, if that's not too much trouble.  If you and @smarnach can coordinate on that, that would be great.

edx/analytics/tasks/reports/reconcile.py
brianhw: Argh!  I wrote a reply to this, but apparently did so on an outdated diff.  Here is what I wrote there:  

 I'm afraid I find this confusing. If we want to determine if the orderitem had no cost, whatever the reason, then we shouldn't need to check line_item_unit_price == 0.0, and just check that line_item_price == 0.0 here. If the goal is to also find a place to test the invariant, then this seems an odd place to do so, in this one case. I would put that somewhere else that's more globally applied, and maybe have it prepend "ERROR_MISMATCH_DISCOUNT_" to the orderitem's audit code. But as coded here, it seems to assume that the quantity is one, which is not necessarily true for discounts coming out of shoppingcart orders. 

I can't remember, but is there a unit test that confirms the ERROR_MISMATCHED_DISCOUNT actually gets returned in the right circumstance? (If you decide to leave it in, that is.) I'm just looking at the diff of new code right now.  

(That last sentence explains why the comment was buried.)

edx/analytics/tasks/reports/reconcile.py
bradenmacdonald: @brianhw Yeah, thinking about this more, this is an odd place to be checking that. I think I'll take your first suggestion and just strip this out so that we "just check that line_item_price == 0.0 here" then set NO_COST.

P.S. GitHub has a bug and is showing comments on diffs in this PR as "outdated" even though they're not. I'm not sure why but it could be because of the other PRs that build upon this.

bradenmacdonald: @brianhw Ok, I just pushed cf7635533e66374ec5ca76c52ea98be6362d8df8 to implement that change.

bradenmacdonald:
@smarnach Please rebase your PR on top of cf7635533e66374ec5ca76c52ea98be6362d8df8 as discussed :)

brianhw:
Great.  I think those changes look good.  I've manually put this on my branch off of Sven's PR, and have it running.  Hopefully it won't take all day, since I should just be rerunning the order, reconcile, and finance report tasks.  And then we can finally merge both PRs.

brianhw:
Looks good to me. :+1:

brianhw:
Actually a squash would be good as well.

bradenmacdonald:
Thanks so much for for the review and testing, @brianhw !

Squashed cf7635533e66374ec5ca76c52ea98be6362d8df8 down to b82643fe5bb4b7ecf0b2de890c07e78ebbd6891f

PR #220, opened 2016-03-21
Add White Label partner short codes.

smarnach:
This patch tags all imported orders with a partner short code to distinguish edX orders from order for White Label partners.  For orders coming from Otto, we use the partner information for each orderitem directly.  For orders coming from ShoppingCart, the partner short code is left empty in `OrderTableTask` and filled in using heuristics based on the course ID during the order reconciliation.

This patch introduces a new configuration setting to map organization IDs to corresponding Otto partner short codes for ShoppingCart orders.  Before merging this PR the production configuration needs to be updated with a suitable mapping.

The changes in this PR have been rebased on top of #218, making it difficult to review the code here.  Please use the fake PR open-craft#2 for review purposes.
#### Preparation

You need a working analytics devstack with a current version of ecommerce (after edx/ecommerce#630 was merged).

To test orders for differen partners, you should set up at least two different sites for Otto and the LMS as described in edx/ecommerce#630.  Make sure you have a working Otto payment processor for each site (PayPal seems to be easier to set up) to be able to place orders.  ShoppingCart uses a dummy payment processor on the devstack, so you can place orders without further setup.
#### Place a few orders in ShoppingCart
1. Add one or two empty courses with a start date in the past in Studio.
2. Go to http://localhost:8000/admin/course_modes/coursemode/ and create paid course modes for them.
3. Create a new user and buy paid seats in these courses (easiest in incognito window).
#### Place a few orders in Otto
1. Add one or two empty courses with a start date in the past in Studio.
2. Go to http://localhost:8002/courses/ and add paid options for these courses.
3. Create a new LMS user for each site, and buy a seat in the new courses for each site.
#### Test the pipeline

Now you should have enough data to meaningfully test the pipeline.  To avoid having to configure access to the payment processor for the pipeline, it's easiest to comment out the [dependency on `PaymentTask` in reconcile.py](https://github.com/edx/edx-analytics-pipeline/blob/6c3ad11a7364eef48c656cb0fa3239fcade19841/edx/analytics/tasks/reports/reconcile.py#L125-L127), commit this change to a temporary branch and push that branch to Github.  On your host machine, activate the virtualenv of your edx-analytics-pipeline repo and run

```
WHEEL_URL=http://edx-wheelhouse.s3-website-us-east-1.amazonaws.com/Ubuntu/precise
remote-task --vagrant-path path/to/analyticstack/ --remote-name devstack \
    --repo https://github.com/open-craft/edx-analytics-pipeline.git \
    --branch name-of-your-test-branch --override-config ${PWD}/config/devstack.cfg \
    --wheel-url $WHEEL_URL --wait ReconciledOrderTransactionTableTask \
    --local-scheduler --import-date=$(date -I) --n-reduce-tasks=1
```

On your devstack, run `hive` as user `hadoop` and run a query like

```
select
    partner_short_code, order_line_item_price, 
    order_user_email, order_course_id, order_processor
    from reconciled_order_transactions;
```

and verify the output.


openedx-webhooks:
Thanks for the pull request, @smarnach! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-analytics-pipeline&number=220

e-kolpakov:
@smarnach testing instructions and sample data would be really helpful - looks like I'll spend much more time generating the data than actually testing changes.

smarnach:
@e-kolpakov There will be a small amount of data included in the acceptance tests, which I'll add to the PR as soon as I managed to run them.

smarnach:
@e-kolpakov The unit tests for the reconcilation task work fine, by the way, and you can run them with `make test` (after activating the virtualenv).

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -325,7 +350,7 @@ def _get_orderitem_audit_code(self, orderitem, trans_balance, trans_audit_codes)
         elif trans_balance == 0.0:
             audit_code = "{}_WAS_REFUNDED".format(audit_code)
         elif (trans_balance == -1 * orderitem.line_item_price and
-                ('REFUND_AGAIN' in trans_audit_codes or 'REFUND_AGAIN_STATUS_NOT_REFUNDED' in trans_audit_codes)):
+              ('REFUND_AGAIN' in trans_audit_codes or 'REFUND_AGAIN_STATUS_NOT_REFUNDED' in trans_audit_codes)):
```
e-kolpakov, line 372: @smarnach looks like accidental edit - and likely should trigger pylint/pep8 validation

edx/analytics/tasks/reports/tests/test_reconcile.py
```diff
@@ -224,6 +227,24 @@ def test_no_transaction(self):
             'transaction_fee_per_item': None,
         })
 
+    @data(
+        ('otto', 'course-v1:MITx+15.071x_3+1T2016', 'EDX', 'EDX'),
+        ('otto', 'course-v1:MITx+15.071x_3+1T2016', '', EDX_PARTNER_SHORT_CODE),
+        ('otto', 'course-v1:MITPE+12345+1T2016', 'MITPE', 'MITPE'),
+        ('shoppingcart', 'course-v1:MITx+15.071x_3+1T2016', '', EDX_PARTNER_SHORT_CODE),
```
e-kolpakov: @smarnach I've been told to try to avoid mentioning edX clients in publicly available places, and this is indeed publibly available place; so, please remove MITPE from the test samples.

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -97,6 +98,14 @@ def __new__(cls, *args, **kwargs):
     '9918',
 )
 
+# Short codes (= organization IDs) of partners with data in ShoppingCart
+SHOPPINGCART_PARTNERS = (
+    'MITPE',
```
e-kolpakov: @smarnach can it be a setting or task parameter instead of a hardcoded value? The problem is, it exposes client identifier in a publicly available place, and edX have some concerns about it (I've mentioned it in slack and got a bit lectured on that).

smarnach: I wondered whether this is public information, and searched for it on the web.  I found that the domain name is mitprofessionalx.edx.org, and that there are MIT web pages mentioning that this is a collaboration with edX, so I figured it is public information.  I'm happy to move this to a setting, but I'm not sure how to make sure it gets the right values.  I'll ask on the upstream ticket.

edx/analytics/tasks/reports/reconcile.py
smarnach, line 372: @e-kolpakov It's intentional.  Pylint complained about the old version and insisted two spaces need to be removed from this line.

smarnach:
@eugeny I've updated the fake PR at open-craft#2 with the changes in response to your comments.  I'm not quite sure what the best way to review this is, but I think I should close this PR and open a new one that includes both Braden's and my changes, and continue the review on the fake PR that target Braden's branch.  Do you have any other suggestion?

Eugeny:
@smarnach what I don't even

smarnach:
@Eugeny Sorry, tagged the wrong person.

edx/analytics/tasks/reports/orders_import.py
```diff
@@ -157,6 +161,8 @@ def insert_query(self):
                     -- The EDX-1XXXX identifier is used to find transactions associated with this order
                     o.number AS payment_ref_id
 
+                    partner.short_code AS partner_short_code,
```
brianhw: Note that the comma at the end needs to be moved to the previous line.

edx/analytics/tasks/reports/orders_import.py
```diff
@@ -251,6 +260,10 @@ def insert_query(self):
                     IF(oi.status = 'refunded', oi.qty * oi.unit_cost, NULL) AS refunded_amount,
                     IF(oi.status = 'refunded', oi.qty, NULL) AS refunded_quantity,
                     oi.order_id AS payment_ref_id
+
+                    -- The partner short code is extracted from the course ID during order reconcilation.
+                    '' AS partner_short_code,
```
brianhw: Note that the comma at the end needs to be moved to the previous line.

edx/analytics/tasks/reports/orders_import.py
smarnach: @brianhw Thanks, I'm aware of this.  This branch is out of date, since I rebased the changes on top of #218.  The latest version of the changes can be seen in the fake PR open-craft#2.  (I don't know a good way to handle pull requests that depend on other PRs.)

I will force push the other branch over this branch here for now.  However, this PR will show a combination of the changes from #218 and this PR, so for review purposes it will be easier to use the other PR.

e-kolpakov:
@smarnach I can confirm that this PR does what it advertises, but I was unable to test it with multiple WL sites - my local install does not have relevant data yet. I'll use analytics sandbox tomorrow to check this last thing.

brianhw:
I've been running this along with ProfEd transaction data using:

```
[financial-reports]
shoppingcart-partners = {"MITProfessionalX": "MITPE"}
```

So far, things look fine.  But we've now got data loaded into Vertica for a before-after comparison that we hope to do soon (as soon as CAT-1 bugs and the like stop distracting us).

e-kolpakov:
@smarnach :+1:, but needs rebasing

smarnach:
@e-kolpakov This is up to date with #218.  I will rebase once that PR has been updated, to avoid having to do the conflict resolution twice.

smarnach:
@douglashall @brianhw This is ready for your review.  For code review, it's probably easier to look at the fake PR open-craft#2 (targetting the branch this PR was based off).  The acceptance test data isn't complete yet, but I expect to get that sorted out today.

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -116,6 +125,20 @@ class ReconcileOrdersAndTransactionsTask(ReconcileOrdersAndTransactionsDownstrea
     """
 
     output_root = luigi.Parameter()
+    shoppingcart_partners = luigi.Parameter(
+        config_path={'section': 'financial-reports', 'name': 'shoppingcart-partners'},
+        description="JSON string containing a dictionary mapping organization IDs of White Label partners "
+        "with data in ShoppingCart to the corresponding Otto partner short code.  The short code to be used "
+        "for other organization IDs must be given as value for the key \"DEFAULT\".",
```
brianhw: nit:  add something like: ", to have a default that is not NULL."

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -189,7 +225,12 @@ def _extract_transactions(self, values):
         transactions = []
         for (record_type, fields) in values:
             if record_type == 'OrderItemRecord':
-                orderitems.append(OrderItemRecord(*fields))
+                orderitem = OrderItemRecord(*fields)
+                if not orderitem.partner_short_code:
+                    orderitem = orderitem._replace(  # pylint: disable=no-member,protected-access
+                        partner_short_code=self._get_partner(orderitem.course_id),
+                    )
```
brianhw, line 232: A small note:  since we have ORDERITEM_FIELD_INDICES, we could just update the partner_short_code fields value directly (by also referencing the course_id directly), and then create the OrderItemRecord directly from that, without having to use a _replace call.  The difference is where the remapping of '\N' values in partner_short_code is done before or after the call to self._get_partner.

edx/analytics/tasks/reports/reconcile.py
```diff
@@ -132,19 +155,27 @@ def mapper(self, line):
...
+            # Convert Hive null values ('\\N') in fields like 'product_detail':
+            defaults = (
+                ('product_detail', ''),
+                ('refunded_amount', '0.0'),
+                ('refunded_quantity', '0'),
+                ('discount_amount', '0.0'),
+                ('coupon_id', None),
+                ('voucher_id', None),
+                ('voucher_code', ''),
+                ('partner_short_code', ''),
```
brianhw, line 169: Just so I understand that this is deliberate:  we map NULL ('\N') to an empty-length string here, rather than leaving as is, and passing as NULL to the output.  By remapping to "" here, we cause the value to be updated to the default (because of the "if not orderitem.partner_short_code" clause).   But the default value could be None if not specified, so then all the nulls end up as null.  Okay.

edx/analytics/tasks/reports/reconcile.py
brianhw, line 232: And I guess it's easier to have that mapping occur in OrderItemRecord along with other things, rather than accessing the "raw" value here.  Okay.

edx/analytics/tasks/reports/tests/test_reconcile.py
```diff
@@ -224,6 +230,24 @@ def test_no_transaction(self):
...
+    )
+    @unpack
+    def test_partner_short_code(self, order_processor, course_id, partner_short_code, expected_short_code):
+        orderitem = self.create_orderitem(
+            order_processor=order_processor,
+            course_id=course_id,
+            partner_short_code=partner_short_code,
+        )
+        self._check_output([orderitem], {'partner_short_code': expected_short_code})
+
```
brianhw, line 253: Is it possible to add a test for when the partner_short_code default is not set?  E.g. if the mapping is entirely missing?

edx/analytics/tasks/reports/reconcile.py
smarnach, line 232: @brianhw The \N values are replaced in the mapper, and _get_partner() is called in the reducer, so all \N values should be gone by the time we get here.  Using the indices from ORDERITEM_FIELD_INDICES makes sense now we have it (we didn't when I implemented this.)

I don't understand your second comment.

edx/analytics/tasks/reports/reconcile.py
smarnach, line 169: I added this mainly for consistency.  The existing code replaced all values that could be \N with some "empty" value.  Since all values were strings, I assumed this is some kind of requirement, but if `None` also works, we could use that as well (the result will be exactly the same, since both the empty string and `None` are falsy).

edx/analytics/tasks/reports/tests/test_reconcile.py
smarnach, line 253: The `ReconcileOrdersAndTransactionsTask` errors out if the `shoppingcart_partners` configuration is missing completely, which is intentional.  I don't think it is possible to test this in the unit tests.

I will add a test for the case that the `shoppingcart_partners` mapping is empty (and thus lacking the `DEFAULT` key).

edx/analytics/tasks/reports/reconcile.py
brianhw, line 169: Right.  It doesn't really matter either way, as long as it gets mapped to something falsy in the reducer.  (For other values that are not used here, it's okay for a value to remain as "\N", because it's eventually being output to a Hive table.)

edx/analytics/tasks/reports/reconcile.py
brianhw, line 232: Sorry, I thought both of these were in the mapper, so that order mattered.  Please ignore my second comment.

edx/analytics/tasks/reports/tests/test_reconcile.py
brianhw, line 253: Thanks.

smarnach:
@brianhw I think I addressed your notes (see the latest commit and my comments).  I also rebased on top of #218 again, and I'm still working on the analytics test data.

brianhw:
@smarnach Looks good.  @bradenmacdonald has made some more changes to #218 and I would like to run one last production test run with both your branches together.

smarnach:
@brianhw @bradenmacdonald I've rebased on top of #218 again, so this branch contains the latest version of both PRs combined.

brianhw:
Looks good to me. :+1:

brianhw:
Actually this looks like it could also benefit from a squash.

smarnach:
@brianhw Thanks for your help with this!

PR #222, opened 2016-03-24
Purge deleted records from vertica

mulby:
@brianhw - this should be a "good enough" band-aid fix for the deleted records issue for the short term until we figure out a better solution.

Note that I run this _before_ we re-create the LAPs. I have already run a test build that did what I expected it to.


mulby:
@HassanJaveed84 FYI

mulby:
@cpennington - I think I managed to save your LAPs... however, they will be unavailable for about 8 hours every day during the table rebuild. We will work on solving that problem more comprehensively when we can actually focus on this work.

brianhw:
LGTM.  :+1:

HassanJaveed84:
:+1:

PR #225, opened 2016-03-31
Change how metric ranges are specified.

HassanJaveed84:

I'm still thinking on how to better specify the different range values, I'm open to suggestions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,8 +62,8 @@ coverage-local: test-local
 	pep8 edx > pep8.report || echo "Not pep8 clean"
 
 	# Compute pylint quality
-	pylint -f parseable edx --msg-template "{path}:{line}: [{msg_id}({symbol}), {obj}] {msg}" > pylint.report || echo "Not pylint clean"
 	diff-quality --violations=pylint --html-report diff_quality_pylint.html pylint.report
+	pylint -f parseable edx --msg-template "{path}:{line}: [{msg_id}({symbol}), {obj}] {msg}" > pylint.report || echo "Not pylint clean"
 
 coverage: test coverage-local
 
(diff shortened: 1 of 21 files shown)


mulby:
The more I think about this, I'm now concerned that it's impossible for a value of "inf" to be anything other than "normal". Could it ever be categorized as "high"?

@dsjen @dan-f @brianhw @shnayder  - we are codifying our discussion of the metric ranges here and I'm still not convinced we have a solution that does what we want. Feel free to weigh in.

To make things easier to reason about let's say the percentiles are actually 10 and 90 (easier for in your head math).

Now if I had an input that was `[2]*8 + [float('inf')]*2`. This would mean that 80% of the learners had a ratio of 2 and the other 20% had an undefined value. In this case the "inf" value is really in the "high" bucket, however, our current logic would just make the normal range `2 <= value < inf` and not define a "high" range. In this case I guess I would actually expect:

normal: `2 <= value < inf`
high: `value == inf`

It would only land in the "high" bucket if it was actually equal to float('inf').

More context here: https://openedx.atlassian.net/browse/AN-6897 (note that I made this ticket public)

mulby:
@HassanJaveed84 to provide more context, these ranges are used to identify users who are outside of the "normal" range for a particular metric, such as the ratio of problem attempts to problems completed correctly.

It is possible for users to perform 10 attempts and not get any problems correct, which can result in these float('inf') values (10/0).

mulby:
More interesting test cases (assuming 10th, 90th):

input: `[1]*10 + [float('inf')]`
In this case the users with the value of `1` should be "normal" and the user with the value of `float('inf')` should be "high". Not sure how to construct the ranges so that would be the case.

shnayder:
Product note: it'd be ok by me to special case 'inf' and compute percentiles of the rest. The point is to highlight the top and bottom ranges -- the exact formula doesn't matter. We'll likely want to tweak parameters once we get some feedback anyway.

PR #226, opened 2016-04-01
Ansible apt module does not work on debian squeeze

omarkhan:
The `python-apt` package is required for the ansible apt module to work, but it is not available in the debian squeeze package repositories. So we have to install git with a shell command instead.

### DIFF
--- a/share/task.yml
+++ b/share/task.yml
@@ -96,7 +96,7 @@
         - "{{ log_dir }}"
 
     - name: make sure git is available on the Debian server
-      apt: pkg=git state=present
+      command: apt-get install -q -y git
       sudo: True
       when: ansible_distribution in common_debian_variants
 


openedx-webhooks:
Thanks for the pull request, @omarkhan! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=226&repo=edx%2Fedx-analytics-pipeline

brianhw:
Sorry to be confused, but what changed in the yumify-task that would require this to be changed?  I presume that this has been working for a long time now.

brianhw:
Interesting!  I thought I had run the branch on an old 2.4.11 configuration without problem.  Could this be something that changed as a result of the expiration of support for Squeeze back a month ago?  

Anyway, if this works then :+1: .

omarkhan:
I'm not sure if or how this was working before, but I could not get it to
work with the 2.4.11 AMI without this change.

PR #229, opened 2016-04-07
Identify EMR server using master_public_dns_name instead of master_private_ip

pomegranited:
Allows access to the EMR master node by servers (e.g. devstacks) that sit outside the AWS VPC.

Without this change, EMR services used in task.yml are referenced by their AWS internal IP addresses, which are not accessible outside of the VPC.

**JIRA tickets**: Cleanup for [OLIVE-22](https://openedx.atlassian.net/browse/OLIVE-20), [OC-1512](https://tasks.opencraft.com/browse/OC-1512)

**Dependencies**: None

**Sandbox URL**: 

http://52.20.136.133:8080/ - ping me for login credentials.

**Testing instructions**:

See [edx/configuration PR #2939](https://github.com/edx/configuration/pull/2939) for full test instructions.

Expected result is the remote task will run successfully (if your configuration is correct).

Without this page, the EMR provisioning fails with AWS access errors:

```
Running command = ['/edx/var/jenkins/workspace/AnswerDistributionWorkflow/venvs/analytics-tasks/share/edx.analytics.tasks/ec2.py', '--refresh-cache']
...
RuntimeError: Unable to refresh ansible inventory cache.
```

**Reviewers**
- [ ] @jbzdak 
- [ ] @mulby 
- [ ] @e0d

### DIFF
--- a/share/ec2.ini
+++ b/share/ec2.ini
@@ -1,7 +1,7 @@
 [ec2]
 regions=us-east-1
 destination_variable=route53_domain_name
-backup_destination_variable=private_ip_address
+backup_destination_variable=public_dns_name
 cache_path=/tmp
 cache_max_age=300
 regions_exclude=
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -6,5 +6,4 @@ httplib2==0.9 		# MIT
 paramiko==1.14.0 	# LGPL
 pycrypto==2.6.1 	# public domain
 ansible==1.4.5          # GPL v3 License
-boto==2.22.1            # MIT
-
+boto==2.39.0            # MIT


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-analytics-pipeline&number=229

share/ec2.ini
```diff
@@ -1,7 +1,7 @@
 [ec2]
 regions=us-east-1
 destination_variable=route53_domain_name
-backup_destination_variable=private_ip_address
+backup_destination_variable=public_dns_name
```
jbzdak, line 4: @pomegranited Could you explain `backup_destination_variable` actually does? I couldn't find (well Google couldnt ;)) a reference to it outside edx repositories.

pomegranited, line 4: @jbzdak From what I can see, [share/ec2.ini](https://github.com/open-craft/edx-analytics-pipeline/blob/jill/olive-22-emr-master_public_dns_name/share/ec2.ini) is the configuration file for the [share/ec2.py](https://github.com/open-craft/edx-analytics-pipeline/blob/jill/olive-22-emr-master_public_dns_name/share/ec2.py) script, which fetches the list of EMR instances and assembles them into an ansible inventory format.  

For each of the instances found, ec2.py checks to see if the configured `destination_variable` and `backup_destination_variable` attributes are present on the instance, and uses the first one it finds as the address for that host.  I don't think we have Route53 domain names set up on our AWS, and for our EMR instances, the `backup_destination_variable` is used instead.  The [upstream ec2.ini file uses `backup_destination_variable=private_ip_address`](https://github.com/edx/edx-analytics-pipeline/blob/master/share/ec2.ini#L4), and the private IP is not accessible outside the VPC, so our vagrant instances can't access it.  This change uses `backup_destination_variable=public_dns_name` instead, which is accessible outside the VPC.

However, [Gabe says that's not a good thing to do](https://github.com/edx/edx-analytics-configuration/pull/31#issuecomment-206945980), so I'm going to close this PR :)

PR #231, opened 2016-04-08
Add non-Debian support and patch Luigi.

brianhw:
This PR includes two important fixes.  The first allows the pipeline to deploy to more modern versions of AWS EMR clusters.  The second permits the pipeline to find metadata for packages that have different names than their egg-info name.

EMR clusters after 2.x switched to RHEL from Debian Squeeze.
This PR provides support for deploying to such clusters.
- Add yum option to Makefile.
- Identify system requirements for installation by yum.
- Don't use Debian wheel settings if platform is not debian.
- Change hadoop variables in Luigi config.

(Note that the Ansible apt module does not work on debian squeeze.  The python-apt module required by ansible is no longer in the package repositories.)

The second change points to a patched version of Luigi, intended to get the pipeline working again with versions of opaque-keys and ccx-keys that are now available on PyPi as edx-opaque-keys and edx-ccx-keys.  With the fix, we can also pin to use the new versions of these packages.

@mulby @HassanJaveed84

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,9 +27,14 @@ develop-local: uninstall
 	python setup.py install_data
 
 system-requirements:
+ifeq (,$(wildcard /usr/bin/yum))
 	sudo apt-get update -q
 	# This is not great, we can't use these libraries on slave nodes using this method.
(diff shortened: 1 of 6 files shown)


mulby:
LGTM

PR #233, opened 2016-04-11
Fix finance acceptance test on Vertica

brianhw:
Start with fix type in coupon fixture data.

@mulby @HassanJaveed84


mulby:
LGTM

brianhw:
Fixing this problem resulted in a successful acceptance test run.

PR #234, opened 2016-04-11
Roster acceptance test.

HassanJaveed84:
@mulby @brianhw 

Up for initial review, I think we should add more data to the input event files. 
Right now, only one test user has an enrollment event.


config/test.cfg
```diff
@@ -114,3 +114,11 @@ cybersource_merchant_ids = test
 
 [videos]
 dropoff_threshold = 0.05
+
+[elasticsearch]
+host = https://search-acceptance-test-kkpt4pnwnib3ywbu3ovsselg6y.us-east-1.es.amazonaws.com/
```
mulby: We should pass this into the acceptance test framework in the ACCEPTANCE_TEST_CONFIG instead of committing it to the repo.

mulby: https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/tests/acceptance/__init__.py#L149

edx/analytics/tasks/tests/acceptance/test_module_engagement.py
```diff
@@ -19,6 +20,53 @@ class ModuleEngagementAcceptanceTest(AcceptanceTestCase):
...
+            else:
+                self.upload_tracking_log(self.EMPTY_INPUT_FILE, fake_date)
+
+        self.execute_sql_fixture_file('load_auth_user_for_internal_reporting_user.sql')
+        self.execute_sql_fixture_file('load_auth_userprofile.sql')
+        self.execute_sql_fixture_file('load_course_groups_courseusergroup.sql')
+        self.execute_sql_fixture_file('load_course_groups_courseusergroup_users.sql')
+
+        self.task.launch([
+            'ModuleEngagementRosterIndexTask',
```
mulby: We should use this entry point:

https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/module_engagement.py#L1189

And also verify the metric ranges since both are required for Insights to work properly.

mulby:
Directionally, this is perfect. I've suggested a couple of tweaks.

I also think having a couple of users would be very useful. This is the only coverage we have of the hive query, so I think it's worth adding a little test complexity to give us more assurance that query continues to do what we want it to.

HassanJaveed84:
@mulby @brianhw can you please review this while I add in more data.

mulby:
@HassanJaveed84 I think the double-run issue may be related to --overwrite. I would remove it from your test.

edx/analytics/tasks/module_engagement.py
```diff
@@ -1187,6 +1187,7 @@ def document_generator(self, lines):
 
 @workflow_entry_point  # pylint: disable=missing-docstring
 class ModuleEngagementWorkflowTask(ModuleEngagementDownstreamMixin,
+                                   OverwriteOutputMixin,
```
mulby: This mixin messes with the `complete()` method for the task. We should just manually define an overwrite parameter here...

HassanJaveed84: ModuleEngagementSummaryMetricRangesMysqlTask defines an overwrite with a default value of True. RosterIndexTask has a default of False for overwrite. 
Do we need to define two different overwrite parameters for this workflow ?

mulby: I think it's OK to not pass overwrite to the metric ranges, since we pretty much always want that to overwrite.

We need to do some architectural thinking about overwrite... the current approach is very error prone and non-intuitive.

mulby:
@HassanJaveed84 I think I fixed the run-twice issue in this PR: https://github.com/edx/edx-analytics-pipeline/pull/238

It looks like we both fixed the overwrite parameter... I'll strip the change out of my branch.

edx/analytics/tasks/tests/acceptance/test_module_engagement.py
```diff
@@ -19,7 +22,119 @@ class ModuleEngagementAcceptanceTest(AcceptanceTestCase):
...
+        self.execute_sql_fixture_file('load_auth_user_for_internal_reporting_user.sql')
+        self.execute_sql_fixture_file('load_auth_userprofile.sql')
+        self.execute_sql_fixture_file('load_course_groups_courseusergroup.sql')
+        self.execute_sql_fixture_file('load_course_groups_courseusergroup_users.sql')
+
+        self.task.launch([
+            'ModuleEngagementWorkflowTask',
+            '--date', '2015-04-17',
+            '--n-reduce-tasks', str(self.NUM_REDUCERS),
+            '--overwrite'
```
mulby: I would remove "--overwrite" here since that's not really the standard way this will be invoked.

edx/analytics/tasks/tests/acceptance/test_module_engagement.py
```diff
@@ -19,7 +22,119 @@ class ModuleEngagementAcceptanceTest(AcceptanceTestCase):
     INPUT_FILE = 'module_engagement_acceptance_tracking_{date}.log'
     NUM_REDUCERS = 1
 
-    def test_engagement(self):
+    def test_roster_generation(self):
```
mulby, line 24: We should add a decorator around this that prevents it from being run locally, much like the decorators for vertica, s3 etc.

edx/analytics/tasks/tests/acceptance/test_module_engagement.py
```diff
@@ -19,7 +22,119 @@ class ModuleEngagementAcceptanceTest(AcceptanceTestCase):
...
+            ('edX/DemoX/Demo_Course_2', datetime.date(2015, 4, 10), datetime.date(2015, 4, 17), 'problem_attempts_per_completed', 'normal', None, None),
+            ('edX/DemoX/Demo_Course_2', datetime.date(2015, 4, 10), datetime.date(2015, 4, 17), 'problem_attempts_per_completed', 'high', None, None),
+            ('edX/DemoX/Demo_Course_2', datetime.date(2015, 4, 10), datetime.date(2015, 4, 17), 'problems_attempted', 'low', 0, 1.0),
+            ('edX/DemoX/Demo_Course_2', datetime.date(2015, 4, 10), datetime.date(2015, 4, 17), 'problems_attempted', 'normal', 1.0, 1.0),
+            ('edX/DemoX/Demo_Course_2', datetime.date(2015, 4, 10), datetime.date(2015, 4, 17), 'problems_attempted', 'high', 1.0, None),
+            ('edX/DemoX/Demo_Course_2', datetime.date(2015, 4, 10), datetime.date(2015, 4, 17), 'videos_viewed', 'low', 0, 1.0),
+            ('edX/DemoX/Demo_Course_2', datetime.date(2015, 4, 10), datetime.date(2015, 4, 17), 'videos_viewed', 'normal', 1.0, 1.0),
+            ('edX/DemoX/Demo_Course_2', datetime.date(2015, 4, 10), datetime.date(2015, 4, 17), 'videos_viewed', 'high', 1.0, None),
+        ])
```
mulby: we should also verify the contents of the `module_engagement` table.

mulby: We might be able to pull a bunch of that logic from the `test_engagement` test below. Once we do so, we can remove that test.

mulby:
@HassanJaveed84 - I made a pass over this test. Generally looks good. I had a few comments that I think we should address before merging.

edx/analytics/tasks/tests/acceptance/__init__.py
```diff
@@ -169,7 +169,10 @@ def setUp(self):
             'course-structure': {
                 'api_root_url': 'acceptance.test',
                 'access_token': 'acceptance'
-            }
+            },
+            'elasticsearch': {
+                'host': self.config['elasticsearch_host']
```
mulby: this should be an optional config setting... if it's not specified than this section shouldn't be included (kind of like 'vertica_creds_url')

edx/analytics/tasks/tests/acceptance/__init__.py
```diff
@@ -247,3 +272,17 @@ def execute_sql_fixture_file(self, sql_file_name, database=None):
...
+
+    def reset_es(self):
+        response = self.elasticsearch_client.indices.get_aliases(name='*{}'.format(self.elasticsearch_alias))
+        for index, alias_info in response.iteritems():
+            for alias in alias_info.keys():
+                if alias.endswith(self.elasticsearch_alias):
+                    self.elasticsearch.indices.delete(index=index)
+
+        #get documents from the marker index that have their target_index set to current alias
+        query = {"query": {"wildcard": {"target_index": '*_{}'.format(self.elasticsearch_alias)}}}
```
HassanJaveed84: We can change our test alias to be of the form  'random_string' + '_roster' and then use a prefix query here instead of wildcard.

edx/analytics/tasks/tests/acceptance/__init__.py
```diff
@@ -247,3 +272,17 @@ def execute_sql_fixture_file(self, sql_file_name, database=None):
...
+        response = self.elasticsearch_client.search(index='index_updates', body=query)
```
HassanJaveed84: This won't return all results, but there would be only one document anyway.
We can also use 'scroll' instead.

edx/analytics/tasks/tests/acceptance/__init__.py
```diff
@@ -247,3 +272,17 @@ def execute_sql_fixture_file(self, sql_file_name, database=None):
...
+
+        for doc in response['hits']['hits']:
+            self.elasticsearch_client.delete(index='index_updates', doc_type='marker', id=doc['_id'])
```
HassanJaveed84: We can change this to use bulk delete api.

edx/analytics/tasks/tests/acceptance/__init__.py
```diff
@@ -208,6 +229,9 @@ def setUp(self):
         self.hive = hive.HiveService(self.task, self.config, database_name)
         self.vertica = vertica.VerticaService(self.config, schema)
 
+        # This supports only aws es clusters for now.
+        self.elasticsearch_client = elasticsearch.Elasticsearch(hosts=[self.config['elasticsearch_host']], connection_class=AwsHttpConnection)
```
mulby: I think it would make sense to move most of this logic to a new "ElasticsearchService" similar to the HiveService and VerticaService etc.

It could contain all of the logic around deleting the indexes and index_updates document etc, as well as making it easier for the test to run queries etc.

edx/analytics/tasks/tests/acceptance/__init__.py
```diff
@@ -7,8 +7,14 @@
...
 import unittest
 
+try:
+    import elasticsearch
+except ImportError:
+    elasticsearch = None
+
 from edx.analytics.tasks.tests.acceptance.services import fs, db, task, hive, vertica
 from edx.analytics.tasks.url import url_path_join, get_target_from_url
+from edx.analytics.tasks.util.aws_elasticsearch_connection import AwsHttpConnection
```
mulby: Ideally this would be configurable, eventually we would like to support running this acceptance tests on devstack which would want to use the default connection class.

edx/analytics/tasks/tests/acceptance/__init__.py
```diff
@@ -247,3 +272,17 @@ def execute_sql_fixture_file(self, sql_file_name, database=None):
             database = self.import_db
         log.debug('Executing SQL fixture %s on %s', sql_file_name, database.database_name)
         database.execute_sql_file(os.path.join(self.data_dir, 'input', sql_file_name))
+
+    def reset_es(self):
```
mulby: I'm actually OK with this method as is. Just put in comments that call out the fact that there should only be 1 document most of the time.

These reset() methods don't need to be terribly performant so I wouldn't worry about optimizing them very much.

edx/analytics/tasks/tests/acceptance/services/es.py
```diff
@@ -0,0 +1,31 @@
...
+except ImportError:
+    elasticsearch = None
+
+from edx.analytics.tasks.util.aws_elasticsearch_connection import AwsHttpConnection
+
+
+class ElasticsearchService(object):
+
+    def __init__(self, config, alias_suffix):
+         self.elasticsearch_client = elasticsearch.Elasticsearch(hosts=[config['elasticsearch_host']], connection_class=eval(config.get('elasticsearch_connection_class', 'AwsHttpConnection')))
```
HassanJaveed84: Is there a way to substitute eval for something else ?
If the connection_class is configurable, how do we dynamically import it ?

edx/analytics/tasks/tests/acceptance/services/es.py
```diff
@@ -0,0 +1,31 @@
...
+         self.alias_suffix = alias_suffix
+
+    @property
+    def client(self):
```
HassanJaveed84: Should there be some sort of a search method in this service that invokes ES API instead of using this property in tests ?

edx/analytics/tasks/tests/acceptance/services/es.py
mulby: I just had a hardcode mapping:

``` python
if config.get('elasticsearch_connection_class') == 'aws':
    connection_class = AwsHttpConnection
```

Not pretty, but probably fine for test code.

edx/analytics/tasks/tests/acceptance/services/es.py
mulby: I think the way you have it is fine!

edx/analytics/tasks/tests/acceptance/__init__.py
```diff
@@ -207,6 +222,7 @@ def setUp(self):
         self.task = task.TaskService(self.config, task_config_override, self.identifier)
         self.hive = hive.HiveService(self.task, self.config, database_name)
         self.vertica = vertica.VerticaService(self.config, schema)
+        self.es = es.ElasticsearchService(self.config, elasticsearch_alias_suffix)
```
mulby: nit, I would prefer you pass in the entire alias name here instead of just the suffix

edx/analytics/tasks/tests/acceptance/services/es.py
mulby: You can probably do:

``` python
if config.get('elasticsearch_connection_class') == 'aws':
    connection_class = AwsHttpConnection
else:
    connection_class = None
```

edx/analytics/tasks/tests/acceptance/test_module_engagement.py
```diff
@@ -19,56 +22,180 @@ class ModuleEngagementAcceptanceTest(AcceptanceTestCase):
...
 
-        # Verify the output
-        with self.export_db.cursor() as cursor:
-            cursor.execute(
-                'SELECT date, course_id, username, entity_type, entity_id, event, count FROM module_engagement'
-                ' ORDER BY date, course_id, username, entity_type, entity_id, event ASC'
-            )
-            results = cursor.fetchall()
+        #es_host = self.task.default_config_override['elasticsearch']['host']
+        #es_client = elasticsearch.Elasticsearch(hosts=[es_host], connection_class=AwsHttpConnection)
```
mulby: OK to remove these, I like the new way better.

edx/analytics/tasks/tests/acceptance/test_module_engagement.py
```diff
@@ -19,56 +22,180 @@ class ModuleEngagementAcceptanceTest(AcceptanceTestCase):
...
 
-        april_thirteenth = datetime.date(2015, 4, 13)
-        expected_first_day = [
-            (april_thirteenth, 'course-v1:edX+DemoX+Demo_Course_2015', 'honor', 'discussion', 'cba3e4cd91d0466b9ac50926e495b76f', 'contributed', 3),
-            (april_thirteenth, 'course-v1:edX+DemoX+Demo_Course_2015', 'honor', 'video', '8c0028eb2a724f48a074bc184cd8635f', 'viewed', 1),
-            (april_thirteenth, 'edX/DemoX/Demo_Course', 'honor', 'video', 'i4x-edX-DemoX-video-8c0028eb2a724f48a074bc184cd8635f', 'viewed', 2),
-            (april_thirteenth, 'edX/DemoX/Demo_Course_2', 'honor', 'video', 'i4x-edX-DemoX-video-8c0028eb2a724f48a074bc184cd8635f', 'viewed', 1),
-            (april_thirteenth, 'edX/DemoX/Demo_Course_2', 'staff', 'video', 'i4x-edX-DemoX-video-8c0028eb2a724f48a074bc184cd8635f', 'viewed', 1),
+        roster_alias = self.task.default_config_override['module-engagement']['alias']
```
mulby: I would prefer we move this logic to the ElasticsearchService

edx/analytics/tasks/elasticsearch_load.py
```diff
@@ -150,6 +151,8 @@ def init_local(self):
                 )
             )
 
+        self.attempted_removal = True
```
brianhw, line 156: This may be obscure enough that it would benefit from a comment.

edx/analytics/tasks/module_engagement.py
```diff
@@ -1249,15 +1248,18 @@ class ModuleEngagementWorkflowTask(ModuleEngagementDownstreamMixin,
         description=ModuleEngagementRosterIndexTask.obfuscate.description
     )
 
+    overwrite = luigi.BooleanParameter(default=False)
```
brianhw, line 1292: This also needs a comment, so I have a clue as to why it's here.  (As opposed to including OverwriteOutputMixin.)

edx/analytics/tasks/tests/acceptance/services/es.py
```diff
@@ -0,0 +1,31 @@
...
+        return self.elasticsearch_client
+
+    def reset(self):
+        response = self.elasticsearch_client.indices.get_aliases(name='*{}'.format(self.alias_suffix))
+        for index, alias_info in response.iteritems():
+            for alias in alias_info.keys():
+                if alias.endswith(self.alias_suffix):
+                    self.elasticsearch.indices.delete(index=index)
+
+        #get documents from the marker index that have their target_index set to current alias
```
brianhw: nit: space after #, init cap, end with period.

edx/analytics/tasks/tests/acceptance/test_module_engagement.py
```diff
@@ -19,56 +22,180 @@ class ModuleEngagementAcceptanceTest(AcceptanceTestCase):
...
+        query = {"query": {"match_all": {}}}
+        response = self.es.client.search(index=roster_alias, body=query)
+
+        self.maxDiff = None
```
brianhw: If this is just for debugging, consider removing it....

edx/analytics/tasks/elasticsearch_load.py
mulby, line 156: I just pushed a commit to add a comment here.

@HassanJaveed84 you can pull my commit when you pick this up again. I figured since this was part of my original patch, I should fix it.

edx/analytics/tasks/module_engagement.py
mulby, line 1292: Done.

edx/analytics/tasks/tests/acceptance/test_module_engagement.py
```diff
@@ -19,56 +22,180 @@ class ModuleEngagementAcceptanceTest(AcceptanceTestCase):
...
-            'ModuleEngagementIntervalTask',
-            '--interval', '2015-04-16',
-            '--n-reduce-tasks', str(self.NUM_REDUCERS),
+        with self.export_db.cursor() as cursor:
+            cursor.execute('SELECT course_id, start_date, end_date, metric, range_type, low_value, high_value FROM module_engagement_metric_ranges')
+            results = cursor.fetchall()
+
+        self.assertItemsEqual([
+            row for row in results
+        ], [
```
brianhw: nit:  I would put the list comprehension on a  single line:  
self.assertItemsEqual(
     [row for row in results],
    [
...
    ]
)

brianhw:
Finished a pass -- only nits found.

HassanJaveed84:
Updated some stuff:
- The acceptance test generates the full alias and passes to ElasticsearchService.
- Simplified ElasticsearchService's reset as it has full alias(uses match instead of wildcard)
- Made an 'alias' property in ElasticsearchService(the acceptance test uses it).

edx/analytics/tasks/tests/acceptance/test_module_engagement.py
HassanJaveed84: Done

edx/analytics/tasks/tests/acceptance/test_module_engagement.py
HassanJaveed84: Done.

edx/analytics/tasks/tests/acceptance/services/es.py
```diff
@@ -0,0 +1,42 @@
+try:
```
mulby: nit: I think it would be clearer if we called the service "elasticsearch" instead of "es"

mulby:
One nit. This is ready to merge as far as I'm concerned.

:+1:

brianhw:
LGTM.  :+1:

HassanJaveed84:
Had some merge conflicts during rebase(in module_engagement.py). 
Just want to config that I resolved them correctly before merging.

mulby:
:+1: rebase looks good to me

PR #238, opened 2016-04-14
Fix multiple runs of tasks when executing with --overwrite

mulby:
Ensure the roster task sets `self.attempted_removal` without this the overwrite mixin will not recognize the task as complete. This was causing the task to run twice, however, since it was intitialized once, the first time it runs it was deleting the "old" index and the second time `self.indexes_for_alias` still contained the old index, so it tried to clean it up _again_. This was causing the error we were seeing.

Reviewers:
- [x] @HassanJaveed84 
- [x] @brianhw


HassanJaveed84:
LGTM :+1:

brianhw:
LGTM too. :+1:

PR #239, opened 2016-04-15
Ability to pass extra args to virtualenv command.

mtyaka:
Allows you to pass arguments such as `--system-site-packages` to the `virtualenv` command when creating the virtual environment on EMR nodes.

Using system site-packages lowers pipeline deployment time for around 3 minutes according to my tests. I believe the speedup comes from using system scipy, which is a transitive dependency.

Additional 2-3 minute speedup can be achieved by changing numpy requirement in requirements/default.txt to 1.10.4, which is the version that comes pre-installed on EMR 4.4.0 nodes.
This PR does not update numpy to 1.10.4, we will probably want to do that in a separate pull request.

cf. [OLIVE-29](https://openedx.atlassian.net/browse/OLIVE-29)

### DIFF
--- a/edx/analytics/tasks/launchers/remote.py
+++ b/edx/analytics/tasks/launchers/remote.py
@@ -40,6 +40,7 @@ def main():
     parser.add_argument('--sudo-user', help='execute the shell command as this user on the cluster', default='hadoop')
     parser.add_argument('--workflow-profiler', choices=['pyinstrument'], help='profiler to run on the launch-task process', default=None)
     parser.add_argument('--wheel-url', help='url of the wheelhouse', default=None)
+    parser.add_argument('--virtualenv-extra-args', help='additional arguments passed to virtualenv command when creating the virtual environment', default=None)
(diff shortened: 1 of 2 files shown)


edx-webhook:
Thanks for the pull request, @mtyaka! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-analytics-pipeline&number=239

brianhw:
LGTM.  :+1:

pomegranited:
:+1: Tested with [PR 35](https://github.com/edx/edx-analytics-configuration/pull/35#issuecomment-210725402).

PR #242, opened 2016-04-19
overwrite N days of module engagement history

mulby:
Given the incremental nature of this workflow, we need to handle "late" events. The short term solution for this problem is to simply reprocess the last N days of data to pick up any stragglers.

Reviewers:
- [x] @brianhw 
- [ ] @HassanJaveed84

### DIFF
--- a/edx/analytics/tasks/module_engagement.py
+++ b/edx/analytics/tasks/module_engagement.py
@@ -57,9 +57,20 @@ class ModuleEngagementRecord(Record):
                                                      ' date.')
 
 
-class ModuleEngagementDownstreamMixin(
-    WarehouseMixin, MapReduceJobTaskMixin, EventLogSelectionDownstreamMixin
-):
+class OverwriteFromDateMixin(object):
+    """Supports overwriting a subset of the data to compensate for late events."""
+
+    overwrite_from_date = luigi.DateParameter(
+        description='This parameter is passed down to the module engagement model which will overwrite data from a date'
+                    ' in the past up to the end of the interval. Events are not always collected at the time that they'
+                    ' are emitted, sometimes much later. By re-processing past days we can gather late events and'
+                    ' include them in the computations.',
+        significant=False
+    )
+
+
+class ModuleEngagementDownstreamMixin(WarehouseMixin, MapReduceJobTaskMixin, EventLogSelectionDownstreamMixin,
+                                      OverwriteFromDateMixin):
     """Common parameters and base classes used to pass parameters through the engagement workflow."""
 
     # Required parameter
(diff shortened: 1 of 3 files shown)


edx/analytics/tasks/module_engagement.py
```diff
@@ -1248,16 +1274,25 @@ class ModuleEngagementWorkflowTask(ModuleEngagementDownstreamMixin,
...
     )
+    overwrite_from_date = None
+    overwrite_n_days = luigi.IntParameter(
+        config_path={'section': 'module_engagement', 'name': 'overwrite_n_days'},
+        significant=False,
+        default=3
+    )
 
     def requires(self):
+        overwrite_from_date = self.date - datetime.timedelta(days=self.overwrite_n_days)
```
brianhw, line 1293: Usually we put such parameter calculations into **init**().  Is there any issue with this being set too late?  (I guess it's not being used elsewhere, so it may be okay.  And more importantly, it's not resetting _itself_ every time.)

brianhw:
Looks good so far.

mulby:
@brianhw I added in the descriptions for this parameter, anything left to do before merging?

edx/analytics/tasks/module_engagement.py
```diff
@@ -293,19 +302,27 @@ class ModuleEngagementIntervalTask(MapReduceJobTaskMixin, EventLogSelectionDowns
                                    OverwriteOutputMixin, luigi.WrapperTask):
     """Compute engagement information over a range of dates and insert the results into Hive and MySQL"""
 
+    overwrite_from_date = luigi.DateParameter(
+        description='',
```
brianhw: This doesn't have a description.  Any reason that ModuleEngagementIntervalTask doesn't inherit from ModuleEngagementDownstreamMixin?

mulby: Fixed. Yeah, it actually needs the interval parameter, it could probably be refactored to share more code, but I believe that DRY isn't always better.

edx/analytics/tasks/module_engagement.py
```diff
@@ -1248,16 +1277,26 @@ class ModuleEngagementWorkflowTask(ModuleEngagementDownstreamMixin,
         default=False,
         description=ModuleEngagementRosterIndexTask.obfuscate.description
     )
+    overwrite_from_date = None
+    overwrite_n_days = luigi.IntParameter(
+        config_path={'section': 'module_engagement', 'name': 'overwrite_n_days'},
+        significant=False,
+        default=3,
+        description=ModuleEngagementDownstreamMixin.overwrite_from_date.description
```
brianhw: I don't think the description of overwrite_from_date works very well as a description also of overwrite_n_days.  (Particularly the mention of "will overwrite data from this day up until end of the interval.")  If you do aim to reword it into a common description, I would suggest it be put into a constant, with a comment saying that it serves two purposes, and then include that.  Doing this kind of include is pretty brittle.

mulby: I'm not convinced the increased brittleness is actually worth having an entire set of separate constants.

I've reworded it though.

brianhw: I'm not convinced that an entire set is needed.  I count only one separate constant, but one that would be used in three places.  My issue with doing it this way is that one has no idea in the original comment that it is being used elsewhere.  If it were pulled out, that possibility would be more obvious.

mulby: Yup, I get the increased brittleness for the reason you specified. That said, with any given parameter, it's not clear where exactly it is used given the tangled web of inheritance at play here. This definitely isn't a step in the right direction, but I see it as no worse than our current state.

Note that I'm using the mechanism to share descriptions in many locations in this workflow, so I would need to create a bunch of constants.

brianhw:
Not happy about the way descriptions are shared.  

I've thumbed it though.   :+1:

mulby:
@brianhw this is ready for final review

brianhw:
LGTM.   Squash and merge.  :+1:

PR #243, opened 2016-04-20
Fix event-export acceptance tests on EMR 4.5.0

brianhw:
Compare log files with assertEventLogEqual.

This is instead of using 'diff' for event files, since events may not be
in the same order, and may not have the dict keys in the same order.
Use new method for testing event exports, both regular and per-course.

@mulby @HassanJaveed84


mulby:
:+1:

PR #244, opened 2016-04-22
Changed elasticsearch test service name.

HassanJaveed84:
@mulby should I run an acceptance test to confirm the fix ?


mulby:
:+1: 

Yeah I would start one up. If it gets passed the point where it failed last time then we can merge this.

HassanJaveed84:
Merging this in, just verified.

PR #245, opened 2016-04-25
Support missing ES config in acceptance tests and allow empty overwrite in acceptance tests

mulby:
The first change should allow the "elasticsearch_host" to be omitted from the configuration dictionary, which will simply disable the service.

The second change fixes the broken roster engagement test.

Reviewers:
- [ ] @HassanJaveed84 
- [ ] @brianhw

### DIFF
--- a/edx/analytics/tasks/tests/acceptance/services/elasticsearch_service.py
+++ b/edx/analytics/tasks/tests/acceptance/services/elasticsearch_service.py
@@ -14,8 +14,12 @@ class ElasticsearchService(object):
         else:
             connection_class = None
 
-        self._elasticsearch_client = elasticsearch.Elasticsearch(hosts=[config['elasticsearch_host']], connection_class=connection_class)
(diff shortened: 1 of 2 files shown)


mulby:
tests now pass

mulby:
http://jenkins-ci.analytics.edx.org/view/ad-hoc/job/edx-analytics-pipeline-acceptance-test-manual/795/

HassanJaveed84:
LGTM 👍

brianhw:
LGTM. :+1: 

But just so I understand:  In order to get any tests to work at all, I added "elasticsearch_host" and "elasticsearch_connection_class" as I had found them defined in the master.  I expect the latter being missing would not have made a difference on the non-ES tests.  And the change here I assume will allow those tests that don't use ES to run.  But what is the expected behavior for the LA test?  Will it get skipped if there is no ES host, or will it fail?

mulby:
It will get skipped

PR #247, opened 2016-04-27
New task to aggregate questions for tags distribution

dmitry-viskov:
There is a new task to aggregate questions according the related tags:

```
export UNIQUE_NAME=$(date +%Y-%m-%dT%H_%M_%SZ)

launch-task TagsDistributionWorkflow \
          --local-scheduler \
          --src hdfs://localhost:9000/data/ \
          --include '*tracking.log*' \
          --dest hdfs://localhost:9000/edx-analytics-pipeline/output/tags_distribution_raw/$UNIQUE_NAME/data \
          --name $UNIQUE_NAME \
          --output-root hdfs://localhost:9000/edx-analytics-pipeline/output/tags_distribution/ \
          --marker hdfs://localhost:9000/edx-analytics-pipeline/output/tags_distribution_raw/$UNIQUE_NAME/marker \
          --n-reduce-tasks 1 \
          --skip-setup \
          --question-tags 'difficulty,learning_outcome'
```

it is very similar to the existed `AnswerDistributionWorkflow` task. The main difference in the required `question-tags` param.
On the output after task will be executed in database should be the new table `tags_distribution` with the columns:

| id | course_id | module_id | total_submissions | correct_submissions | problem_display_name | difficulty | learning_outcome | created |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| 1001 | ... | ... | 121 | 57 | question name | Hard | Learned everything | 2016-04-27 00:34:14 |

@mulby could you please take a look. If you OK with suggested approach I'll add tests to cover the new functionality


openedx-webhooks:
Thanks for the pull request, @StraNNiKK! I've created [OSPR-1247](https://openedx.atlassian.net/browse/OSPR-1247) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

mulby:
@StraNNiKK we had originally discussed simply adding columns to the answer_dist table, did that approach not pan out as well as we had hoped?

dmitry-viskov:
@mulby hi. I thought about it. Yes primarily we discussed only to add additional columns and modify the existed `AnswerDistributionWorkflow` task. But there is one corner case that may cause troubles. I mean in the case if somebody decide to change the tag in CMS for some question. 
For example we have a radio button question with 4 possible answers. And at the some point of time somebody decided to change the difficulty tag from the `Easy` to `Medium`. Because the existed `AnswerDistributionWorkflow` task aggregates data according the unique `answer_id` value there may be situation when nobody answered to the first radio variant. So in the result table `answer_distribution` there could be situation like this one:

| id | ... | module_id | part_id | ... | difficulty |
| --- | --- | --- | --- | --- | --- |
| 1001 | ... | radio_question_module_id | 72503148547db5be12dc_2_1 | ... | Easy |
| 1002 | ... | radio_question_module_id | 72503148547db5be12dc_2_2 | ... | Medium |
| 1003 | ... | radio_question_module_id | 72503148547db5be12dc_2_3 | ... | Medium |
| 1004 | ... | radio_question_module_id | 72503148547db5be12dc_2_4 | ... | Medium |

Due to this there may be problems in the new server API (that calculates total and correct submissions) what tag value should be used for the question with `radio_question_module_id`.

To solve this problem we need to take tag from the last question event (not answer event). But as I could see it could be done only with creation the new aggregation task.

What do you think about all this?

mulby:
@StraNNiKK This case is already handled by the existing code, note that it first identifies the most recent answer to the problem here:

https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/answer_dist.py#L285

It then pulls out a bunch of metadata from that most recent answer here:

https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/answer_dist.py#L296

It then uses that metadata to actually construct the row here:

https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/answer_dist.py#L355

You could follow a very similar parttern for the difficulty and outcome.

dmitry-viskov:
@mulby 

> This case is already handled by the existing code, note that it first identifies the most recent answer to the problem here:
> 
> https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/answer_dist.py#L285

yes. I saw this. Are you sure that it is the most recent answer **to the problem** ?
because it seems that it is the most recent answer for the specific variant
https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/answer_dist.py#L250
because as I could see the aggregation key is:

```
(edX/DemoX/Demo_Course, i4x-edX-DemoX-problem-c554538a57664fac80783b99d9d6da7c_2_1),
```

so it seems that it is the most recent answer to `i4x-edX-DemoX-problem-c554538a57664fac80783b99d9d6da7c_2_1`

mulby:
Ah, I see your point, if there are multiple variants that could result in multiple different values for the tag.

mulby:
@brianhw do you see a better way to handle this?

mulby:
Assuming @brianhw doesn't see a better way to handle this, I'll give feedback on the proposal as it stands right now:

1) Do you need a CSV to be generated? That CSV is exposed on the instructor dashboard, but is a legacy pattern. We would prefer that the information is exposed in Insights and provide a CSV download mechanism via the patterns already established in Insights.
2) We try to avoid subclassing workflows. Workflows are intended to pull together a group of tasks for a specific purpose. I would recommend creating your own workflow that doesn't share code with the answer distribution workflow.

This from a brief look at the code, I'll take a deeper look after Brian weighs in.

dmitry-viskov:
> 1) Do you need a CSV to be generated?

No. I don't think that we need it. Ok. I'll remove this from code

> 2) We try to avoid subclassing workflows.

So as I understand you correctly I shouldn't use `AnswerDistributionWorkflowBase` as a parent for `AnswerDistributionWorkflow` and the new one `TagsDistributionWorkflow` ?

mulby:
> So as I understand you correctly I shouldn't use AnswerDistributionWorkflowBase as a parent for AnswerDistributionWorkflow and the new one TagsDistributionWorkflow ?

Correct.

dmitry-viskov:
@mulby I've updated the PR according your comments.

If there are no any disagreements about the current approach I'm going to cover it with tests

brianhw:
Sorry I didn't see this sooner.  I'm out of time today, but I'll review it tomorrow.

edx/analytics/tasks/answer_dist.py
```diff
@@ -548,6 +562,82 @@ def normalize(value):
             return unicode(answer_value)
 
 
+class TagsDistributionPerCourseMixin(AnswerTaggingMixin):
```
mulby: I don't think this needs to be a mixin, it can just be included inline in `TagsDistributionPerCourse`, can't it?

edx/analytics/tasks/answer_dist.py
```diff
@@ -548,6 +562,82 @@ def normalize(value):
...
+        """
+        Args:  tab-delimited values in a single text line
+
+        Yields:  (course_id, problem_id), (timestamp, answer_id, answer_data)
+
+        """
+        course_id, answer_id, date, answer_data = line.split('\t')
+        answer = json.loads(answer_data)
+        problem_id = answer.get('problem_id')
+        yield (course_id, problem_id), (date, answer_id, answer_data)
```
mulby: I don't think you actually care about "answer_id". Given a capa problem with multiple parts, this will uniquely identify one of the parts (aka questions) in the problem. I think you want to aggregate by problem_id which represents the complete problem.

edx/analytics/tasks/answer_dist.py
```diff
@@ -548,6 +562,82 @@ def normalize(value):
...
+    def mapper(self, line):
+        """
+        Args:  tab-delimited values in a single text line
+
+        Yields:  (course_id, problem_id), (timestamp, answer_id, answer_data)
+
+        """
+        course_id, answer_id, date, answer_data = line.split('\t')
+        answer = json.loads(answer_data)
+        problem_id = answer.get('problem_id')
```
mulby: If this is `None`, you probably want to just filter out the record

edx/analytics/tasks/answer_dist.py
```diff
@@ -548,6 +562,82 @@ def normalize(value):
...
+
+        for _timestamp, answer_id, value_string in reversed(values):
+            answer = json.loads(value_string)
+            if answer_id not in answer_id_list:
+                answer_id_list.append(answer_id)
+
+            attempt_category = answer.get('attempt_category')
+            correct = answer.get('correct')
+
+            if attempt_category == 'last':
```
mulby: You can (and should) perform this filter in the mapper.

``` python
def mapper(line):
    course_id, answer_id, date, answer_data = line.split('\t')
    answer = json.loads(answer_data)
    if answer.get('attempt_category') != 'last':
        return None
    # ...
```

edx/analytics/tasks/answer_dist.py
```diff
@@ -548,6 +562,82 @@ def normalize(value):
...
+                total_count += 1
+                if correct:
+                    correct_count += 1
+
+        result['Total submissions'] = total_count / len(answer_id_list)
```
mulby: Just a note, this will not actually count "submissions". The `ProblemCheckEvent` task only focuses on the first and last submission of each learner to each problem. Some problems may allow multiple submissions (say 5). The 2nd, 3rd and 4th submissions are excluded from the analysis here.

The more I look at this, I'm wondering if it wouldn't be simpler for you to just construct your own task and not try to reuse the answer distribution code, you probably want something like this:

``` python
class TagDistributionTask(EventLogSelectionMixin, MapReduceJobTask):

    def mapper(self, line):
        value = self.get_event_and_date_string(line)
        if value is None:
            return
        event, date_string = value

        if event.get('event_type') != 'problem_check' and event.get('event_source') != 'server':
            return

        timestamp = eventlog.get_event_time_string(event)
        if timestamp is None:
            return

        course_id = eventlog.get_course_id(event)
        if not course_id:
            return

        event_data = eventlog.get_event_data(event)
        if event_data is None:
            return

        problem_id = event_data.get('problem_id')
        if not problem_id:
            return

        is_correct = event_data.get('success') == 'correct'
        asides_data = event.get('context', {}).get('module', {}).get('asides', [])
        for aside in asides_data:
            if aside.get('type') != 'TaggingAside':
                continue

            for tag_name, tag_value in aside.get('tags', {}).iteritems():
                yield (course_id, problem_id, tag_name), (timestamp, tag_value, is_correct, 1)

        course_id, problem_id = key

        latest_timestamp = None
        latest_tag_value = None
        num_correct = 0
        total = 0
        for timestamp, tag_value, is_correct, count in values:
            if latest_timestamp is None or timestamp > latest_timestamp:
                latest_timestamp = timestamp
                latest_tag_value = tag_value

            if is_correct:
                num_correct += 1

            total +=1

        yield (course_id, problem_id, tag_name, tag_value), (num_correct, total)
```

mulby:
@StraNNiKK - if you do end up implementing completely different than answer dist, I would recommend reviewing module_engagement.py for more recent patterns. The answer_dist.py code is one of the first workflows we wrote and the patterns in here are quite old.

brianhw:
I agree with Gabe that if your needs are sufficiently simpler than answer-dist, there's a lot of benefit to decoupling your code from the complexity in the answer-dist codebase (not to mention the brittleness due to its age).  
And if your needs are as simple as what would be met by Gabe's TagDistributionTask, that sounds like a much better approach.  However, I'm not entirely clear if it meets your needs, because I'm not sure what you are trying to count.  Are you counting number of users who made a submission for a given problem, or are you counting the number of total submissions, regardless whether they come from the same user or from different users?  If counting users, then you will likely need two tasks:  one to get the last response for a user, so you know if it should be counted as correct or not; and a second to sum and to determine the most recent tags to apply per problem (i.e. from the most recent user).
But if you are working at the problem level, and not at the individual "answers" within a problem, it would be good to skip the complications arising from finding those answers, dealing with variants, and all that.

dmitry-viskov:
@mulby @brianhw thanks for the helpful comments!
you're right - seems that it is simpler to construct the separate task. I've created the new file `tags_dist.py` where put all logic.
Does it now looks better?

@brianhw 

> Are you counting number of users who made a submission for a given problem, or are you counting the number of total submissions, regardless whether they come from the same user or from different users?

I'm counting the total and correct submissions for the each pair: problem_id + connected tag value

edx/analytics/tasks/tags_dist.py
```diff
@@ -0,0 +1,180 @@
...
+    """
+    output_root = luigi.Parameter(
+        description='Directory to store the output in.',
+    )
+
+
+class TagsDistributionPerCourse(EventLogSelectionMixin, MapReduceJobTask):
+    """Calculates tags distribution."""
+
+    output_root = luigi.Parameter()
```
brianhw: I would just use the mixin above, rather than redefining it.

edx/analytics/tasks/tags_dist.py
```diff
@@ -0,0 +1,180 @@
...
+from edx.analytics.tasks.mysql_load import MysqlInsertTask, MysqlInsertTaskMixin
+from edx.analytics.tasks.decorators import workflow_entry_point
+from edx.analytics.tasks.pathutil import EventLogSelectionDownstreamMixin, EventLogSelectionMixin
+
+
+import logging
+log = logging.getLogger(__name__)
+
+
+class BaseTagsDistributionDownstreamMixin(EventLogSelectionDownstreamMixin):
```
brianhw: nit: maybe leave off the Base.   The "Downstream" seems to provide appropriate indication.

edx/analytics/tasks/tags_dist.py
```diff
@@ -0,0 +1,180 @@
...
+        #asides_data = event.get('context', {}).get('module', {}).get('asides', [])
+        #for aside in asides_data:
+        #    if aside.get('type') != 'TaggingAside':
+        #        continue
+        #
+        #    for tag_name, tag_value in aside.get('tags', {}).iteritems():
+        #        yield (course_id, problem_id, tag_name), (timestamp, tag_value, is_correct, 1)
+
+        saved_tags = event_data.get('saved_tags', {})
+        yield (course_id, problem_id, problem_display_name), (timestamp, saved_tags, is_correct)
```
brianhw: I'm not clear why you want the problem_display_name in the key here.  If the display_name of a problem changes, do you want to calculate a separate set of submission counts?  Otherwise I would just put it in the value and pull out in the same way you get the tags for the problem.

edx/analytics/tasks/tags_dist.py
```diff
@@ -0,0 +1,180 @@
...
+        num_correct = 0
+        num_total = 0
+
+        latest_timestamp = None
+        latest_tags = None
+
+        for timestamp, saved_tags, is_correct in values:
+            if latest_timestamp is None or timestamp > latest_timestamp:
+                latest_timestamp = timestamp
+                latest_tags = saved_tags.copy() if saved_tags else None
```
brianhw, line 99: If you thought the values weren't going to be too numerous to hold them all in memory, you could sort the values, which should order them by timestamp.   Then you can get latest_tags by just pulling the values[-1] out.

edx/analytics/tasks/tags_dist.py
```diff
@@ -0,0 +1,180 @@
...
+        if not latest_tags:
+            return
+        else:
+            for tag_key, tag_val in latest_tags.iteritems():
+                yield (course_id, problem_id, problem_display_name, tag_key, tag_val), (num_total, num_correct)
+
+
+class TagsDistributionToMySQLTaskWorkflow(
+        BaseTagsDistributionDownstreamMixin,
+        MapReduceJobTaskMixin,
```
brianhw, line 134: You could probably move MapReduceJobTaskMixin to TagsDistributionDownstreamMixin.

edx/analytics/tasks/tags_dist.py
dmitry-viskov: @brianhw Not quite understand what mixin should be used here.
I've tried to use `MultiOutputMapReduceJobTask` but in this case hadoop task every time failed with an error:

```
2016-04-29 00:22:11,697 INFO 9191 [luigi-interface] hadoop.py:265 - 16/04/29 00:22:11 INFO streaming.StreamJob: Output directory: /edx-analytics-pipeline/output/tags_distribution/marker/-5885320816231369411-temp-2016-04-29T00-21-46.018133
/var/lib/analytics-tasks/devstack/venv/local/lib/python2.7/site-packages/luigi/util.py:289: UserWarning: Keyword argument fail_if_exists is deprecated, use raise_if_exists
  .format(old_name, new_name))
2016-04-29 00:22:17,720 INFO 9191 [luigi-interface] worker.py:296 - [pid 9191] Worker Worker(salt=476060231, host=precise64, username=hadoop, pid=9191) done      TagsDistributionPerCourse(output_root=hdfs://localhost:9000/edx-analytics-pipeline/output/tags_distribution/, source=('hdfs://localhost:9000/data/',), interval=2012-01-01-2020-01-01, expand_interval=2 days, 0:00:00, pattern=('.*tracking.log.*',))
2016-04-29 00:22:17,721 INFO 9191 [luigi-interface] worker.py:282 - [pid 9191] Worker Worker(salt=476060231, host=precise64, username=hadoop, pid=9191) running   TagsDistributionToMySQLTaskWorkflow(source=('hdfs://localhost:9000/data/',), interval=2012-01-01-2020-01-01, expand_interval=2 days, 0:00:00, pattern=('.*tracking.log.*',), database=reports, credentials=/edx/etc/edx-analytics-pipeline/output.json, output_root=hdfs://localhost:9000/edx-analytics-pipeline/output/tags_distribution/, marker=hdfs://localhost:9000/edx-analytics-pipeline/output/tags_distribution/marker, overwrite=True)
cat: `/edx-analytics-pipeline/output/tags_distribution/marker/-5885320816231369411': Is a directory
2016-04-29 00:22:20,725 ERROR 9191 [luigi-interface] worker.py:304 - [pid 9191] Worker Worker(salt=476060231, host=precise64, username=hadoop, pid=9191) failed    TagsDistributionToMySQLTaskWorkflow(source=('hdfs://localhost:9000/data/',), interval=2012-01-01-2020-01-01, expand_interval=2 days, 0:00:00, pattern=('.*tracking.log.*',), database=reports, credentials=/edx/etc/edx-analytics-pipeline/output.json, output_root=hdfs://localhost:9000/edx-analytics-pipeline/output/tags_distribution/, marker=hdfs://localhost:9000/edx-analytics-pipeline/output/tags_distribution/marker, overwrite=True)
  File "/var/lib/analytics-tasks/devstack/venv/local/lib/python2.7/site-packages/luigi/worker.py", line 292, in _run_task
    task.run()
  File "/var/lib/analytics-tasks/devstack/venv/local/lib/python2.7/site-packages/edx/analytics/tasks/mysql_load.py", line 330, in run
    self.insert_rows(cursor)
  File "/var/lib/analytics-tasks/devstack/venv/local/lib/python2.7/site-packages/edx/analytics/tasks/mysql_load.py", line 292, in insert_rows
    for row_count, row in enumerate(self.rows(), start=1):
  File "/var/lib/analytics-tasks/devstack/venv/local/lib/python2.7/site-packages/edx/analytics/tasks/mysql_load.py", line 173, in rows
    for line in fobj:
  File "/var/lib/analytics-tasks/devstack/venv/local/lib/python2.7/site-packages/luigi/format.py", line 119, in __iter__
    self._finish()
  File "/var/lib/analytics-tasks/devstack/venv/local/lib/python2.7/site-packages/luigi/format.py", line 85, in _finish
    raise RuntimeError('Error reading from pipe. Subcommand exited with non-zero exit status %s.' % self._process.returncode)
RuntimeError: Error reading from pipe. Subcommand exited with non-zero exit status 1.
```

edx/analytics/tasks/tags_dist.py
dmitry-viskov: I've changed value to `TagsDistributionDownstreamMixin`

edx/analytics/tasks/tags_dist.py
dmitry-viskov: you are right. fixed

edx/analytics/tasks/tags_dist.py
dmitry-viskov, line 134: done

edx/analytics/tasks/tags_dist.py
dmitry-viskov, line 99: I thought about it. But I don't see any significant profit from it. Because anyway I need a `for in` cycle to calculate number of correct submissions.

dmitry-viskov:
I've added unit tests to cover new task

edx/analytics/tasks/tags_dist.py
brianhw: @StraNNiKK Sorry not to be clearer, and this is a pretty small nit.  You have just defined TagsDistributionDownstreamMixin above this, which defines the output_root parameter.  You should be able to just include TagsDistributionDownstreamMixin in your definition of TagsDistributionPerCourse, and remove this line.

edx/analytics/tasks/tags_dist.py
```diff
@@ -0,0 +1,172 @@
...
+        if event_data is None:
+            return
+
+        problem_id = event_data.get('problem_id')
+        if not problem_id:
+            return
+
+        is_correct = event_data.get('success') == 'correct'
+
+        problem_display_name = event.get('context').get('module', {}).get('display_name', None)
```
mulby: You shouldn't need to keep track of the problem display name, this can be pulled by Insights when it displays the UI. We only pulled it through the pipeline for answer distribution because it also generates a CSV file that is sent directly to instructors.

edx/analytics/tasks/tags_dist.py
```diff
@@ -0,0 +1,172 @@
...
+            if is_correct:
+                num_correct += 1
+
+            num_total +=1
+
+        if not latest_tags:
+            return
+        else:
+            for tag_key, tag_val in latest_tags.iteritems():
+                yield (course_id, problem_id, latest_problem_display_name, tag_key, tag_val), (num_total, num_correct)
```
mulby: I recommend defining a record to represent this data type.

See: https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/module_engagement.py#L44

It can also be used to generate the MySQL schema in the next task and generate documentation.

edx/analytics/tasks/tags_dist.py
```diff
@@ -0,0 +1,172 @@
...
+    def indexes(self):
+        return [
+            ('course_id',),
+            ('module_id',),
+            ('course_id', 'module_id'),
+        ]
+
+
+@workflow_entry_point
+class TagsDistributionWorkflow(
```
mulby: I don't think this class is necessary you can just decorate the MySQL insert task with `@workflow_entry_point`

mulby:
Also - please add in an acceptance test for this functionality.

nedbat:
@mulby tell @lamagnifica if there are doc changes needed, kthxbye.

dmitry-viskov:
@mulby @brianhw hi. I've updated the PR according your comments and added an acceptance test:

```
cd /var/lib/analytics-tasks/devstack/repo/scripts
./run_acceptance.sh --test test_tags_dist.py:TagsDistributionAcceptanceTest
```

edx/analytics/tasks/tests/acceptance/test_tags_dist.py
```diff
@@ -0,0 +1,160 @@
...
+import datetime
+
+from edx.analytics.tasks.url import url_path_join
+from edx.analytics.tasks.tests.acceptance import AcceptanceTestCase
+
+
+class TagsDistributionAcceptanceTest(AcceptanceTestCase):
+    """Acceptance tests for Tags Distribution Task -> MySQL"""
+
+    INPUT_FILE = 'tags_dist_acceptance_tracking.log'
```
mulby, line 14: I would prefer you include less data in this file. Maybe 100 events or so is fine.

edx/analytics/tasks/tags_dist.py
```diff
@@ -0,0 +1,168 @@
...
+        TagsDistributionDownstreamMixin,
+        EventLogSelectionDownstreamMixin,
+        MapReduceJobTaskMixin,
+        MysqlInsertTask,
+        luigi.WrapperTask):
+    """Calculate tags distribution and output to files and to database."""
```
mulby: This should include more information. It will be visible here: http://edx-analytics-pipeline-reference.readthedocs.io/en/latest/workflow_entry_point.html

You can see other examples there. There are Makefile targets for generating the docs that you can use locally.

Consider explaining the various input parameters, how the task is intended to be run, why it exists, related dependencies in edx-platform etc.

edx/analytics/tasks/tests/acceptance/test_tags_dist.py
dmitry-viskov, line 14: ok. I've reduced the number of the test events

dmitry-viskov:
@mulby Updated the PR according your comments

dmitry-viskov:
@mulby hi. what is the next with this PR ?

mulby:
@StraNNiKK I assume you updated this based on our changes to the event emission code?

mulby:
Assuming the location of the tags in the event is correct, this looks good to me. :+1: 

I'll wait for approval from @brianhw before merging this.

dmitry-viskov:
> I assume you updated this based on our changes to the event emission code?

quite true

dmitry-viskov:
@brianhw could you please take a look

edx/analytics/tasks/tags_dist.py
```diff
@@ -0,0 +1,186 @@
...
+
+@workflow_entry_point
+class TagsDistributionWorkflow(
+        TagsDistributionDownstreamMixin,
+        EventLogSelectionDownstreamMixin,
+        MapReduceJobTaskMixin,
+        MysqlInsertTask,
+        luigi.WrapperTask):
+    """
+    This task could be used to calculate total and correct submissions
```
brianhw: nit:   Change to " This task calculates total and correct submissions for each unique pair....".

edx/analytics/tasks/tags_dist.py
```diff
@@ -0,0 +1,186 @@
...
+    for each unique pair: problem id + connected tag.
+    It makes sense to use this task only if:
+
+    - xblock asides are enabled in your CMS and LMS settings (it could be done through the Django admin panel)
+    - you have already fill MySQL tables `tagging_tagcategories` and `tagging_tagavailablevalues` with necessary tags values
```
brianhw: nit:  s/fill/populated/

edx/analytics/tasks/tags_dist.py
```diff
@@ -0,0 +1,186 @@
...
+    - you have already marked necessary questions with some tags
```
brianhw: nit:  not quite clear.  Do you mean "You have already marked some questions with necessary tags."?  I'm not clear what "necessary" means here.

edx/analytics/tasks/tags_dist.py
```diff
@@ -0,0 +1,186 @@
...
+
+    Only in this case after someone answer the question the assigned tags will be included in the emitted event
+    and this kind of distribution could be calculated.
```
brianhw: How does this sound:  "Once the LMS is properly configured, the assigned tags will be included in the emitted event each time a learner answers a tagged question.  This task can then calculate aggregates based on these tagged events."

edx/analytics/tasks/tags_dist.py
```diff
@@ -0,0 +1,186 @@
...
+
+    The results of this task will be stored in the separate MySQL table in the `reports` DB
+    and will be used in the analytics dashboard to categorize questions by tags and to measure
+    how students in the LMS perform on different learning outcomes.
+    """
```
brianhw, line 153: How about:  "The results of this task will be stored in the "tags_distribution" table in the resultstore database.
When enabled in Insights, this table is used to display counts for a given set of tags.  When the tags are used to label learning outcomes for individual problems, Insights can provide a sense of how learners in the LMS perform on different learning outcomes."   This is the first time you mentioned "learning outcomes", and it seems like it could be one way of using tags.  Also, not everyone will be using this in Insights.

brianhw:
The code looks good!  Just a few suggestions to improve the documentation.

dmitry-viskov:
@brianhw big thanks for the assistance! I've updated the PR according your comments. now it looks better?

dmitry-viskov:
I can't look what is wrong with acceptance tests. For me this URL is unavailable: http://jenkins-ci.analytics.edx.org/job/edx-analytics-pipeline-acceptance-auto-pr/681/

brianhw:
Yes, I'm trying to run the tests on our Jenkins.  The unit tests did pass (though with a few minor pylint and pep8 issues), but I think our acceptance test job is having trouble with the fork.

dmitry-viskov:
@brianhw I've updated PR and fixed some little defects that found locally with `pep8` and `pylint` commands. If I can change something else please let me know.

brianhw:
Acceptance test ran fine.  I had to blow away some directories on our acceptance-test cluster in order to get it to run your fork, but that worked, and I have now merged it.  Thanks for your contribution!

brianhw:
@lamagnifica Let's follow up on whether there are documentation changes that need to be made.

PR #248, opened 2016-05-02
WIP: Add country_code to RDX events.

HassanJaveed84:
@mulby @brianhw


edx/analytics/tasks/events_obfuscation.py
```diff
@@ -333,6 +337,15 @@ def _obfuscate_event(self, event):
             else:
                 event['event'] = event_data
 
+        ip_address = event.get('ip')
```
HassanJaveed84, line 340: I'm not sure what would be best place to include this logic. 
This piece of code is not obfuscating anything, but rather adding additional info.

brianhw, line 340: I think this is a fine place.  The 'ip' entry is being obfuscated, in that it is getting remapped.  Just not in-place, because of the change in representation.  And this has to be done before the 'ip' value is actually removed, so it can't be done later than line 351.

mulby:
Looks great to me so far.

edx/analytics/tasks/events_obfuscation.py
```diff
@@ -297,6 +288,7 @@ def _obfuscate_event(self, event):
 
         # Clean or remove values from context.
         if 'context' in event:
+            event['context'].pop('username', None)
```
brianhw: I'm not sure I agree that we should actually remove this.  I took a look at eight Berkeley courses, and found one that had usernames in the context.  And I looked at the code above, and I don't see a bug.  I was concerned that what he was reporting was caused by a problem in the obfuscation code, but I'm more likely to believe it's a property of the original data.  And if we start to make decisions about cleaning this stuff out, then I'm not sure where to stop.

HassanJaveed84:
@mulby  I tried a different approach than what we discussed earlier. 
@brianhw

HassanJaveed84:

edx/analytics/tasks/user_location.py
```diff
@@ -29,6 +29,36 @@ class GeolocationMixin(object):
...
 
+class GeolocationTask(object):
+
+    geoip = None
+
+    def geolocation_data_target(self):
+        """Defines target from which geolocation data can be read."""
+        raise NotImplementedError
+
+    def init_reducer(self):
```
mulby, line 40: maybe we should call `super().init_reducer()` here? same for final_reducer?

HassanJaveed84, line 40: Done.

HassanJaveed84, line 40: calling super for final_reducer results in an exception
TypeError: 'NotImplementedType' object is not callable.
I'm guessing that's because by default final_reducer is NotImplemented
https://github.com/edx/luigi/blob/edx/release/luigi/hadoop.py#L579

brianhw:
LGTM.   :+1:   (Squash first, of course.)

mulby:
:+1: from me as well

PR #249, opened 2016-05-04
Hassan/rdx country code 2

HassanJaveed84:
@mulby @brianhw 
I've moved out the geolocation bit into a util class in this PR. 
I'm not sure which approach is better, would appreciate some feedback.


HassanJaveed84:
Closing this as we are going to stick with the original PR.

PR #250, opened 2016-05-10
WIP: Pipeline load d_user_course_certificate

HassanJaveed84:
@mulby @brianhw


mulby:
Looking good so far

edx/analytics/tasks/load_internal_reporting_certificates.py
```diff
@@ -0,0 +1,89 @@
...
+        return None
+
+    @property
+    def columns(self):
+        return [
+            ('user_id', 'INTEGER NOT NULL'),
+            ('course_id', 'VARCHAR(200) NOT NULL'),
+            ('is_certified', 'INTEGER'),
+            ('enrollment_mode', 'VARCHAR(200)'),
+            ('final_grade', 'VARCHAR(5)')
```
brianhw, line 82: Out of curiosity, what is final_grade?  Just a letter?  (E.g.  A, B, C, ...)  Or a number between 0 and 100?  

Sizes:  (5) seems small.  And course_id(200) is on the smaller side too, since it's 256 in the LMS.  Probably doesn't matter in reality, and we can worry about standardizing and updating later.

edx/analytics/tasks/load_internal_reporting_certificates.py
```diff
@@ -0,0 +1,89 @@
...
+            , if(status='downloadable', 1, 0) as is_certified
+            , mode as enrollment_mode
+            , grade as final_grade
+            FROM certificates_generatedcertificate
+            """
+
+
+class LoadInternalReportingCertificatesToWarehouse(WarehouseMixin, VerticaCopyTask):
+
+    date = luigi.DateParameter()
```
brianhw, line 59: It will be helpful to add doc to this.  I don't know what this is supposed to represent.  Presumably the date of the Sqoop export.

brianhw:
LGTM so far, too.

edx/analytics/tasks/load_internal_reporting_certificates.py
HassanJaveed84, line 82: From what I've seen so far, it is a number between 0.0 to 1.0

edx/analytics/tasks/load_internal_reporting_certificates.py
HassanJaveed84, line 59: This task does not use ExternalURL as the insert_source_task but LoadInternalReportingCertificatesTableHive(which requires ImportGeneratedCertificatesTask). 

This date parameter is used in the partition path spec.

HassanJaveed84:
Added acceptance test. 
TODO: docstrings.

edx/analytics/tasks/database_imports.py
```diff
@@ -838,6 +838,23 @@ def columns(self):
...
+        return 'certificates_generatedcertificate'
+
+    @property
+    def columns(self):
+        return [
+            ('user_id', 'INT'),
+            ('course_id', 'STRING'),
+            ('grade', 'STRING'),
+            ('status', 'STRING'),
+            ('mode', 'STRING'),
```
brianhw, line 854: Based on our discussion after standup, does it make sense to also add in here the created_date and modified_date, if these exist on the table?  I think we can pull more here even if we don't immediately add it to the Vertica table, if we expect to do so in the near future.

While we're at it, are there other fields we should be grabbing now?  No others look useful, really, except maybe distinction and name.

edx/analytics/tasks/load_internal_reporting_certificates.py
brianhw, line 59: Right.   The partition path spec for ImportGeneratedCertificatesTask being the date that the Sqoop export of that table was performed, if it exists.   If it doesn't exist, it will create a partition with the specified date, rather than today's date.  

Which continues to be a weakness of the date parameter for Sqoop exports.  I'm wondering, off topic here, if it makes sense to just have Sqoop exports fail if the date specified is not found and is not today, rather than pretending that today's data should be given the specified date.  I would really like to be able to trust this partition date more.  Any issue with my creating a tech debt story for this?

edx/analytics/tasks/load_internal_reporting_certificates.py
brianhw, line 82: I see.  It's this size in the original LMS table.   Fine.

edx/analytics/tasks/load_internal_reporting_certificates.py
```diff
@@ -0,0 +1,89 @@
...
+    @property
+    def auto_primary_key(self):
+        """No automatic primary key here; user's id is enough."""
+        return None
+
+    @property
+    def columns(self):
+        return [
+            ('user_id', 'INTEGER NOT NULL'),
+            ('course_id', 'VARCHAR(200) NOT NULL'),
```
brianhw: I changed my mind.  Is it possible to make this 255, to match the original table, and many of our other uses of course_id?  The sooner we make this consistent, the better, I figure.

brianhw:
Still looks good.

edx/analytics/tasks/database_imports.py
mulby, line 854: I think we wait to do this once we have parity with pentaho.

mulby:
Looking good so far. This feels like it's almost ready for final review?

edx/analytics/tasks/load_internal_reporting_certificates.py
HassanJaveed84: Done.

edx/analytics/tasks/load_internal_reporting_certificates.py
HassanJaveed84, line 59: I think as well that this deserves some story points, please do.

HassanJaveed84:

mulby:
:+1:

edx/analytics/tasks/database_imports.py
brianhw, line 854: My point is only that adding this to the import is independent of the Hive query below that needs to match pentaho.

edx/analytics/tasks/load_internal_reporting_certificates.py
```diff
@@ -0,0 +1,97 @@
...
+
+import luigi
+import logging
+
+from edx.analytics.tasks.vertica_load import VerticaCopyTask
+from edx.analytics.tasks.util.hive import HiveTableFromQueryTask, WarehouseMixin, HivePartition
+from edx.analytics.tasks.database_imports import ImportGeneratedCertificatesTask
+
+log = logging.getLogger(__name__)
+
```
brianhw, line 13: nit: does pep8 require another blank line here?

brianhw:
LGTM.  👍 after checking pep8 and squashing.

PR #251, opened 2016-05-10
do not overwrite the hive table

mulby:
I was seeing the following pattern:
1. Some days are already computed, so all luigi does is create a hive table and hive partitions for them.
2. For the days that were already computed "overwrite=False" which was being passed down to the table creation task.
3. Since some days are recomputed, they were requiring a table task with "overwrite=True", which hashes differently and causes separate tasks to be run. In this case the table is dropped before creation. At this point, however, there are several partitions already in the table, so they get dropped too.
4. When the ModuleEngagementIntervalTask goes to evaluate it's `complete()` method, it finds a bunch of missing partitions and raises the following error:

```
  File "/var/lib/analytics-tasks/automation/venv/lib/python2.7/site-packages/luigi/worker.py", line 288, in _run_task
    raise RuntimeError('Unfulfilled %s at run time: %s' % (deps, ', '.join(missing)))
RuntimeError: Unfulfilled dependencies at run time: ModuleEngagementPartitionTask...
```
1. It then re-runs the partition tasks which re-create their partitions and now everyone is happy and the workflow progresses.

Reviewers:
- [ ] @brianhw 
- [ ] @HassanJaveed84

### DIFF
--- a/edx/analytics/tasks/module_engagement.py
+++ b/edx/analytics/tasks/module_engagement.py
@@ -242,7 +242,6 @@ class ModuleEngagementPartitionTask(ModuleEngagementDownstreamMixin, HivePartiti
     def hive_table_task(self):
         return ModuleEngagementTableTask(
             warehouse_path=self.warehouse_path,
-            overwrite=self.overwrite,
         )
 
     @property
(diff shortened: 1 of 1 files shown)


brianhw:
So I assume that this fixes the incremental processing, by allowing the decision as to what gets run to be done at a higher level.  But does it do the right thing when one wants to overwrite?  I assume that we don't have a way to do that.  Is that the intention, and that we would try to solve that later?

PR #252, opened 2016-05-11
Support explicit decimal precision and scale in Hive >0.13

mulby:
This should resolve the issue we were seeing where decimal values were being interpreted as integers in Hive 1.0.0 but appeared to be working properly in Hive 0.11.

EMR 4.X is using hive 1.0.0 which corresponds to version number 0.14.1. We are currently using 0.11 in production. Apparently the handling of DECIMAL types changed in 0.13.

https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=27838462#LanguageManualTypes-Decimals

If you specify the type as simply "DECIMAL" in 0.11 it "does the right thing", but in 0.13 it assumes you mean "decimal(10,0)" which appears to translate to "a 10 digit integer". This rounds off any fractional part of the number. So "25.60" was being rounded up to 26.

> With the changes in the Decimal data type in Hive 0.13.0, the pre-Hive 0.13.0 columns (of type "decimal") will be treated as being of type decimal(10,0).  What this means is that existing data being read from these tables will be treated as 10-digit integer values, and data being written to these tables will be converted to 10-digit integer values before being written. To avoid these issues, Hive users on 0.12 or earlier with tables containing Decimal columns will be required to migrate their tables, after upgrading to Hive 0.13.0 or later.

This should fix the finance report w/o vertica acceptance tests on Hive 1.0.0.

Reviewers:
- [x] @brianhw 
- [x] @HassanJaveed84 

FYI:
- @bradenmacdonald

### DIFF
--- a/edx/analytics/tasks/util/hive.py
+++ b/edx/analytics/tasks/util/hive.py
@@ -21,6 +21,42 @@ def hive_database_name():
     return get_config().get('hive', 'database', 'default')
 
 
+def hive_version():
+    """
+    Returns the version of Hive that is declared in the configuration file. Defaults to 0.11 if it's not specified.
+
+    Returns: A tuple with each index representing a part of the version. For example: version="0.11.0.1" would return
+    (0, 11, 0, 1). The 0 indexed integer is the most significant part of the version number.
+    """
+    version_str = luigi.configuration.get_config().get('hive', 'version', '0.11')
+    return tuple([int(x) for x in version_str.split('.')])
+
+
+def hive_decimal_type(precision, scale):
+    """
+    Return the appropriate DECIMAL type declaration depending on the Hive version.
+
+    In versions >0.12, the syntax for declaring DECIMAL field types was changed. In Hive >0.12 using the DECIMAL type
+    without a precision or scale value defaults to the equivalent of DECIMAL(10, 0). This declaration only supports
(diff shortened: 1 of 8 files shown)


mulby:
Note I don't actually like this approach very much. I'm looking into other options now.

brianhw:
I guess (12,2) will work well enough for the values we have.  For basic money amounts, we expect all values to be good to two places.  I worry that the "per-item" values might encounter new rounding error due to the introduction of the two-decimal limit.   Just curious how you came up with 12,2?

mulby:
I took those values from the MySQL schema for these columns.

edx/analytics/tasks/util/hive.py
```diff
@@ -21,6 +21,19 @@ def hive_database_name():
     return get_config().get('hive', 'database', 'default')
 
 
+def hive_schema_version():
+    version_str = luigi.configuration.get_config().get('hive', 'schema_version', '0.11')
```
brianhw: docstring here and below?

mulby:
@brianhw and @HassanJaveed84 this is ready for final review

brianhw:
LGTM.  👍

mulby:
@HassanJaveed84 - can you review this when you get a chance?

HassanJaveed84:
👍 as well.

PR #253, opened 2016-05-12
In obfuscation, check for IPv6 addresses.

brianhw:
We are getting IPv6 addresses, and asking for location throws a GeoIPException with the message "Invalid database type; expected IPv4 address".  Just avoid doing so by checking for colons in ip_address, and skipping output in such cases.

@mulby @HassanJaveed84

### DIFF
--- a/edx/analytics/tasks/events_obfuscation.py
+++ b/edx/analytics/tasks/events_obfuscation.py
@@ -338,13 +338,12 @@ class ObfuscateCourseEventsTask(ObfuscatorMixin, GeolocationMixin, GeolocationTa
                 event['event'] = event_data
 
         ip_address = event.get('ip')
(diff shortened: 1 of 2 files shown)


mulby:
:+1:

edx/analytics/tasks/events_obfuscation.py
```diff
@@ -339,12 +339,12 @@ def _obfuscate_event(self, event):
 
         ip_address = event.get('ip')
         country_code = "UNKNOWN"
```
HassanJaveed84: We can probably remove this line as `event.update({'augmented': {'country_code': country_code}})` is now under the if condition.

HassanJaveed84:
👍

PR #254, opened 2016-05-19
Fix wheel override

mulby:
## Steps to reproduce

Attempt to setup a new analytics devstack using the [official instructions](http://edx.readthedocs.io/projects/edx-installing-configuring-and-running/en/latest/installation/analytics/index.html).
## Symptom

When you go to run the pipeline it will fail with an error message that looks like this:

```
  File "/var/lib/analytics-tasks/devstack/venv/bin/launch-task", line 6, in <module>
    from edx.analytics.tasks.launchers.local import main
  File "/var/lib/analytics-tasks/devstack/venv/local/lib/python2.7/site-packages/edx/analytics/tasks/launchers/local.py", line 16, in <module>
    import cjson
ImportError: /var/lib/analytics-tasks/devstack/venv/local/lib/python2.7/site-packages/cjson.so: undefined symbol: PyUnicodeUCS2_DecodeUnicodeEscape
```
## Root cause
- When we went to update the pipeline to support both EMR 4.X and EMR 2.4.11 we added support for installing the pipeline on either RHEL variants or Debian variants. This was necessary since the base AMI for EMR 4.X is amazon linux, which is a RHEL variant.
- When adding this support we renamed "install_env" to "install_env_debian". The remote-task script was explicitly overriding "install_env" with the value passed into the CLI using the `--wheel-url` parameter. Since that variable was no longer used by ansible, the `--wheel-url` parameter value was not actual being used by ansible.
## Context

Storing wheels on S3 dramatically speeds up installation times since compiling numpy, scipy, and pandas takes a very long time. Since these libraries rely on system-wide shared libraries, the compiled wheels have to match the runtime environment very closely. You can see this play out in the Traceback above. I haven't done a full analysis of the issue, but I suspect that the result of [this ifndef](https://github.com/python/cpython/blob/2.7/Include/unicodeobject.h#L146) statement results in different compiled code on ubuntu precise and debian squeeze. So if you try to run using a debian squeeze wheel on ubuntu precise, it will be very unhappy (like it is above). One work around for this issue is to simply pass in `--wheel-url ""`, which forces pip to recompile those libraries on the machine that will be running the code. However, since `--wheel-url` no longer had any effect, it became impossible to install to any other debian variant than "squeeze".
## Other notes

I was unable to merge this for a while since my original commit broke the Debian squeeze case (EMR 2.4.11) which is what we run in production. I just found the bug this morning and revised the patch accordingly.
## Summary of the fix

Define top level "wheel_url" and "wheel_pyver" variables. This simplifies the override logic and allows them to be overridden independently.
## Reviewers
- [ ] @pomegranited 
- [x] @brianhw

### DIFF
--- a/share/task.yml
+++ b/share/task.yml
@@ -37,11 +37,7 @@
       - hostname: github.com
         public_key: 'ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAq2A7hRGmdnm9tUDbO9IDSwBK6TbQa+PXYPCPy6rbTrTtw7PHkccKrpp0yVhp5HdEIcKr6pLlVDBfOLX9QUsyCOV0wzfjIJNlGEYsdlLJizHhbn2mUjvSAHQqZETYP81eFzLQNnPHt4EVVUh7VfDESU84KezmD5QlWpXLmvU31/yMf+Se8xhHTvKSCZIFImWwoG6mbUoWf9nzpIoaSjB+weqqUUmpaaasXVal72J+UX2B+2RPW3RcT0eOzQgqlJL3RKrTJvdsjE3JEAvGq3lGHSZXy28G3skua2SmVi/w4yCE6gbODqnTWlg7+wC604ydGXA8VJiS5ap43JXiUFFAaQ=='
     - local_log_dir: build/logs
-    - install_env_debian:
-        # EMR runs a modified version of Debian 6 (squeeze) on early AMIs, before switching their own RHEL variant.
-        WHEEL_URL: http://edx-wheelhouse.s3-website-us-east-1.amazonaws.com/Debian/squeeze
-        # Match the version of Python define in virtualenv_python above.
-        WHEEL_PYVER: 2.7
+    - wheel_pyver: 2.7
     - common_debian_variants:
       - Ubuntu
       - Debian
@@ -148,13 +144,18 @@
       shell: >
         . {{ working_venv_dir }}/bin/activate && make install
         chdir={{ working_repo_dir }}
-      environment: install_env_debian
+      environment:
+        WHEEL_URL: "{{ wheel_url | default('http://edx-wheelhouse.s3-website-us-east-1.amazonaws.com/Debian/squeeze') }}"
+        WHEEL_PYVER: "{{ wheel_pyver }}"
       when: ansible_distribution in common_debian_variants
 
     - name: virtualenv initialized on RHEL
       shell: >
         . {{ working_venv_dir }}/bin/activate && make install
         chdir={{ working_repo_dir }}
+      environment:
+        WHEEL_URL: "{{ wheel_url | default('http://edx-wheelhouse.s3-website-us-east-1.amazonaws.com/Amazon/emr4') }}"
+        WHEEL_PYVER: "{{ wheel_pyver }}"
       when: ansible_distribution in common_redhat_variants
 
     - name: logging configured
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 
 # If a wheel repository is defined, then have pip use that.  But don't require the use of wheel.
-ifdef WHEEL_PYVER
+ifneq ($(strip $(WHEEL_URL)),)
 	PIP_INSTALL = pip install --use-wheel --find-links=$$WHEEL_URL/Python-$$WHEEL_PYVER --allow-external mysql-connector-python
 else
 	PIP_INSTALL = pip install --allow-external mysql-connector-python
--- a/edx/analytics/tasks/launchers/remote.py
+++ b/edx/analytics/tasks/launchers/remote.py
@@ -154,11 +154,8 @@ def convert_args_to_extra_vars(arguments, uid):
         extra_vars['secure_config_branch'] = arguments.secure_config_branch
     if arguments.secure_config:
         extra_vars['secure_config'] = arguments.secure_config
-    if arguments.wheel_url:
-        extra_vars['install_env'] = {
-            'WHEEL_URL': arguments.wheel_url,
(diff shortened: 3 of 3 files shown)


mulby:
With this fix applied you should be able to run

``` bash
remote-task --wheel-url ""
```

Which should build all packages from scratch instead of using the Debian squeeze wheels.

brianhw:
LGTM.  👍

mulby:
@nedbat we will likely want to get this PR into Eucalyptus.

mulby:
Without it, single-instance analytics installs are broken.

share/task.yml
```diff
@@ -148,13 +144,18 @@
...
+        WHEEL_URL: "{{ wheel_url | default('http://edx-wheelhouse.s3-website-us-east-1.amazonaws.com/Debian/squeeze') }}"
+        WHEEL_PYVER: "{{ wheel_pyver }}"
       when: ansible_distribution in common_debian_variants
 
     - name: virtualenv initialized on RHEL
       shell: >
         . {{ working_venv_dir }}/bin/activate && make install
         chdir={{ working_repo_dir }}
+      environment:
+        WHEEL_URL: "{{ wheel_url | default('http://edx-wheelhouse.s3-website-us-east-1.amazonaws.com/Amazon/emr4') }}"
```
mulby, line 157: Note: this doesn't exist yet... but when wheels are built it will default to pulling them from this location for RHEL variants

mulby:
Single instance installs based on Ubuntu 12.04 (the default) can use `--wheel-url http://edx-wheelhouse.s3-website-us-east-1.amazonaws.com/Ubuntu/precise` to speed up installations.

mulby:
@bradenmacdonald are you or someone else from opencraft available to review this PR? Both Brian and Hassan are out of the office this week and this is breaking most open source installs.

bradenmacdonald:
@mulby, @pomegranited is on call this week; she should be able to review.

share/task.yml
```diff
@@ -148,13 +144,18 @@
       shell: >
         . {{ working_venv_dir }}/bin/activate && make install
         chdir={{ working_repo_dir }}
-      environment: install_env_debian
+      environment:
+        WHEEL_URL: "{{ wheel_url | default('http://edx-wheelhouse.s3-website-us-east-1.amazonaws.com/Debian/squeeze') }}"
+        WHEEL_PYVER: "{{ wheel_pyver }}"
```
pomegranited, line 149: @mulby These lines makes it impossible to install without a wheel.  If `wheel_url` is an empty string, it will be overridden by the default.

Also, it looks like it's the `WHEEL_PYVER` env var that determines whether the [Makefile defines pip install using the wheel or not](https://github.com/edx/edx-analytics-pipeline/blob/master/Makefile#L3-L4).

So even with `--wheel-url ""`, the remote pip install command still look like this:

```
/var/lib/analytics-tasks/analyticstack/venv/bin/python2.7 /var/lib/analytics-tasks/analyticstack/venv/bin/pip install --use-wheel --find-links=http://edx-wheelhouse.s3-website-us-east-1.amazonaws.com/Debian/squeeze/Python-2.7 --allow-external mysql-connector-python -U -r requirements/default.txt
```

pomegranited:
Thanks for this fix, @mulby ! It's almost ready.

Using wheels works, but I'm afraid that not using wheels doesn't.  See step 3 below.
- [ ] I tested this on my analyticstack (created 14 Jul).  Verification steps:
- Verified the described `import cjson` error on master when `launch-task` ran.
  
  ```
  export WHEEL_URL=http://edx-wheelhouse.s3-website-us-east-1.amazonaws.com/Ubuntu/precise
  remote-task --wheel-url $WHEEL_URL --vagrant-path ..  --remote-name analyticstack --override-config ${PWD}/config/devstack.cfg --wait TotalEventsDailyTask --interval 2016 --output-root hdfs://localhost:9000/output/ --local-scheduler
  ```
- Verified that using the PR branch ran sucessfully.  Note: had to remove the installed `venv` directory, otherwise pip wasn't reinstalling packages from the wheel.
  
  ```
  vagrant ssh -c "rm -rf /var/lib/analytics-tasks/analyticstack/venv/"
  remote-task --branch gabe/fix-wheel-override --wheel-url $WHEEL_URL --vagrant-path ..  --remote-name analyticstack --override-config ${PWD}/config/devstack.cfg --wait TotalEventsDailyTask --interval 2016 --output-root hdfs://localhost:9000/output/ --local-scheduler
  ```
- **FAILED** `remote-task --wheel-url ""` failed with `import cjson` error.  Please [see comments](https://github.com/edx/edx-analytics-pipeline/pull/254/files#r71648771).
- [x] I read through the code
- [x] Includes documentation: bugfix; no documentation change required.

mulby:
Great catch @pomegranited - I'll investigate. The ansible docs were a bit misleading about "default()", I thought it would only use the default value if it was undefined, apparently it will use it if it's empty or undefined.

mulby:
OK - it looks like default() was behaving as expected, however, the conditional in remote.py was a little to vague:

``` python
if arguments.wheel_url:
```

That was evaluating to `False` when an empty `--wheel-url` was being passed in. I'm testing all three scenarios again now.

mulby:
Ready for re-review @pomegranited

pomegranited:
@mulby Perfect!  :+1:  when tests pass.
- `remote-task --wheel-url ""` now succeeds, and builds all the pips from source.
- Also re-verified the build with `--wheel-url` set.

nedbat:
@mulby can you make another PR into the open-release/eucalyptus/master branch to get this in Eucalyptus? Thanks.

nedbat:
I've cherry-picked this into Eucalyptus.

PR #256, opened 2016-05-23
revise metric range logic

mulby:
Changes:
- a normal(0, 'inf') range will be emitted for every metric type for each course even if no users actually fall into that metric range (for example, no-one has actually watched any videos in the course, the video metric would still have a range of 0->inf).
- Improved handling of the edge cases where the N/A value is actually a threshold between the various ranges.

Todo:
- [x] Update acceptance tests
- [x] Look over docstrings for any needed updates
- [x] pep8/pylint

Reviewers:
- [x] @brianhw 
- [ ] @HassanJaveed84 

FYI: @dan-f @dsjen


brianhw:
LGTM so far.

brianhw:
LGTM. 👍

mulby:
@HassanJaveed84 when you get a chance can you review this?

HassanJaveed84:
LGTM as well. 👍

PR #257, opened 2016-05-27
Single workflow for pentaho loading.

HassanJaveed84:
@mulby @brianhw please review while I validate it.

### DIFF
--- a/edx/analytics/tasks/load_warehouse.py
+++ b/edx/analytics/tasks/load_warehouse.py
@@ -1,0 +1,84 @@
+"""
+Workflow to load the warehouse, this serves as a replacement for pentaho loading.
+"""
+import logging
+import luigi
+
+from edx.analytics.tasks.load_internal_reporting_certificates import LoadInternalReportingCertificatesToWarehouse
+from edx.analytics.tasks.load_internal_reporting_country import LoadInternalReportingCountryToWarehouse
+from edx.analytics.tasks.load_internal_reporting_course import LoadInternalReportingCourseToWarehouse
+from edx.analytics.tasks.load_internal_reporting_user_activity import LoadInternalReportingUserActivityToWarehouse
+from edx.analytics.tasks.load_internal_reporting_user_course import LoadInternalReportingUserCourseToWarehouse
+from edx.analytics.tasks.load_internal_reporting_user import LoadInternalReportingUserToWarehouse
+from edx.analytics.tasks.course_catalog import DailyLoadSubjectsToVerticaTask
+
+from edx.analytics.tasks.util.hive import WarehouseMixin
+
+log = logging.getLogger(__name__)
+
+
+class LoadWarehouse(WarehouseMixin, luigi.WrapperTask):
+    """Runs the workflow needed to load warehouse."""
+
+    date = luigi.DateParameter()
+
+    n_reduce_tasks = luigi.Parameter()
+
+    # We are not using VerticaCopyTaskMixin as OverwriteOutputMixin changes the complete() method behavior.
+    schema = luigi.Parameter(
+        config_path={'section': 'vertica-export', 'name': 'schema'},
+        description='The schema to which to write.',
+    )
+    credentials = luigi.Parameter(
+        config_path={'section': 'vertica-export', 'name': 'credentials'},
+        description='Path to the external access credentials file.',
+    )
+
+    overwrite = luigi.BooleanParameter(default=False)
+
+    def requires(self):
+        kwargs = {
+            'schema': self.schema,
+            'credentials': self.credentials,
+            'overwrite': self.overwrite,
+            'warehouse_path': self.warehouse_path,
+        }
+
+        yield (
+            LoadInternalReportingCertificatesToWarehouse(
+                date=self.date,
+                **kwargs
+            ),
+            LoadInternalReportingCountryToWarehouse(
+                date=self.date,
+                n_reduce_tasks=self.n_reduce_tasks,
+                **kwargs
+            ),
+            LoadInternalReportingCourseToWarehouse(
+                date=self.date,
+                n_reduce_tasks=self.n_reduce_tasks,
+                **kwargs
+            ),
+            LoadInternalReportingUserCourseToWarehouse(
+                date=self.date,
+                n_reduce_tasks=self.n_reduce_tasks,
+                **kwargs
+            ),
+            LoadInternalReportingUserActivityToWarehouse(
+                date=self.date,
+                n_reduce_tasks=self.n_reduce_tasks,
+                **kwargs
+            ),
+            LoadInternalReportingUserToWarehouse(
+                date=self.date,
+                n_reduce_tasks=self.n_reduce_tasks,
+                **kwargs
+            ),
+            DailyLoadSubjectsToVerticaTask(
+                date=self.date,
+                **kwargs
+            )
+        )
+
+    def output(self):
+        return [task.output() for task in self.requires()]
--- a/edx/analytics/tasks/location_per_course.py
+++ b/edx/analytics/tasks/location_per_course.py
@@ -16,13 +16,14 @@ from edx.analytics.tasks.url import ExternalURL, get_target_from_url, url_path_j
 from edx.analytics.tasks.user_location import BaseGeolocation, GeolocationMixin
 from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
 from edx.analytics.tasks.util import eventlog
-from edx.analytics.tasks.util.hive import hive_database_name
+from edx.analytics.tasks.util.hive import WarehouseMixin, hive_database_name
 from edx.analytics.tasks.decorators import workflow_entry_point
 
 log = logging.getLogger(__name__)
 
 
 class LastCountryOfUserMixin(
+        WarehouseMixin,
         MapReduceJobTaskMixin,
         EventLogSelectionDownstreamMixin,
         GeolocationMixin,
@@ -31,10 +32,7 @@ class LastCountryOfUserMixin(
     Defines parameters for LastCountryOfUser task and downstream tasks that require it.
 
     """
-    user_country_output = luigi.Parameter(
-        config_path={'section': 'last-country-of-user', 'name': 'user_country_output'},
-        description='Location of the resulting Hadoop output.',
-    )
+    pass
 
 
 class LastCountryOfUser(LastCountryOfUserMixin, EventLogSelectionMixin, BaseGeolocation, MapReduceJobTask):
@@ -55,7 +53,8 @@ class LastCountryOfUser(LastCountryOfUserMixin, EventLogSelectionMixin, BaseGeol
     def output(self):
         return get_target_from_url(
             url_path_join(
-                self.user_country_output,
+                self.warehouse_path,
+                'last_country_of_user',
                 'dt={0}/'.format(self.interval.date_b.strftime('%Y-%m-%d'))  # pylint: disable=no-member
             )
         )
@@ -115,7 +114,7 @@ class ImportLastCountryOfUserToHiveTask(LastCountryOfUserMixin, ImportIntoHiveTa
 
     @property
     def table_location(self):
-        return self.user_country_output
+        return url_path_join(self.warehouse_path, 'last_country_of_user')
 
     @property
     def table_format(self):
@@ -138,9 +137,22 @@ class ImportLastCountryOfUserToHiveTask(LastCountryOfUserMixin, ImportIntoHiveTa
             pattern=self.pattern,
             geolocation_data=self.geolocation_data,
             overwrite=self.overwrite,
-            user_country_output=self.user_country_output,
         )
 
+
+class ExternalLastCountryOfUserToHiveTask(ImportLastCountryOfUserToHiveTask):
+
+    interval = None
+    date = luigi.DateParameter()
+
+    @property
+    def partition_date(self):
+        return self.date.isoformat()  # pylint: disable=no-member
+
+    def requires(self):
+        yield ExternalURL(
+            url=url_path_join(self.warehouse_path, 'last_country_of_user', self.partition.path_spec) + '/'
+        )
 
 class InsertToMysqlLastCountryOfUserTask(LastCountryOfUserMixin, MysqlInsertTask):
     """
@@ -168,7 +180,6 @@ class InsertToMysqlLastCountryOfUserTask(LastCountryOfUserMixin, MysqlInsertTask
             pattern=self.pattern,
             geolocation_data=self.geolocation_data,
             overwrite=self.overwrite,
-            user_country_output=self.user_country_output,
         )
 
 
@@ -259,7 +270,6 @@ class QueryLastCountryPerCourseWorkflow(LastCountryOfUserMixin, QueryLastCountry
                 pattern=self.pattern,
                 geolocation_data=self.geolocation_data,
                 overwrite=self.overwrite,
-                user_country_output=self.user_country_output,
             ),
             InsertToMysqlLastCountryOfUserTask(
                 mapreduce_engine=self.mapreduce_engine,
@@ -269,7 +279,6 @@ class QueryLastCountryPerCourseWorkflow(LastCountryOfUserMixin, QueryLastCountry
                 pattern=self.pattern,
                 geolocation_data=self.geolocation_data,
                 overwrite=self.overwrite,
-                user_country_output=self.user_country_output,
             ),
             # We can't make explicit dependencies on this yet, until we
             # solve the multiple-credentials problem, as well as the split-kwargs
@@ -337,6 +346,5 @@ class InsertToMysqlCourseEnrollByCountryWorkflow(
             pattern=self.pattern,
             geolocation_data=self.geolocation_data,
             overwrite=self.overwrite,
-            user_country_output=self.user_country_output,
             course_country_output=self.course_country_output,
         )
--- a/edx/analytics/tasks/tests/test_location_per_course.py
+++ b/edx/analytics/tasks/tests/test_location_per_course.py
@@ -153,7 +153,6 @@ class ImportLastCountryOfUserToHiveTestCase(unittest.TestCase):
         """Provides minimum args for instantiating ImportLastCountryOfUserToHiveTask."""
         return {
             'interval': Year.parse('2013'),
-            'user_country_output': 's3://output/path',
         }
 
     def test_query_with_date_interval(self):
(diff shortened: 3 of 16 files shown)


edx/analytics/tasks/load_warehouse.py
```diff
@@ -0,0 +1,80 @@
...
+            ),
+            LoadInternalReportingUserCourseToWarehouse(
+                date=self.date,
+                n_reduce_tasks=self.n_reduce_tasks,
+                **kwargs
+            ),
+            InternalReportingUserActivityWorkflow(
+                interval=self.interval,
+                n_reduce_tasks=self.n_reduce_tasks,
+                history_schema=self.history_schema,
```
brianhw: I'm not sure I understand the difference between the schema and the history schema.  I had imagined that we would want all the contents of these going to the same schema if they came from the same workflow.  We didn't do that at present because different pipelines were doing the work:  pentaho and analytics pipeline.  So I guess do we want to exactly match the layout of the current jobs, or do the consolidation now into a single schema?   I would vote for the single schema.   @mulby do you think otherwise?

brianhw:
Looks good so far.  Question about the number of schema.  And then wrapping with a validation and rename step?  I see that comes in a separate story.

edx/analytics/tasks/load_warehouse.py
mulby: Single schema is my vote as well.

HassanJaveed84:
@gabe @brianhw please review.

edx/analytics/tasks/load_warehouse.py
```diff
@@ -0,0 +1,81 @@
...
+from edx.analytics.tasks.util.hive import WarehouseMixin
+
+import logging
+log = logging.getLogger(__name__)
+
+
+class LoadWarehouse(WarehouseMixin, luigi.WrapperTask):
+
+    date = luigi.DateParameter()
+    interval = luigi.DateIntervalParameter()
```
mulby: It's not clear to me what this interval is used for and how it relates to "date".

Is it actually used? Given that we aren't actually generating the user activity data, what is it used for?

edx/analytics/tasks/load_warehouse.py
```diff
@@ -0,0 +1,81 @@
...
+
+    n_reduce_tasks = luigi.Parameter()
+
+    user_country_output = luigi.Parameter(
+        config_path={'section': 'last-country-of-user', 'name': 'user_country_output'},
+        description='Location for intermediate output of location_per_course task.',
+    )
+
+    schema = luigi.Parameter()
+    credentials = luigi.Parameter()
```
mulby: Can we read these two from the config file by default?

mulby:
Looking for a little clarification about the parameters. Overall looking great!

edx/analytics/tasks/load_warehouse.py
HassanJaveed84: I think we should change LoadInternalReportingUserActivityToWarehouse to accept a date parameter instead of an interval.

However, LoadInternalReportingUserToWarehouse passes the interval down to ImportLastCountryOfUserToHiveTask. We can also change LoadInternalReportingUserToWarehouse to use ExternalURL(right now it requires AggregateInternalReportingUserTableHive).

mulby: > I think we should change LoadInternalReportingUserActivityToWarehouse to accept a date parameter instead of an interval.

Agreed!

> However, LoadInternalReportingUserToWarehouse passes the interval down to ImportLastCountryOfUserToHiveTask. We can also change LoadInternalReportingUserToWarehouse to use ExternalURL(right now it requires AggregateInternalReportingUserTableHive).

Also agreed, let's just assume the last_country_of_user folder has already been created and reference it with an ExternalURL.

HassanJaveed84: AggregateInternalReportingUserTableHive does join between 'auth_user', 'auth_userprofile' and 'last_country_of_user' to produce 'internal_reporting_user' hive table which then gets loaded in Vertica.

edx/analytics/tasks/load_warehouse.py
HassanJaveed84: Done.

edx/analytics/tasks/load_warehouse.py
```diff
@@ -0,0 +1,92 @@
...
+
+import logging
+log = logging.getLogger(__name__)
+
+
+class LoadWarehouse(WarehouseMixin, luigi.WrapperTask):
+
+    date = luigi.DateParameter()
+
+    interval = luigi.DateIntervalParameter(
```
HassanJaveed84: This is passed down to ImportLastCountryOfUserToHiveTask from LoadInternalReportingUserToWarehouse.

edx/analytics/tasks/load_warehouse.py
```diff
@@ -0,0 +1,92 @@
...
+        description='A date_interval object containing the interval over which to pull data for user location. '
+        'Should usually be from the beginning of the Open edX installation to the present day '
+        '(i.e. through the previous day).',
+    )
+
+    n_reduce_tasks = luigi.Parameter()
+
+    user_country_output = luigi.Parameter(
```
HassanJaveed84: Right now I don't see this defined in our secure repo.

HassanJaveed84:
@mulby @brianhw This is ready for a final review.

edx/analytics/tasks/load_warehouse.py
```diff
@@ -0,0 +1,96 @@
...
+from edx.analytics.tasks.load_internal_reporting_course import LoadInternalReportingCourseToWarehouse
+from edx.analytics.tasks.load_internal_reporting_user_activity import LoadInternalReportingUserActivityToWarehouse
+from edx.analytics.tasks.load_internal_reporting_user_course import LoadInternalReportingUserCourseToWarehouse
+from edx.analytics.tasks.load_internal_reporting_user import LoadInternalReportingUserToWarehouse
+from edx.analytics.tasks.course_catalog import DailyLoadSubjectsToVerticaTask
+from edx.analytics.tasks.vertica_load import VerticaCopyTaskMixin
+
+from edx.analytics.tasks.util.hive import WarehouseMixin
+
+import logging
```
brianhw: nit:  shouldn't this appear earlier?   pylint?

edx/analytics/tasks/load_warehouse.py
```diff
@@ -0,0 +1,96 @@
...
+class LoadWarehouse(WarehouseMixin, luigi.WrapperTask):
+    """Runs the workflow needed to load warehouse."""
+
+    date = luigi.DateParameter()
+
+    interval = luigi.DateIntervalParameter(
+        description='A date_interval object containing the interval over which to pull data for user location. '
+        'Should usually be from the beginning of the Open edX installation to the present day '
+        '(i.e. through the previous day).',
+    )
```
brianhw: I'm wondering if this can be handled differently, more in the way that enrollments are handled.  This interval should presumably end with whatever the value is for "date".   And if the start date should be the "beginning", then that could be set in config similarly to how "[enrollments] interval_start = 2013-11-01" is set.  (But with a different value, in our case.)

edx/analytics/tasks/load_warehouse.py
```diff
@@ -0,0 +1,96 @@
...
+            LoadInternalReportingCountryToWarehouse(
+                date=self.date,
+                n_reduce_tasks=self.n_reduce_tasks,
+                **kwargs
+            ),
+            LoadInternalReportingCourseToWarehouse(
+                run_date=self.date,
+                n_reduce_tasks=self.n_reduce_tasks,
+                **kwargs
+            ),
```
brianhw, line 61: Would it be hard to switch this to take "date" as an argument, instead of "run_date"?  That way, along with a switch from "interval" to "date" in "LoadInternalReportingUserToWarehouse", the same parameter can be passed to all.

edx/analytics/tasks/load_warehouse.py
mulby: Bigger question, do we need this? I thought we were going to be referencing this data using an ExternalURL?

edx/analytics/tasks/load_warehouse.py
brianhw: I also don't see it in test.cfg.   I wasn't sure, then, how this worked in an acceptance test, but then I realized that this doesn't have an acceptance test.  Next question:  should it?

edx/analytics/tasks/load_warehouse.py
HassanJaveed84, line 61: I'll make that change.

edx/analytics/tasks/load_warehouse.py
HassanJaveed84: Currently AggregateInternalReportingUserTableHive(which builds internal_reporting_user) defines requires as:

``` python
def requires(self):
        """
        This task reads from auth_user, auth_user_profile, and last_country_of_user, so require that they be
        loaded into Hive (via MySQL loads into Hive or via the pipeline as needed).
        """
        return [ImportAuthUserTask(overwrite=self.overwrite, destination=self.warehouse_path),
                ImportAuthUserProfileTask(overwrite=self.overwrite, destination=self.warehouse_path),
                ImportLastCountryOfUserToHiveTask(overwrite=self.overwrite,
                                                  interval=self.interval,
                                                  user_country_output=self.user_country_output,
                                                  n_reduce_tasks=self.n_reduce_tasks)]
```

If 'last_country_of_user' is already built, we can possibly use ExternalURL here like this:

``` python
def requires(self):
    partition_location=url_path_join(self.warehouse_path, 'last_country_of_user', HivePartition('dt', self.date.isoformat()).path_spec) + '/'
    return [ImportAuthUserTask(overwrite=self.overwrite, destination=self.warehouse_path),
            ImportAuthUserProfileTask(overwrite=self.overwrite, destination=self.warehouse_path),
            ExternalURL(url=partition_location)]
```

Does that seem reasonable ?

HassanJaveed84: I'm still unclear on how to proceed here.

brianhw: My understanding was that the guiding principle was that if tasks were already calculated as part of existing workflows, that this should depend on the results of those runs through ExternalURL rather than providing a competing way to run the same tasks.  Any given computational task would then be part of only one workflow.  

What I'm not sure above is whether the last-country data also needs to be loaded into Hive in order to be joined as part of the user data calculation.  The ExternalURL approach makes sense when we are simply loading S3 files directly into Vertica.  I can't quite remember, but I thought this was a step that calculated a Hive table (and thus writing to S3) based on data in Hive.  And ExternalURL will not get data into Hive.  So I doubt that the ExternalURL replacement will actually work.

HassanJaveed84: Right now, user-location-by-course Jenkins job executes ImportCountryWorkflow, which further runs ImportLastCountryOfUserToHiveTask as a dependency in the chain. The data is output in edx-analytics-scratch as it is output path is calculated using INTERMEDIATE_OUTPUT_ROOT from the Jenkins job.

HassanJaveed84: Should we stick with ImportLastCountryOfUserToHiveTask in the dependency, and change how we set interval(like you suggested) ?

brianhw: Are there any examples in the code of loading into Hive without running the underlying task?  I'll also take a look after I get the research exports restarted.

mulby: I just added something like this to the module engagement workflow. See the way that it depends on course enrollment.

mulby: https://github.com/edx/edx-analytics-pipeline/pull/260

HassanJaveed84:
@brianhw @mulby this is ready for review again.

brianhw:
Looking good.  I now know what's bothering me about the user_country location as temporary storage.  It's no longer a transient intermediate result if it is being referenced by another query in Hive.  We want the intermediate stuff to be data that we can just delete.  But if it's going to be joined the way it is now, that data should be in the warehouse.  
So I think we can simplify the arguments a little if we change the intermediate table to use warehouse-dir to define its location, and then let defaults set it for the workflow.  
Does that seem reasonable?

HassanJaveed84:
It does, I've made the changes, please review.

edx/analytics/tasks/location_per_course.py
```diff
@@ -115,7 +113,7 @@ def columns(self):
 
     @property
     def table_location(self):
-        return self.user_country_output
+        return url_path_join(self.warehouse_path, 'user_location')
```
brianhw: I think this should have the same name in the warehouse as in Hive:  so "last_country_of_user" instead of "user_location".

brianhw:
Just one comment about naming the "table" in S3.  Otherwise looks good.

HassanJaveed84:
@mulby @brianhw I've updated some tests to reflect the parameter changes.

edx/analytics/tasks/tests/test_location_per_course.py
```diff
@@ -197,7 +196,7 @@ def test_no_overwrite(self):
     def test_requires(self):
         task = ImportLastCountryOfUserToHiveTask(**self._get_kwargs())
         required_task = task.requires()
-        self.assertEquals(required_task.output().path, 's3://output/path/dt=2014-01-01')
+        self.assertEquals(required_task.output().path, 's3://fake/warehouse/last_country_of_userdt=2014-01-01')
```
brianhw: Is there a slash missing here?

edx/analytics/tasks/load_warehouse.py
```diff
@@ -0,0 +1,91 @@
...
+from edx.analytics.tasks.load_internal_reporting_user import LoadInternalReportingUserToWarehouse
+from edx.analytics.tasks.course_catalog import DailyLoadSubjectsToVerticaTask
+from edx.analytics.tasks.vertica_load import VerticaCopyTaskMixin
+
+from edx.analytics.tasks.util.hive import WarehouseMixin
+import logging
+log = logging.getLogger(__name__)
+
+
+class LoadWarehouse(WarehouseMixin, luigi.WrapperTask):
```
brianhw, line 20: Any reason not to use VerticaCopyTaskMixin?  Should take care of schema, credentials, overwrite, and would add read_timeout.    It seems you have already included it in the above.

edx/analytics/tasks/load_warehouse.py
```diff
@@ -0,0 +1,91 @@
...
+    """Runs the workflow needed to load warehouse."""
+
+    date = luigi.DateParameter()
+
+    interval_start = luigi.DateParameter(
+        config_path={'section': 'warehouse_load', 'name': 'interval_start'},
+    )
```
brianhw: This parameter is not very clear, as to how it relates to "warehouse_load" and what the interval that it should be starting.  My understanding is that this is only for specifying the interval used for calculating LastCountryOfUser.  We could move an interval_start kind of parameter down into that, in the same way that is done in enrollments.py.  But we're not actually running that task anymore, since we're loading its output through an ExternalURL in ExternalLastCountryOfUserToHiveTask.  In your implementation of the latter task, you redefined the requires() method, so that it no longer explicitly needs the interval to be provided.  But it implicitly uses the interval still, to define the partition_date property, and even more implicitly because it's required by the LastCountryOfUserMixin from which it is derived.  
I'm proposing that ExternalLastCountryOfUserToHiveTask be made to not require interval.  I think it would be enough to define interval = None at the class level, define a date parameter, and to redefine the partition_date property to use self.date instead of self.interval.  
With that, we would no longer need this parameter, and would not need to pass any interval down.  All subtasks would just use date.

edx/analytics/tasks/load_warehouse.py
```diff
@@ -0,0 +1,91 @@
...
+from edx.analytics.tasks.load_internal_reporting_country import LoadInternalReportingCountryToWarehouse
+from edx.analytics.tasks.load_internal_reporting_course import LoadInternalReportingCourseToWarehouse
+from edx.analytics.tasks.load_internal_reporting_user_activity import LoadInternalReportingUserActivityToWarehouse
+from edx.analytics.tasks.load_internal_reporting_user_course import LoadInternalReportingUserCourseToWarehouse
+from edx.analytics.tasks.load_internal_reporting_user import LoadInternalReportingUserToWarehouse
+from edx.analytics.tasks.course_catalog import DailyLoadSubjectsToVerticaTask
+from edx.analytics.tasks.vertica_load import VerticaCopyTaskMixin
+
+from edx.analytics.tasks.util.hive import WarehouseMixin
+import logging
```
brianhw: nit: I assume this ordering of imports will trigger pylint complaints.  Move import up.

edx/analytics/tasks/location_per_course.py
```diff
@@ -138,10 +137,16 @@ def requires(self):
...
-            user_country_output=self.user_country_output,
         )
 
 
+class ExternalLastCountryOfUserToHiveTask(ImportLastCountryOfUserToHiveTask):
+
+    def requires(self):
+        yield ExternalURL(
+            url=url_path_join(self.warehouse_path, 'last_country_of_user', self.partition.path_spec) + '/'
+        )
```
brianhw, line 155: See comments above about how to make this work with self.date instead of self.interval.

brianhw:
Sorry for being so incremental in my change requests.  I _think_ that this is my last.  Definitely better and better. But I think it would be clearer to get rid of the interval at the top level, without requiring invasive changes.

edx/analytics/tasks/load_warehouse.py
HassanJaveed84, line 20: Did not use VerticaCopyTaskMixin as it inherits from OverwriteOutputMixin, which I thought could change the behavior of the 'complete' method for this task. Is it okay to do so ?

edx/analytics/tasks/load_warehouse.py
HassanJaveed84: Done.

HassanJaveed84:
I've made the changes. DailyLoadSubjectsToVerticaTask still accepts a 'run_date' parameter, should we change it to 'date' as well ?

brianhw:
Yes, please.  Also, it would be good to fix the unit tests so that the pep8/pylint testing can run (and be addressed).  I'm finding that the SET TIMEZONE TO 'GMT' addition is throwing off some VerticaCopyTaskTest tests.

HassanJaveed84:
- Unit tests are passing.
- DailyLoadSubjectsToVerticaTask now accepts 'date' parameter instead of 'run_date'.
- Added in a comment explaining why we are not using VerticaCopyTaskMixin.

brianhw:
LGTM.  👍

HassanJaveed84:
@mulby can you please review this when you get a chance ?

mulby:
LGTM :+1:

PR #258, opened 2016-05-31
WIP: Fix d_course.

HassanJaveed84:
@mulby @brianhw 

This attempts to fix d_course job using the new API(http://edx.readthedocs.io/projects/edx-platform-api/en/latest/courses/courses.html#get-a-list-of-courses). 

The API does not return 'run' and 'course' fields in the response JSON.

### DIFF
--- a/edx/analytics/tasks/tests/test_course_information.py
+++ b/edx/analytics/tasks/tests/test_course_information.py
@@ -116,11 +116,10 @@ class TestCourseInformation(unittest.TestCase):
         input_data = {
             "results": [
                 {
-                    "id": "foo",
+                    "id": "bar/foo/2T2015",
                     "name": "Foo",
                     "org": "bar",
-                    "course": "Baz",
-                    "run": "2T2015",
+                    "number": "foo",
                     "start": "2015-08-24T00:00:00Z",
                     "end": "2016-08-25T00:00:00Z"
                 }
@@ -131,10 +130,10 @@ class TestCourseInformation(unittest.TestCase):
         # We expect to see this course with the mock structure information.
         self.assertEquals(data.shape[0], 1)
         expected = {
-            'course_id': 'foo',
+            'course_id': 'bar/foo/2T2015',
             'course_name': 'Foo',
             'course_org_id': 'bar',
-            'course_number': 'Baz',
+            'course_number': 'foo',
             'course_run': '2T2015',
             'course_start': '2015-08-24T00:00:00+00:00',
             'course_end': '2016-08-25T00:00:00+00:00'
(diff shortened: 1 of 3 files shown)


edx/analytics/tasks/tests/test_course_information.py
```diff
@@ -131,7 +130,7 @@ def test_single_course(self):
         # We expect to see this course with the mock structure information.
         self.assertEquals(data.shape[0], 1)
         expected = {
-            'course_id': 'foo',
+            'course_id': 'bar/foo/2T2015',
             'course_name': 'Foo',
             'course_org_id': 'bar',
             'course_number': 'Baz',
```
brianhw: Does it matter that the course_number here doesn't match the course_id?  I guess number is returned separate from the course_id, but it's a bit confusing.

edx/analytics/tasks/tests/test_course_information.py
```diff
@@ -131,7 +130,7 @@ def test_single_course(self):
         # We expect to see this course with the mock structure information.
         self.assertEquals(data.shape[0], 1)
         expected = {
-            'course_id': 'foo',
```
brianhw, line 134: I'm assuming that any of the old values listed here for course_id would be the values for course_number.  But what was the course_number representing before?  The run?

HassanJaveed84, line 134: Previously the 'course'  in the JSON response was being stored as 'course_number' in Vertica. 
The newer API does not have a 'course' field, but instead comes with a 'number' field(which is the same as course). 

Regarding old values for course_id being the values for course_number, not necessarily as previously the course_number was set to the value of 'course' in the JSON.

edx/analytics/tasks/tests/test_course_information.py
HassanJaveed84: Change to match the course_id.

HassanJaveed84:

mulby:
:+1:

brianhw:
LGTM.  👍

HassanJaveed84:
@mulby @brianhw I'm not sure what we decided on how to proceed with this story.

mulby:
@HassanJaveed84 I was hoping we could try to figure out why the change to timestamptz was necessary. Is there a way to load the data that matches the other tables (even though that method appears to be incorrect)?

edx/analytics/tasks/vertica_load.py
```diff
@@ -421,6 +421,8 @@ def run(self):
             # that would commit the transaction.
             self.init_copy(connection)
 
+            connection.cursor().execute("SET TIMEZONE TO 'GMT';")
```
HassanJaveed84, line 424: @mulby @brianhw this produces the same timestamps as compared to production schema.

HassanJaveed84:
@mulby @brianhw I reverted the TIMESTAMPTZ to TIMESTAMP. Setting Vertica's timezone to 'GMT' before loading produces the same timestamps as production schema.

mulby:
@HassanJaveed84 looks reasonable to me. Thanks for looking into the timestamp issue!

:+1:

PR #259, opened 2016-05-31
Fail the module engagement workflow if the enrollment data is missing

mulby:
Reviewers:
- [x] @brianhw 
- [x] @HassanJaveed84


brianhw:
LGTM.  👍 

But it reminds me that we encode table names outside of tasks pretty freely, and it might be good at some point if there were some encapsulation.

mulby:
Acceptance tests passed.

http://jenkins-ci.analytics.edx.org/view/ad-hoc/job/edx-analytics-pipeline-acceptance-test-manual/862/

HassanJaveed84:
LGTM 👍 
Quick question, previously we specified CourseEnrollment as a dependency, and I think the Jenkins build  for module_engagement is triggered to run after the enrollments job. Out of curiosity would Jenkins schedule module engagement if enrollments build fails ?

mulby:
Currently it will not schedule the module engagement build unless the enrollment job succeeds.

PR #260, opened 2016-06-02
Create missing enrollment hive table

mulby:
When running on a different cluster than the enrollment workflow we need to create the Hive table. The acceptance test passes because the two jobs are run in succession, so the Hive table persists unfortunately.

@brianhw 
@HassanJaveed84


edx/analytics/tasks/module_engagement.py
```diff
@@ -985,6 +985,14 @@ def columns(self):
         return ModuleEngagementRosterRecord.get_hive_schema()
 
 
+class ExternalCourseEnrollmentTableTask(CourseEnrollmentTableTask):
+
```
brianhw: Shouldn't this really be moved into the location where CourseEnrollmentTableTask is defined?

mulby: Done

mulby:
@brianhw this is ready for review again

brianhw:
LGTM. 👍

PR #261, opened 2016-06-06
Upgrade numpy to it's latest version

muzaffaryousaf:
We need to upgrade this to keep the numpy versions in-sync across different repos. 
Platform PR: https://github.com/edx/edx-platform/pull/12444


muzaffaryousaf:
@andy-armstrong @nedbat Different version in this repo would't complaint about VersionConflict, Do you think we should update the numpy version here ?

andy-armstrong:
@muzaffaryousaf I'd check with @mulby as I'm not familiar with this repo.

mulby:
This repo is installed in a separate virtualenv from edx-platform, so I would be surprised if you need to update this.

muzaffaryousaf:
@mulby I've just updated this because it was using the older version, no other reason. I realized it later that we're not using it in edx-platform.

PR #263, opened 2016-06-13
counting link clicks per course per day

sanfordstudent:
Reviewers:
Two of
- [ ] @brianhw  
- [x] @nasthagiri 
- [x] @mulby 

FYI @stroilova 

For [TNL-4538](https://openedx.atlassian.net/browse/TNL-4538).

This first PR creates a single MapReduce job, which outputs the number of link clicks and the number of external link clicks (ie to links outside the courseware) for each course on each day on which that course has at least one link clicked event.

When this PR is finished, I propose that I move the remaining queries into separate tickets.


edx/analytics/tasks/link_clicked.py
```diff
@@ -0,0 +1,98 @@
...
+
+log = logging.getLogger(__name__)
+LINK_CLICKED = 'edx.ui.lms.link_clicked'
+
+
+class LinkClickedTask(EventLogSelectionMixin, MapReduceJobTask):
+    """
+    Produce a data set that shows how many users clicked out of the
+    courseware in each course.
+    """
```
nasthagiri: Update docstring since this now also counts when users clicked within the courseware.

edx/analytics/tasks/link_clicked.py
```diff
@@ -0,0 +1,98 @@
...
+        if course_id is None or not opaque_key_util.is_valid_course_id(course_id):
+            log.error("encountered explicit link_clicked event with invalid course_id: %s", event)
+            return
+
+        target_url = event_data.get('target_url')
+        if target_url is None:
+            log.error("encountered explicit link_clicked event with no target_url: %s", event)
+
+        is_internal = 0
+        if re.match(r'https?://courses.edx.org', target_url):
```
nasthagiri: Is there a setting or environment variable for the LMS server?  Couldn't this code be run on open edX instances?

edx/analytics/tasks/link_clicked.py
```diff
@@ -0,0 +1,98 @@
...
+            is_internal = 1
+
+        datestamp = eventlog.timestamp_to_datestamp(timestamp)
+        yield (course_id), (datestamp, is_internal)
```
nasthagiri: Why convert the time value to a date value?  Wouldn't the loss of information prevent us from sorting the event stream by time?

edx/analytics/tasks/link_clicked.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def reducer(self, key, values):
+        """Emit the number of events for each course"""
+        course_id = key
+        date_to_click_count = {}
+        date_to_external_click_count = {}
+        for val in values:
+            date, is_internal = val
+            if not date_to_click_count.has_key(date):
```
nasthagiri: Consider using a `defaultdict` type instead.  It will help reduce these conditional initialization code.

edx/analytics/tasks/link_clicked.py
nasthagiri: I guess for now, since you are only counting the event, sorting by time is unneeded.

setup.cfg
```diff
@@ -68,6 +68,7 @@ edx.analytics.tasks =
     data_obfuscation   = edx.analytics.tasks.data_obfuscation:ObfuscatedCourseDumpTask
     events_obfuscation = edx.analytics.tasks.events_obfuscation:ObfuscateCourseEventsTask
     obfuscation = edx.analytics.tasks.obfuscation:ObfuscatedCourseTask
+    link-clicked = edx.analytics.tasks.link_clicked:LinkClickedTask
```
nasthagiri: Can we provide a better namespace for this task since it's specific to just LMS, and even further, just to the courseware within LMS?  For example, LMSCoursewareLinkClickedTask.

edx/analytics/tasks/link_clicked.py
sanfordstudent: That was my thought. Though I suppose we could always bucket by day when we generate the report and leave the full timestamp here.

edx/analytics/tasks/tests/test_link_clicked.py
```diff
@@ -0,0 +1,135 @@
...
+        line = 'this is garbage'
+        self.assert_no_map_output_for(line)
+
+    def test_unparseable_enrollment_event(self):
+        line = 'this is garbage but contains edx.ui.lms.link_clicked'
+        self.assert_no_map_output_for(line)
+
+    def test_link_clicked_event_count_per_course(self):
+        line = self.create_event_log_line()
+        self.assert_single_map_output(line, self.course_id, (self.datestamp, 0))
```
nasthagiri: For which parameter is the 0 integer being passed?

sanfordstudent: is_internal.

setup.cfg
sanfordstudent: Absolutely. I meant to update that name but neglected to do so.

edx/analytics/tasks/link_clicked.py
sanfordstudent: Thanks! That's very helpful

edx/analytics/tasks/tests/test_link_clicked.py
nasthagiri: nvm.  This is the value for "internal", right?

edx/analytics/tasks/link_clicked.py
sanfordstudent: I'm going to leave the full timestamp. Having the timestamp instead of the date will add little-to-no effort to later visualization of this data.

brianhw:
I'm wondering about the pivot from per-date aggregation to per-timestamp aggregation.  I understood the original intention was to calculate the former, and the conversion from timestamp to datestamp was appropriate.  (The datestamp was already returned as _date_string, since it's so often used, so you didn't actually have to calculate it.)
If you now want to postpone the aggregation, then I would just output the individual events.  Aggregating counts per timestamp seems like it will rarely count more than one event for a given millisecond increment.  
But if you don't aggregate here, then where?  I would be concerned that the event stream for these events will be rather large, if you're looking to shove this into the Mysql result store to be served up by the data api.

sanfordstudent:
@brianhw that's a good point and one I neglected to consider. The size of the event stream with no per-date aggregation will indeed be quite large. I don't think we have any current intent to view the events of a single day in order, so the full timestamp is unnecessary.

edx/analytics/tasks/lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,102 @@
...
+    Produce a data set that shows how many users clicked to a new page in each course.
+    Includes whether the click was within the edX courseware or to an external site.
+    """
+
+    output_root = luigi.Parameter()
+
+    enable_direct_output = True
+
+    def mapper(self, line):
+        value = self.get_event_and_date_string(line)
```
brianhw, line 37: An optimization we find useful with single-event tasks like this is to look for the event in the line first:  e.g. "if LINK_CLICKED not in line:  return".  That saves the time parsing for a lot of events that will be discarded.

edx/analytics/tasks/lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,102 @@
...
+
+        if value is None:
+            return
+        event, _date_string = value
```
brianhw: You have the _date_string here.  I would change this to "date_string", and that below in the yield statement instead of the timestamp_to_datestamp() call.

edx/analytics/tasks/lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,102 @@
...
+            log.error("encountered event with no event_type: %s", event)
+            return
+
+        if event_type != LINK_CLICKED:
+            return
+
+        timestamp = eventlog.get_event_time_string(event)
+        if timestamp is None:
+            log.error("encountered event with bad timestamp: %s", event)
+            return
```
brianhw: If you use date_string, then you don't need timestamp here at all....

edx/analytics/tasks/lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,102 @@
...
+
+        course_id = event.get('context').get('course_id')
+        if course_id is None or not opaque_key_util.is_valid_course_id(course_id):
+            log.error("encountered explicit link_clicked event with invalid course_id: %s", event)
+            return
+
+        target_url = event_data.get('target_url')
+        if target_url is None:
+            log.error("encountered explicit link_clicked event with no target_url: %s", event)
```
brianhw, line 68: If these are none, you probably want to return?  I don't know what urlparse will do with None, but it's probably not very useful?  (Here and for current_url.)

edx/analytics/tasks/lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,102 @@
...
+
+        current_url = event_data.get('current_url')
+        if target_url is None:
+            log.error("encountered explicit link_clicked event with no current_url: %s", event)
+
+        # a link is considered "internal" when it does not navigate away from the current host
```
brianhw: nit:  complete sentences, por favor.  Comments should start with cap and end with period.

edx/analytics/tasks/lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,102 @@
...
+        is_internal = 0
+        if urlparse(target_url).netloc == urlparse(current_url).netloc:
```
brianhw: Does urlparse raise exceptions on bad data?  And will its return value always have a .netloc attribute?

sanfordstudent: @brianhw per https://docs.python.org/2/library/urlparse.html#urlparse.urlparse I don't believe it does. It will just leave the members of the result object empty. I can check to ensure that the two URLs actually have netloc members before this line.

edx/analytics/tasks/lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,102 @@
...
+            is_internal = 1
+
+        yield (course_id), (eventlog.timestamp_to_datestamp(timestamp), is_internal)
```
brianhw: I would suggest making the key here (course_id, date_string), and the value could just be (is_internal), or it could be (1, is_internal).  The reducer can then sum the number of entries (or the 1) and sum the is_internal values, and then output that as the two counts for the (course_id, date_string).

edx/analytics/tasks/tests/test_lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,136 @@
...
+        self.default_event_template = 'link_clicked_event'
+
+        self.expected_key = (self.course_id)
+
+    def test_non_link_event(self):
+        line = 'this is garbage'
+        self.assert_no_map_output_for(line)
+
+    def test_unparseable_enrollment_event(self):
+        line = 'this is garbage but contains edx.ui.lms.link_clicked'
```
brianhw: nit: you can create this line using LINK_CLICKED, to make it clearer.

brianhw:
Finished a pass.  Great progress!

edx/analytics/tasks/lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def run(self):
+        output_target = self.output()
+        if not self.complete() and output_target.exists():
+            output_target.remove()
+
+        super(LMSCoursewareLinkClickedTask, self).run()
+
+class PushToVerticaLMSCoursewareLinkClickedTask(VerticaCopyTask):
+    """Push the LMS courseware link clicked task data to Vertica."""
```
sanfordstudent, line 119: @brianhw here's the task to push the data to Vertica.

edx/analytics/tasks/lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,134 @@
...
+            log.error("encountered explicit link_clicked event with invalid course_id: %s", event)
+            return
+
+        target_url = event_data.get('target_url')
+        if target_url is None:
+            log.error("encountered explicit link_clicked event with no target_url: %s", event)
+            return
+
+        current_url = event_data.get('current_url')
+        if target_url is None:
```
nasthagiri: s/target_url/current_url

edx/analytics/tasks/lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,134 @@
...
+            log.error("encountered explicit link_clicked event with no current_url: %s", event)
+            return
+
+        # A link is considered "internal" when it does not navigate away from the current host.
+        current_loc = urlparse(current_url).netloc
+        target_loc = urlparse(target_url).netloc
+        if current_loc == "" or target_loc == "":
+            return
```
nasthagiri: Add log statement before escaping out of the method.

edx/analytics/tasks/lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,134 @@
...
+
+        current_url = event_data.get('current_url')
+        if target_url is None:
+            log.error("encountered explicit link_clicked event with no current_url: %s", event)
+            return
+
+        # A link is considered "internal" when it does not navigate away from the current host.
+        current_loc = urlparse(current_url).netloc
+        target_loc = urlparse(target_url).netloc
+        if current_loc == "" or target_loc == "":
```
nasthagiri: Would we have events with relative URLs that won't have a `netloc`?

nasthagiri: If so, we should still handle those as internal links.

edx/analytics/tasks/lms_courseware_link_clicked.py
nasthagiri: Why return in this case?  Everything else succeeded.  Is this case that severe that the event needs to be skipped?

edx/analytics/tasks/lms_courseware_link_clicked.py
sanfordstudent: Oh, true. In that case the netloc would be `""`. And re: the below comment, the case is not very severe, no. We can probably just go ahead and still log the event.

edx/analytics/tasks/lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,133 @@
...
+        if event_type is None:
+            log.error("encountered event with no event_type: %s", event)
+            return
+
+        if event_type != LINK_CLICKED:
+            return
+
+        event_data = eventlog.get_event_data(event)
+        if event_data is None:
+            return
```
nasthagiri, line 54: Did you want to log here as well, since it's not expected?

sanfordstudent, line 54: Yes- sorry for these oversights. There should definitely be logging here as well.

edx/analytics/tasks/lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,133 @@
...
+        return get_target_from_url(url_path_join(self.output_root, '_SUCCESS')).exists()
+
+    def run(self):
+        output_target = self.output()
+        if not self.complete() and output_target.exists():
+            output_target.remove()
+
+        super(LMSCoursewareLinkClickedTask, self).run()
+
+class PushToVerticaLMSCoursewareLinkClickedTask(VerticaCopyTask):
```
nasthagiri, line 118: Is there a good way to implement unit tests for this class?

nasthagiri:
👍  Looks good to me.  Let's see what @brianhw has to say about the Vertica task.

edx/analytics/tasks/lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,134 @@
...
+        course_id, datestamp = key
+        date_to_click_count = defaultdict(int)
+        date_to_external_click_count = defaultdict(int)
+        for val in values:
+            is_external = val
+            date_to_click_count[datestamp] += 1
+            date_to_external_click_count[datestamp] += is_external
+
+        for date in date_to_click_count.keys():
+            yield (course_id, date, date_to_external_click_count[date], date_to_click_count[date])
```
brianhw: All values passed to the reducer share the same key value, so there is (now) only one date.  You don't need defaultdict here or any lookup for keys.  Init to 0, and remove the [datestamp] indexing.  And you could replace val with is_external, for that matter.   And have a single yield statement.

edx/analytics/tasks/lms_courseware_link_clicked.py
brianhw, line 118: No, unfortunately.   The only way we have for testing Vertica workflows is with acceptance tests.   Which is the next step to take here!

edx/analytics/tasks/lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,134 @@
...
+class PushToVerticaLMSCoursewareLinkClickedTask(VerticaCopyTask):
+    """Push the LMS courseware link clicked task data to Vertica."""
+    output_root = luigi.Parameter()
+    interval = luigi.DateIntervalParameter()
+    n_reduce_tasks = luigi.Parameter()
+    events_list_file_path = luigi.Parameter(default=None)
+
+    @property
+    def table(self):
+        return "event_type_distribution"
```
brianhw: Use your own table! : )

edx/analytics/tasks/lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,134 @@
...
+            output_target.remove()
+
+        super(LMSCoursewareLinkClickedTask, self).run()
+
+class PushToVerticaLMSCoursewareLinkClickedTask(VerticaCopyTask):
+    """Push the LMS courseware link clicked task data to Vertica."""
+    output_root = luigi.Parameter()
+    interval = luigi.DateIntervalParameter()
+    n_reduce_tasks = luigi.Parameter()
+    events_list_file_path = luigi.Parameter(default=None)
```
brianhw: You don't need this parameter.

edx/analytics/tasks/lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,134 @@
...
+            ('link_clicked_events', 'INT'),
+        ]
+
+    @property
+    def insert_source_task(self):
+        return LMSCoursewareLinkClickedTask(
+            output_root=self.output_root,
+            interval=self.interval,
+            n_reduce_tasks=self.n_reduce_tasks,
+            events_list_file_path=self.events_list_file_path
```
brianhw: You shouldn't use this parameter -- it's not a parameter of LMSCoursewareLinkClickedTask.

setup.cfg
```diff
@@ -68,6 +68,9 @@ edx.analytics.tasks =
     data_obfuscation   = edx.analytics.tasks.data_obfuscation:ObfuscatedCourseDumpTask
     events_obfuscation = edx.analytics.tasks.events_obfuscation:ObfuscateCourseEventsTask
     obfuscation = edx.analytics.tasks.obfuscation:ObfuscatedCourseTask
+    lms_courseware_link_clicked = edx.analytics.tasks.lms_courseware_link_clicked:LMSCoursewareLinkClickedTask
+    push_to_vertica_lms_courseware_link_clicked = edx.analytics.tasks.lms_courseware_link_clicked:PushToVerticaLMSCoursewareLinkClickedTask
```
brianhw, line 72: You only need to add one per file, and it's only used to get the file loaded and the tasks within the file imported.

brianhw:
Looks good!  You're ready to start looking at adding an acceptance test.  Let me know if you need help with that.

edx/analytics/tasks/tests/acceptance/__init__.py
```diff
@@ -54,6 +54,7 @@ def when_geolocation_data_available(function):
 
 def when_vertica_available(function):
     config = get_test_config()
+    print config
```
nasthagiri: don't forget to remove the `print` statements.  ;)

sanfordstudent: I definitely will. I'm just pushing what I have so far so that Gabe can help me figure out why none of my map inputs in my acceptance test are making it out of the mapper!

nasthagiri: ok. may the force be with you.

sanfordstudent:
Note that I will remove the commit that makes this work locally (but breaks production) before merging, obviously :)

edx/analytics/tasks/lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,140 @@
...
+        target_loc = urlparse(target_url).netloc
+
+        is_external = current_loc != target_loc and target_loc is not ""
+
+        yield (course_id, date_string), (is_external)
+
+    def reducer(self, key, values):
+        """
+        Emits the number of clicks for each course for each day with a click, and how many of the clicks
+        were to external links.
```
nasthagiri: Nit: remove "to".

edx/analytics/tasks/lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,140 @@
...
+        """
+        course_id, datestamp = key
```
nasthagiri: What is the type of `datestamp`?  Isn't it a `string` since that's what the `mapper` returns?  If so, rename to `date_string` for consistency and clarity.

edx/analytics/tasks/lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,140 @@
...
+        click_count = 0
+        external_click_count = 0
+        for is_external in values:
+            click_count += 1
+            external_click_count += is_external
```
nasthagiri, line 97: Isn't `is_external` a `boolean` instead of an `integer`?

sanfordstudent, line 97: Yes, but the comparison functions return 1 or 0. I can change this for clarity.

edx/analytics/tasks/tests/acceptance/fixtures/input/lms_courseware_link_clicked_acceptance_tracking.log
```diff
@@ -0,0 +1,20 @@
...
+{"username": "", "event_source": "browser", "name": "edx.ui.lms.link_clicked", "accept_language": "en-US,en;q=0.8", "time": "2016-06-05T16:14:53.168930+00:00", "agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/50.0.2661.102 Safari/537.36", "page": "http://courses.example.com/a/", "host": "precise64", "session": "", "referer": "http://courses.example.com/a/", "context": {"user_id": null, "org_id": "edX", "course_id": "course-v1:edX+DemoX+Demo_Course", "path": "/event"}, "ip": "10.0.2.2", "event": {"target_url": "http://courses.example.com/internal/", "current_url": "http://courses.example.com/a/"}, "event_type": "edx.ui.lms.link_clicked"}
```
nasthagiri, line 20: Add events with `target_url` equal to (assuming we run into each of these):
- `http://courses.example.com` (top-level page, and not a sub-page)
- `https://courses.example.com`  (SSL schema)
- `courses.example.com/a` (without HTTP schema)
- `/`
- `#`

Also, is the `current_url` always fully formed?

edx/analytics/tasks/tests/acceptance/test_lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,59 @@
+"""
+End to end test of the lms courseware link clicked events per day table loading task.
```
nasthagiri: Nit: consider adding a comma, here or there, in this sentence.  ;)

edx/analytics/tasks/tests/acceptance/test_lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,59 @@
...
+from datetime import date
+
+from edx.analytics.tasks.tests.acceptance import AcceptanceTestCase, when_vertica_available
+
+
+log = logging.getLogger(__name__)
+
+
+class LmsCoursewareLinkClickedAcceptanceTest(AcceptanceTestCase):
+    """End-to-end test of the workflow to load the warehouse's lms_courseware_link_clicked_events table."""
```
nasthagiri: Not sure about the formatting convention in the analytics repo, but the general edX convention is to always have a multi-line docstring.

edx/analytics/tasks/tests/acceptance/fixtures/input/lms_courseware_link_clicked_acceptance_tracking.log
sanfordstudent, line 20: The `current_url` should always be fully formed. The js will never log an event for `#` clicks since they just navigate within the page, so I didn't test those. The rest I'll add!

edx/analytics/tasks/tests/acceptance/test_lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,59 @@
...
+                'output',
+                'acceptance_expected_lms_courseware_link_clicked_events.csv')
+            expected = pandas.read_csv(expected_output_csv, parse_dates=True)
+
+            cursor.execute(
+                "SELECT * FROM {schema}.lms_courseware_link_clicked_events ORDER BY course_id, event_date"
+                    .format(schema=self.vertica.schema_name)
+            )
+            response = cursor.fetchall()
+            lms_courseware_link_clicked_events = pandas.DataFrame(response, columns=['record_number', 'course_id',
```
nasthagiri: Once again, for formatting, the edX convention is to keep the parens on separate lines, and indent 4 spaces in.  Not sure if it's different in this repo.

edx/analytics/tasks/tests/acceptance/test_lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,59 @@
...
+
+        self.validate_output()
+
+    def validate_output(self):
+        """Validates the output, comparing it to a csv of all the expected output from this workflow."""
+        with self.vertica.cursor() as cursor:
+            expected_output_csv = os.path.join(
+                self.data_dir,
+                'output',
+                'acceptance_expected_lms_courseware_link_clicked_events.csv')
```
nasthagiri: Nit: ending paren on new line.

edx/analytics/tasks/tests/acceptance/test_lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,59 @@
...
+        """Tests the workflow for the lms_courseware_link_clicked_events table, end to end."""
+        self.upload_tracking_log(self.INPUT_FILE, self.DATE)
+        self.task.launch([
+            'PushToVerticaLMSCoursewareLinkClickedTask',
+            '--output-root', self.test_out,
+            '--interval', str(2016),
+            '--n-reduce-tasks', str(self.NUM_REDUCERS)
+        ])
+
+        self.validate_output()
```
nasthagiri, line 39: Curious:  Is the `launch` command above asynchronous?  What's the guarantee that the task has completed?

edx/analytics/tasks/tests/acceptance/test_lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,59 @@
...
+                "SELECT * FROM {schema}.lms_courseware_link_clicked_events ORDER BY course_id, event_date"
+                    .format(schema=self.vertica.schema_name)
+            )
+            response = cursor.fetchall()
+            lms_courseware_link_clicked_events = pandas.DataFrame(response, columns=['record_number', 'course_id',
+                                                                                     'event_date',
+                                                                                     'external_link_clicked_events',
+                                                                                     'link_clicked_events'])
+
+            try:  # A ValueError will be thrown if the column names don't match or the two data frames are not square.
```
nasthagiri, line 68: Curious: What does "square" mean here?

nasthagiri, line 68: Or, do you mean, `the same dimensions`?

nasthagiri:
@sanfordstudent I have completed my pass.  I'm glad the acceptance tests are finally working.  I'd love to have an RCA/retro on this work, at some point, to understand where the hurdles were.

edx/analytics/tasks/tests/acceptance/test_lms_courseware_link_clicked.py
sanfordstudent, line 68: "Square" here means the data frames have the same dimensions, or at least that's how I took it based on the comments on other acceptance tests.

sanfordstudent:
@nasthagiri thanks. Sounds good. We can schedule one after this gets merged.

edx/analytics/tasks/lms_courseware_link_clicked.py
mulby, line 97: I think that would help clarify this and is worth the extra verbosity.

edx/analytics/tasks/lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,140 @@
...
+
+        yield (course_id, datestamp, external_click_count, click_count)
+
+    def output(self):
+        return get_target_from_url(self.output_root)
```
mulby, line 105: If `self.output_root` doesn't end with a "/" we should either bail out with a clear error message that the output needs to be a directory, or automatically add one. This is likely the issue you encountered when the acceptance test was trying to read the directory like a file.

mulby, line 105: This is an area of tech debt btw.

edx/analytics/tasks/lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def complete(self):
+        return get_target_from_url(url_path_join(self.output_root, '_SUCCESS')).exists()
```
mulby, line 108: More debt here, we could move this logic to base class or mixin. I wouldn't do it now, mostly just making a note to myself for the retro.

edx/analytics/tasks/tests/acceptance/test_lms_courseware_link_clicked.py
mulby: In general we try to stick to the edX conventions for formatting, but are not draconian about it. If the code is readable and reasonably consistent with the rest of the repo, that's the important part.

edx/analytics/tasks/tests/acceptance/test_lms_courseware_link_clicked.py
mulby, line 39: It's a synchronous subprocess. This would be problematic if something had an infinite loop. We could add in a timeout to the task running service, but we haven't bothered yet, since it hasn't been an issue for us yet.

edx/analytics/tasks/tests/test_lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,122 @@
...
+        line = 'this is garbage but contains {}'.format(LINK_CLICKED)
+        self.assert_no_map_output_for(line)
+
+    def test_link_clicked_event_count_per_course(self):
+        line = self.create_event_log_line()
+        self.assert_single_map_output(line, (self.course_id, self.datestamp), 1)
+
+    def test_internal_link_clicked_event_count(self):
+        line = self.create_event_log_line(event={
+                                          "current_url": "http://courses.edx.org/blah",
```
mulby: shouldn't these be "courses.example.com"?

edx/analytics/tasks/tests/acceptance/fixtures/input/lms_courseware_link_clicked_acceptance_tracking.log
mulby, line 20: It might make more sense to add these to the unit tests? We try to put as much coverage as possible in the unit tests and use the acceptance tests for validating the happy path of data hand-offs between tasks.

sanfordstudent, line 20: I believe most of these cases are already covered by unit tests.

mulby:
@sanfordstudent so far so good, I called out a couple of things, but this is rapidly approaching mergability.

I would love to get your feedback on the new developer experience. I know there are a lot of rough spots, it would be great to have both the comprehensive list as well as a rough prioritization from your perspective. We have put very little effort into improving the new developer experience so this area needs a lot of work. We can then prioritize it along with our other tech debt.

edx/analytics/tasks/tests/acceptance/fixtures/input/lms_courseware_link_clicked_acceptance_tracking.log
nasthagiri, line 20: great.  can you double check?

edx/analytics/tasks/tests/test_lms_courseware_link_clicked.py
sanfordstudent: Theoretically it doesn't matter what the URL is, but for consistency, yes.

edx/analytics/tasks/tests/acceptance/__init__.py
```diff
@@ -147,8 +147,8 @@ def setUp(self):
         self.identifier = m.hexdigest()
         self.test_root = url_path_join(url, self.identifier, self.__class__.__name__)
 
-        self.test_src = url_path_join(self.test_root, 'src')
-        self.test_out = url_path_join(self.test_root, 'out')
+        self.test_src = url_path_join(self.test_root, 'src/')
+        self.test_out = url_path_join(self.test_root, 'out/')
```
sanfordstudent: @mulby I wanted to bring this specific change to your attention. Do you think that this is the right approach for preventing the problem I saw where it was trying to read from a directory? This change did solve my problem.

edx/analytics/tasks/lms_courseware_link_clicked.py
sanfordstudent, line 105: @mulby Do you think that's something that makes sense to do in an **init** function for this task, or to add to the general mapreduce task?

mulby, line 105: I think doing it in the init or directly in this function probably makes sense. Not all mapreduce tasks define an output_root parameter, so I would avoid doing it in the base class.

edx/analytics/tasks/lms_courseware_link_clicked.py
```diff
@@ -0,0 +1,145 @@
...
+            ('course_id', 'VARCHAR(255)'),
+            ('event_date', 'DATE'),
+            ('external_link_clicked_events', 'INT'),
+            ('link_clicked_events', 'INT')
+        ]
+
+    @property
+    def insert_source_task(self):
+        # the output_root must end with a forward slash to indicate to subsequent jobs that it is a directory
+        if not self.output_root.endswith("/"):
```
sanfordstudent: @mulby I decided to move the slash handling here. Since the rest of the acceptance tests worked without my change to the acceptance task's `__init__` function, it felt inappropriate to make my original change there. In this file, I think it makes sense to put it here because this is the entry point by which the task gets run in practice, and it's easiest to update the `output_root` in this one place.

edx/analytics/tasks/lms_courseware_link_clicked.py
```diff
@@ -123,6 +125,9 @@ def columns(self):
 
     @property
     def insert_source_task(self):
+        # the output_root must end with a forward slash to indicate to subsequent jobs that it is a directory
+        if not self.output_root.endswith("/"):
+            self.output_root += "/"
```
mulby: I think we should move this logic to the `output()` method of the LMSCoursewareLinkClickedTask since it's necessary for any other task to consume its output.

sanfordstudent: @mulby updated.

mulby:
@sanfordstudent I kicked off our jenkins tests on this PR since it's pretty much ready to merge, once those are all green we can merge it.

Note that we use a separate jenkins box from edx-platform for running our tests.

mulby:
Looking at your diff cover report there are some missing test cases around malformed events (invalid course ID etc).

You can run `make coverage` and `make coverage-local` to generate the report locally (the latter doesn't install any new python packages and runs much quicker).

mulby:
:+1:

nasthagiri:
Before I give a thumbs, are the acceptance tests supposed to fail?

mulby:
@nasthagiri - heh, definitely not, however, we had a failing test on master that I suspect this branch is based off of. Given that this PR doesn't touch any of the code related to that test, I'm OK merging it as is. We could also rebase it onto the tip of master and re-run the tests, but given the low number of contributors to this repo, I suspect that's more trouble than it's worth.

nasthagiri:
Thanks @mulby  

👍

PR #265, opened 2016-06-21
Pentaho schema management

HassanJaveed84:
@mulby @brianhw please review. 
Not sure if using CredentialFileVerticaTarget for the schema management task is the right way to go about it.


edx/analytics/tasks/load_warehouse.py
```diff
@@ -11,12 +11,93 @@
...
+            log.exception("Rolled back the transaction; exception raised: %s", str(exc))
+            connection.rollback()
+            raise
+        finally:
+            connection.close()
+
+    def output(self):
+        return CredentialFileVerticaTarget(
+            credentials_target=self.input()['credentials'],
+            table='',
```
HassanJaveed84: Not sure what to provide here, this task is not loading a particular table. See comment above about the issue this causes.

edx/analytics/tasks/load_warehouse.py
```diff
@@ -11,12 +11,93 @@
...
 
 from edx.analytics.tasks.util.hive import WarehouseMixin
+from edx.analytics.tasks.url import ExternalURL
 
 log = logging.getLogger(__name__)
 
 
+class SchemaManagementTask(VerticaCopyTaskMixin, luigi.Task):
+
+    completed = False
```
HassanJaveed84: We want to run this task every time as part of the workflow. The typical procedure I see is to mark the task as complete by using touch(which creates an entry in the marker_table). If we do that, we also have to remove the output on overwrite(which has it's own problem as the row in the marker table is stored without a proper 'target_table') and call this task with overwrite=True, otherwise this won't run if the workflow is triggered again.

edx/analytics/tasks/load_warehouse.py
```diff
@@ -36,49 +117,54 @@ class LoadWarehouse(WarehouseMixin, luigi.WrapperTask):
 
     overwrite = luigi.BooleanParameter(default=False)
 
+    marker_schema = luigi.Parameter()
+
     def requires(self):
         kwargs = {
-            'schema': self.schema,
+            'schema': self.schema + '_loading',
+            'marker_schema': self.marker_schema,
```
HassanJaveed84, line 175: Luigi checks for task completeness all the time. When checking for completeness it recursively checks for all dependencies. 
When a warehouse load task completes, we rename the schema name and luigi while checking for completeness complains as the tasks' outputs seems to have disappeared(it's now under a different schema). 
Using a separate schema for the marker_table solves this issue, though I'm not sure if that's the best way to solve this particular problem.

edx/analytics/tasks/util/vertica_target.py
```diff
@@ -130,3 +135,7 @@ def create_marker_table(self):
             else:
                 raise
         connection.close()
+
+    def create_marker_schema(self, connection):
```
HassanJaveed84, line 140: The marker schema might not exist, it can different than the actual schema.

edx/analytics/tasks/vertica_load.py
```diff
@@ -46,6 +46,7 @@ class VerticaCopyTaskMixin(OverwriteOutputMixin):
     read_timeout = luigi.IntParameter(
         config_path={'section': 'vertica-export', 'name': 'read_timeout'}
     )
+    marker_schema = luigi.Parameter(default=None)
```
HassanJaveed84: The marker schema would default to 'schema' if it is set to None.

edx/analytics/tasks/load_warehouse.py
mulby: What if we use the date of the load as a unique identifier for this particular load and rename operation?

We could then tie all of our re-run and overwrite semantics to that date. For example if we re-run the task after it succeeded on the same date, then it would do nothing. If we re-run it with overwrite for the same date, then it would re-run the entire process.

Not sure exactly how this would play out... just an idea.

edx/analytics/tasks/load_warehouse.py
mulby, line 175: I'm OK with this, but we should write a bunch of comments about this in the code since it's not how we handle things elsewhere.

edx/analytics/tasks/load_warehouse.py
```diff
@@ -36,14 +42,112 @@ class LoadWarehouse(WarehouseMixin, luigi.WrapperTask):
...
+    @property
+    def table(self):
+        raise NotImplementedError
+
+    def requires(self):
+        return {
+            'credentials': ExternalURL(self.credentials)
+        }
+
+    def init_touch(self, connection):
```
HassanJaveed84, line 88: This method is exactly the same as defined in vertica_load.py without `connection.cursor().flush_to_query_ready()`

edx/analytics/tasks/load_warehouse.py
```diff
@@ -36,14 +42,112 @@ class LoadWarehouse(WarehouseMixin, luigi.WrapperTask):
...
+    def queries(self):
+        return [
+            "DROP SCHEMA IF EXISTS {schema} CASCADE;".format(schema=self.schema_last),
+            "DROP SCHEMA IF EXISTS {schema} CASCADE;".format(schema=self.schema_loading),
+            "CREATE SCHEMA IF NOT EXISTS {schema}".format(schema=self.schema_loading),
+        ]
+
+    @property
+    def table(self):
+        return 'pre_load_' + self.date.strftime('%Y-%m-%d')
```
HassanJaveed84, line 166: We need to separate the marker table entries for the Pre and post load schema management tasks. The date takes care of the overwriting and existence check behavior. I'm open to suggestion on how to make this logic better.

edx/analytics/tasks/load_warehouse.py
```diff
@@ -64,21 +168,67 @@ def requires(self):
...
         )
 
-    def output(self):
-        return [task.output() for task in self.requires()]
+
+class PostLoadWarehouseTask(SchemaManagementTask):
+
+    n_reduce_tasks = luigi.Parameter()
+
+    priority = -100
```
HassanJaveed84, line 231: I don't see a way to build the dependency chain without using priority. 
If we had dynamic dependencies, we could have used that.

edx/analytics/tasks/load_warehouse.py
```diff
@@ -64,21 +168,67 @@ def requires(self):
...
+            "GRANT USAGE ON SCHEMA {schema} TO analyst;".format(schema=self.schema),
+            "GRANT SELECT ON ALL TABLES IN SCHEMA {schema} TO analyst;".format(schema=self.schema),
+        ]
+
+    @property
+    def table(self):
+        return 'post_load_' + self.date.strftime('%Y-%m-%d')
+
+    def run(self):
+        connection = self.output().connect()
```
HassanJaveed84, line 263: The validation can be moved to another task. 
But I'm curious about the output it would produce if any, we can have a marker table entry or custom logic around complete method for that.

HassanJaveed84:
@mulby please review when you get a chance. I'm currently running the workflow against production data(loading into pipeline_etl schema). However, there are still two dependencies missing:
1. 'internal_reporting_user_activity' needs to be constructed(I thought we would be using UserActivityWorkflow instead of CourseActivityWeeklyTask in the Jenkins job to take care of that; currently scheduled job uses CourseActivityWeeklyTask).
2. 'internal_reporting_user' relies on a change from my previous PR to be released first as I removed 'user_country_output' parameter(it placed the output in a temp location; going forward last_country_of_user data would be placed in warehouse so it can be used to build 'internal_reporting_user').

I've left some comments on the parts I'm not particularly happy about.

edx/analytics/tasks/load_warehouse.py
mulby, line 88: It's probably worth putting in a comment in here that states this fact and explains why we don't want to call flush_to_query_ready in this case

edx/analytics/tasks/load_warehouse.py
```diff
@@ -36,14 +42,112 @@ class LoadWarehouse(WarehouseMixin, luigi.WrapperTask):
...
+
+        try:
+            for query in self.queries:
+                log.debug(query)
+                connection.cursor().execute(query)
+
+            self.attempted_removal = True
+            self.init_touch(connection)
```
mulby: Don't we want to do this before running the query?

edx/analytics/tasks/load_warehouse.py
mulby, line 166: We also probably want to clarify here why this string works here, even though it's not a table.

edx/analytics/tasks/load_warehouse.py
mulby, line 231: OK! Let's stick with priority for now, then.

edx/analytics/tasks/load_warehouse.py
HassanJaveed84: I followed the approach used in vertica_load, it calls init_touch after copying data to the table. 
I'll move this up, before running the queries.

edx/analytics/tasks/load_warehouse.py
mulby, line 263: I trust your judgement on this point. I think I like the idea of a separate task with a marker, but I kind of like the simplicity of the method you use here.

HassanJaveed84:
@mulby this is ready for a final review.

edx/analytics/tasks/load_warehouse.py
```diff
@@ -80,5 +221,84 @@ def requires(self):
...
+                schema_loading=self.schema_loading,
+                table=table
+            )
+            log.debug(query)
+            cursor.execute(query)
+            row = cursor.fetchone()
+            if row is None:
+                log.error('Failed to validate table: %s', table)
+                connection.close()
+                return
```
mulby: Should this raise an exception?

edx/analytics/tasks/load_warehouse.py
```diff
@@ -36,14 +45,146 @@ class LoadWarehouse(WarehouseMixin, luigi.WrapperTask):
...
+
+    @property
+    def queries(self):
+        """
+        Provides queries that are needed to run this task.
+        """
+        raise NotImplementedError
+
+    @property
+    def table(self):
```
mulby: Perhaps call this something else? Like "marker_name" or "identifier"?

HassanJaveed84: Done.

edx/analytics/tasks/load_warehouse.py
HassanJaveed84: Done.

HassanJaveed84:
@mulby I've made two changes:
- Raises Exception in case the validation fails.
- Renamed 'table' property to 'marker_name'

mulby:
:+1:

PR #266, opened 2016-06-22
Fix d_user.

HassanJaveed84:
@gabe @brianhw

### DIFF
--- a/edx/analytics/tasks/location_per_course.py
+++ b/edx/analytics/tasks/location_per_course.py
@@ -149,9 +149,15 @@ class ExternalLastCountryOfUserToHiveTask(ImportLastCountryOfUserToHiveTask):
     def partition_date(self):
         return self.date.isoformat()  # pylint: disable=no-member
 
+    def partition_spec(self):
+        return "{key}={value}".format(
+            key=self.partition.keys()[0],
+            value=self.partition.values()[0],
+        )
+
     def requires(self):
         yield ExternalURL(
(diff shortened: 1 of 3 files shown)


edx/analytics/tasks/location_per_course.py
```diff
@@ -149,9 +149,15 @@ class ExternalLastCountryOfUserToHiveTask(ImportLastCountryOfUserToHiveTask):
...
+    def partition_spec(self):
+        return "{key}={value}".format(
+            key=self.partition.keys()[0],
+            value=self.partition.values()[0],
+        )
+
     def requires(self):
         yield ExternalURL(
-            url=url_path_join(self.warehouse_path, 'last_country_of_user', self.partition.path_spec) + '/'
+            url=url_path_join(self.warehouse_path, 'last_country_of_user', self.partition_spec()) + '/'
```
mulby, line 160: Why can't we use `path_spec` here?

mulby:
Just one question, mostly LGTM. :+1:

edx/analytics/tasks/location_per_course.py
HassanJaveed84, line 160: ImportIntoHiveTableTask defines partition as a simple dictionary:

``` @property
    def partition(self):
        """Provides name of Hive database table partition."""
        # The Luigi hive code expects partitions to be defined by dictionaries.
        return {'dt': self.partition_date}
```

mulby, line 160: ah... tech debt. OK, let's ship it the way you have it then!

PR #269, opened 2016-07-22
Merge master into eucalyptus

mulby:
@nedbat I think it makes the most sense just to merge master into eucalyptus. We are planning on releasing master in the near future anyway (see tag 0.5.8-rc.1).

FYI @HassanJaveed84 @brianhw


nedbat:
Hmm, rather than merging, we can re-create the eucalyptus branch, which might make more sense.

PR #270, opened 2016-07-26
Bump edx-opaque-keys to 0.3.2 (perf fixes)

ormsbee:
@mulby: Bumping OpaqueKeys to 0.3.2

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -3,7 +3,7 @@ argparse==1.2.1 	# Python Software Foundation License
 cffi==1.1.0             # MIT
 ciso8601==1.0.1         # MIT
 cryptography==0.9       # BSD or Apache 2.0
-edx-opaque-keys==0.2.1  # AGPL
+edx-opaque-keys==0.3.2  # AGPL
 edx-ccx-keys==0.1.2     # AGPL
 elasticsearch==1.7.0    # Apache
 enum34==1.0.4           # BSD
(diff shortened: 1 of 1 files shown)


mulby:
@ormsbee It looks like there might be an issue related to `six`

```
  File "/var/lib/analytics-tasks/4298d15021f8a5c6260538d84df3a6fd/venv/bin/launch-task", line 6, in <module>
    from edx.analytics.tasks.launchers.local import main
  File "/var/lib/analytics-tasks/4298d15021f8a5c6260538d84df3a6fd/venv/local/lib/python2.7/site-packages/edx/analytics/tasks/launchers/local.py", line 18, in <module>
    import opaque_keys
  File "/var/lib/analytics-tasks/4298d15021f8a5c6260538d84df3a6fd/venv/local/lib/python2.7/site-packages/opaque_keys/__init__.py", line 14, in <module>
    from six import (
ImportError: cannot import name python_2_unicode_compatible
```

Given that we peg the version at 1.6.1 [here](https://github.com/edx/edx-analytics-pipeline/blob/master/requirements/default.txt#L32) in the data pipeline and that opaque-keys requires >= 1.10.0 [here](https://github.com/edx/opaque-keys/blob/master/setup.py#L10)

When I ran "make develop" in the pipeline I saw it end up with six 1.6.1 not sure exactly how these conflicts are resolved. Assuming six 1.10.0 is backwards compatible, we may be able to just upgrade that at the same time.

Inside a virtualenv you can test this by running "launch-task --help".

ormsbee:
Added the bump to version for `six` as well.

mulby:
@ormsbee - It looks like the new version of six breaks some notoriously brittle code in luigi core (we've had to patch it before). We may want to transition this task over to the analytics team, so that you don't have to setup a full development environment (which will take a non-trivial amount of time).

ormsbee:
@mulby: Understood. Sorry for the trouble -- I'll leave this in your care.

brianhw:
We've managed to upgrade edx-opaque-keys to 0.4, so this is no longer needed.

PR #271, opened 2016-07-28
Prevent duplicate data insertion as part of the module_engagement workflow

mulby:
## Problem
- We overwrite the last 3 days of granular engagement data in S3 and mysql every night by setting the overwrite parameter to `True`.
- This means that task for 4 days ago will have previously always been run with overwrite set to `True` and now, for the first time, be running with overwrite set to `False`.
- Since overwrite is a significant parameter it is used to generate the hash code stored in the `table_updates` table that indicates whether the task has run successfully or not. When its value changes, the computed hash changes and luigi _thinks_ that it hasn't successfully run.
- At this point luigi schedules the MySQL task for execution with overwrite set to `False`.
- When the task executes, it inserts the data from S3 into the MySQL table. Note that the data was already in the table from the last time the task executed when overwrite was set to `True`.
- This results in duplicate data in the table.
## Fix included in this patch
1. Make the overwrite parameter insignificant. It does not affect the content of the output so it never should have been significant to begin with.
2. Change the `update_id()` function so that only the date determines whether the task should re-run or not.
## Reproduction Steps

Run the module engagement workflow on subsequent days when there is at least 4 days of data available for processing.

Find a course_id that is very active and run this query on your result store, substituting <course_id> with your course_id. If you don't have much data in your system, you can remove the where clause that specifies the course_id.

``` sql
select date, max(num_rows) from (select *, count(*) as num_rows from (select course_id, username, date, entity_type, entity_id, event from module_engagement where course_id='<course_id>') a group by 1, 2, 3, 4, 5, 6 having count(*) > 1) b group by date order by date desc;
```

If the above query does not return any rows, then there is no duplicate data in your system. If it returns any rows than those dates have duplicate data.
## Applying the Patch

The easiest way to apply this patch is to write to a new result store database. You can update the `database` key of the `database-export` section of the config file. You then need to update the API configuration to point at that new database once it is fully populated. You will have to run all tasks again using the updated configuration to populate the new database.

If you want to update your existing database in place, you will need to do execute something like the following procedure.

Using this branch, run a task that looks like:

``` bash
  ModuleEngagementIntervalTask --local-scheduler \
     --interval $(date +%Y-%m-%d -d "$FROM_DATE")-$(date +%Y-%m-%d -d "$TO_DATE") \
     --n-reduce-tasks $NUM_REDUCE_TASKS \
     --overwrite-from-date $(date +%Y-%m-%d -d "$FROM_DATE")
```

Where FROM_DATE is set to the first date you have data for and TO_DATE is the day after the last (usually "today").

This task will overwrite all of your duplicated data with de-duplicated data and insert a new record into `table_updates` with the new update_id format.

Note that this task will require fairly expensive I/O operations on the database since there is no index on the date column. This could affect reads issued by your users if you are executing this on a production system. You can avoid this by truncating the table before running the task above, however, that will cause the charts to be unavailable while the data is reinserted.

Clean out any old records from table_updates:

``` sql
DELETE FROM table_updates WHERE target_table='module_engagement' AND update_id NOT LIKE 'ModuleEngagementMysqlTask%'
```

In the future run the workflow using a version of the code that includes the changes from this branch.
## Reviewers
- [x] @HassanJaveed84 
- [x] @brianhw 
## FYI
- [ ] @renevatium - since I think you are running this workflow?
- [ ] @pomegranited - since you are also touching this code (or will be soon)

### DIFF
--- a/edx/analytics/tasks/module_engagement.py
+++ b/edx/analytics/tasks/module_engagement.py
@@ -65,6 +65,7 @@ class OverwriteFromDateMixin(object):
                     ' in the past up to the end of the interval. Events are not always collected at the time that they'
                     ' are emitted, sometimes much later. By re-processing past days we can gather late events and'
                     ' include them in the computations.',
+        default=None,
         significant=False
     )
 
@@ -270,6 +271,25 @@ class ModuleEngagementMysqlTask(ModuleEngagementDownstreamMixin, IncrementalMysq
         default=False,
         config_path={'section': 'module-engagement', 'name': 'allow_empty_insert'},
     )
+    overwrite_hive = luigi.BooleanParameter(
+        default=False,
+        significant=False
+    )
+
+    def update_id(self):
+        """
+        Use the task name and date to uniquely identify this update.
+
+        This means that other significant parameters will be ignored when testing to see if this task is complete.
+        This is a deliberate choice since we really only care about the date. We should only execute this task if the
+        data for that date has never been loaded or if overwrite is True. Any other changes to parameters should not
+        cause a re-insert.
+
+        Without this, changes to any parameter would result in duplicate records being inserted into the database since
+        the hash code would change for the task, so it would be marked incomplete. Since records are only deleted if
+        overwrite is specified, then it was relatively easy to insert duplicate records into the database.
+        """
+        return '{task_name}(date={key})'.format(task_name=self.task_family, key=self.date.isoformat())
 
     @property
     def table(self):
@@ -286,7 +306,8 @@ class ModuleEngagementMysqlTask(ModuleEngagementDownstreamMixin, IncrementalMysq
     @property
     def indexes(self):
         return [
-            ('course_id', 'username', 'date')
+            ('course_id', 'username', 'date'),
+            ('date',),
         ]
 
     @property
@@ -294,8 +315,8 @@ class ModuleEngagementMysqlTask(ModuleEngagementDownstreamMixin, IncrementalMysq
         partition_task = ModuleEngagementPartitionTask(
             date=self.date,
             n_reduce_tasks=self.n_reduce_tasks,
-            overwrite=self.overwrite,
-            overwrite_from_date=self.overwrite_from_date,
+            warehouse_path=self.warehouse_path,
+            overwrite=self.overwrite_hive,
         )
         return partition_task.data_task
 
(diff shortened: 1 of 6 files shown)


edx/analytics/tasks/module_engagement.py
```diff
@@ -271,6 +271,9 @@ class ModuleEngagementMysqlTask(ModuleEngagementDownstreamMixin, IncrementalMysq
         config_path={'section': 'module-engagement', 'name': 'allow_empty_insert'},
     )
 
+    def update_id(self):
```
brianhw, line 279: This might be a good place for a short comment as to what this is doing and why.

brianhw:
LGTM.  One nit to add a comment in the code around update_id, but otherwise 👍 .

mulby:
@HassanJaveed84 - Can you review this when you get a chance?

edx/analytics/tasks/module_engagement.py
mulby, line 279: Good idea. Adding.

HassanJaveed84, line 279: A question, why do we need this when we have made overwrite an insignificant parameter?
Do we expect other significant parameters to change as well ?

HassanJaveed84:
LGTM👍, just that one question that I left in the comments.

edx/analytics/tasks/module_engagement.py
mulby, line 279: We haven't rigorously audited our significant parameters. I'm not very confident that they are all correctly marked significant or insignificant... for example: if the warehouse_path changed, would we want to write a second copy of the records?

brianhw, line 279: Hmm.  I do think that if I run this with a different value of warehouse_path, I want it to create a separate copy of records.  And particularly if the first load of records has been loaded into stage (using dev), doing this would presumably refuse to load the data into prod.  I don't like that idea.   We should at least be treating environment-related stuff as significant, I think, including source and destination.

mulby, line 279: I can't imagine any scenario where we would want to write data for the same date to the result store multiple times. First it should be noted that the `table_updates` table is specific to one result store (and thus, one environment-deployment combo) so the record of the write will only be in the prod-edx (for example) result store. If I then attempted to write data for that same date from the stage-edx warehouse path, I wouldn't want that operation to succeed! Ideally it would raise an error without actually doing anything, but I could see an argument being made around just NOP.

When would we want to insert data for a date _twice_ without overwrite? It seems to me like the only two scenarios that should happen are:
1. We insert data into a date that has never been loaded before.
2. We overwrite data in a date that has been loaded before. In this case we can change whatever parameters we want to load data from wherever etc.

We should never do:
1. We insert data into a date that has been loaded before.

brianhw, line 279: When I run tests now that involves changing a current workflow, I want to be able to run in stage with a different warehouse-path, specific to a change I'm making (particularly a schema-altering change), and I also want to run with a different database name.  That means that the table_updates table is also in that different database.  

Note that all of this is specific to the logic of how a Mysql-loading job determines whether it is complete or not.  And I'm now comfortable with this change being for that specific purpose.  👍

PR #276, opened 2016-08-09
Update user activity workflow.

HassanJaveed84:
@gabe @brianhw I had changed some parameters for the task, and hadn't gotten this work merged, breaking the user-activity job.

### DIFF
--- a/edx/analytics/tasks/load_internal_reporting_user_activity.py
+++ b/edx/analytics/tasks/load_internal_reporting_user_activity.py
@@ -13,6 +13,7 @@ from edx.analytics.tasks.user_activity import UserActivityTableTask
 from edx.analytics.tasks.vertica_load import VerticaCopyTask, VerticaCopyTaskMixin, CredentialFileVerticaTarget
 from edx.analytics.tasks.database_imports import ImportAuthUserTask
 from edx.analytics.tasks.util.hive import HiveTableFromQueryTask, WarehouseMixin, HivePartition
+from edx.analytics.tasks.util.weekly_interval import WeeklyIntervalMixin
 from edx.analytics.tasks.user_activity import CourseActivityWeeklyTask
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


edx/analytics/tasks/load_internal_reporting_user_activity.py
```diff
@@ -229,15 +232,32 @@ class UserActivityWorkflow(luigi.WrapperTask):
     )
     weeks = luigi.IntParameter(default=24)
     n_reduce_tasks = luigi.Parameter()
-    interval = luigi.DateIntervalParameter()
-    pipeline_credentials = luigi.Parameter()
+    credentials = luigi.Parameter()
+
+    @property
+    def interval(self):
```
HassanJaveed84: This part is duplicate of what appears at https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/user_activity.py#L231. We can remove it from there and make the CourseActivityWeeklyTask accept an interval parameter instead if it makes sense. 
Though that would makes things different from CourseActivityMonthlyTask, which does something similar for calculating month.

brianhw:
This may be out of scope for this PR, but I'm wondering if the logic to convert from end_date and week to an interval should be pulled into a mixin in a util.  I don't see a reasonable existing util file -- datetime_util doesn't look right.  But we have at least two places that want to do the same thing, and I can imagine it being used elsewhere.   The only issue I see is when there are competing definitions of interval, making sure that this one overrides the other.

edx/analytics/tasks/load_internal_reporting_user_activity.py
mulby: Or maybe just move this logic to a utility function? I prefer composition over inheritance for this type of logic sharing.

mulby: Brian and I just discussed this, I missed the fact that there are both shared parameters and shared logic. If that's the case, a mixin is probably the way to go.

brianhw:
Brief discussion with Gabe:  he proposed a utility function to do the conversion to an interval, given a number of weeks and end_date.  The definition of the latter parameters and the definition of the interval property would still exist in each class that used it.  

I was proposing that a Mixin would also include the latter two parameters.  But I wasn't sure about the interval property, because that would often have to override an existing parameter in order to get EventSelection to work.   But it could use a utility function as well, provided by the Mixin or directly.  Just a thought.  We could just start with the utility function for now.

mulby:
@HassanJaveed84 don't we need to merge this as well?

HassanJaveed84:
@gabe I refactored some code, please review. Not quite sure where to best place the mixin and if it named appropriately enough.

mulby:
I like the refactor

:+1:

PR #277, opened 2016-08-09
Updated analyze log script for EMR 4x.

HassanJaveed84:
@mulby @brianhw

### DIFF
--- a/edx/analytics/tasks/launchers/analyze/main.py
+++ b/edx/analytics/tasks/launchers/analyze/main.py
@@ -181,7 +181,7 @@ def analyze_overall_execution(parser):
 
 
 def analyze_hadoop_job(starting_message, parser):
-    match = re.match(r'.*?(?P<job_id>job_\d{12}_\d{4})', starting_message.content)
+    match = re.match(r'.*?(?P<job_id>job_\d{13}_\d{4})', starting_message.content)
     job_id = match.group('job_id')
     start_timestamp = starting_message.timestamp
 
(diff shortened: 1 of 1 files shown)


edx/analytics/tasks/launchers/analyze/main.py
```diff
@@ -181,15 +181,15 @@ def analyze_overall_execution(parser):
...
+    match = re.match(r'.*?(?P<job_id>job_\d{13}_\d{4})', starting_message.content)
     job_id = match.group('job_id')
     start_timestamp = starting_message.timestamp
 
     message = starting_message
     while message:
         message = parser.next_message()
 
-        if 'Job complete:' in message.content or 'Ended Job = ' in message.content:
+        if 'completed successfully' in message.content or 'Ended Job = ' in message.content:
```
brianhw, line 192: If 'Job complete:' here got changed to 'completed successfully', should that also be the case on the next line?

brianhw:
General question about backwards compatibility:   do we want this to also work on EMR 2.x output?  If so, one could change the 13 to 12 plus optional 1, and change the "if" to be "if old or new or ...".  That way, one could run this on before/after cases to make comparison more direct.

Also, I found I had to change HiveTableFromQueryTask to HiveTableFromParameterQueryTask.  That should probably be an 'or', in fact.  But just curious if that had any impact on the output on the new jobs.

HassanJaveed84:
I'll make those changes. 

Regarding the impact on the output, it seems to me that task name is only used for constructing parameters.

PR #279, opened 2016-08-18
Make warehouse loading task in sync with pentaho.

HassanJaveed84:

Note that I implemented some custom logic to get the latest partition from internal_reporting_user_activity table. It is similar to what we do in DataObfuscationTask currently.
I'm not sure how to DRY this, I still don't have access to edx-analytics-etl-pentaho repo(Brian mentioned that I could follow the approach for looking up most recent copy of the table from /util/Load_Vertica_From_S3.kjb)

### DIFF
--- a/edx/analytics/tasks/load_internal_reporting_user_course.py
+++ b/edx/analytics/tasks/load_internal_reporting_user_course.py
@@ -5,7 +5,8 @@ import luigi
 import logging
 
 from edx.analytics.tasks.vertica_load import (
-    VerticaCopyTask, VerticaProjection, PROJECTION_TYPE_NORMAL, PROJECTION_TYPE_AGGREGATE
+    VerticaCopyTask, VerticaProjection, PROJECTION_TYPE_NORMAL, PROJECTION_TYPE_AGGREGATE,
+    VerticaCopyTaskMixin, CredentialFileVerticaTarget
 )
 from edx.analytics.tasks.util.hive import WarehouseMixin, HivePartition
 from edx.analytics.tasks.url import url_path_join, ExternalURL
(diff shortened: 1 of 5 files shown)


edx/analytics/tasks/load_internal_reporting_user_course.py
```diff
@@ -172,3 +173,43 @@ def projections(self):
...
+
+    def run(self):
+        connection = self.output().connect()
+        try:
+            cursor = connection.cursor()
+
+            # We mark this task as complete first, since the view creation does an implicit commit.
+            self.output().touch(connection)
+
+            build_view_query = "CREATE VIEW {schema}.{view} AS SELECT * FROM {persistent_schema}.f_user_course".format(
```
mulby, line 198: I don't think this query is actually run? Don't we need something like:

``` python
cursor.execute(build_view_query)
```

HassanJaveed84, line 198: Accidentally removed some code during cleanup, thanks!

mulby:
:+1: ready for squash and merge

mulby:
@brianhw can you review this?

edx/analytics/tasks/load_internal_reporting_user_course.py
```diff
@@ -172,3 +173,45 @@ def projections(self):
  GROUP BY 1, 2, 3, 4;"""
             ),
         ]
+
+class BuildUserCourseView(VerticaCopyTaskMixin, WarehouseMixin, luigi.Task):
+
+    persistent_schema = luigi.Parameter(default='production_persistent')
```
brianhw: This doesn't seem like we can override it from load_warehouse.  So this is always going to point to production_persistent.f_user_course.  Should this be pulled out into a mixin, so that it can be set in LoadWarehouseTask?

brianhw:
Just the mixin?  Is it needed to release, or are we okay with hardcoding of production_persistent?

brianhw:
Otherwise LGTM. 👍

brianhw:
Ah, okay, so the commenting out of the other tasks is temporary, not part of the final change. Otherwise looks good.

HassanJaveed84:
@mulby @brianhw 

BuildUserCourseView now uses VerticaCopyTask as the base class. We want the view to be always built if overwrite is specified, previously BuildUserCourseView task ran only once even with overwrite.

I also made a small change in vertica_load, I believe it was not doing the correct thing before.

Should I pull 'persistent_schema' in to a mixin or follow some other approach(define it in two places separately?)

mulby:
Maybe make it a config parameter and include it in the same mixins as the normal vertica schema?

HassanJaveed84:
@mulby should I merge this in ?

edx/analytics/tasks/vertica_load.py
```diff
@@ -52,6 +52,9 @@ class VerticaCopyTaskMixin(OverwriteOutputMixin):
         'default to the schema value if the value here is None.'
     )
 
+    persistent_schema = luigi.Parameter(
+        config_path={'section': 'vertica-export', 'name': 'persistent_schema'}
```
mulby, line 57: I would put in a default value here.

mulby:
One tiny nit, then this is ready to merge. :+1:

PR #281, opened 2016-08-19
implement adyen payment reporting

zubair-arbi:
SOL-1880
@hasnain-naveed @ibrahimahmed443 @douglashall 
- Implement `Adyen` daily payment reporting
- Add tests config section for `Adyen`
- Add unit tests for `Adyen` reporting
- Update payment acceptance tests for `Adyen` reports
### How to run this locally
- Start your analytics devstack http://edx.readthedocs.io/projects/edx-installing-configuring-and-running/en/latest/installation/analytics/start_analytics.html
- Add settings for `adyen` to your `edx-analytics-pipeline/config/devstack.cfg`

```
[adyen]
company_account = TestAccount
merchant_accounts_list = EdXORG
username = report
password = ************
url = https://ca-test.adyen.com/reports/download
interval_start = 2016-08-19
```
- Run the remote command against your `analytics devstack` from your `edx-analytics-pipeline` for a specific day for example to get report for `2016-08-17` run following command:

``` shell
remote-task --vagrant-path ~/Workspace/analytics-edx/analyticstack/ --remote-name devstack --override-config ${PWD}/config/devstack.cfg --wheel-url $WHEEL_URL --branch origin/zub/sol-1880-implement-adyen-transactions-reporting --wait DailyProcessFromAdyenTask --company-account EdXAccount --merchant-account EdXORG --output-root hdfs://localhost:9000/edx-analytics-pipeline/output/adyen/ --run-date=2016-08-17 --local-scheduler
```
- SSH your `analytics` devstack
- Switch to `hadoop` user

```
sudo su hadoop
```
- Verify that there is respective report available by running this command:

```
hadoop@precise64:/home/vagrant$ hadoop fs -ls /edx-analytics-pipeline/output/adyen/
Found 1 items
drwxr-xr-x   - hadoop supergroup          0 2016-08-31 10:32 /edx-analytics-pipeline/output/adyen/payments
```


zubair-arbi:
@douglashall @HassanJaveed84 please give initial review for this PR.

zubair-arbi:
@brianhw @mulby please take a look at this PR and guide me how to write tests for it.
Since `adyen` provides `tsv` files so we are using `tsv` directly instead of fetching `csv` files firstand then converting them to `tsv` and example `tsv` file is attached on the related JIRA ticket [SOL-1880](https://openedx.atlassian.net/browse/SOL-1880).

edx/analytics/tasks/reports/adyen.py
```diff
@@ -0,0 +1,160 @@
...
+            base_url=self.url,
+            merchant_account=self.merchant_account,
+            report_name=self.REPORT_NAME,
+            report_date=self.run_date.strftime('%Y_%m_%d'),
+            report_format=self.REPORT_FORMAT
+        )
+        return url
+
+
+class IntervalPullFromAdyenTask(DailyPullFromAdyenTask, WarehouseMixin, luigi.WrapperTask):
```
HassanJaveed84: Should this be PullFromAdyenTaskMixin instead of DailyPullFromAdyenTask ?

edx/analytics/tasks/reports/adyen.py
```diff
@@ -0,0 +1,160 @@
...
+            self.output_root = self.warehouse_path
+        if self.interval is None:
+            self.interval = date_interval.Custom(self.interval_start, self.interval_end)
+
+    def requires(self):
+        """Internal method to actually calculate required tasks once."""
+        args = {
+            'company_account': self.company_account,
+            'merchant_account': self.merchant_account,
+            'output_root': self.warehouse_path,
```
HassanJaveed84: I think you can just pass in self.output_root 
`'output_root': self.output_root,`
here as it would default to the warehouse_path if None.

mulby:
@zubair-arbi - First, thanks for building this! We are definitely on the right track here. Here's my initial round of feedback:
1. You can see examples of similar tests for the paypal report pull here: https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/reports/tests/test_paypal.py#L534
2. You will also need to update our acceptance tests to include some fake data here: https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/tests/acceptance/test_financial_reports.py#L30
3. As you develop this, please be _very_ careful not to accidentally upload the password for this account. I recommend copying the devstack.cfg file somewhere outside of the repo say "cp config/devstack.cfg /var/tmp/adyen.cfg", modifying it to include the real password and then running with `LUIGI_CONFIG_PATH=/var/tmp/adyen.cfg launch-task ...`
4. Can you point me to a spec for the format of the TSV that we download from adyen? If it contains certain classes of protected information about the user then we will not want to persist the file to S3, and instead just extract the information we need and save that result. We need the final output to look like this:  https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/reports/cybersource.py#L150.

zubair-arbi:
4) @mulby I have attached a `tsv` file `received_payments_report_2016_08_17.tsv` is on the related [JIRA ticket SOL-1880](https://openedx.atlassian.net/browse/SOL-1880). You can download it and check the fields.

edx/analytics/tasks/reports/adyen.py
zubair-arbi: Yes. Let me change it.

mulby:
@zubair-arbi awesome, thanks for the sample. I've responded on the JIRA ticket.

zubair-arbi:
@mulby @brianhw I have updated this PR to read/use selected fields from actual `adyen` payments report by using the column names. Please review.

zubair-arbi:
@mulby @brianhw @douglashall I have updated this PR by adding a separate task for processing the daily pulled adyen reports and verified that the new task `DailyProcessFromAdyenTask` works locally. Please review.

edx/analytics/tasks/reports/adyen.py
```diff
@@ -0,0 +1,253 @@
...
+    )
+
+    def __init__(self, *args, **kwargs):
+        super(PullFromAdyenTaskMixin, self).__init__(*args, **kwargs)
+
+        config = get_config()
+        section_name = 'adyen'
+        self.username = config.get(section_name, 'username')
+        self.password = config.get(section_name, 'password')
+        self.url = config.get(section_name, 'url')
```
brianhw: I recognize that interval_start is needed in a different class than username, password, and url, but I would move all four out of this mixin and into their respective classes.   I think this mixin should just define the common luigi parameters that need to be passed through.

edx/analytics/tasks/reports/adyen.py
```diff
@@ -0,0 +1,253 @@
...
+        self.interval_start = luigi.DateParameter().parse(config.get(section_name, 'interval_start'))
+
+
+class DailyPullFromAdyenTask(PullFromAdyenTaskMixin, luigi.Task):
+    """ A task that reads out adyen reports "Payment Accounting Report" from a
+    remote Adyen account with a report user and generates a CSV file.
+
+    Subscribe Adyen daily reports "Payment Accounting Report" with CSV format.
+    https://docs.adyen.com/developers/reporting-manual#reportdownload
```
brianhw, line 57: Is there a more specific link that points directly to the payment accounting report?

edx/analytics/tasks/reports/adyen.py
```diff
@@ -0,0 +1,253 @@
...
+        config = get_config()
+        section_name = 'adyen'
+        self.username = config.get(section_name, 'username')
+        self.password = config.get(section_name, 'password')
+        self.url = config.get(section_name, 'url')
+        self.interval_start = luigi.DateParameter().parse(config.get(section_name, 'interval_start'))
+
+
+class DailyPullFromAdyenTask(PullFromAdyenTaskMixin, luigi.Task):
+    """ A task that reads out adyen reports "Payment Accounting Report" from a
```
brianhw: nit: capitalize Adyen

edx/analytics/tasks/reports/adyen.py
```diff
@@ -0,0 +1,253 @@
...
+    remote Adyen account with a report user and generates a CSV file.
+
+    Subscribe Adyen daily reports "Payment Accounting Report" with CSV format.
+    https://docs.adyen.com/developers/reporting-manual#reportdownload
+
+    Pulls are made for only a single day. It allows runs to performed
```
brianhw: nit:  "to performed" => "to be performed".

edx/analytics/tasks/reports/adyen.py
```diff
@@ -0,0 +1,253 @@
...
+
+        # Fetch the report
+        auth = (
+            '{report_username}%Company.{company_account}'.format(
+                report_username=self.username,
+                company_account=self.company_account
+            ), self.password
+        )
+        print "report_url =======> ", self.report_url
+        print "auth =======> ", auth
```
brianhw: I assume in production we will want the print statements to be removed.

edx/analytics/tasks/reports/adyen.py
```diff
@@ -0,0 +1,253 @@
...
+        with self.input().open('r') as input_file:
+            # DictReader class skips the header row and uses it for named indexing.
+            csv_reader = csv.DictReader(input_file, delimiter=',')
+
+            with self.output().open('w') as output_file:
+                for row in csv_reader:
+                    # Output most of the fields from the original source file.
+                    # https://docs.adyen.com/developers/reporting-manual#paymentaccountingreportstructure
+
+                    # Read only settled (sale) or refunded (refund) transaction records
```
brianhw: nit: add trailing period.

edx/analytics/tasks/reports/adyen.py
```diff
@@ -0,0 +1,253 @@
...
+        description='URL of location to write output.',
+    )
+
+    def __init__(self, *args, **kwargs):
+        super(IntervalPullFromAdyenTask, self).__init__(*args, **kwargs)
+        # Provide default for output_root at this level.
+        if self.output_root is None:
+            self.output_root = self.warehouse_path
+        if self.interval is None:
+            self.interval = date_interval.Custom(self.interval_start, self.interval_end)
```
brianhw, line 290: Since interval_start is only needed here, I think this is where it should be read from configuration.

edx/analytics/tasks/reports/adyen.py
zubair-arbi: @brianhw since this class `DailyPullFromAdyenTask` is related to just fetching the subscribed reports so this [report download manual url](https://docs.adyen.com/developers/reporting-manual#reportdownload) from the `reporting-manual` makes more sense here which describes how you can subscribe and automate the report downloading.

For the report structure of `Payment Accounting Report` I have added the [paymentaccountingreportstructure url](https://docs.adyen.com/developers/reporting-manual#paymentaccountingreportstructure) in the method `run` of class `DailyProcessFromAdyenTask` where we process the downloaded report.

edx/analytics/tasks/reports/adyen.py
zubair-arbi: This is just for local testing purpose and will be removed with our finalized commit.

edx/analytics/tasks/reports/adyen.py
zubair-arbi: Done.

edx/analytics/tasks/reports/adyen.py
```diff
@@ -0,0 +1,261 @@
...
+    )
+    output_root = luigi.Parameter(
+        description='URL of location to write output.',
+    )
+
+    def __init__(self, *args, **kwargs):
+        super(PullFromAdyenTaskMixin, self).__init__(*args, **kwargs)
+
+        self.config = get_config()
+        self.config_section_name = 'adyen'
```
mulby: In the cybersource task we made this dynamic because we needed to support different credentials for different accounts, so we needed to have multiple sets of credentials stored in the config file. So it uses section names like "cybersource:account_name".

It looks like, in this case, we don't need account-specific credentials so you could probably just use the standard "default_from_config" kwarg when defining the relevant parameters.

edx/analytics/tasks/reports/adyen.py
```diff
@@ -0,0 +1,261 @@
...
+    Subscribe Adyen daily reports "Payment Accounting Report" with CSV format.
+    https://docs.adyen.com/developers/reporting-manual#reportdownload
+
+    Pulls are made for only a single day. It allows runs to be performed
+    incrementally on a daily tempo.
+
+    """
+    # Date to fetch Adyen report.
+    run_date = luigi.DateParameter(
+        default=date.today(),
```
mulby: This should be UTC today: `datetime.utcnow().date()`

edx/analytics/tasks/reports/adyen.py
```diff
@@ -0,0 +1,261 @@
...
+        description='Default is today.',
+    )
+
+    REPORT_NAME = 'payments_accounting_report'
+    REPORT_FORMAT = 'csv'
+
+    def __init__(self, *args, **kwargs):
+        super(DailyPullFromAdyenTask, self).__init__(*args, **kwargs)
+
+        self.username = self.config.get(self.config_section_name, 'username')
```
mulby: As noted above, these can be made normal parameters unless they are sensitive (like the password). The parameters are printed to the log, so I think it makes sense to read the password directly from the config file, but the others could just be normal parameters.

edx/analytics/tasks/reports/adyen.py
```diff
@@ -0,0 +1,261 @@
...
+        self.password = self.config.get(self.config_section_name, 'password')
+        self.url = self.config.get(self.config_section_name, 'url')
+
+    def requires(self):
```
mulby: You should be able to omit this.

edx/analytics/tasks/reports/adyen.py
```diff
@@ -0,0 +1,261 @@
...
+        auth = (
+            '{report_username}%Company.{company_account}'.format(
+                report_username=self.username,
+                company_account=self.company_account
+            ), self.password
+        )
+        print "report_url =======> ", self.report_url
+        print "auth =======> ", auth
+
+        response = requests.get(self.report_url, auth=auth)
```
mulby: We should probably wrap this in some retry logic. I recently added a retry decorator that support exponential backoff:

https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/util/retry.py

You can see a usage here:

https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/util/edx_api_client.py#L88

Not sure what responses they can give us that are "fatal" vs. "retryable", but we should distinguish between them so that the job doesn't fail if the service is temporarily unavailable.

edx/analytics/tasks/reports/adyen.py
```diff
@@ -0,0 +1,261 @@
...
+                    if transaction_type not in ['Refunded', 'Settled']:
+                        continue
+
+                    booking_date = datetime.strptime(
+                        row['Booking Date'], '%Y-%m-%d %H:%M:%S'
+                    ).replace(tzinfo=timezone(row['TimeZone']))
+
+                    result = [
+                        # date (ISO8601 formatted transaction creation date)
+                        booking_date.date().isoformat(),
```
mulby, line 225: We should convert this to UTC before outputting it.

edx/analytics/tasks/reports/adyen.py
```diff
@@ -0,0 +1,261 @@
...
+        return get_target_from_url(url_with_filename)
+
+    @property
+    def report_url(self):
+        """Generate the url to download a report from a Adyen account.
+
+        Example URL:
+            https://ca-test.adyen.com/reports/download/MerchantAccount/EdXORG/payments_accounting_report_2016_08_17.csv
+        """
+        url = '{base_url}/MerchantAccount/{merchant_account}/{report_name}_{report_date}.{report_format}'.format(
```
mulby, line 166: I'm hoping/assuming that this report_date is UTC?

requirements/pre.txt
```diff
@@ -1 +1 @@
-pip==1.5.6             # MIT // AN-4322
+pip==8.1.2             # MIT
```
mulby: This change seems unrelated to this PR and kind of risky. Is it needed?

zubair-arbi: Previously I was getting issues in installing packages `cffi` and `cryptography` with errors like:

```
build/temp.macosx-10.11-intel-2.7/_openssl.c:433:10: fatal error: 'openssl/opensslv.h' file not found
#include <openssl/opensslv.h>
         ^
1 error generated.
error: command 'cc' failed with exit status 1
```

and

```
File "/Users/zubairafzal/Workspace/analytics-edx/edx-analytics-pipeline/venv/lib/python2.7/site-packages/cryptography/hazmat/bindings/openssl/binding.py", line 76, in wrapper
    ffi.def_extern(name=name, **kwargs)(func)
AttributeError: 'CompiledFFI' object has no attribute 'def_extern'
```

So I updated `pip` too while updating some other packages. 

This update is not trivial but I was really surprised to see that we are using such and old version `1.5.6` while the latest version is `8.1.2` with lots of changes and improvements. My suggestion would be to update our packages on consistent basis. On my local machine I had no issues using this version but I you think that is not necessary or we can have a separate PR then I can do as you suggest.

edx/analytics/tasks/reports/adyen.py
zubair-arbi: Updated

edx/analytics/tasks/reports/adyen.py
zubair-arbi: According to your suggestion I have update the PR to use UTC everywhere.

edx/analytics/tasks/reports/adyen.py
zubair-arbi: I have added retry logic, please review.

edx/analytics/tasks/reports/adyen.py
zubair-arbi: Done

edx/analytics/tasks/reports/adyen.py
zubair-arbi: Updated

edx/analytics/tasks/reports/adyen.py
zubair-arbi: Done

edx/analytics/tasks/reports/adyen.py
zubair-arbi: I have now moved this as a file level setting `ADYEN_CONFIG_SECTION`.

edx/analytics/tasks/reports/adyen.py
```diff
@@ -0,0 +1,305 @@
...
+
+    """
+    # Date to fetch Adyen report.
+    run_date = luigi.DateParameter(
+        default=datetime.utcnow().date(),
+        description='Default is today, UTC.',
+    )
+    username = luigi.Parameter(
+        default_from_config={'section': ADYEN_CONFIG_SECTION, 'name': 'username'},
+        description='Username for the Adyen account, for which data is being gathered for.',
```
brianhw: nit:  either "for which" or "gathered for", but not "for" in both places.

edx/analytics/tasks/reports/adyen.py
```diff
@@ -0,0 +1,305 @@
...
+                return False
+
+            return error_response.status_code in retry_on
+
+        @retry(should_retry=should_retry_fetch_report, timeout=timeout_seconds)
+        def fetch_report_with_retry():
+            """Fetches Adyen report, using an exponential
+            back-off to retry recoverable, failed requests.
+            """
+
```
brianhw: nit: remove blank line.

edx/analytics/tasks/reports/adyen.py
```diff
@@ -0,0 +1,305 @@
...
+            report_response = requests.get(self.report_url, auth=auth)
+            return report_response
+
+        response = fetch_report_with_retry()
+        return response
+
```
brianhw, line 158: nit: check PEP8.  It should complain there is an extra line here.

edx/analytics/tasks/reports/adyen.py
```diff
@@ -0,0 +1,305 @@
...
+
+    @property
+    def report_url(self):
+        """Generate the url to download a report from a Adyen account.
```
brianhw: nit: " from an Adyen account" instead of "from a ..."

edx/analytics/tasks/reports/adyen.py
```diff
@@ -0,0 +1,305 @@
...
+
+        Example URL:
+            https://ca-test.adyen.com/reports/download/MerchantAccount/EdXORG/payments_accounting_report_2016_08_17.csv
+        """
+        url = '{base_url}/MerchantAccount/{merchant_account}/{report_name}_{report_date}.{report_format}'.format(
+            base_url=self.url,
+            merchant_account=self.merchant_account,
+            report_name=self.REPORT_NAME,
+            report_date=self.run_date.strftime('%Y_%m_%d'),
```
brianhw, line 170: nit:  we tend to use isoformat() for dates, which use dashes instead of underlines.

edx/analytics/tasks/reports/adyen.py
```diff
@@ -0,0 +1,305 @@
...
+            csv_reader = csv.DictReader(input_file, delimiter=',')
+
+            with self.output().open('w') as output_file:
+                for row in csv_reader:
+                    # Output most of the fields from the original source file.
+                    # https://docs.adyen.com/developers/reporting-manual#paymentaccountingreportstructure
+
+                    # Read only settled (sale) or refunded (refund) transaction records.
+                    transaction_type = row['Record Type']
+                    if transaction_type not in ['Refunded', 'Settled']:
```
brianhw: Looks like these should really be the keys of TRANSACTION_TYPE_MAP, rather than redefined here.

edx/analytics/tasks/reports/adyen.py
```diff
@@ -0,0 +1,305 @@
...
+    requests.codes.request_timeout,         # HTTP Status Code 408
+    requests.codes.too_many_requests,       # HTTP Status Code 429
+    requests.codes.service_unavailable,     # HTTP Status Code 503
+    requests.codes.gateway_timeout          # HTTP Status Code 504
+)
+DEFAULT_TIMEOUT_SECONDS = 7200
+TRANSACTION_TYPE_MAP = {
+    'Settled': 'sale',
+    'Refunded': 'refund'
+}
```
brianhw, line 35: It would be good to add a comment here that these are the transaction states that we consider relevant.  (And maybe list the other states that are not considered relevant.)

requirements/pre.txt
brianhw: It won't work on our system.  I tried it in a different branch, and the pip upgrade left the system in an inconsistent state that required deleting the virtualenv and starting over.  

I share your pain regarding the Cryptography upgrade.  That currently breaks for me as well, unless it's upgraded to at least 1.0.  But I think all that should be handled in a separate pull request, since it's not at all related to this PR.

brianhw: I'm betting it's _still_ related to https://openedx.atlassian.net/browse/AN-4322, as that comment suggests.  So it's really related to the version of virtualenv that we use on our Jenkins.

brianhw (commented):
Looking good.  A few nits, and reverting the requirements.

edx/analytics/tasks/reports/adyen.py
zubair-arbi: added comment

edx/analytics/tasks/reports/adyen.py
zubair-arbi: done

edx/analytics/tasks/reports/adyen.py
zubair-arbi: removed extra line

edx/analytics/tasks/reports/adyen.py
zubair-arbi: changed

edx/analytics/tasks/reports/adyen.py
zubair-arbi: Since this is the making of external `adyen` report url which needs a specific format of date so cannot modify it.

edx/analytics/tasks/reports/adyen.py
zubair-arbi: good catch. changed

zubair-arbi:
@brianhw @mulby @douglashall I have addressed your feedback and also added tests for different scenarios. Please review.

edx/analytics/tasks/reports/adyen.py
```diff
@@ -0,0 +1,309 @@
...
+import logging
+
+import luigi
+from luigi import date_interval
+from luigi.configuration import get_config
+from pytz import timezone, utc
+import requests
+
+from edx.analytics.tasks.url import get_target_from_url
+from edx.analytics.tasks.url import url_path_join
```
brianhw: nit: you can put these on the same line.

edx/analytics/tasks/reports/adyen.py
brianhw: Don't forget! : )

edx/analytics/tasks/reports/adyen.py
```diff
@@ -0,0 +1,309 @@
...
+        super(DailyPullFromAdyenTask, self).__init__(*args, **kwargs)
+
+        config = get_config()
+        self.password = config.get(ADYEN_CONFIG_SECTION, 'password')
+
+    def run(self):
+        self.remove_output_on_overwrite()
+
+        # Fetch the report
+        response = self.fetch_report(timeout_seconds=self.REPORT_FETCH_TIMEOUT_SECONDS)
```
brianhw, line 91: nit: not clear on the value of overriding the default value for timeout_seconds here, only to set it to the same value.

edx/analytics/tasks/reports/adyen.py
brianhw: Yes, my mistake.  It's not our output.

requirements/test.txt
```diff
@@ -7,5 +7,6 @@ nose-ignore-docstring==0.2
 nose==1.3.4
 pep8==1.6.0
 pylint==1.5.0
+responses==0.5.1
```
brianhw: This has a number of dependencies, and it would be good to freeze them as well.   These include requests, and cookies.  The version for requests should be the same as in default.txt, otherwise a newer version may be installed.  It appears this also wasn't done for six, Jinja2, and python-dateutil -- not related to your change, but if you could also freeze those to their default.txt values, that would be great.

edx/analytics/tasks/reports/tests/test_adyen.py
```diff
@@ -0,0 +1,153 @@
...
+TEST_URL = 'https://ca-test.adyen.com/reports/download'
+TEST_DATA_DIR = os.path.join(os.path.dirname(__file__), 'test_data/')
+
+
+@ddt
+class TestAdyenDailyPullFromAdyenTask(unittest.TestCase):
+
+    DEFAULT_DATE = '2016-08-17'
+
+    @responses.activate
```
brianhw: I'm not really familiar with either, but is there a particular advantage to using responses rather than httpretty as was used in test_paypal.py?  It seems a little odd to use two different packages if they both accomplish roughly the same kind of testing of payment provider APIs.

edx/analytics/tasks/reports/tests/test_adyen.py
```diff
@@ -0,0 +1,153 @@
...
+        self.task = DailyPullFromAdyenTask(
+            run_date=luigi.DateParameter().parse(self.DEFAULT_DATE),
+            output_root='/fake/output',
+            company_account='TestCompany',
+            merchant_account='TestMerchant',
+        )
+        self.output_target = FakeTarget()
+        self.task.output = MagicMock(return_value=self.output_target)
+        dummy_adyen_report_path = os.path.join(TEST_DATA_DIR, 'adyen/2016-08/adyen_TestMerchant_20160817.csv')
+        self.dummy_adyen_report = open(dummy_adyen_report_path, 'r').read()
```
brianhw, line 41: This is the first case of having an external data fixture in our unit tests.   Looking in our other tests here in reports, all the others inline the data as a multiline string.

edx/analytics/tasks/reports/tests/test_adyen.py
```diff
@@ -0,0 +1,153 @@
...
+        # verify that the task's output value is same as our dummy adyen
+        # report file content
+        output_target_value = self.output_target.value.strip()
+        self.assertEqual(output_target_value, self.dummy_adyen_report)
+
+    @data(
+        requests.codes.request_timeout,         # HTTP Status Code 408
+        requests.codes.too_many_requests,       # HTTP Status Code 429
+        requests.codes.service_unavailable,     # HTTP Status Code 503
+        requests.codes.gateway_timeout          # HTTP Status Code 504
```
brianhw: nit:  I've no idea if this would work, but can you use *DEFAULT_RETRY_STATUS_CODES here?  Just curious....

brianhw (commented):
Only nits on the code, but a few more comments on the unit tests.

edx/analytics/tasks/reports/adyen.py
zubair-arbi: updated

edx/analytics/tasks/reports/adyen.py
zubair-arbi: I have update the method `fetch_report` to take a required argument (previously optional) `timeout_seconds` to clear the confusion. This will help us in testing the task `DailyPullFromAdyenTask` with custom `timeout_seconds`.

requirements/test.txt
zubair-arbi: I have pinned tests dependencies `cookies==2.2.1`, `Jinja2==2.7.3`, `python-dateutil==2.2`, `requests==2.7.0` and `six==1.6.1` w.r.t to our default requirements.

edx/analytics/tasks/reports/tests/test_adyen.py
zubair-arbi: I have to use `responses` in order to mock the secure http request (fetching `adyen` reports) since `httpretty` was unable to mock those calls and I couldn't find a way to achieve this other than monkey patching the `httpretty`. I know this seems redundant to use two different libraries for same purpose but I wanted to write tests close to our prod settings. Suggestions are welcome.

edx/analytics/tasks/reports/tests/test_adyen.py
zubair-arbi: I added external data fixture to increase readability (since adding multiple records with different record types can take significant space and also considering individual records to be too long) and improve the unit tests updates in future easily by separating our externally dependent data from our internally processed content.

edx/analytics/tasks/reports/tests/test_adyen.py
zubair-arbi: Yes, this works 👍

edx/analytics/tasks/reports/tests/test_adyen.py
mulby: We just test with "http://" urls elsewhere in the code. Could you do that here to avoid pulling in the new dependency?

brianhw:
LGTM.  👍 after squashing.

edx/analytics/tasks/reports/tests/test_adyen.py
```diff
@@ -0,0 +1,146 @@
...
+        # responses status codes are retry-able
+        latest_requests = httpretty.HTTPretty.latest_requests
+        self.assertEqual(len(latest_requests), 2)
+
+        # verify that the task's output value is same as our dummy adyen
+        # report file content
+        output_target_value = self.output_target.value.strip()
+        self.assertEqual(output_target_value, self.dummy_adyen_report)
+
+    @patch('edx.analytics.tasks.reports.adyen.DailyPullFromAdyenTask.REPORT_FETCH_TIMEOUT_SECONDS', 0.5)
```
mulby: We try _very_ hard to keep our unit tests running very quickly, we currently are able to run ~1700 tests in 9 seconds. I would strongly prefer we don't introduce a 500 ms delay into one of our tests.

I would either not test this logic, or do something like I did here: https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/util/tests/test_edx_api_client.py#L99

Note that it patches `time.sleep()` and `datetime.utcnow()`

mulby (changes requested):
I'd like to remove the `sleep` calls in the unit tests before merging this in.

edx/analytics/tasks/reports/tests/test_adyen.py
zubair-arbi: @mulby with your suggestion we would have to use timeout of value atleast `0.03` seconds so that there are http calls for adyen reports greater than 1. So now I have removed that test check and replaced timeout with `0.001` seconds which should be fine with our test time target.

zubair-arbi:
@mulby @brianhw @douglashall @HassanJaveed84  I have squashed and rebased my PR with master and it's ready for final review. There are some unit tests failing related to `CourseEnrollmentTask` with error `UnknownParameterException` for `output_root` which seems unrelated to my changes.

I have couple of questions too:
1. 
   how we add prod config settings for `adyen`?
2. 
   would merging this PR now would have any negative affect? since we haven't integrated `Adyen` in out system.

zubair-arbi:
@brianhw @mulby The PR is ready for final review. Both unit tests and acceptance tests are passing now.

brianhw:
Prod config settings are stored in a separate private repo that we use for storing secure information.  We would need to add Adyen sections to this, to at least define 'merchant_accounts_list' as an empty list.  When we wanted to actually start pulling information from Adyen, we would add entries to 'merchant_accounts_list' and 'company_account'.

Those changes should probably be done in a branch of the secure repo, and then we would run payment-pulling (or all finance) with a different warehouse_path to keep the results separate from production.

mulby (approved):

mulby:
@zubair-arbi - this looks good now, but we need to update our production configuration before merging this to master so that it doesn't break production.

Can you create an analytics ticket that specifies the procedure we can use to obtain the proper reporting credentials for Adyen? We can then follow that procedure, update our production config and merge this to master.

zubair-arbi:
@mulby thanks for follow up. I have created an [analytics ticket AN-8051](https://openedx.atlassian.net/browse/AN-8051) for providing procedure to get production configuration settings. The steps aren't defined yet, I have asked @douglashall for help on describing this procedure.

mulby:
After some discussion on AN-8051, we've decided to close this until the project is re-prioritized.

PR #283, opened 2016-08-24
Create a program mapping table in the data warehouse

mulby:


mulby:
@bderusha - I posted this PR early so that you can follow along in the development process. So far this is what I did:
1. Run launch-task on my devstack passing in custom values for --api-root-url and --api-access-token using my own personal access token (still trying to figure out how to get one for a programmatic client, I have an open DevOps ticket). I ran the task ExtractProgramCourseTask since my devstack isn't configured to insert into Vertica.
2. I created and ran jenkins job in our scheduler that can run this workflow and actually insert into Vertica.

If things look good, I'll refactor this a bit, fixture with unit tests and add an acceptance test.

Outstanding questions:
- As we pull multiple pages of data, is it possible we actually get an inconsistent view of the complete data set?
- Should we implement a throttling mechanism?
- Will the API let us know when we are hitting it too hard so that we can back-off?

I'm also planning on adding additional tasks here that parse raw course_run json blobs and extract d_course_seats (mapping from course_ids to information about the seats they have available) and a d_course table that exposes most of the other fields returned here. We can just pull the raw JSON once and then generate all of those tables from this data.

mulby:
Note that this will cache the pull for the day. If we ran this task multiple times in the same day it would just use the cached value unless we specify "--overwrite"

edx/analytics/tasks/load_internal_reporting_course_catalog.py
```diff
@@ -0,0 +1,153 @@
...
+    """Call the course catalog API and place the resulting JSON on S3."""
+
+    def run(self):
+        self.remove_output_on_overwrite()
+        offset = 0
+        with self.output().open('w') as output_file:
+            response = self.get_page_of_data(offset)
+            offset += self.write_response_to_output(response, output_file)
+            while response.get('next') is not None:
+                response = self.get_page_of_data(offset)
```
bderusha: Why are we manually doing all of this with the offset?  Can't we just used the `next` link provided by the API?

mulby: Hah, oops, I ported this from some old code and didn't notice the redundancy. Yeah, this should absolutely just be using the "next" url.

mulby: Fixed.

mulby:
@brianhw

edx/analytics/tasks/load_internal_reporting_course_catalog.py
```diff
@@ -0,0 +1,155 @@
...
+
+        parsed_response = response.json()
+        for course in parsed_response.get('results', []):
+            output_file.write(json.dumps(course))
+            output_file.write('\n')
+
+        return parsed_response.get('next')
+
+    def output(self):
+        """Output is in the form {warehouse_path}/courses_raw/dt={CCYY-MM-DD}/course_structure.json"""
```
brianhw: nit: comment is wrong.

edx/analytics/tasks/load_internal_reporting_course_catalog.py
```diff
@@ -0,0 +1,155 @@
...
+                                          "course_catalog.json")
+        return get_target_from_url(url_with_filename)
+
+
+class ProgramCourseRecord(Record):
+    program_id = StringField(nullable=False, length=36)
+    program_type = StringField(nullable=False, length=32)
+    program_title = StringField(nullable=True, length=255)
+    catalog_course = StringField(nullable=False, length=255)
+    catalog_course_title = StringField(nullable=True, length=255)
```
brianhw, line 91: Not sure, but this may need to be larger?

brianhw:
I think that retrying on 408 or 429 might be reasonable, but I don't know if we have to deal with that right away.  (I wrote some retry code for course_structure, but I don't think that it ever was actually needed.  The 50x I was getting turned out to be unrecoverable, so retrying accomplished nothing there.

As for throttling, I saw a comment on Adam's loading of neo4j from course_blocks that suggested that slowing things down was desirable, as well as loading from a replica to not affect caching.  I don't think this has the same impact.

edx/analytics/tasks/load_internal_reporting_course_catalog.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def output(self):
+        date_string = self.date.strftime('dt=%Y-%m-%d')  # pylint: disable=no-member
+        url_with_filename = url_path_join(self.warehouse_path, 'course_catalog_raw', date_string,
+                                          "course_catalog.json")
+        return get_target_from_url(url_with_filename)
+
+
+class ProgramCourseRecord(Record):
+    program_id = StringField(nullable=False, length=36)
```
bderusha, line 87: I don't know if this is important, but programs and courses will have a `status` attribute denoting if they are published, retired, deleted, etc.  Is that information relevant to this?

mulby: Probably!

I haven't yet pulled the data from the programs endpoint... I was planning on (eventually) pulling that into a separate d_program table that you could join with this mapping table using the program's UUID.

edx/analytics/tasks/load_internal_reporting_course_catalog.py
mulby: I don't think so?

https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/models.py#L235

mulby:
@HassanJaveed84 and @brianhw - this is ready for review

config/devstack.cfg
```diff
@@ -83,4 +83,4 @@ host = http://172.17.0.1:9201/
 
 [module-engagement]
 alias = roster
-number_of_shards = 5
+number_of_shards = 5
```
brianhw: Any reason for making this change?

brianhw: Should there in fact be a course-catalog-api section here too?

edx/analytics/tasks/load_internal_reporting_course_catalog.py
```diff
@@ -0,0 +1,166 @@
...
+        is_list=True,
+        description="A list of partner short codes that we should fetch data for."
+    )
+    api_root_url = luigi.Parameter(
+        config_path={'section': 'course-catalog-api', 'name': 'api_root_url'},
+        description="The base URL for the course catalog API. This URL should look like"
+                    "https://catalog-service.example.com/api/v1/"
+    )
+    api_page_size = luigi.IntParameter(
+        config_path={'section': 'course-catalog-api', 'name': 'page_size'}, significant=False,
```
brianhw: nit: why isn't "significant" on a separate line?

edx/analytics/tasks/load_internal_reporting_course_catalog.py
brianhw: Okay.  I may have been thinking of description.

edx/analytics/tasks/load_internal_reporting_course_catalog.py
```diff
@@ -0,0 +1,166 @@
...
+        default=100,
+        description="The number of records to request from the API in each HTTP request."
+    )
+
+
+class PullCourseCatalogAPIData(LoadInternalReportingCourseCatalogMixin, luigi.Task):
+    """Call the course catalog API and place the resulting JSON on S3."""
```
brianhw: There's actually nothing I see here that is specific to S3.  "on S3" => "in an output file"?

edx/analytics/tasks/load_internal_reporting_course_catalog.py
```diff
@@ -0,0 +1,166 @@
...
+        for program in course_run.get('programs', []):
+            record = ProgramCourseRecord(
+                program_id=program['uuid'],
+                program_type=program['type'],
+                program_title=program.get('title'),
+                catalog_course=course_run['course'],
+                catalog_course_title=course_run['title'],
+                course_id=course_run['key'],
+                org_id=get_org_id_for_course(course_run['key']),
+                partner_short_code=course_run.get('partner_short_code')
```
brianhw, line 135: So is this a conscious indication of what is required to be included by the API and what is optional?  As written, all fields are required except partner_short_code and program_title, which use get() calls to provide default behavior.  Just checking...

edx/analytics/tasks/tests/acceptance/test_internal_reporting_course_catalog.py
```diff
@@ -0,0 +1,58 @@
...
+            cursor.execute(
+                "SELECT program_id,program_type,program_title,catalog_course,catalog_course_title,course_id,org_id,"
+                "partner_short_code FROM {schema}.d_program_course ORDER BY course_id ASC".format(
+                    schema=self.vertica.schema_name
+                )
+            )
+            response = cursor.fetchall()
+            d_program_course = pandas.DataFrame(response, columns=['program_id', 'program_type', 'program_title',
+                                                                   'catalog_course', 'catalog_course_title',
+                                                                   'course_id', 'org_id', 'partner_short_code'])
```
brianhw, line 53: nit: can we get the columns from the ProgramCourseRecord, instead of hardcoding here?

edx/analytics/tasks/util/tests/test_edx_api_client.py
```diff
@@ -0,0 +1,189 @@
...
+@httpretty.activate
+class EdxApiClientTestCase(unittest.TestCase):
+    """Test the client"""
+
+    def setUp(self):
+        self.current_time = datetime(2016, 1, 1, 0, 0, 0, 0)
+        self.time_offset = 0
+
+        class MockDateTime(datetime):
+            """A fake datetime that returns the time we need it two when called."""
```
brianhw: nit: what does this mean?

edx/analytics/tasks/util/retry.py
```diff
@@ -0,0 +1,74 @@
...
+log = logging.getLogger(__name__)
+
+
+class RetryTimeoutError(Exception):
+    """Raised when repeated attempts fail to succeed before the specified timeout expires."""
+    pass
+
+
+def retry_on_any_exception(_exc):
+    """Retry regardless of the type of exception raised or it's data."""
```
brianhw: nit:  "its"

edx/analytics/tasks/util/hive.py
```diff
@@ -65,6 +65,13 @@ class WarehouseMixin(object):
         description='A URL location of the data warehouse.',
     )
 
+    def hive_partition_path(self, table_name, partition_value, partition_key='dt'):
+        """pass"""
```
brianhw: nit:   TBD. : )

edx/analytics/tasks/util/edx_api_client.py
```diff
@@ -0,0 +1,166 @@
...
+
+    Sample configuration that can be included in the configuration file:
+
+    [edx-rest-api]
+    auth_url = https://example.com/oauth2/v1/access_token
+    client_id = 1234567890
+    client_secret = this-should-be-a-secret
+
+    Arguments:
+        auth_url (str): The full URL of of the authentication endpoint. It should look like
```
brianhw: nit:  "of" appears twice.

edx/analytics/tasks/util/edx_api_client.py
```diff
@@ -0,0 +1,166 @@
...
+                      retry_on=DEFAULT_RETRY_STATUS_CODES, pagination_key='next'):
+        """
+        Fetches a paginated resource.
+
+        Arguments:
+            url (str): The URL of the resource.
+            params (dict): This is a dictionary of key-value pairs that are URL encoded and injected into the query
+                string when making the request.
+            timeout_seconds (float): When requesting a page, keep retrying unless this much time has elapsed. Raise an
+                error if a successful response cannot be fetched in this amount of time.
```
brianhw: I don't know if it's important to clarify here, but the timeout applies to each page, not to the total fetching from the resource.

edx/analytics/tasks/util/edx_api_client.py
```diff
@@ -0,0 +1,166 @@
...
+
+log = logging.getLogger(__name__)
+
+DEFAULT_RETRY_STATUS_CODES = (
+    requests.codes.request_timeout,
+    requests.codes.too_many_requests,
+    requests.codes.service_unavailable,
+    requests.codes.gateway_timeout
+)
+DEFAULT_TIMEOUT = 7200
```
brianhw: It would be good to note here what the units are.  I assume seconds, rather than milliseconds.  

It would also be convenient to have the status codes following the hardcoded values as comments, just to know.

brianhw:
Nothing substantial -- I think only doc nits.  LGTM. 👍

config/devstack.cfg
mulby: I'll omit this section until we have a better idea of where this service would live in the devstack, what port etc.

edx/analytics/tasks/load_internal_reporting_course_catalog.py
mulby: done

edx/analytics/tasks/load_internal_reporting_course_catalog.py
mulby: done

edx/analytics/tasks/load_internal_reporting_course_catalog.py
mulby: Correct.

edx/analytics/tasks/tests/acceptance/test_internal_reporting_course_catalog.py
mulby: I actually prefer the duplication since it should give us pause when changing it. In general I feel that acceptance tests should break if a breaking change is introduced.

edx/analytics/tasks/util/tests/test_edx_api_client.py
mulby: Hopefully clarified.

edx/analytics/tasks/util/retry.py
mulby: done

edx/analytics/tasks/util/hive.py
mulby: heh, oops! done.

edx/analytics/tasks/util/edx_api_client.py
mulby: done

edx/analytics/tasks/util/edx_api_client.py
mulby: I had hoped to convey that, but now see that it's still somewhat ambiguous. I've tried to clarify the wording.

edx/analytics/tasks/util/edx_api_client.py
mulby: Good idea, done.

edx/analytics/tasks/util/tests/test_edx_api_client.py
```diff
@@ -0,0 +1,193 @@
...
+class EdxApiClientTestCase(unittest.TestCase):
+    """Test the client"""
+
+    def setUp(self):
+        self.current_time = datetime(2016, 1, 1, 0, 0, 0, 0)
+        self.time_offset = 0
+
+        class MockDateTime(datetime):
+            """
+            A fake datetime that allows us the test to control the return value from the utcnow() method.
```
brianhw: nitty nit:  "allows us the test to control":  is it "us" or "the test", or something else?

brianhw:
LGTM. 👍

PR #284, opened 2016-08-24
Hotfix enrollment and malformed course_ids

mulby:
The primary fix here is to reject events that contain these malformed course_ids that end with "\n" characters. Longer term we should probably clean these course_ids, but this felt like the smallest scope patch that would keep our jobs running. This will introduce a new source of error into our metrics, however, given the small number of enrollments (~200) that are affected by this issue, I think that's acceptable.

Note that enrollment on master is broken, so this patch also includes a fix for that issue.

Reviewer: @HassanJaveed84 
FYI: @adampalay @dsjen @stroilova @katymyw @brianhw

### DIFF
--- a/edx/analytics/tasks/util/opaque_key_util.py
+++ b/edx/analytics/tasks/util/opaque_key_util.py
@@ -22,6 +22,10 @@ def is_valid_course_id(course_id):
     """
     Determines if a course_id from an event log is possibly legitimate.
     """
+    if course_id and course_id[-1] == '\n':
+        log.error("Found course_id that ends with a newline character '%s'", course_id)
+        return False
+
     try:
         _course_key = CourseKey.from_string(course_id)
         return True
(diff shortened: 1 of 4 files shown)


edx/analytics/tasks/util/opaque_key_util.py
```diff
@@ -22,6 +22,10 @@ def is_valid_course_id(course_id):
     """
     Determines if a course_id from an event log is possibly legitimate.
     """
+    if course_id[-1] == '\n':
```
adampalay: @mulby , can this ever be an empty string? b/c this would throw an error

mulby: Great catch. Fixed.

adampalay:
👍

HassanJaveed84:
LGTM 👍

mulby:
@HassanJaveed84 I've merged in this branch, so I'm all set from a release standpoint. Note that we shouldn't delete the branch from the repo, though, until after the release since the jobs are running off this branch for the time being.

PR #285, opened 2016-08-25
handle course_ids with newline characters in them

mulby:
This is one of those moments when I wish I had a canonicalization stage implemented :frowning: 

I searched for places in the code where we were calling `opaque_key_util.is_valid_course_id()` and introduced this normalization before it. I'm not 100% confident this will cover all cases where we parse course_ids from events, but it will certainly cover most of them. Let me know if you think of any others or another approach we could use to find others.

My plan is to:
1. Baseline the number of occurrences of this error message: https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/util/opaque_key_util.py#L26 when running the enrollment workflow.
2. Apply this patch and run it in stage.
3. Check the occurrences of that message on the fixed run - it should drop to zero.
4. Deploy the fix to prod.
5. Download logs for each production job after the fix and see they contain that error message.

Reviewer: @HassanJaveed84 @adampalay 
FYI: @adampalay @dsjen @stroilova @katymyw @brianhw

### DIFF
--- a/edx/analytics/tasks/answer_dist.py
+++ b/edx/analytics/tasks/answer_dist.py
@@ -133,7 +133,7 @@ class ProblemCheckEventMixin(object):
         event = json.loads(event_string)
 
         # Get context information:
-        course_id = event.get('context').get('course_id')
+        course_id = eventlog.get_course_id(event)
         timestamp = event.get('timestamp')
         problem_id = event.get('problem_id')
         problem_display_name = event.get('context').get('module', {}).get('display_name', None)
@@ -979,7 +979,7 @@ def get_problem_check_event(line):
     # contain the org and course name, but not the run.)  Course_id
     # information could be found from other events, but it would
     # require expanding the events being selected.
-    course_id = problem_data.get('context').get('course_id')
+    course_id = eventlog.get_course_id(event)
     if course_id is None:
         log.error("encountered explicit problem_check event with missing course_id: %s", event)
         return None
--- a/edx/analytics/tasks/lms_courseware_link_clicked.py
+++ b/edx/analytics/tasks/lms_courseware_link_clicked.py
@@ -53,13 +53,8 @@ class LMSCoursewareLinkClickedTask(EventLogSelectionMixin, MapReduceJobTask):
             log.error("encountered explicit link_clicked event with no event data: %s", event)
             return
 
-        context = event.get('context')
-        if not context:
-            log.error("encountered explicit link_clicked event with no context: %s", event)
-            return
-
-        course_id = context.get('course_id')
-        if course_id is None or not opaque_key_util.is_valid_course_id(course_id):
+        course_id = eventlog.get_course_id(event)
+        if course_id is None:
             log.error("encountered explicit link_clicked event with invalid course_id: %s", event)
             return
 
--- a/edx/analytics/tasks/load_internal_reporting_course.py
+++ b/edx/analytics/tasks/load_internal_reporting_course.py
@@ -12,7 +12,7 @@ import ciso8601
 
 from opaque_keys.edx.keys import CourseKey
 
-from edx.analytics.tasks.util.opaque_key_util import is_valid_course_id
+from edx.analytics.tasks.util.opaque_key_util import is_valid_course_id, normalize_course_id
 from edx.analytics.tasks.url import get_target_from_url
 from edx.analytics.tasks.url import url_path_join
 from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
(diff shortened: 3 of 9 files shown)


edx/analytics/tasks/load_internal_reporting_course.py
```diff
@@ -99,7 +99,7 @@ def run(self):
                         else:
                             cleaned_end_string = ciso8601.parse_datetime(end_string)
 
-                        course_id = course.get('id', '\N')
+                        course_id = normalize_course_id(course.get('id', '\N'))
```
adampalay, line 102: why is this the default value?

adampalay:
👍

edx/analytics/tasks/load_internal_reporting_course.py
mulby: That's the character code that Hive uses to represent NULL values in text files.

adampalay: cool

edx/analytics/tasks/answer_dist.py
```diff
@@ -979,7 +979,7 @@ def get_problem_check_event(line):
     # contain the org and course name, but not the run.)  Course_id
     # information could be found from other events, but it would
     # require expanding the events being selected.
-    course_id = problem_data.get('context').get('course_id')
+    course_id = opaque_key_util.normalize_course_id(problem_data.get('context').get('course_id'))
```
brianhw: This is one of two places in answer_dist where we get the course_id.  We get it here to use as part of the key, but then its also pulled out in the reducer (or actually in _generate_answers()).  But I'm not seeing this in the prod answer_distribution table, even though I've confirmed there were problem_check events on 8/23 and 8/24 that had 'n' characters appended in the events loaded to Vertica.

brianhw: That is, I'm not seeing any funny course_id's in answer_distribution.  Perhaps they are being dropped at present?

brianhw: Note that these places could make use of the eventlog.get_course_id() method that is used for much of our more recent non-enrollment tasks, like module_engagement and user_activity and video and such.

edx/analytics/tasks/lms_courseware_link_clicked.py
```diff
@@ -58,7 +58,7 @@ def mapper(self, line):
             log.error("encountered explicit link_clicked event with no context: %s", event)
             return
 
-        course_id = context.get('course_id')
+        course_id = opaque_key_util.normalize_course_id(context.get('course_id'))
```
brianhw: nit: This too should probably be using the eventlog.get_course_id().

edx/analytics/tasks/load_internal_reporting_course.py
brianhw: Is this just being defensive, or do we expect there to have been problems with this data?   I looked at the raw exports for 8/23 and 8/24 and didn't see a problem with it.

edx/analytics/tasks/answer_dist.py
brianhw: Right.  They are being dropped by the is_valid_course_id() call below.  But with the code change, they would get through, but not get processed correctly, and would also not be caught by the logging, since the second location doesn't bother to check that the resulting course_id is valid.

brianhw:
Answer_dist needs to be fixed -- it will not appear broken, but it won't work correctly.  

Optional, but it would be good to fix it and another one (or two) with calls to eventlog.get_course_id(), so that we're more consistent with getting course_id from context.

We could also add a new eventlog method that gets the course_id from payload, and use that for the enrollment and navigation_click, to centralize that logic too.

edx/analytics/tasks/load_internal_reporting_course.py
mulby: I don't know, but I suspect it's just being defensive. I can't imagine pulling a record that didn't have a course ID...

edx/analytics/tasks/answer_dist.py
mulby: Fixed. Great catch on the second course_id usage.

edx/analytics/tasks/lms_courseware_link_clicked.py
mulby: Agreed. Done.

edx/analytics/tasks/lms_courseware_link_clicked.py
```diff
@@ -53,12 +53,7 @@ def mapper(self, line):
...
             return
 
-        context = event.get('context')
-        if not context:
-            log.error("encountered explicit link_clicked event with no context: %s", event)
-            return
-
-        course_id = context.get('course_id')
+        course_id = eventlog.get_course_id(event)
         if course_id is None or not opaque_key_util.is_valid_course_id(course_id):
```
brianhw: nit: get_course_id() includes a call to is_valid_course_id(), so we can just check for None here and skip the logging as well (and just return).

brianhw:
LGTM.  👍

PR #286, opened 2016-08-26
Make elasticsearch load more tunable and resiliant

mulby:
@HassanJaveed84 here is the branch I'm currently working on

### DIFF
--- a/edx/analytics/tasks/module_engagement.py
+++ b/edx/analytics/tasks/module_engagement.py
@@ -1342,6 +1342,12 @@ class ModuleEngagementWorkflowTask(ModuleEngagementDownstreamMixin, ModuleEngage
 
     # Don't use the OverwriteOutputMixin since it changes the behavior of complete() (which we don't want).
     overwrite = luigi.BooleanParameter(default=False, significant=False)
+    throttle = luigi.FloatParameter(
+        config_path={'section': 'module-engagement', 'name': 'throttle'},
+        description=ElasticsearchIndexTask.throttle.description,
+        default=0.75,
+        significant=False
+    )
 
     def requires(self):
         overwrite_from_date = self.date - datetime.timedelta(days=self.overwrite_n_days)
@@ -1353,6 +1359,7 @@ class ModuleEngagementWorkflowTask(ModuleEngagementDownstreamMixin, ModuleEngage
             n_reduce_tasks=self.n_reduce_tasks,
             overwrite_from_date=overwrite_from_date,
             overwrite=self.overwrite,
+            throttle=self.throttle
         )
         yield ModuleEngagementSummaryMetricRangesMysqlTask(
             date=self.date,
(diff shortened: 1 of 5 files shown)


mulby:
@HassanJaveed84 can you review this when you get a chance?

HassanJaveed84:
LGTM 👍

mulby:
@brianhw can you review this today?

edx/analytics/tasks/module_engagement.py
```diff
@@ -1342,6 +1342,14 @@ class ModuleEngagementWorkflowTask(ModuleEngagementDownstreamMixin, ModuleEngage
...
     # Don't use the OverwriteOutputMixin since it changes the behavior of complete() (which we don't want).
     overwrite = luigi.BooleanParameter(default=False, significant=False)
+    throttle = luigi.FloatParameter(
+        default=0.1,
+        significant=False,
+        description='Wait this many seconds between batches of records submitted to the cluster to be indexed. This can'
+                    ' be used to tune the indexing process, allowing the cluster to successfully "keep up" with the'
+                    ' loader. Note that often the hadoop cluster can load records much more quickly than the cluster'
+                    ' can index them, which eventually causes queues to overflow within the elasticsearch cluster.'
+    )
```
brianhw, line 1350: I'm sorry if we settled this before, but I'm confused by the redefinition of this here, rather than the inclusion of ElasticsearchIndexTaskMixin.  The latter mixin class has defined this parameter, but it also has many other parameters, and it is only used to define ElasticsearchIndexTask.  If we don't have the right parameters in the mixin to use downstream, then it should be split up.  But it should just work to include the mixin here, and pass only the throttle down.

mulby: This is likely to spiral into a philosophical debate... I find the tangle of mixins to be very hard to read in the code. I'd rather have some duplication than have to hunt through a complex hierarchy to figure out what parameter the task has.

Note that in this case I also want to specialize the settings for this parameter for this workflow (which I just put into the latest commit).

As a compromise, I've at least pulled the description from the parent parameter.

brianhw:
The rest looks fine.  I commented on the throttle while you were changing it.  The big change I see is that you put it in config files as a module_enrollment throttle setting, rather than an elastic-search one.  So I'm a little confused about the use of ElasticsearchIndexTaskMixin.  It currently serves no purpose, as it's only used in ElasticsearchIndexTask.

mulby:
That's fair, it looks like realistically, we'll almost always want to override those parameters. I'll just remove the Mixin for now.

mulby:
@brianhw I think I've addressed your comment. Can you take another pass?

brianhw:
LGTM.  👍

brianhw:
Was there a reason this wasn't merged?  Are you working on this elsewhere?

mulby:
no I merged it on the CLI and for some reason github is calling it closed. It's in master.

PR #287, opened 2016-08-30
 Incremental enrollment.

HassanJaveed84:
@mulby @brianhw 
I started with a different approach where I used Hive partition on date to store daily enrollment. But I wasn't sure how to run a mapreduce over a hive table output(all partitions). 

Currently, I just store daily enrollment on a warehouse path and another mapreduce job generates the enrollment like we do now. It's incremental in sense that we would have a daily task that just processes one days of event logs, and another that would use all the incremental task outputs and produces what we have now.

Is there a way I can put daily enrollment into hive partitions and query across all the partitions ?


edx/analytics/tasks/pathutil.py
```diff
@@ -145,10 +145,10 @@ class EventLogSelectionTask(EventLogSelectionDownstreamMixin, luigi.WrapperTask)
 
     def __init__(self, *args, **kwargs):
         super(EventLogSelectionTask, self).__init__(*args, **kwargs)
-        self.interval = DateInterval(
-            self.interval.date_a - self.expand_interval,
-            self.interval.date_b + self.expand_interval
-        )
+        # self.interval = DateInterval(
```
HassanJaveed84: Hack to get one day's of event logs. Can we specify some value for expand_interval to achieve that ?

edx/analytics/tasks/enrollments.py
```diff
@@ -69,13 +89,65 @@ def mapper(self, line):
...
+
+    output_root = luigi.Parameter()
+
+    interval = luigi.DateIntervalParameter(
+        description='The range of dates to export logs for.',
+    )
+
+    def requires(self):
+        yield ExternalURL(
+            url=url_path_join(self.warehouse_path, 'course_enrollment_incremental') + '/*/'
```
HassanJaveed84: Was not sure how to best read a recursive structure, so I just used this for testing.

HassanJaveed84:
@mulby @brianhw this is ready for review again. Do we need to add unit tests ?

edx/analytics/tasks/enrollments.py
```diff
@@ -69,13 +90,87 @@ def mapper(self, line):
...
+        default=None,
+        description='A URL to a file location that can store the complete set of input files. '
+        'A manifest file is required by Hadoop if it hits the OS limit on the length '
+        'of the command to run when launching the job using the Hadoop CLI. '
+        'This file will be written to if it doesn\'t exist, and read from if it already does.',
+    )
+
+    def requires(self):
+        source = url_path_join(self.warehouse_path, 'course_enrollment_incremental')
+        yield PathSetTask([source], self.include, self.manifest)
```
brianhw: Sorry, I'm not clear how this is working.  This is meant to grab the days that have already been processed, presumably to be able to pass on to the downstream tasks the list of days to use as input.  Is this dynamic?  Does it need to be diffed with the overwrite interval below?  Or is there an implicit assumption that because it is yielded first, it will be run after the CourseEnrollmentDailyTask is executed below, and will therefore also list them in the resulting input (twice)? Does Hadoop do a dedupe of files passed to it?

HassanJaveed84: I overlooked this when I introduced the overwrite_n_days. This is definitely an issue. 
We can override `input_hadoop` method, but we would need to wrap it around our logic of converting to manifest if necessary. 
Probably someting like:

``` python
def input_hadoop(self):
    manifest_id = str(hash(self)).replace('-', 'n')
    return convert_to_manifest_input_if_necessary(manifest_id, PathSetTask([source]).output())
```

What we do currently in  MapReduceJobTask:

``` python
def input_hadoop(self):
    manifest_id = str(hash(self)).replace('-', 'n')
    return convert_to_manifest_input_if_necessary(manifest_id, super(MapReduceJobTask, self).input_hadoop())
```

From the super class:

``` python
def requires_hadoop(self):
    return self.requires()  # default impl

def input_hadoop(self):
    return luigi.task.getpaths(self.requires_hadoop())
```

This seems hackish, I'm not sure if there is a better way of doing it. Dynamic requirements would have definitely helped. What do you think ?

edx/analytics/tasks/enrollments.py
```diff
@@ -8,20 +8,41 @@
...
+        description='The range of dates to export logs for.',
+    )
+    output_root = luigi.Parameter()
+
+    def requires(self):
+        for date in [d for d in self.interval]:  # pylint: disable=not-an-iterable
+            day_interval = luigi.date_interval.Custom(date, date + datetime.timedelta(days=1))
+            yield CourseEnrollmentDailyTask(
+                interval=day_interval,
+                n_reduce_tasks=10,
```
mulby: This should be parameterized

edx/analytics/tasks/enrollments.py
```diff
@@ -8,20 +8,41 @@
...
+class GenerateIncrementalEnrollmentTask(luigi.WrapperTask):
+
+    interval = luigi.DateIntervalParameter(
+        description='The range of dates to export logs for.',
+    )
+    output_root = luigi.Parameter()
+
+    def requires(self):
+        for date in [d for d in self.interval]:  # pylint: disable=not-an-iterable
+            day_interval = luigi.date_interval.Custom(date, date + datetime.timedelta(days=1))
```
mulby: Could you use `luigi.date_interval.Date`?

edx/analytics/tasks/enrollments.py
```diff
@@ -69,13 +90,87 @@ def mapper(self, line):
         yield (course_id, user_id), (timestamp, event_type, mode)
 
     def reducer(self, key, values):
+        course_id, user_id = key
+        for value in values:
+            timestamp, event_type, mode = value
+            yield course_id, user_id, timestamp, event_type, mode
+
+    def output(self):
+        return get_target_from_url(url_path_join(self.output_root, 'course_enrollment_incremental', self.interval.date_b.isoformat()))
```
mulby: to make this hive compatible we could use this method to generate these paths: https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/util/hive.py#L68

edx/analytics/tasks/enrollments.py
```diff
@@ -69,13 +90,87 @@ def mapper(self, line):
...
+
+    def complete(self):
+        return get_target_from_url(url_path_join(self.output_root, 'course_enrollment_incremental', self.interval.date_b.isoformat(), '_SUCCESS')).exists()
```
mulby: We may want to extract a mixin for these `enable_direct_output` patterns.

edx/analytics/tasks/enrollments.py
mulby: Yeah I'd prefer to not duplicate that hashing and manifest conversion logic if possible.

Is there someway we could constrain the PathSetTask so that it didn't include the paths that we are overwriting?

Maybe we could override `requires_hadoop()` to de-duplicate? That also feels hacky...

mulby (commented):
Generally looking good. The big puzzle is the PathSetTask + overwrite situation.

edx/analytics/tasks/enrollments.py
```diff
@@ -7,32 +7,31 @@
...
+
 log = logging.getLogger(__name__)
 DEACTIVATED = 'edx.course.enrollment.deactivated'
 ACTIVATED = 'edx.course.enrollment.activated'
 MODE_CHANGED = 'edx.course.enrollment.mode_changed'
 
 
-class CourseEnrollmentTask(EventLogSelectionMixin, MapReduceJobTask):
-    """Produce a data set that shows which days each user was enrolled in each course."""
+class CourseEnrollmentMixin(object):
```
brianhw: nit:  I suggest renaming this to CourseEnrollmentMapperMixin, so it's clearer that's what it provides?

edx/analytics/tasks/enrollments.py
```diff
@@ -7,32 +7,31 @@
...
 
     output_root = luigi.Parameter()
```
brianhw: nit: does this need to be defined here?  It's not part of the mapping functionality.

edx/analytics/tasks/enrollments.py
```diff
@@ -66,7 +63,107 @@ def mapper(self, line):
...
+    def output(self):
+        return get_target_from_url(url_path_join(self.output_root, 'course_enrollment_incremental', self.interval.date_b.isoformat()))
+
+    def complete(self):
+        return get_target_from_url(url_path_join(self.output_root, 'course_enrollment_incremental', self.interval.date_b.isoformat(), '_SUCCESS')).exists()
+
+    def run(self):
+        output_target = self.output()
+        if not self.complete() and output_target.exists():
+            output_target.remove()
```
brianhw: Is it possible to add a comment here explaining why this is needed?  In order for run() to be called in the first place, it had to not be complete, so there was no marker file.  So the purpose here is to clean up any partial data that may have been left around from previous failures, right?

HassanJaveed84: Actually I was thinking that maybe we can remove this part(along with the complete method) and also remove `enable_direct_output = True` from this task. Thoughts ?

edx/analytics/tasks/enrollments.py
```diff
@@ -66,7 +63,107 @@ def mapper(self, line):
...
+    )
+
+    def requires(self):
+        source = url_path_join(self.warehouse_path, 'course_enrollment_incremental')
+        yield PathSetTask([source], self.include, self.manifest)
+
+        overwrite_interval = luigi.date_interval.Custom(self.interval.date_b - datetime.timedelta(days=self.overwrite_n_days), self.interval.date_b)
+
+        for date in overwrite_interval:
+            day_interval = luigi.date_interval.Custom(date, date + datetime.timedelta(days=1))
```
brianhw: nit: I found it confusing looking at the CourseEnrollmentDailyTask, and seeing nothing in it that was really daily.  The "daily" logic is really here.  Would it make sense to pass the date to the DailyTask, and have it do this conversion to define the interval?  Or is that too hard to do with existing log-selection code?

edx/analytics/tasks/enrollments.py
```diff
@@ -66,7 +63,107 @@ def mapper(self, line):
...
+    def output_path_for_key(self, key):
+        date_string = key
+        return url_path_join(
+            self.output_root,
+            'course_enrollment_incremental',
+            date_string,
+            'course_enrollment_{date}'.format(
+                date=date_string,
+            )
+        )
```
brianhw, line 92: An issue with interoperability of history and daily:  I would expect that it should not matter whether I used history to create data for a sequence of days, or whether I do a "daily" for each of those days.  But while the daily relies on _SUCCESS markers, I don't see them being produced by the history task.   Maybe it makes sense to not rely on the _SUCCESS markers here, and just create a single file per day (with a single reducer).

edx/analytics/tasks/enrollments.py
```diff
@@ -66,7 +63,107 @@ def mapper(self, line):
...
+        default=None,
+        description='A URL to a file location that can store the complete set of input files. '
+        'A manifest file is required by Hadoop if it hits the OS limit on the length '
+        'of the command to run when launching the job using the Hadoop CLI. '
+        'This file will be written to if it doesn\'t exist, and read from if it already does.',
+    )
+
+    def requires(self):
+        source = url_path_join(self.warehouse_path, 'course_enrollment_incremental')
+        yield PathSetTask([source], self.include, self.manifest)
```
brianhw: I'm interested in following up on the PathSetTask filtering.  There are a couple of possibilities here.
- Add an argument to PathSetTask that defines an additional arbitrary filter function that can be passed in.  In this case, that function could be defined to pull the date out of each path (i.e. by matching on "dt=(\d\d\d\d-\d\d-\d\d)"), and excluding those that are within the overwrite interval.
- Materialize the list of paths returned by PathTaskSet without converting to a manifest file (i.e. manifest=None), then remove those from the list that are within the overwrite interval.  Then call the convert-to-manifest on that list. 
- Instead of PathSetTask, modify and generalize EventLogSelectionTask.  A better name might be PathSelectionByDateTask, since its approach doesn't have to be only for Event logs.   The only thing that needs to be changed is the format of the date string that is extracted and parsed.  The interval passed in would be the non-overlap interval.  The include argument would provide a pattern that defines "date".    EventLogSelectionMixin would define the date_pattern property to be  '%Y%m%d' when it instantiates PathSelectionByDateTask (instead of EventLogSelectionTask).  Here (i.e. CourseEnrollmentTask.requires()) we could also use PathSelectionByDateTask, but pass in a different date_pattern property, as well as an include pattern, source directory and interval.  

Having written all that, I probably like the last idea the most.

brianhw: At this level, we could later add contiguity validation.  That is, for developers we just leave it this way, and they get the data that the patterns find.  But we could add a validation flag that checks against an overall interval and returns an error if a day is found to be missing.

brianhw: Actually, to be more precise, how about:  "PathSelectionByDateIntervalTask" ?

edx/analytics/tasks/enrollments.py
brianhw: Seems to me now, for this case, all we need is a CoruseEnrollmentDataTask that takes an interval.  To calculate history, we call it with an interval.  To calculate the overwrite, we call it with the overlap_interval and with overwrite=True.  But either case it does the same thing, which is to collect all the enrollments events for a given day into a single file in that date's partition.

edx/analytics/tasks/tests/test_enrollments.py
```diff
@@ -90,16 +92,16 @@ def test_missing_user_id(self):
...
 
     def test_good_unenroll_event(self):
         line = self.create_event_log_line(event_type=DEACTIVATED)
-        expected_value = (self.timestamp, DEACTIVATED, 'honor')
+        expected_value = (self.course_id, self.user_id, self.timestamp, DEACTIVATED, 'honor')
         self.assert_single_map_output(line, self.expected_key, expected_value)
 
 
-class CourseEnrollmentTaskMapTest(InitializeLegacyKeysMixin, unittest.TestCase):
+class CourseEnrollmentTaskLegacyMapTest(InitializeLegacyKeysMixin, CourseEnrollmentTaskMapTest, unittest.TestCase):
```
brianhw, line 103: I didn't check to see if this were deliberate, but before this class redefined the previous class, and didn't provide any methods, so no mapping tests were running at all.  This now runs the tests twice, once with each key config.

brianhw: Heh, it's been broken for years now.  :)

edx/analytics/tasks/enrollments.py
HassanJaveed84: I like the idea of PathSelectionByDateIntervalTask.

edx/analytics/tasks/enrollments.py
```diff
@@ -7,32 +7,29 @@
...
 MODE_CHANGED = 'edx.course.enrollment.mode_changed'
 
 
-class CourseEnrollmentTask(EventLogSelectionMixin, MapReduceJobTask):
-    """Produce a data set that shows which days each user was enrolled in each course."""
-
-    output_root = luigi.Parameter()
-
-    enable_direct_output = True
+class CoruseEnrollmentDataTask(OverwriteOutputMixin, WarehouseMixin, EventLogSelectionMixin, MultiOutputMapReduceJobTask):
```
brianhw: nit: spelling of Coruse. : )

edx/analytics/tasks/tests/test_enrollments.py
```diff
@@ -20,7 +21,7 @@ class CourseEnrollmentTaskMapTest(MapperTestMixin, InitializeOpaqueKeysMixin, un
     Tests to verify that event log parsing by mapper works correctly.
     """
     def setUp(self):
-        self.task_class = CourseEnrollmentTask
+        self.task_class = CourseEnrollmentHistoryTask
```
brianhw: This just needs to be the data task, or one of the data tasks.   I initially tried with the existing mixin, but needed init_local to be defined, so this worked just as well.

edx/analytics/tasks/enrollments.py
```diff
@@ -66,7 +63,81 @@ def mapper(self, line):
...
+
+        yield CourseEnrollmentDataTask(
+            interval=overwrite_interval,
+            n_reduce_tasks=self.n_reduce_tasks,
+            output_root=course_enrollment_data_root_path,
+            warehouse_path=self.warehouse_path,
+            overwrite=True,
+        )
+
+        path_selection_interval = luigi.date_interval.Custom(self.interval.date_a, overwrite_interval.date_a - datetime.timedelta(days=1))
```
brianhw: I thought the end of an interval was exclusive, so that this should just be overwrite_interval.date_a.

edx/analytics/tasks/enrollments.py
```diff
@@ -66,7 +63,84 @@ def mapper(self, line):
...
+
+        course_enrollment_data_root_path = url_path_join(self.warehouse_path, 'course_enrollment_data')
+
+        yield CourseEnrollmentDataTask(
+            interval=overwrite_interval,
+            n_reduce_tasks=self.n_reduce_tasks,
+            output_root=course_enrollment_data_root_path,
+            warehouse_path=self.warehouse_path,
+            overwrite=True,
+        )
```
HassanJaveed84, line 187: MultiOutputMapReduceJobTask returns the marker as the output

``` python
    def output(self):
        marker_url = url_path_join(self.marker, str(hash(self)))
        return get_target_from_url(marker_url)
```

Which in turn gets sent as input to this task. I was thinking of somehow storing the output paths for all the keys and providing a method like 'output_targets' that would return the targets from those output paths from the MultiOutputMapReduceJobTask, which then can be used here.

HassanJaveed84: Oh wait, would that even work ?

HassanJaveed84: Actually now that I think of it, we would probably need to use UncheckedExternalURL instead of target.

edx/analytics/tasks/enrollments.py
```diff
@@ -66,7 +64,92 @@ def mapper(self, line):
...
+            self.interval.date_a,
+            overwrite_interval.date_a
+        ))
+
+        yield PathSelectionByDateIntervalTask(
+            source=[course_enrollment_data_root_path],
+            interval=path_selection_interval,
+            pattern=['.*?course_enrollment_data_(?P<date>\\d{4}-\\d{2}-\\d{2})'],
+            date_pattern='%Y-%m-%d',
+        )
```
HassanJaveed84, line 205: PathSelectionByDateIntervalTask currently expands the interval. We don't want to be doing that for this case. Does it make sense to include a boolean parameter in this task that toggles the behavior ?

HassanJaveed84:
A couple of questions regarding:
1. The output produced by MultiOutputMapReduceJobTask(which is used as input for CourseEnrollmentTask).
2. Expand interval logic.

edx/analytics/tasks/enrollments.py
brianhw: So output() of the multi would be used as the input() of the task that requires this task.  I'm wondering if what makes sense is to have complete() check the marker file's existence, but for output() to return the output_root as a target.  That seems "right" to me, but may be hugely disruptive.    The second way would be to change the definition of the input() for the job that requires this, so that it uses UncheckedExternalURL as you suggest for each of the dates that would be overwritten.  Nothing else comes to mind right now.

edx/analytics/tasks/pathutil.py
```diff
@@ -144,11 +150,12 @@ class EventLogSelectionTask(EventLogSelectionDownstreamMixin, luigi.WrapperTask)
...
-        self.interval = DateInterval(
-            self.interval.date_a - self.expand_interval,
-            self.interval.date_b + self.expand_interval
-        )
+        super(PathSelectionByDateIntervalTask, self).__init__(*args, **kwargs)
+        if self.should_expand:
+            self.interval = DateInterval(
+                self.interval.date_a - self.expand_interval,
+                self.interval.date_b + self.expand_interval
+            )
```
brianhw: Instead of defining "should_expand", does it just make sense to set expand_interval to zero when this case of PathSelectionByDateIntervalTask is instantiated, overriding the default?  And then we just test if self_expand_interval > 0 here.

edx/analytics/tasks/enrollments.py
```diff
@@ -294,6 +379,11 @@ class CourseEnrollmentTableDownstreamMixin(WarehouseMixin, EventLogSelectionDown
         'Default is today, UTC.',
     )
 
+    overwrite_n_days = luigi.IntParameter(
+        significant=False,
+        #default=3,
```
brianhw: I know it's convenient to not have to make configuration changes yet, but down the road, do we think it makes sense to put this is a config file?   One possibility is to have it read from config now, and since it's not there, it will fail.  Note that it's no different than the current situation, except it would allow a value to be put in config for running tests.  This is the one thing that really interferes with just running unit and acceptance tests.

edx/analytics/tasks/enrollments.py
```diff
@@ -66,7 +64,93 @@ def mapper(self, line):
...
+        'The default value is [\'*\'].',
+    )
+
+    manifest = luigi.Parameter(
+        default=None,
+        description='A URL to a file location that can store the complete set of input files. '
+        'A manifest file is required by Hadoop if it hits the OS limit on the length '
+        'of the command to run when launching the job using the Hadoop CLI. '
+        'This file will be written to if it doesn\'t exist, and read from if it already does.',
+    )
```
brianhw, line 151: Is this used anymore?

edx/analytics/tasks/enrollments.py
HassanJaveed84: I agree, changing the output() seems the correct thing to do, but it will be disruptive.

edx/analytics/tasks/pathutil.py
HassanJaveed84: That is what I initially thought of doing, but what I could not figure out was how to pass a zero value for the expand_interval through command line. I don't think we need to test if expand_interval > 0 as the interval value would remain the same in case it is zero.

brianhw:
Not sure I understand the split into requires_hadoop and requires_local.  Luigi will still combine these to determine the dependencies, and offhand that seems a little off.  Can you change your current requires_hadoop() back to requires(), and then just change input_hadoop() to do what your requires_local() does (wrapped in a getpaths call, like the default input_hadoop())?

HassanJaveed84:
Sorry, that was like a test change. We already override input_hadoop() in MapReduceJobTask to create the manifest if necessary.

I was thinking of changing such that:

``` python
def requires_local(self):
    return CourseEnrollmentDataTask(
            interval=overwrite_interval,
            output_root=course_enrollment_data_root_path,
            overwrite=True,
        )

def requires_hadoop(self):
    path_task =  PathSelectionByDateIntervalTask(
            source=[course_enrollment_data_root_path],
            interval=path_selection_interval,
            pattern=['.*?course_enrollment_data_(?P<date>\\d{4}-\\d{2}-\\d{2})'],
            expand_interval=datetime.timedelta(),
            date_pattern='%Y-%m-%d',
    )
    return {
        'path_task': path_task,
        'data_tasks': self.requires_local().output_targets() 
    }
```

so that way when luigi builds the dependency graph, it will schedule CourseEnrollmentDataTask from requires_local() and PathSelectionByDateIntervalTask from requires_hadoop(), and would only use PathSelectionByDateIntervalTask plus the UncheckedExternalUrls as hadoop input.

HassanJaveed84:
@mulby @brianhw please review, especially the CourseEnrollmentTask. There is a bunch of logic which defines the requirements and hadoop input. I can create a new PR if it makes sense.

brianhw:
I see.  So it would schedule them all:  the DataTask, the outputs of the path selection task, and the UncheckedExternalUrls as well (i.e. the union of _local and _hadoop).  But we don't mind throwing the external URLs to the scheduler because they should be cheap to evaluate.

edx/analytics/tasks/enrollments.py
```diff
@@ -66,7 +69,105 @@ def mapper(self, line):
...
+        MultiOutputMapReduceJobTask returns marker as output.
+        This method returns the external tasks which can then be used as input in other jobs.
+        """
+
+        output_tasks = []
+        for date in self.interval:
+            url = self.output_path_for_key(date.isoformat())
+            output_tasks.append(UncheckedExternalURL(url))
+
+        return output_tasks
```
HassanJaveed84: When this method is used as hadoop input, things can break if there is missing data for a date as we use an UncheckedExternalURL.

edx/analytics/tasks/tests/acceptance/test_enrollments.py
```diff
@@ -24,6 +24,7 @@ def test_demographic_trends(self):
             'ImportEnrollmentsIntoMysql',
             '--interval', '2014-08-01-2014-08-06',
             '--n-reduce-tasks', str(self.NUM_REDUCERS),
+            '--overwrite-n-days', '5',
```
HassanJaveed84, line 27: As there would be no pre-computed history for the acceptance test, we need to overwrite all days within the interval.

edx/analytics/tasks/enrollments.py
```diff
@@ -66,7 +73,105 @@ def mapper(self, line):
...
+    def output_path_for_key(self, key):
+        date_string = key
+        return url_path_join(
+            self.hive_partition_path('course_enrollment_data', date_string),
+            'course_enrollment_data_{date}'.format(
+                date=date_string,
+            ),
+        )
+
+    def output_tasks(self):
```
mulby: nit: We often talk about targets when talking about outputs, maybe name this something like "downstream_input_tasks"?

edx/analytics/tasks/enrollments.py
mulby: We may want to note this with a comment in the code as well.

edx/analytics/tasks/enrollments.py
```diff
@@ -66,7 +73,105 @@ def mapper(self, line):
...
+
+class CourseEnrollmentTask(WarehouseMixin, MapReduceJobTask):
+    """Produce a data set that shows which days each user was enrolled in each course."""
+
+    output_root = luigi.Parameter()
+
+    enable_direct_output = True
+
+    interval = luigi.DateIntervalParameter(
+        description='The range of dates to export logs for.',
```
mulby: nit: this description should be removed or updated

edx/analytics/tasks/enrollments.py
```diff
@@ -66,7 +73,105 @@ def mapper(self, line):
...
+
+    def multi_output_reducer(self, _date_string, values, output_file):
+        for value in values:
+            output_file.write('\t'.join([str(field) for field in value]))
+            output_file.write('\n')
+
+    def output_path_for_key(self, key):
+        date_string = key
+        return url_path_join(
+            self.hive_partition_path('course_enrollment_data', date_string),
```
mulby: nit: maybe name this "enrollment_events"

edx/analytics/tasks/enrollments.py
```diff
@@ -368,6 +479,7 @@ def required_table_tasks(self):
                 interval=self.interval,
                 pattern=self.pattern,
```
mulby, line 492: We should probably remove source and pattern from these as well since they aren't passed through to the base tasks anymore: https://github.com/edx/edx-analytics-pipeline/pull/287/files?diff=unified#diff-04eea63db0b65ed376d0067fbd1eae0cL332

edx/analytics/tasks/pathutil.py
```diff
@@ -132,8 +132,12 @@ class EventLogSelectionDownstreamMixin(object):
         'the right file in order for them to be processed.',
     )
 
+    date_pattern = luigi.Parameter(
+        default='%Y%m%d',
```
mulby, line 136: nit: we should add a description for this parameter

edx/analytics/tasks/tests/test_enrollments.py
```diff
@@ -109,7 +111,7 @@ class CourseEnrollmentTaskReducerTest(ReducerTestMixin, unittest.TestCase):
     """
     def setUp(self):
         self.task_class = CourseEnrollmentTask
-        super(CourseEnrollmentTaskReducerTest, self).setUp()
+        # super(CourseEnrollmentTaskReducerTest, self).setUp()
```
mulby: This should probably be either uncommented out, or removed.

mulby (commented):
Looking good so far, only a couple of questions.

edx/analytics/tasks/enrollments.py
brianhw: Or "course_enrollment_events".   And actually changing the filename below similarly.

edx/analytics/tasks/enrollments.py
```diff
@@ -5,34 +5,41 @@
...
+class CourseEnrollmentDataTask(
+        OverwriteOutputMixin,
+        WarehouseMixin,
+        EventLogSelectionMixin,
+        MultiOutputMapReduceJobTask):
+    """
+    Task to extract enrollment events from eventlogs over a given interval.
+    This would produce a different output file for each day within the interval
+    containing that day's enrollment events only.
+    """
```
brianhw, line 35: nit: The description is pretty clear about enrollment events here.  So I'm thinking that the task name should be CourseEnrollmentEventsTask, to better reflect what it produces rather than how it serves as an input to another task.

edx/analytics/tasks/enrollments.py
brianhw: I like the suggestion.  I initially didn't, but I think that having "output" in the name gets confusing relative to output() itself.  (Maybe that's what Gabe means by "talk about targets"?)   So this seems descriptive enough.

edx/analytics/tasks/enrollments.py
brianhw: Is that really acceptable?  We need for there to be an enrollment event for every day in the overlap period?  One alternative is to wrap the reducer() to actually check all urls that this would output, and create empty versions of any missing files.

edx/analytics/tasks/enrollments.py
```diff
@@ -66,7 +73,105 @@ def mapper(self, line):
...
+
+        output_tasks = []
+        for date in self.interval:
+            url = self.output_path_for_key(date.isoformat())
+            output_tasks.append(UncheckedExternalURL(url))
+
+        return output_tasks
+
+    def run(self):
+        self.remove_output_on_overwrite()
```
brianhw, line 111: At the moment, this only removes the marker file.   Do we also want it to remove the files pointed to by output_tasks()?  (If so, then define a remove_output_on_overwrite() method on this class that could call the wrapped version and do the extra cleanup.)

edx/analytics/tasks/enrollments.py
```diff
@@ -294,6 +399,12 @@ class CourseEnrollmentTableDownstreamMixin(WarehouseMixin, EventLogSelectionDown
         'Default is today, UTC.',
     )
 
+    overwrite_n_days = luigi.IntParameter(
+        config_path={'section': 'enrollments', 'name': 'overwrite_n_days'},
+        significant=False,
+        default=3,
```
brianhw: I don't think this should have a hard-coded default.  I think it's preferable for it to require an explicit value either on the command line or config file.  For tests, it could use the test.cfg value.

brianhw: Also, should this have a description here?

edx/analytics/tasks/enrollments.py
brianhw: This description also appears in the DownloadMixin.  Does it make sense for this class to use the DownloadMixin to define these parameters (including output_root and overwrite_n_days) here as well?  I'm assuming that the **init** logic will still do the right thing in that case, with the super() getting the interval defined first.

edx/analytics/tasks/enrollments.py
```diff
@@ -66,7 +73,105 @@ def mapper(self, line):
...
+
+        path_selection_interval = DateIntervalParameter().parse('{}-{}'.format(
+            self.interval.date_a,
+            self.overwrite_from_date,
+        ))
+
+        path_selection_task = PathSelectionByDateIntervalTask(
+            source=[self.course_enrollment_data_root_path],
+            interval=path_selection_interval,
+            pattern=['.*?course_enrollment_data_(?P<date>\\d{4}-\\d{2}-\\d{2})'],
```
brianhw: ... and if the filename changes, this pattern would need to change.  

nit: Not sure if it makes sense to move the pattern to the CourseEnrollmentDataTask class, to make the dependency more obvious....

edx/analytics/tasks/tests/test_enrollments.py
```diff
@@ -51,7 +52,8 @@ def setUp(self):
         }
         self.default_event_template = 'enrollment_event'
 
-        self.expected_key = (self.course_id, self.user_id)
+        # self.expected_key = (self.course_id, self.user_id)
```
brianhw: Remove the commented code now.

edx/analytics/tasks/tests/test_enrollments.py
brianhw: Removed.  Calling it will fail because it can't instantiate the task without an extra argument.

brianhw (commented):
LGTM.  Mostly nits, though a more substantial question about missing files.

edx/analytics/tasks/enrollments.py
HassanJaveed84: CourseEnrollmentTask would actually fail if any of the urls given as input do not exist.
For example if we run the whole workflow with an overwrite interval that goes back to 2013-11-01. The job would actually fail as the multi output mapreduce job(CourseEnrollmentDataTask) would not create any output for 2013-11-01-2013-11-10, but this method would still return and UncheckedExternalURL for those dates.

mulby: Ah, so we need to manually create empty folders for those dates that we expect to be empty? I think I actually like this best of all... if we end up with missing data the job will fail!

edx/analytics/tasks/enrollments.py
HassanJaveed84: I'm not sure about output_root being part of DownstreamMixin. ImportEnrollmentsIntoMysql uses DownstreamMixin, so unless we provide a default value for the output_root, we would need to pass in output_root to the workflow. But in CourseEnrollmentTableTask we use the partition location as the output_root for CourseEnrollmentTask:

``` python
    def requires(self):
        return CourseEnrollmentTask(
            output_root=self.partition_location,
            interval=self.interval,
            overwrite_n_days=self.overwrite_n_days,
        )
```

edx/analytics/tasks/enrollments.py
HassanJaveed84: Do we need to create empty folders ? The behavior would be the same regardless.

edx/analytics/tasks/enrollments.py
HassanJaveed84: Changed to course_enrollment_events. Also changed the class name to 'CourseEnrollmentEventsTask'

edx/analytics/tasks/enrollments.py
HassanJaveed84: Done.

edx/analytics/tasks/enrollments.py
HassanJaveed84: They are now passed through to CourseEnrollmentEventsTask.

edx/analytics/tasks/pathutil.py
HassanJaveed84: Done. Please review.

edx/analytics/tasks/tests/test_enrollments.py
HassanJaveed84: Done.

edx/analytics/tasks/enrollments.py
HassanJaveed84: Done.

edx/analytics/tasks/enrollments.py
HassanJaveed84: I am not sure if that is needed, I believe the task would overwrite any existing files in the output.
The reducer for MultiOutputMapReduceJobTask opens the target in 'w' mode, which should truncate any existing file.

edx/analytics/tasks/enrollments.py
HassanJaveed84: Removed the default, added description.

edx/analytics/tasks/enrollments.py
HassanJaveed84: Not sure how would that look like.  How should we specify this regex in CourseEnrollmentDataTask ?

edx/analytics/tasks/tests/test_enrollments.py
HassanJaveed84: Done.

HassanJaveed84:
This is ready for another review. I've addressed most of the comments, only a few clarifications. 
If the path stuff looks good, I can start a history job that computes the daily partitions, outputting to the prod warehouse.

edx/analytics/tasks/pathutil.py
```diff
@@ -132,8 +132,14 @@ class EventLogSelectionDownstreamMixin(object):
...
+        description='The format of the date as it appears in the source file name. Note that this correlates with the '
+        'named capture group for date in the pattern parameter.',
+    )
+
 
-class EventLogSelectionTask(EventLogSelectionDownstreamMixin, luigi.WrapperTask):
+class PathSelectionByDateIntervalTask(EventLogSelectionDownstreamMixin, luigi.WrapperTask):
```
HassanJaveed84, line 143: Should EventLogSelectionDownstreamMixin be renamed to PathSelectionDownstreamMixin ?
We use EventLogSelectionDownstreamMixin through the codebase, so would need to update accordingly.

edx/analytics/tasks/enrollments.py
brianhw: It's a nit, but I would just return the string value as it appears above, as the value of CourseEnrollmentDataTask.DATE_FROM_FILEPATH_REGEX or something similar.

edx/analytics/tasks/enrollments.py
```diff
@@ -66,7 +72,147 @@ def mapper(self, line):
...
+            pattern=self.pattern,
+            n_reduce_tasks=self.n_reduce_tasks,
+            output_root=self.course_enrollment_events_root,
+            warehouse_path=self.warehouse_path,
+            overwrite=True,
+        )
+
+    def requires_hadoop(self):
+        # We want to pass in the historical data as well as the output of CourseEnrollmentEventsTask to the hadoop job.
+        # CourseEnrollmentEventsTask returns the marker as output, that is why we need custom logic to pass the output
```
brianhw: nit:  "as output, that is why we need" => "as output, so we need"....

edx/analytics/tasks/enrollments.py
```diff
@@ -66,7 +72,147 @@ def mapper(self, line):
...
+
+        for date in self.interval:
+            url = self.output_path_for_key(date.isoformat())
+            target = get_target_from_url(url)
+            if not target.exists():
+                target.open("w").close() # touch the file
+
+
+class CourseEnrollmentTableDownstreamMixin(WarehouseMixin, EventLogSelectionDownstreamMixin, MapReduceJobTaskMixin):
+    """All parameters needed to run the CourseEnrollmentTableTask task."""
```
brianhw: Actually this is now for the CourseEnrollmentTask, not (just) the CourseEnrollmentTableTask.  Probably don't need "Table" in the mixin name either.

edx/analytics/tasks/enrollments.py
```diff
@@ -66,7 +72,147 @@ def mapper(self, line):
...
+
+    # Make the interval be optional:
+    interval = luigi.DateIntervalParameter(
+        default=None,
+        description='The range of dates to export logs for. '
```
brianhw: nit: I guess we kinda are exporting logs here, but I am guessing this is cut-and-pasted from event export.

edx/analytics/tasks/enrollments.py
```diff
@@ -66,7 +72,147 @@ def mapper(self, line):
...
+        'If not specified, `interval_start` and `interval_end` are used to construct the `interval`.',
+    )
+
+    # Define optional parameters, to be used if 'interval' is not defined.
+    interval_start = luigi.DateParameter(
+        config_path={'section': 'enrollments', 'name': 'interval_start'},
+        significant=False,
+        description='The start date to export logs for.  Ignored if `interval` is provided.',
```
brianhw: nit: same about "export logs"....

edx/analytics/tasks/pathutil.py
brianhw: I'm actually okay with EventLogSelectionDownstreamMixin being used in most places that it is.  What we might want to do is define PathSelectionDownstreamMixin that overrides some parameter definitions for the general case.  Or vice versa -- that EventLogSelectionDownstreamMixin provides the parameter definitions that use different config categories and different default values (e.g. date_pattern).

edx/analytics/tasks/tests/acceptance/test_enrollments.py
brianhw: Still true?

brianhw (commented):
Small nits in a quick pass.

edx/analytics/tasks/tests/acceptance/test_enrollments.py
HassanJaveed84: Yes, as CourseEnrollmentTask uses PathSelectionByDateIntervalTask to select files prior to dates for which the files are overwritten. In the acceptance test we would have any precomputed partitions.

edx/analytics/tasks/enrollments.py
HassanJaveed84: Done.

edx/analytics/tasks/enrollments.py
HassanJaveed84: Done.

edx/analytics/tasks/enrollments.py
HassanJaveed84: Changed.

edx/analytics/tasks/enrollments.py
HassanJaveed84: Done.

edx/analytics/tasks/enrollments.py
HassanJaveed84: Done.

HassanJaveed84:
I've added some logic to generate empty output files for missing dates. Please review.

edx/analytics/tasks/enrollments.py
```diff
@@ -66,7 +74,147 @@ def mapper(self, line):
...
+                date=date_string,
+            ),
+        )
+
+    def downstream_input_tasks(self):
+        """
+        MultiOutputMapReduceJobTask returns marker as output.
+        This method returns the external tasks which can then be used as input in other jobs.
+        Note that this method does not verify the existence of the underlying urls, it assumes that
+        there is output data for all the dates appearning within the interval. Any MapReduce job
```
brianhw: nit:  appearning => appearing.  Also put a period after "underlying urls", and capitalize the following "It".   And I'd suggest changing "output data" to "an output file for every date within the interval".

edx/analytics/tasks/enrollments.py
```diff
@@ -66,7 +74,147 @@ def mapper(self, line):
...
+            url = self.output_path_for_key(date.isoformat())
+            tasks.append(UncheckedExternalURL(url))
+
+        return tasks
+
+    def run(self):
+        self.remove_output_on_overwrite()
+        super(CourseEnrollmentEventsTask, self).run()
+
+        for date in self.interval:
```
brianhw, line 116: I would add a comment here that matches the docstring for downstream_input_tasks(), saying that it makes sure that a file exists for each date in the interval, because downstream tasks will require that they exist.

edx/analytics/tasks/enrollments.py
```diff
@@ -66,7 +74,147 @@ def mapper(self, line):
...
+            url = self.output_path_for_key(date.isoformat())
+            target = get_target_from_url(url)
+            if not target.exists():
+                target.open("w").close()  # touch the file
+
+
+class CourseEnrollmentDownstreamMixin(WarehouseMixin, EventLogSelectionDownstreamMixin, MapReduceJobTaskMixin):
+    """All parameters needed to run the CourseEnrollmentTableTask task."""
```
brianhw: nit:  CourseEnrollmentTask.

edx/analytics/tasks/enrollments.py
```diff
@@ -66,7 +74,147 @@ def mapper(self, line):
...
+        significant=False,
+        description='The end date to extract enrollments events for.  Ignored if `interval` is provided. '
+        'Default is today, UTC.',
+    )
+
+    overwrite_n_days = luigi.IntParameter(
+        config_path={'section': 'enrollments', 'name': 'overwrite_n_days'},
+        significant=False,
+        description='This parameter is passed down to CourseEnrollmentTask which will overwrite course enrollment '
+                    ' events for the most recent n days.'
```
brianhw, line 150: nit:  "This parameter is used by CourseEnrollmentTask which will overwrite ..."

edx/analytics/tasks/enrollments.py
```diff
@@ -21,18 +23,24 @@
...
+        EventLogSelectionMixin,
+        MultiOutputMapReduceJobTask):
+    """
+    Task to extract enrollment events from eventlogs over a given interval.
+    This would produce a different output file for each day within the interval
+    containing that day's enrollment events only.
+    """
 
-    enable_direct_output = True
+    FILEPATH_PATTERN = '.*?course_enrollment_data_(?P<date>\\d{4}-\\d{2}-\\d{2})'
```
brianhw: nit: add a brief comment that this should match output files defined by output_path_for_key().

edx/analytics/tasks/pathutil.py
brianhw: I take that back, at this point.  It turns out that PathSelectionDownstreamMixin is not needed yet, so no point in defining it yet.

edx/analytics/tasks/pathutil.py
```diff
@@ -132,8 +132,14 @@ class EventLogSelectionDownstreamMixin(object):
         'the right file in order for them to be processed.',
     )
 
+    date_pattern = luigi.Parameter(
+        default='%Y%m%d',
+        description='The format of the date as it appears in the source file name. Note that this correlates with the '
+        'named capture group for date in the pattern parameter.',
```
brianhw: I might just slip in somewhere here that it's particularly the "source event file name", that this is really intended for selecting event logs.

edx/analytics/tasks/tests/acceptance/test_enrollments.py
brianhw: It should be okay for there to not be any precomputed partitions.  The key is that the overwrite must extend back to where the data is.  I think it's 8/1, so we need 5 to reach back from 8/6.  But it should still work if we were to run with an interval starting 2014-07-01, or whatever.

brianhw (approved):
Looks good.  Just nits, then okay.  👍

mulby (approved):
Looks great!

edx/analytics/tasks/tests/acceptance/test_enrollments.py
HassanJaveed84: Yup, you are right. Do we need to make any changes here ?

mulby:
It looks like this PR broke the module engagement workflow acceptance test, see this failure:

http://jenkins-ci.analytics.edx.org/job/edx-analytics-pipeline-acceptance-auto-pr/707/testReport/junit/edx.analytics.tasks.tests.acceptance.test_module_engagement/ModuleEngagementAcceptanceTest/test_roster_generation/

```
  File "/var/lib/analytics-tasks/0d7fa8e3bccb9a8f027f477b5dc74570/venv/bin/launch-task", line 10, in <module>
    sys.exit(main())
  File "/var/lib/analytics-tasks/0d7fa8e3bccb9a8f027f477b5dc74570/venv/local/lib/python2.7/site-packages/edx/analytics/tasks/launchers/local.py", line 67, in main
    luigi.run()
  File "/var/lib/analytics-tasks/0d7fa8e3bccb9a8f027f477b5dc74570/venv/local/lib/python2.7/site-packages/luigi/interface.py", line 451, in run
    tasks = interface.parse(cmdline_args, main_task_cls=main_task_cls)
  File "/var/lib/analytics-tasks/0d7fa8e3bccb9a8f027f477b5dc74570/venv/local/lib/python2.7/site-packages/luigi/interface.py", line 309, in parse
    return self.parse_task(cmdline_args, main_task_cls)
  File "/var/lib/analytics-tasks/0d7fa8e3bccb9a8f027f477b5dc74570/venv/local/lib/python2.7/site-packages/luigi/interface.py", line 304, in parse_task
    task = task_cls.from_str_params(params, Register.get_global_params())
  File "/var/lib/analytics-tasks/0d7fa8e3bccb9a8f027f477b5dc74570/venv/local/lib/python2.7/site-packages/luigi/task.py", line 378, in from_str_params
    value = param.parse_from_input(param_name, params_str[param_name])
  File "/var/lib/analytics-tasks/0d7fa8e3bccb9a8f027f477b5dc74570/venv/local/lib/python2.7/site-packages/luigi/parameter.py", line 252, in parse_from_input
    (param_name, "--" + param_name.replace('_', '-')))
luigi.parameter.MissingParameterException: No value for 'overwrite_n_days' (--overwrite-n-days) submitted and no default value has been assigned.
```

It runs the enrollment workflow so I suspect it's not being triggered properly.

The student engagement workflow has a similar issue http://jenkins-ci.analytics.edx.org/job/edx-analytics-pipeline-acceptance-auto-pr/707/testReport/junit/edx.analytics.tasks.tests.acceptance.test_student_engagement/StudentEngagementAcceptanceTest/test_student_engagement/:

```
2016-09-23 05:32:41,918 ERROR 25538 [luigi-interface] worker.py:173 - Luigi unexpected framework error while scheduling StudentEngagementCsvFileTask(output_root=s3://edx-analytics-test-output/automation/0d7fa8e3bccb9a8f027f477b5dc74570/StudentEngagementAcceptanceTest/out/daily, source=('s3://edx-analytics-test-output/automation/0d7fa8e3bccb9a8f027f477b5dc74570/StudentEngagementAcceptanceTest/src',), interval=2015-04-06-2015-04-20, expand_interval=2 days, 0:00:00, pattern=('.*tracking.log-(?P<date>\\d{8}).*\\.gz', '.*tracking.notalog-(?P<date>\\d{8}).*\\.gz'), date_pattern=%Y%m%d, warehouse_path=s3://edx-analytics-test-output/automation/0d7fa8e3bccb9a8f027f477b5dc74570/StudentEngagementAcceptanceTest/warehouse, interval_type=daily)
  File "/var/lib/analytics-tasks/0d7fa8e3bccb9a8f027f477b5dc74570/venv/local/lib/python2.7/site-packages/luigi/worker.py", line 196, in add
    for next in self._add(current):
  File "/var/lib/analytics-tasks/0d7fa8e3bccb9a8f027f477b5dc74570/venv/local/lib/python2.7/site-packages/luigi/worker.py", line 249, in _add
    deps = task.deps()
  File "/var/lib/analytics-tasks/0d7fa8e3bccb9a8f027f477b5dc74570/venv/local/lib/python2.7/site-packages/luigi/hadoop.py", line 629, in deps
    return luigi.task.flatten(self.requires_hadoop()) + luigi.task.flatten(self.requires_local())
  File "/var/lib/analytics-tasks/0d7fa8e3bccb9a8f027f477b5dc74570/venv/local/lib/python2.7/site-packages/luigi/task.py", line 586, in flatten
    for result in struct:
  File "/var/lib/analytics-tasks/0d7fa8e3bccb9a8f027f477b5dc74570/venv/local/lib/python2.7/site-packages/edx/analytics/tasks/student_engagement.py", line 439, in requires
    CourseEnrollmentTableTask(**kwargs_for_enrollment),
  File "/var/lib/analytics-tasks/0d7fa8e3bccb9a8f027f477b5dc74570/venv/local/lib/python2.7/site-packages/luigi/task.py", line 100, in __call__
    param_values = cls.get_param_values(params, args, kwargs)
  File "/var/lib/analytics-tasks/0d7fa8e3bccb9a8f027f477b5dc74570/venv/local/lib/python2.7/site-packages/luigi/task.py", line 315, in get_param_values
    raise parameter.MissingParameterException("%s: requires the '%s' parameter to be set" % (exc_desc, param_name))
MissingParameterException: CourseEnrollmentTableTask... overwrite_n_days
```

HassanJaveed84:
The issue is that CourseEnrollmentTableTask now requires an overwrite_n_days parameter which can be read from config if present.

```
    overwrite_n_days = luigi.IntParameter(
        config_path={'section': 'enrollments', 'name': 'overwrite_n_days'},
        significant=False,
        description='This parameter is used by CourseEnrollmentTask which will overwrite course enrollment '
                    ' events for the most recent n days.'
    )
```

We should set a default in test.cfg but I'm not sure what is an appropriate value. Two weeks ? so it would be 14.

mulby:
Sounds reasonable to me!

HassanJaveed84:
Should I merge this in ?

mulby:
@HassanJaveed84 I forgot that this test is still failing on this branch: http://jenkins-ci.analytics.edx.org/view/pull%20requests/job/edx-analytics-pipeline-acceptance-auto-pr/708/testReport/edx.analytics.tasks.tests.acceptance.test_student_engagement/StudentEngagementAcceptanceTest/test_student_engagement/

HassanJaveed84:
@mulby this is due to incorrect value of overwrite_n_days being passed down from the test config. It was set to 14. Student engagement acceptance test uses an interval or 2 weeks but relies on course_enrollment data from all time. The interval is 2015-04-06-2015-04-20 but there are enrollments events in the input log file on dates like 2014-08-01, but those never got captured as the overwrite logic only extracted 14 days of enrollment events(due to overwrite_n_days being 14). Thoughts on what should be done here ?

mulby:
@HassanJaveed84 I'm open to ideas... it might make sense to have student engagement depend on enrollment via ExternalCourseEnrollmentTableTask like we do in module_engagement instead of the direct dependency. We aren't actually running these tasks in production, so we could also remove them maybe?

HassanJaveed84:
Fixed by updating the input tracking log to student_engagement acceptance test. The enrollments events now fall within the two weeks of overwrite interval.

mulby:
Perfect, squash, merge and release when ready.

PR #289, opened 2016-09-09
fix a couple of acceptance testing annoyances

mulby:
@mattdrayer @bderusha @douglashall 

My demo today was running of this branch, it fixes some issues with DISABLE_RESET_STATE. I'll try to get this merged in soon, but in the meantime I recommend including this commit in any branches you work on.

Reviewers: @brianhw @HassanJaveed84

### DIFF
--- a/edx/analytics/tasks/pathutil.py
+++ b/edx/analytics/tasks/pathutil.py
@@ -67,9 +67,10 @@ class PathSetTask(luigi.Task):
                     source = url_path_join(src, path)
                     yield ExternalURL(source)
             elif src.startswith('hdfs'):
-                # TODO: implement exclude_zero_length to match S3 case.
-                for source in luigi.hdfs.listdir(src, recursive=True):
-                    if any(fnmatch.fnmatch(source, include_val) for include_val in self.include):
+                for source, size in luigi.hdfs.listdir(src, recursive=True, include_size=True):
+                    if not self.include_zero_length and size == 0:
+                        continue
+                    elif any(fnmatch.fnmatch(source, include_val) for include_val in self.include):
                         yield ExternalURL(source)
             else:
                 # Apply the include patterns to the relative path below the src directory.
@@ -292,3 +293,17 @@ class EventLogSelectionMixin(EventLogSelectionDownstreamMixin):
         except KeyError:
             self.incr_counter('Event', 'Missing Time Field', 1)
             return None
+
+    def get_map_input_file(self):
+        """Get the name of the input file from Hadoop."""
+        # Hadoop sets an environment variable with the full URL of the input file. This url will be something like:
+        # s3://bucket/root/host1/tracking.log.gz. In this example, assume self.source is "s3://bucket/root".
+        try:
+            return os.environ['mapreduce_map_input_file']
+        except KeyError:
+            try:
+                # Older versions of Hadoop support a deprecated key, so also try that.
+                return os.environ['map_input_file']
+            except KeyError:
+                log.warn('mapreduce_map_input_file not defined in os.environ, unable to determine input file path')
+                return ''
--- a/edx/analytics/tasks/tests/acceptance/test_financial_reports.py
+++ b/edx/analytics/tasks/tests/acceptance/test_financial_reports.py
@@ -10,11 +10,9 @@ import luigi
 import pandas
 from pandas.util.testing import assert_frame_equal, assert_series_equal
 
-from edx.analytics.tasks.tests import unittest
-from edx.analytics.tasks.tests.acceptance import AcceptanceTestCase, when_vertica_available, when_vertica_not_available, get_jenkins_safe_url
-from edx.analytics.tasks.url import url_path_join, get_target_from_url
+from edx.analytics.tasks.tests.acceptance import AcceptanceTestCase, when_vertica_available, when_vertica_not_available
+from edx.analytics.tasks.url import url_path_join
 from edx.analytics.tasks.reports.reconcile import LoadInternalReportingOrderTransactionsToWarehouse
-from edx.analytics.tasks.pathutil import PathSetTask
 
 log = logging.getLogger(__name__)
 
@@ -26,6 +24,9 @@ class FinancialReportsAcceptanceTest(AcceptanceTestCase):
 
     def setUp(self):
         super(FinancialReportsAcceptanceTest, self).setUp()
+
+        if not self.should_reset_state:
+            return
 
         for input_file_name in ('paypal.tsv', 'cybersource_test.tsv'):
             src = url_path_join(self.data_dir, 'input', input_file_name)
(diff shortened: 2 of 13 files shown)


edx/analytics/tasks/tests/acceptance/test_financial_reports.py
```diff
@@ -100,7 +104,8 @@ def test_end_to_end_without_vertica(self):
         output_targets = PathSetTask([output_root], ['*']).output()
         raw_output = ""
         for output_target in output_targets:
-            output_target = get_target_from_url(get_jenkins_safe_url(output_target.path))
+            if isinstance(output_target, S3HdfsTarget):
```
brianhw: Is this the appropriate test?  Or is it enough instead to check that output_target has a "path" attribute?

mulby: The get_jenkins_safe_url method is used for converting S3 targets to S3+https targets. HdfsTargets also have the path property but it is simply `/foo/bar`. `get_target_from_url` is looking for it to start with `hdfs://` otherwise it thinks it's a local file.

Basically, this logic for fixing S3 urls is kicking in on HDFS targets and accidentally converting them to local file targets.

brianhw: That's all pretty confusing.  Is there some way to rationalize this by modifying get_target_from_url or get_jenkins_safe_url, or should this just be commented here?

edx/analytics/tasks/tests/acceptance/test_financial_reports.py
```diff
@@ -27,6 +28,9 @@ class FinancialReportsAcceptanceTest(AcceptanceTestCase):
     def setUp(self):
         super(FinancialReportsAcceptanceTest, self).setUp()
 
+        if not self.should_reset_state:
+            return
```
brianhw, line 29: I'm assuming that this is just a speed-up.  And that we could add this to any setUp() override that seems to do some form of significant work.  But that this is the only one we really care about right now.  (Someday if we can do timing analysis of acceptance tests to see what the breakdown really is....)

mulby: Unfortunately it's more than a speed-up. This code will attempt to reload tables that already exist without clearing them out first. This causes all kinds of conflicts. It also tries to create new files in HDFS (which fails since they already exist). Given that the intent is to not reset the state, however, it seemed appropriate to just skip all of that.

edx/analytics/tasks/tests/acceptance/test_financial_reports.py
mulby: I agree, I may need to think a bit more about how to solve this more elegantly. At a minimum I'll be sure to put in a comment.

edx/analytics/tasks/tests/acceptance/test_financial_reports.py
brianhw: There are lots of tests that upload files.  Are those likely to fail as well?   If that's the case, one possibility is to put the test into the upload_file/upload_tracking_log methods.

edx/analytics/tasks/tests/acceptance/test_financial_reports.py
brianhw: One quick observation.  We always call get_target_from_url() on the output of get_jenkins_safe_url().  Maybe all this logic can be encapsulated in a single method.  Especially if the same HDFS logic should be used in those other cases:  removing the root test directory, and checking geolocation data.

And this is an issue for acceptance tests because it's executed on the host instead of a remote cluster?  I'm wondering if get_local_url would be a clearer name than get_jenkins_safe_url, as it sounds like it's really separate from Jenkins and more about S3.

mulby: I like that idea a lot.

edx/analytics/tasks/tests/acceptance/test_financial_reports.py
mulby: I also like this idea.

brianhw:
See PR #295 for modifications to this branch based on comments here.

brianhw:
Merged changes in and rebased.  I'll now run acceptance tests.

brianhw:
Improve acceptance test framework.
- Refactor handling of paths with s3 schemas.
- Factor out get_targets_from_remote_path.
- Add exclude-zero-length support for hdfs.
- Remove s3 limitation from tests for event export, answer dist, student engagement, and enrollment validation.
- Event obfuscation requires geolocation, so flag it properly.
- For event export, use  mapreduce_map_input_file, since map_input_file is deprecated.

Acceptance tests pass both locally and on Jenkins.

@mulby @HassanJaveed84

edx/analytics/tasks/pathutil.py
```diff
@@ -292,3 +293,17 @@ def get_event_time(self, event):
         except KeyError:
             self.incr_counter('Event', 'Missing Time Field', 1)
             return None
+
+    def get_map_input_file(self):
+        """Get the name of the input file from Hadoop."""
+        # Hadoop sets an environment variable with the full URL of the input file. This url will be something like:
+        # s3://bucket/root/host1/tracking.log.gz. In this example, assume self.source is "s3://bucket/root".
+        try:
+            return os.environ['mapreduce_map_input_file']
```
mulby, line 302: I assume this is set in Hadoop 2.X but not in 1.X?

If so, it would be nice to have that info in the comment above!

edx/analytics/tasks/tests/acceptance/test_financial_reports.py
mulby: Did you decide not to move this into the common methods? I'm OK with this for now... we could always add it in later if we find we are working around it a bunch...

mulby (commented):
This generally looks good to me, however, I'm apparently not allowed to approve my own PR...

:+1: 

We can just merge this once @HassanJaveed84 approves it as well.

mulby:
Oh, one last thought, can we remove the code for the "when_s3_available" decorator?

brianhw:
I can remove the use of the when_s3_available() decorator, because the two remaining places are also protected by relevant decorators that prevent those tests from being run anyway  (because either the exporter or vertica is also required).  But is there any utility to the decorator itself?

edx/analytics/tasks/pathutil.py
brianhw: It seems to be more complicated than that, in that in EMR we are running an even more recent version of Hadoop than on analyticstack, and yet it seemed to support the old value.

brianhw: Actually the comment (such as it is) is in the code right below.  I left it vague because I didn't know when the value changed.

brianhw:
@HassanJaveed84 Does this look okay?

HassanJaveed84:
Looks great! 👍

HassanJaveed84 (approved):

mulby:
:tada:

PR #291, opened 2016-09-13
Revert "PR review: updates module engagement acceptance tests to add new fields"

mulby:
This reverts commit bf5743df468109c9d3a31eaedeb4c05ea370711f.

@pomegranited - I discovered in our production data that some of these fields contain separator characters ("\t" or "\n"). I've been thinking a bit about what to do in such scenarios, considering the following options:
1. We use different separator characters and tell Hive about it (Hive uses 0x01 as the field separator by default... we could do the same and we do for our sqoop exports).
2. We make record "smarter", having it escape separator characters it finds in the data that is being stored in fields (and subsequently unescapes it when it reads it back). The downside of this approach is that Hive processing will see escaped data.

In theory we could use both approaches. I'm also open to other ideas. @brianhw - do you have an opinion about this?

Short term, I'm going to revert this so that it doesn't block the release and we should open a new PR to merge this in once this issue is fixed.

### DIFF
--- a/edx/analytics/tasks/tests/acceptance/fixtures/input/load_auth_userprofile.sql
+++ b/edx/analytics/tasks/tests/acceptance/fixtures/input/load_auth_userprofile.sql
@@ -1,6 +1,6 @@
 CREATE TABLE `auth_userprofile` ( `id` int(11) NOT NULL, `user_id` int(11) NOT NULL, `name` varchar(255) NOT NULL, `language` varchar(255) NOT NULL, `location` varchar(255) NOT NULL, `meta` longtext NOT NULL, `courseware` varchar(255) NOT NULL, `gender` varchar(6), `mailing_address` longtext, `year_of_birth` int(11), `level_of_education` varchar(6), `goals` longtext, `allow_certificate` tinyint(1) NOT NULL, `country` varchar(2), `city` longtext, PRIMARY KEY (`id`), UNIQUE KEY `user_id` (`user_id`) );
 
-INSERT INTO `auth_userprofile` (`id`,`user_id`,`name`,`language`,`location`,`meta`,`courseware`,`gender`,`mailing_address`,`year_of_birth`,`level_of_education`,`goals`,`allow_certificate`,`country`,`city`) VALUES (1,1,'honor','es-ES','Europe','','course.xml','m','Luna, 10 - 3, 28300 ARANJUEZ',1984,'a','Me encanta aprender.',1,'ES', 'Madrid');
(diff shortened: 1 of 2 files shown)


mulby:
Oops - this is a partial revert, I've corrected this with #292

brianhw:
We already do some ad hoc cleaning so far, as we apply a strip() to username and more recently to course_id to deal with data in which we don't generally expect to find newlines, and don't expect them to be meaningful.  I'm not sure the assumption is quite right, as these are actually functioning as identifiers, so the whitespace is probably actually meaningful.  Another place is in answer distribution, where we do clean up on the HTML and such that we get for problem labels.

Generally I would prefer to be agnostic about newlines/tabs and such, but in order to do escaping, one generally also has to escape backslashes, and that could be more disruptive.  However, once we get beyond the raw events, I don't think we need to be nearly as committed to preserving whitespace.  In such cases, I suspect it would be good enough to normalize whitespace down to single spaces or similar.

pomegranited:
@mulby Thank you for recognizing the potential impact on production data before this change was deployed!  That would have been hard to repair.

I can't think of many cases where tabs would be structural to the fields we deal with, so I'd say they could safely be replaced with a single space, as @brianhw  suggests.  Also, there's a lot of pipeline code that assumes a `\t` separator (e.g. [UsersPerCountry.mapper](https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/user_location.py#L250), [ModuleEngagementDataTask.reducer](https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/module_engagement.py#L197), [CourseEnrollmentChangesPerDay.mapper](https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/course_enroll.py#L136)), which isn't good practice, but changing the hive delimiter would require a careful overhaul of the pipeline, and maybe [luigi.hadoop too](https://github.com/edx/luigi/blob/master/luigi/hadoop.py#L820)?

And if I understand correctly, [Luigi hadoop implicitly assumes the records are newline-separated](https://github.com/edx/luigi/blob/master/luigi/hadoop.py#L795)?  So if we decide to keep them, then we'll have to serialize/deserialize them out of the raw data.

For the user profile fields, I don't think we need to keep the newlines either.  I'd be fine with just doing a whitespace sanitation on all the input data before it goes into hive/sql.

mulby:
I like the whitespace substitution idea. I agree that is probably the most simple approach.

You can see I did something similar in the Hive query using regex_replace to sanitize the "name" field.

PR #293, opened 2016-09-15
replace separators when serializing a record to a separated string

mulby:
This is a hotfix for the newly released d_program_course table. A bunch of course titles are now being returned from the catalog API as `<p>Some Title</p\n`.

I verified that they were not returned this way previously. For the time being I've delayed stripping out the HTML tags and instead just focused on the "\n" since it's a delimiter.

Reviewers:
- [ ] @brianhw 
- [ ] @HassanJaveed84

### DIFF
--- a/edx/analytics/tasks/util/record.py
+++ b/edx/analytics/tasks/util/record.py
@@ -224,7 +224,7 @@ class Record(object):
         new_attribute_values.update(kwargs)
         return self.__class__(**new_attribute_values)
 
-    def to_string_tuple(self, null_value=DEFAULT_NULL_VALUE):
+    def to_string_tuple(self, string_encoder=None):
         """
         Convert the record into a tuple of UTF-8 encoded byte strings.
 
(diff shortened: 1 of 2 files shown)


mulby:
The urgency of this has somewhat diminished since the most recent refresh of the catalog metadata removed the paragraph tags and newlines etc.

brianhw:
I'm not sure about whether a dict is the right approach here.  In particular, I'm thinking that what we really want to pass in is an encoder, and possibly to allow a decoder to be specified on the other end.  That would allow us to choose to preserve the newlines if they were considered important.  To pursue this, I would imagine that the encoder/decoder would be defined when the Record's field is being defined.   We would then be able to provide backslash encoding/decoding for some fields (such as those pesky usernames, for example, or similar things for which preservation should be important), and provide a lossy whitespace-stripper for encoding, and a no-op for decoding.  Perhaps the whitespace-stripper even becomes the default for text.

In terms of these specific low-level record methods, it may be that a dict is adequate in cases (like whitespace) where order doesn't matter.  And that an ordered dict can be passed in in cases (like backslash-encoding) where order does matter.

mulby:
Dealing with this elsewhere. Closing for now.

PR #295, opened 2016-09-26
Brian/fix finance acceptance test

brianhw:
@mulby This PR to your branch includes the revisions that I was suggesting on your PR to master.  I went a bit further, while I was at it, to find other places where access to S3 from the test might be occurring -- or should/could be occurring. 

I have verified that this branch passes acceptance tests on Jenkins.  However, I have not been successful at verifying tests on an analyticstack.

### DIFF
--- a/edx/analytics/tasks/event_exports.py
+++ b/edx/analytics/tasks/event_exports.py
@@ -1,7 +1,6 @@
 """Group events by institution and export them for research purposes"""
 
 import logging
-import os
 import gzip
 from collections import defaultdict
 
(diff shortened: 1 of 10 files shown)


brianhw:
I have verified that existing tests work on analyticstack, including the finance test.  The new answer-dist-to-mysql test also works now.  I could not verify other tests due to virtual memory errors on my analyticstack.

brianhw:
Tests now work on analyticstack, including a number that were previously disabled.    I think the first step is to merge this to Gabe's branch, and then we can review it more fully and then merge that branch to master.  @mulby @HassanJaveed84

mulby (approved):
woohoo! Let's do it.

LGTM

PR #296, opened 2016-09-26
Create a summarized enrollment table

mulby:
Generate a smaller version of the `course_enrollment` model that just includes summary information for each user in each course.

Todo:
- [x] Acceptance Test
- [x] Pep8/Pylint
- [x] Docstrings

Reviewers: @HassanJaveed84 @brianhw

### DIFF
--- a/edx/analytics/tasks/enrollments.py
+++ b/edx/analytics/tasks/enrollments.py
@@ -9,18 +9,21 @@ from luigi.parameter import DateIntervalParameter
 
 from edx.analytics.tasks.database_imports import ImportAuthUserProfileTask
 from edx.analytics.tasks.mapreduce import MapReduceJobTaskMixin, MapReduceJobTask, MultiOutputMapReduceJobTask
-from edx.analytics.tasks.pathutil import PathSelectionByDateIntervalTask, EventLogSelectionDownstreamMixin, EventLogSelectionMixin
+from edx.analytics.tasks.pathutil import PathSelectionByDateIntervalTask, EventLogSelectionDownstreamMixin, \
+    EventLogSelectionMixin
 from edx.analytics.tasks.url import get_target_from_url, url_path_join, ExternalURL, UncheckedExternalURL
 from edx.analytics.tasks.util import eventlog, opaque_key_util
 from edx.analytics.tasks.util.hive import WarehouseMixin, HiveTableTask, HivePartition, HiveQueryToMysqlTask
 from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
 from edx.analytics.tasks.decorators import workflow_entry_point
-
+from edx.analytics.tasks.util.record import Record, StringField, IntegerField, BooleanField, DateTimeField
 
 log = logging.getLogger(__name__)
 DEACTIVATED = 'edx.course.enrollment.deactivated'
 ACTIVATED = 'edx.course.enrollment.activated'
 MODE_CHANGED = 'edx.course.enrollment.mode_changed'
+ENROLLED = 1
+UNENROLLED = 0
 
 
 class CourseEnrollmentEventsTask(
@@ -171,6 +174,9 @@ class CourseEnrollmentTask(CourseEnrollmentDownstreamMixin, MapReduceJobTask):
         self.course_enrollment_events_root = url_path_join(self.warehouse_path, 'course_enrollment_events')
 
     def requires_local(self):
+        if self.overwrite_n_days == 0:
+            return []
+
         overwrite_interval = DateIntervalParameter().parse('{}-{}'.format(
             self.overwrite_from_date,
             self.interval.date_b
@@ -204,10 +210,14 @@ class CourseEnrollmentTask(CourseEnrollmentDownstreamMixin, MapReduceJobTask):
             date_pattern='%Y-%m-%d',
         )
 
-        return {
+        requirements = {
             'path_selection_task': path_selection_task,
-            'downstream_input_tasks': self.requires_local().downstream_input_tasks(),
         }
+
+        if self.overwrite_n_days > 0:
+            requirements['downstream_input_tasks'] = self.requires_local().downstream_input_tasks()
+
+        return requirements
 
     def mapper(self, line):
         (
@@ -308,8 +318,6 @@ class DaysEnrolledForEvents(object):
 
     """
 
-    ENROLLED = 1
-    UNENROLLED = 0
     MODE_UNKNOWN = 'unknown'
 
     def __init__(self, course_id, user_id, interval, events):
@@ -341,7 +349,7 @@ class DaysEnrolledForEvents(object):
 
         # Before we start processing events, we can assume that their current state is the same as it has been for all
         # time before the first event.
-        self.state = self.previous_state = self.UNENROLLED
+        self.state = self.previous_state = UNENROLLED
         self.mode = self.MODE_UNKNOWN
 
     def days_enrolled(self):
@@ -409,10 +417,10 @@ class DaysEnrolledForEvents(object):
         """
         self.mode = self.event.mode
 
-        if self.state == self.ENROLLED and self.event.event_type == DEACTIVATED:
-            self.state = self.UNENROLLED
-        elif self.state == self.UNENROLLED and self.event.event_type == ACTIVATED:
-            self.state = self.ENROLLED
+        if self.state == ENROLLED and self.event.event_type == DEACTIVATED:
+            self.state = UNENROLLED
+        elif self.state == UNENROLLED and self.event.event_type == ACTIVATED:
+            self.state = ENROLLED
         elif self.event.event_type == MODE_CHANGED:
             pass
         else:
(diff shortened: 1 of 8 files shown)


edx/analytics/tasks/enrollments.py
```diff
@@ -583,6 +585,72 @@ def columns(self):
...
+              first_credit_enrollment_date
+            FROM (
+              SELECT
+                course_id,
+                user_id,
+                LAST_VALUE(mode) OVER user_course AS current_enrollment_mode,
+                LAST_VALUE(at_end) OVER user_course AS current_enrollment_is_active,
+                FIRST_VALUE(mode) OVER user_course AS first_enrollment_mode,
+                FIRST_VALUE(date) OVER user_course AS first_enrollment_date,
+                MAX(CASE WHEN change = -1 THEN date ELSE NULL END) OVER user_course AS last_unenrollment_date,
```
brianhw: What about edge case where user enrolls and unenrolls on the first day?  Actually, I guess there are several challenges with that scenario.   If there was a verified or credit enrollment on that day, followed by an unenrollment on the same day, we also miss the first_XXX_enrollment_date values below.  That is, all three dates will end up being NULL.

edx/analytics/tasks/load_internal_reporting_user_course.py
```diff
@@ -208,3 +212,47 @@ def run(self):
...
+        return 'd_user_course'
+
+    @property
+    def default_columns(self):
+        """List of tuples defining name and definition of automatically-filled columns."""
+        return None
+
+    @property
+    def auto_primary_key(self):
+        """The warehouse schema defines an auto primary key called row_number for this table."""
```
brianhw: nit: it doesn't define, since this is None.

edx/analytics/tasks/load_internal_reporting_user_course.py
```diff
@@ -208,3 +212,47 @@ def run(self):
...
+    date = luigi.DateParameter()
+
+    @property
+    def partition(self):
+        """The table is partitioned by date."""
+        return HivePartition('dt', self.date.isoformat())  # pylint: disable=no-member
+
+    @property
+    def insert_source_task(self):
+        return ExternalURL(url=self.hive_partition_path('course_enrollment_summary', self.date))
```
brianhw, line 221: So this is decoupled, but expects that enrollment has been run earlier in the same day.  And it's not simply loading the most recent version of enrollment summary that may be present.  So if that enrollment job fails, this task here will also fail, and therefore the whole warehouse load fails.  And maybe that's desirable. 

One possibility is to allow this a little wiggle-room:  that it looks for date, and if it doesn't exist, then return the previous day.

brianhw (commented):
First round of reviews.

edx/analytics/tasks/enrollments.py
mulby: Correct. Open to suggestions here.

edx/analytics/tasks/load_internal_reporting_user_course.py
mulby: Yeah, I considered borrowing that logic from user activity, however, I'm concerned about it silently using very stale data without us knowing about it.

We are missing a "strong warning" where it raises an alert that we actually see but proceeds with the load of the stale data.

edx/analytics/tasks/load_internal_reporting_user_course.py
mulby: copypasta!

mulby:
@brianhw and @HassanJaveed84 this is ready for review again

edx/analytics/tasks/enrollments.py
```diff
@@ -204,11 +210,15 @@ def requires_hadoop(self):
...
         )
 
-        return {
+        requirements = {
             'path_selection_task': path_selection_task,
-            'downstream_input_tasks': self.requires_local().downstream_input_tasks(),
         }
 
+        if self.overwrite_n_days > 0:
+            requirements['downstream_input_tasks'] = self.requires_local().downstream_input_tasks()
```
brianhw, line 218: This is fine.  I just thought it would automatically work:  that is, an n_days=0 would result in a zero-day interval, which would return an empty list for downstream_input_tasks(), which would have no effect on the requirements when flattened.  So that wasn't the case?

edx/analytics/tasks/enrollments.py
```diff
@@ -465,6 +473,142 @@ def requires(self):
...
+    current_enrollment_mode = StringField(
+        length=100,
+        nullable=False,
+        description='The last mode seen on an activation or mode change event.'
+    )
+    current_enrollment_is_active = BooleanField(
+        nullable=False,
+        description='True if the user is currently enrolled as of the end of the interval.'
+    )
+    first_enrollment_mode = StringField(length=100, nullable=True, description='The mode the user first enrolled with.')
```
brianhw, line 490: Is it still the case that enrollments are a multi-state transition?  I thought at one point we were seeing the enrollment event, which was then followed immediately by the mode-change event that reflected the actual, requested mode rather than a default mode.  If that were so, we'd be getting a lot of default values here instead of a lot of requested values.

edx/analytics/tasks/enrollments.py
```diff
@@ -465,6 +473,142 @@ def requires(self):
...
+            last_unenrollment_time=self.format_timestamp(last_unenroll_event),
+            first_verified_enrollment_time=self.format_timestamp(first_event_by_mode.get('verified')),
+            first_credit_enrollment_time=self.format_timestamp(first_event_by_mode.get('credit')),
+            end_time=DateTimeField().deserialize_from_string(self.interval.date_b.isoformat())
+        )
+        yield record.to_string_tuple()
+
+    @staticmethod
+    def format_timestamp(event):
+        """Given an event, return a datetime object for it's timestamp."""
```
brianhw: nit:  it's => its

edx/analytics/tasks/enrollments.py
```diff
@@ -465,6 +473,142 @@ def requires(self):
...
+                        self.incr_counter('Enrollment State', 'Enrolled Activation Mode Changed', 1)
+                elif state == UNENROLLED:
+                    if event.event_type == DEACTIVATED:
+                        self.incr_counter('Enrollment State', 'Unenrolled Deactivation', 1)
+                    elif event.event_type == MODE_CHANGED:
+                        self.incr_counter('Enrollment State', 'Unenrolled Mode Change', 1)
+
+        if first_enroll_event is None:
+            # The user only has deactivate and mode change events... that's odd, just throw away the record.
+            return
```
brianhw, line 565: we should add a counter here too.

edx/analytics/tasks/enrollments.py
```diff
@@ -465,6 +473,142 @@ def requires(self):
...
+                    # If we see multiple activation events in a row, consider the first one to be the first enrollment.
+                    state = ENROLLED
+                    if first_enroll_event is None:
+                        first_enroll_event = event
+
+                if event.event_type == MODE_CHANGED:
+                    if event.mode == most_recent_mode:
+                        self.incr_counter('Enrollment State', 'Redundant Mode Change', 1)
+
+                # The most recent mode is computed from the activation and mode changes, if we see a different mode
```
brianhw: nit:  "mode changes, if" => "mode changes. If"

edx/analytics/tasks/enrollments.py
```diff
@@ -465,6 +473,142 @@ def requires(self):
...
+                # on the deactivation event, it is ignored. It's unclear in many of these cases which event to trust,
+                # so fairly arbitrary decisions have been made.
```
brianhw, line 545: Umm, this sounds a bit scary.  Hopefully this doesn't happen too often.  If it does, we should probably try to understand this better.  But based on the experience of enrollment event synthesis arising from enrollment validation (which should effectively make these cases disappear), the numbers are less than 10 a day, and usually much less.  (And for Edge, it's pretty much zero.)

edx/analytics/tasks/tests/acceptance/__init__.py
```diff
@@ -283,3 +291,35 @@ def assertEventLogEqual(self, expected_filepath, actual_filepath):
...
                 actual = sorted([json.loads(eventline) for eventline in actual_output_file])
                 self.assertListEqual(expected, actual)
+
+    @staticmethod
+    def read_dfs_directory(url):
+        output_targets = PathSetTask([url], ['*']).output()
+        raw_output = []
+        for output_target in output_targets:
+            if isinstance(output_target, S3HdfsTarget):
+                output_target = get_target_from_url(get_jenkins_safe_url(output_target.path))
```
brianhw, line 300: I have a refactoring of this in PR#295, waiting to go into your fix-acceptance-test branch.  Fortunately, it won't be hard to patch it into this.

mulby:
@brianhw, @HassanJaveed84 ping

edx/analytics/tasks/tests/test_enrollments.py
```diff
@@ -9,8 +9,8 @@
     CourseEnrollmentEventsTask,
     DEACTIVATED,
     ACTIVATED,
-    MODE_CHANGED
-)
+    MODE_CHANGED,
+    CourseEnrollmentSummaryTask)
```
brianhw: Actually, the parentheses on a separate line should conform more with our style guide.  Did it break pylint?

brianhw (approved):
No real blockers.

edx/analytics/tasks/load_internal_reporting_user_course.py
brianhw: If we wanted to, however, we can define what counts as "slightly stale but acceptable" versus "very stale".   The f_user_activity load is blind, loading whatever is latest regardless of how stale it is.  Instead, one could require that it be within a day of the current date.  Or I was imagining just looking for current date, and if not present, using the previous day.  If that is not present, then failing.

edx/analytics/tasks/enrollments.py
mulby: Yeah, I saw it actually try to run a job with an interval that looked like "2016-01-01-2016-01-01" it was not the intended behavior, so I put in this check.

edx/analytics/tasks/enrollments.py
mulby: My gut here says that the verified enrollment will be captured by the first_verified_enrollment_time and we can make a judgement call at analysis time as to whether it's an "upgrade" or an "initial enrollment".

edx/analytics/tasks/enrollments.py
mulby: I hope so as well... was going to check the counters once this is running in prod.

edx/analytics/tasks/enrollments.py
mulby: Good call, adding.

edx/analytics/tasks/enrollments.py
mulby: Fixed

edx/analytics/tasks/enrollments.py
mulby: fixed

edx/analytics/tasks/tests/acceptance/__init__.py
mulby: yeah, I tried to limit the scope of my change, considered pulling in that whole branch but decided it was scope creep

edx/analytics/tasks/tests/test_enrollments.py
mulby: Oops, pycharm did this automatically, fixing.

brianhw:
LGTM. 👍

HassanJaveed84:
LGTM 👍 
Would we still maintain f_user_course ?

HassanJaveed84 (approved):

PR #300, opened 2016-10-05
Make output_root a non parameter.

HassanJaveed84:
@brianhw @mulby this task does not use output_root, so made this a non-param on Brian's suggestion.

### DIFF
--- a/edx/analytics/tasks/enrollments.py
+++ b/edx/analytics/tasks/enrollments.py
@@ -40,6 +40,8 @@ class CourseEnrollmentEventsTask(
     # FILEPATH_PATTERN should match the output files defined by output_path_for_key().
     FILEPATH_PATTERN = '.*?course_enrollment_events_(?P<date>\\d{4}-\\d{2}-\\d{2})'
 
+    output_root = ''
+
     def mapper(self, line):
         value = self.get_event_and_date_string(line)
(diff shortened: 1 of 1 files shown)


mulby (approved):
Ah, right, because we just use the warehouse path to generate the output path.

:+1:

edx/analytics/tasks/enrollments.py
```diff
@@ -40,6 +40,8 @@ class CourseEnrollmentEventsTask(
     # FILEPATH_PATTERN should match the output files defined by output_path_for_key().
     FILEPATH_PATTERN = '.*?course_enrollment_events_(?P<date>\\d{4}-\\d{2}-\\d{2})'
 
+    output_root = ''
```
brianhw, line 44: It would be good to have a comment on this, noting why it's here.

brianhw:
Add a comment, and this should be good to go. 👍

brianhw:
LGTM 👍

brianhw (approved):
LGTM. 👍   (Trying again to approve.)

HassanJaveed84:
Forgot to squash before merging 😞

brianhw:
@HassanJaveed84 This seems to have broken master's unit tests.  Can you take a look?

PR #301, opened 2016-10-05
Replace d_course fields from course catalog API.

HassanJaveed84:
@mulby @brianhw please review. 

I'm having trouble getting the acceptance test to pass. 

I've a sample output from the test at https://gist.github.com/HassanJaveed84/00f4626a9ff4b376fafb1ce65668d822. 

I am having similar issue while validating course_seats. The data from the generated report looks identical. Can it be a whitespace issue ?


HassanJaveed84:
This appears to be related to DATETIME field, as the test for course_seat passed after removing 'course_seat_upgrade_deadline' from the comparison. I'm still investigating..

edx/analytics/tasks/tests/acceptance/test_internal_reporting_course_catalog.py
```diff
@@ -52,7 +54,44 @@ def validate_output(self):
...
-            try:  # A ValueError will be thrown if the column names don't match or the two data frames are not square.
-                self.assertTrue(all(d_program_course == expected))
-            except ValueError:
-                self.fail("Expected and returned data frames have different shapes or labels.")
+            self.assert_data_frames_equal(d_program_course, expected)
+
+    def validate_course_seat(self):
+        with self.vertica.cursor() as cursor:
+            expected_output_csv = os.path.join(self.data_dir, 'output', 'acceptance_expected_d_course_seat.csv')
+            expected = pandas.read_csv(expected_output_csv, parse_dates=[4])
```
HassanJaveed84, line 66: Unable to get pandas.read_csv  to parse the date with parse_dates=True, the expected data had a 'str' type for the date column before this change.

HassanJaveed84:
pandas.read_csv with parse_dates=True does not seem to return parsed dates but instead just strings.

edx/analytics/tasks/load_internal_reporting_course_catalog.py
```diff
@@ -13,9 +13,10 @@
...
 from edx.analytics.tasks.url import url_path_join
 from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
-from edx.analytics.tasks.vertica_load import VerticaCopyTask
+from edx.analytics.tasks.vertica_load import VerticaCopyTask, VerticaCopyTaskMixin
 from edx.analytics.tasks.util.hive import WarehouseMixin
-from edx.analytics.tasks.util.record import Record, StringField
+from edx.analytics.tasks.util.record import Record, StringField, FloatField, DateTimeField, IntegerField
+
 
 # pylint: disable=anomalous-unicode-escape-in-string
```
mulby, line 22: We also want to start using the "exclude_utm=1" parameter on our requests to the catalog now, this will prevent UTM codes from being appended to the marketing_url field.

This is referencing line 63 but github won't let me comment there since it hasn't changed :stuck_out_tongue: 

See this for reference: https://prod-edx-discovery.edx.org/api-docs/#!/course_runs/Course_Run_list

edx/analytics/tasks/load_internal_reporting_course_catalog.py
```diff
@@ -165,3 +176,142 @@ def table(self):
...
+    end_time = DateTimeField(nullable=True)
+    enrollment_start_time = DateTimeField(nullable=True)
+    enrollment_end_time = DateTimeField(nullable=True)
+    content_language = StringField(nullable=True, length=255)
+    pacing_type = StringField(nullable=True, length=255)
+    level_type = StringField(nullable=True, length=255)
+    availability = StringField(nullable=True, length=255)
+    org_id = StringField(nullable=False, length=255)
+    partner_short_code = StringField(nullable=True, length=255)
+    marketing_url = StringField(nullable=True, length=1024)
```
mulby, line 250: We also want to pull min_effort, max_effort, and modified.

We may want to consider renaming "modified" since I'm pretty sure it indicates the last time the catalog was refreshed. It could mean a bunch of other stuff (last time the course content changed, last time any of its metadata changed etc), but I don't think it does.

Maybe @katymyw or @stroilova have opinions about what it should be named?

edx/analytics/tasks/tests/acceptance/fixtures/output/acceptance_expected_d_course.csv
```diff
@@ -1,4 +1,4 @@
...
+course-v1:AcceptanceX+Testing101x+2T2016,AcceptanceX+Testing101x,All about acceptance testing!,2016-06-01 00:00:00,2016-09-01 00:00:00,2016-03-01 00:00:00,2016-06-01 00:00:00,en-us,self_paced,Introductory,Archived,AcceptanceX,openedx,https://www.edx.org/course/all-about-acceptance-testing-edx-testing101-2t2016?utm_source=analyticsauto&utm_medium=affiliate_partner
+course-v1:AcceptanceX+Testing101x+3T2016,AcceptanceX+Testing101x,All about acceptance testing!,2016-09-01 00:00:00,2016-12-01 00:00:00,2016-06-01 00:00:00,2016-09-01 00:00:00,en-us,instructor_paced,Intermediate,Current,AcceptanceX,openedx,https://www.edx.org/course/all-about-acceptance-testing-edx-testing101-3t2016?utm_source=analyticsauto&utm_medium=affiliate_partner
+course-v1:AcceptanceX+Testing102x+1T2017,AcceptanceX+Testing102x,All about acceptance testing Part 3!,2016-12-01 00:00:00,2017-02-01 00:00:00,2016-09-01 00:00:00,2016-12-01 00:00:00,en-us,instructor_paced,Advanced,Upcoming,AcceptanceX,openedx,https://www.edx.org/course/all-about-acceptance-testing-edx-testing-part-3-102-1t2017?utm_source=analyticsauto&utm_medium=affiliate_partner
```
mulby: The utm_source and utm_medium stuff should disappear once we add the "exclude_utm" parameter to our API call

mulby (commented):
We should also ensure the old d_course loading code doesn't run as part of this PR.

Probably delete the code for the old d_course loading method.

edx/analytics/tasks/load_internal_reporting_course_catalog.py
katymyw: @mulby do you know who could confirm what exactly that field means? Agreed 'modified' is not the most intuitive if that is what it means.

edx/analytics/tasks/util/record.py
```diff
@@ -286,16 +288,18 @@ def from_string_tuple(cls, string_tuple, null_value=DEFAULT_NULL_VALUE):
             null_value (string): When an input string has this value it is set to None in the constructed Record.
 
         """
+        if string_decoder is None:
+            string_decoder = HiveTsvEncoder()
```
brianhw, line 291: Is there a reason this is defined in here, and not as a default value above?

edx/analytics/tasks/tests/acceptance/test_internal_reporting_course_catalog.py
```diff
@@ -52,7 +54,44 @@ def validate_output(self):
...
+                "SELECT course_id,course_seat_type,course_seat_price,course_seat_currency,"
+                "course_seat_upgrade_deadline,course_seat_credit_provider,course_seat_credit_hours FROM"
+                " {schema}.d_course_seat ORDER BY course_id ASC".format(
+                    schema=self.vertica.schema_name
+                )
+            )
+            response = cursor.fetchall()
+            d_course_seat = pandas.DataFrame(response, columns=['course_id', 'course_seat_type', 'course_seat_price',
+                                                                   'course_seat_currency', 'course_seat_upgrade_deadline',
+                                                                   'course_seat_credit_provider', 'course_seat_credit_hours'])
```
brianhw: Sorry if this is redundant from previous reviews, but it seems we could define this list of columns, and use here and in the cursor.execute call above.  Even if you don't want to share the definition with the code under test.

edx/analytics/tasks/tests/acceptance/test_internal_reporting_course_catalog.py
```diff
@@ -52,7 +54,44 @@ def validate_output(self):
...
+                "marketing_url FROM {schema}.d_course ORDER BY course_id ASC".format(
+                    schema=self.vertica.schema_name
+                )
+            )
+            response = cursor.fetchall()
+            d_course = pandas.DataFrame(response, columns=['course_id', 'catalog_course', 'catalog_course_title',
+                                                                   'start_time', 'end_time','enrollment_start_time',
+                                                                   'enrollment_end_time', 'content_language',
+                                                                   'pacing_type', 'level_type', 'availability', 'org_id',
+                                                                   'partner_short_code', 'marketing_url'])
```
brianhw: Same thing here:  looks like the same values on 84-86.

edx/analytics/tasks/load_internal_reporting_course_catalog.py
```diff
@@ -165,3 +176,142 @@ def table(self):
...
+    def table(self):
+        return 'd_course'
+
+    @property
+    def columns(self):
+        return CourseRecord.get_sql_schema()
+
+
+class LoadInternalReportingCourseCatalogToWarehouse(LoadInternalReportingCourseCatalogMixin, VerticaCopyTaskMixin, luigi.WrapperTask):
+
```
brianhw, line 314: docstring....

edx/analytics/tasks/load_internal_reporting_course_catalog.py
```diff
@@ -165,3 +176,142 @@ def table(self):
...
+    start_time = DateTimeField(nullable=True)
+    end_time = DateTimeField(nullable=True)
+    enrollment_start_time = DateTimeField(nullable=True)
+    enrollment_end_time = DateTimeField(nullable=True)
+    content_language = StringField(nullable=True, length=255)
+    pacing_type = StringField(nullable=True, length=255)
+    level_type = StringField(nullable=True, length=255)
+    availability = StringField(nullable=True, length=255)
+    org_id = StringField(nullable=False, length=255)
+    partner_short_code = StringField(nullable=True, length=255)
```
brianhw: partner_short_code is actually 255?  I thought it was 8.

edx/analytics/tasks/load_internal_reporting_course_catalog.py
```diff
@@ -86,16 +87,18 @@ class ProgramCourseRecord(Record):
...
     org_id = StringField(nullable=False, length=255)
     partner_short_code = StringField(nullable=True, length=8)
 
 
-class ExtractProgramCourseTask(LoadInternalReportingCourseCatalogMixin, luigi.Task):
-    """Process the information from the course structure API and write it to a tsv."""
+class BaseCourseMetadataTask(LoadInternalReportingCourseCatalogMixin, luigi.Task):
+    """Process the information from the API and write it to a tsv."""
+
+    table_name = 'override_me'
```
brianhw, line 102: Should this be a property that raises a NotImplementedError, the same way other methods do?

edx/analytics/tasks/load_internal_reporting_course_catalog.py
```diff
@@ -165,3 +176,142 @@ def table(self):
...
+    level_type = StringField(nullable=True, length=255)
+    availability = StringField(nullable=True, length=255)
+    org_id = StringField(nullable=False, length=255)
+    partner_short_code = StringField(nullable=True, length=255)
+    marketing_url = StringField(nullable=True, length=1024)
+
+
+class ExtractCourseTask(BaseCourseMetadataTask):
+
+    table_name = 'course_catalog'
```
brianhw, line 258: when to use a property?

edx/analytics/tasks/load_internal_reporting_course_catalog.py
brianhw: We also leave out a bunch of descriptive stuff, like video, image, announcement, short_description, full_description, and also ignore staff.  Do we at some point want to pull in transcript_languages at some point?  I thought there was interest, for example, in whether the availability of other languages boosted course activity.

brianhw (commented):
Looks good, mostly nits.

mulby:
@katymyw I talked this over with e-comm. They said it should change when
the core metadata in the course run model changes. Note this only takes
into account the top level fields in the model, not any linked models like
course seats. In practice this seems to be updated whenever the catalog
refreshes its data.

edx/analytics/tasks/util/record.py
HassanJaveed84: I think if we define it as a default value above, we would need to move HiveTsvEncoder definition before this class(Record).

edx/analytics/tasks/tests/acceptance/test_internal_reporting_course_catalog.py
```diff
@@ -29,30 +29,77 @@ def test_internal_reporting_user_course(self):
...
+    def validate_course(self):
+        with self.vertica.cursor() as cursor:
+            expected_output_csv = os.path.join(self.data_dir, 'output', 'acceptance_expected_d_course.csv')
+            expected = pandas.read_csv(expected_output_csv, parse_dates=[3, 4, 5, 6])
+
+            columns = [
+                'course_id', 'catalog_course', 'catalog_course_title', 'start_time', 'end_time',
+                'enrollment_start_time', 'enrollment_end_time', 'content_language', 'pacing_type',
+                'level_type', 'availability', 'org_id', 'partner_short_code', 'marketing_url',
+                'min_effor', 'max_effort', 'modified',
```
brianhw: typo on min_effort.

edx/analytics/tasks/util/record.py
brianhw: I didn't know that.

brianhw (commented):
nit

edx/analytics/tasks/load_internal_reporting_course_catalog.py
HassanJaveed84: I've pulled in min_effort,max_effort and modified(as modified until we have a consensus on what it should be renamed to). I've only seen null values for min_effort and max_effort from the json response. Do we need to pull in any more fields ?

edx/analytics/tasks/tests/acceptance/fixtures/output/acceptance_expected_d_course.csv
HassanJaveed84: Done.

edx/analytics/tasks/load_internal_reporting_course_catalog.py
HassanJaveed84: Done.

edx/analytics/tasks/tests/acceptance/test_internal_reporting_course_catalog.py
HassanJaveed84: Done.

edx/analytics/tasks/tests/acceptance/test_internal_reporting_course_catalog.py
HassanJaveed84: Done.

edx/analytics/tasks/tests/acceptance/fixtures/output/acceptance_expected_d_course.csv
```diff
@@ -1,4 +1,4 @@
...
+course-v1:AcceptanceX+Testing101x+2T2016,AcceptanceX+Testing101x,All about acceptance testing!,2016-06-01 00:00:00,2016-09-01 00:00:00,2016-03-01 00:00:00,2016-06-01 00:00:00,en-us,self_paced,Introductory,Archived,AcceptanceX,openedx,https://www.edx.org/course/all-about-acceptance-testing-edx-testing101-2t2016,2,4,2016-09-07 18:38:32.207371
+course-v1:AcceptanceX+Testing101x+3T2016,AcceptanceX+Testing101x,All about acceptance testing!,2016-09-01 00:00:00,2016-12-01 00:00:00,2016-06-01 00:00:00,2016-09-01 00:00:00,en-us,instructor_paced,Intermediate,Current,AcceptanceX,openedx,https://www.edx.org/course/all-about-acceptance-testing-edx-testing101-3t2016,3,5,2016-09-07 18:38:32.207371
+course-v1:AcceptanceX+Testing102x+1T2017,AcceptanceX+Testing102x,All about acceptance testing Part 3!,2016-12-01 00:00:00,2017-02-01 00:00:00,2016-09-01 00:00:00,2016-12-01 00:00:00,en-us,instructor_paced,Advanced,Upcoming,AcceptanceX,openedx,https://www.edx.org/course/all-about-acceptance-testing-edx-testing-part-3-102-1t2017,2,5,2016-09-07 18:38:32.207371
```
HassanJaveed84: The modified field in the JSON response has strings like:
2016-10-06T00:45:22.555138Z

Wanted to clarify if we want to drop anything out of that ?

edx/analytics/tasks/load_internal_reporting_course_catalog.py
mulby: We should probably double check all of these. I did this by logging into the discovery service database and looking at the course_metadata_courserun table.

mulby:
@HassanJaveed84 - after some discussion, we decided we want exclude the "modified" field since we don't fully understand what it represents at the moment and don't want to confuse users.

edx/analytics/tasks/load_internal_reporting_course_catalog.py
mulby: If you login to the MySQL database backing the discovery service you can see the schema for all of these fields. I was lazy and used 255 but we probably want to more closely match the actual values in the tables.

edx/analytics/tasks/load_internal_reporting_course_catalog.py
mulby: when it's dynamic? This can easily be overridden by a property if the subclass wanted to do something special...

edx/analytics/tasks/util/record.py
mulby: Also, it has the nice property that other methods can use the default `string_decoder=None` and simply pass that value along without having to replicate the default value in their own signature.

mulby (commented):
This is mostly looking good to me, just need to tweak the sizes of some of the string columns.

edx/analytics/tasks/load_internal_reporting_course_catalog.py
HassanJaveed84: I have been able to verify most of these fields, but I can't seem to find some fields like 'marketing_url' and availability. I have looked up multiple tables related to course_metadata in the discovery service db.

HassanJaveed84:
I've removed the 'modified' field and updated some fields. A bunch of pep8/pylint stuff.

edx/analytics/tasks/util/record.py
```diff
@@ -235,15 +235,16 @@ def to_string_tuple(self, null_value=DEFAULT_NULL_VALUE):
             null_value (str): The string to use to represent None if a nullable field has a None value.
```
mulby: nit: I think the docstrings in this file need to be updated to reference the `string_encoder` instead of the `null_value`

edx/analytics/tasks/util/record.py
```diff
@@ -494,6 +519,14 @@ def validate(self, value):
                 validation_errors.append('The string length exceeds the maximum allowed')
         return validation_errors
 
+    def serialize_to_string(self, value):
+        """Returns a unicode string representation of a value for this field."""
+        try:
+            return unicode(value, encoding=getattr(self, 'encoding', 'utf8'))
```
mulby, line 525: hrmmm, we may want to pull this change out of this PR. I don't remember if it's actually necessary or not, and I don't think there are any tests for it.

It allows you to change the encoding for fields by passing in parameters when they are declared:

``` python
foo = StringField(encoding='latin-1')
```

mulby (approved):
A couple of nits, but no blockers.

edx/analytics/tasks/util/record.py
HassanJaveed84: Removing this fails the unit tests for record.py(https://github.com/edx/edx-analytics-pipeline/blob/b3626f8d35bb48dfac3a0f50a91214aa40f597bc/edx/analytics/tasks/util/tests/test_record.py#L161)

brianhw (approved):
LGTM. 👍

edx/analytics/tasks/load_internal_reporting_course_catalog.py
```diff
@@ -165,3 +177,159 @@ def table(self):
...
+        }
+        yield LoadInternalReportingCourseToWarehouse(**kwargs)
+        yield LoadInternalReportingCourseSeatToWarehouse(**kwargs)
+        yield LoadInternalReportingProgramCourseToWarehouse(**kwargs)
+
+    def output(self):
+        return [t.output() for t in self.requires()]
+
+    def complete(self):
+        return all(r.complete() for r in luigi.task.flatten(self.requires()))
```
HassanJaveed84, line 333: An unfulfilled dependency issue is caused by OverwriteOutputMixin. We should not include OverwriteOutputMixin for this wrapper task. Should we split out the Mixin instead of defining a custom 'complete' method here ?

edx/analytics/tasks/load_internal_reporting_course_catalog.py
```diff
@@ -86,16 +88,18 @@ class ProgramCourseRecord(Record):
     """Represents a course run within a program."""
     program_id = StringField(nullable=False, length=36)
```
HassanJaveed84, line 89: program_id is set to value of program's uuid. I see that as type of uuid in course_metadata_program table as char(32). However, the json response has uuid of length 36 from what I've seen.

brianhw: That seems plausible.  The json would include the dashes, but a type of "uuid" wouldn't need to store the dashes.

dsjen:
FYI, I believe this PR removed `edx.analytics.tasks.load_internal_reporting_course` which is needed by [`test_course_information.py`](https://github.com/edx/edx-analytics-pipeline/blob/dc442b5885b8f20e7d2227a06082246d9cce2a57/edx/analytics/tasks/tests/test_course_information.py#L20).

PR #302, opened 2016-10-06
Set output_root as None.

HassanJaveed84:
@mulby @brianhw I inadvertently broke the unit tests for enrollments. Need to set this to None instead of an empty string.


mulby (approved):

brianhw (approved):
LGTM. 👍

PR #305, opened 2016-10-20
AN-7880 UserActivityTask: process new forum event_types

thallada:
This will enable processing forum events in the weekly course activity task. I tested this locally with some (slightly modified) [example forum events from student_forum_activity.log](https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/tests/acceptance/fixtures/input/student_forum_activity.log) in my tracking.log.

Some things I wasn't able to confirm:
- If these event_types are really what the forums emits now (I can't get forums working on my devstack)
- ~~If I'm missing or including and event I shouldn't in DISCUSSION_ACTIVITIES. It depends on how we want to define what a "discussion activity" is.~~ Nevermind, I talked to Katy and now I'm copying what the learner roster considers "discussion activity"

### DIFF
--- a/edx/analytics/tasks/tests/test_user_activity.py
+++ b/edx/analytics/tasks/tests/test_user_activity.py
@@ -4,6 +4,7 @@ Tests for tasks that collect enrollment events.
 """
 import datetime
 import json
+from ddt import ddt, data, unpack
 from edx.analytics.tasks.tests.map_reduce_mixins import ReducerTestMixin, MapperTestMixin
 
 from luigi import date_interval
@@ -22,6 +23,7 @@ from edx.analytics.tasks.tests import unittest
 from edx.analytics.tasks.tests.opaque_key_mixins import InitializeOpaqueKeysMixin, InitializeLegacyKeysMixin
 
 
+@ddt
 class UserActivityTaskMapTest(InitializeOpaqueKeysMixin, MapperTestMixin, unittest.TestCase):
     """
     Tests to verify that event log parsing by mapper works correctly.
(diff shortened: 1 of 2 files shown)


edx/analytics/tasks/tests/test_user_activity.py
```diff
@@ -113,7 +113,7 @@ def test_problem_event(self):
         self.assertEquals(event, expected)
 
     def test_post_forum_event(self):
-        line = self.create_event_log_line(event_source='server', event_type='blah/blah/threads/create')
+        line = self.create_event_log_line(event_source='server', event_type='edx.forum.thread.created')
```
mulby: given the logic in the mapper, I'm assuming there are other types of events that should be counted as well? Can you enumerate the ones we know about here?

thallada: Good point, I'll test those events too.

dsjen (approved):
Nice to see!  LGTM, although I don't know much about the pipeline...yet.  👍

edx/analytics/tasks/tests/test_user_activity.py
```diff
@@ -113,12 +113,33 @@ def test_problem_event(self):
...
         event = tuple(self.task.mapper(line))
         expected = (((self.course_id, self.username, self.expected_date_string, ACTIVE_LABEL), 1),
                     ((self.course_id, self.username, self.expected_date_string, POST_FORUM_LABEL), 1))
         self.assertEquals(event, expected)
 
+    def test_forum_response_event(self):
```
mulby: I think we could make this more DRY with ddt!

thallada: I've never used ddt before... it's pretty convenient!

thallada:
Can someone on the pipeline team give this a review / thumbs up when they get a chance?

brianhw (approved):
LGTM.  👍    Thanks, Tyler!

thallada:
Thanks Brian! :ship:

PR #306, opened 2016-10-24
Remove traceback info from retry log message.

brianhw:
Not sure if this is a great idea in general, but there are cases where we want to retry and we don't really need the traceback information to clutter our logs.  This is particularly true for calls to API's where we are making a request and get back a particular error code.  

@mulby @HassanJaveed84


mulby (approved):
LGTM

PR #309, opened 2016-10-27
Added course metadata and enrollment summary tables.

dsjen:
I'll be adding acceptance tests next and this is hopefully in a good enough state for an early review at the very least.  Please feel free to review!

Changes include:
- In `load_internal_reporting_course_catalog`, wrapped Extract\* classes into `HivePartionTask` and created analogous `BareHiveTableTask`
- Added `course_enrollment_summary.py` that includes the job to create a course summary sql table with course metadata and recent enrollment data

@thallada @HassanJaveed84 @brianhw @mulby


edx/analytics/tasks/course_enrollment_summary.py
```diff
@@ -0,0 +1,203 @@
...
+    Record,
+    StringField,
+)
+
+
+class CourseEnrollmentSummaryRecord(Record):
+    """Recent enrollment summary and metadata for a course."""
+    course_id = StringField(length=255, nullable=False)
+    catalog_course_title = StringField(nullable=True, length=255,
+                                       normalize_whitespace=True)
```
brianhw: nit: line length should be 120, so I would expect this to fit on a single line.

edx/analytics/tasks/course_enrollment_summary.py
```diff
@@ -0,0 +1,203 @@
...
+)
+from edx.analytics.tasks.load_internal_reporting_course_catalog import (
+    ExtractCoursePartitionTask,
+    ExtractProgramCoursePartitionTask,
+    LoadInternalReportingCourseCatalogMixin,
+)
+
+from edx.analytics.tasks.util.hive import (
+    BareHiveTableTask,
+    hive_database_name,
```
brianhw: nit: Not necessarily your problem, but I'm surprised there's a global called "hive_database_name".  Not sure what our naming conventions are for such things, but this just looks odd.

edx/analytics/tasks/course_enrollment_summary.py
brianhw: Then again, at some point all of these will (hopefully) have description strings as well, which will change the formatting dramatically.

edx/analytics/tasks/course_enrollment_summary.py
```diff
@@ -0,0 +1,203 @@
...
+class CourseEnrollmentSummaryRecord(Record):
+    """Recent enrollment summary and metadata for a course."""
+    course_id = StringField(length=255, nullable=False)
+    catalog_course_title = StringField(nullable=True, length=255,
+                                       normalize_whitespace=True)
+    start_time = DateTimeField(nullable=True)
+    end_time = DateTimeField(nullable=True)
+    count = IntegerField(nullable=True)
+    count_change_7_days = IntegerField(nullable=True)
+    cumulative_count = IntegerField(nullable=True)
```
brianhw: Not sure how important the column ordering is, but it's determined by the order of fields here.  Seems odd to have the aggregates in the middle of the metadata, and to have non-nullable fields (like program_id, enrollment_mode, catalog_course) at the end, far away from the key (the course_id).

edx/analytics/tasks/load_internal_reporting_course_catalog.py
```diff
@@ -252,7 +293,48 @@ class CourseRecord(Record):
...
+    @property
+    def table(self):
+        return 'course_catalog'
+
+    @property
+    def columns(self):
+        return CourseRecord.get_hive_schema()
+
+
+class ExtractCoursePartitionTask(LoadInternalReportingCourseCatalogMixin, HivePartitionTask):
```
brianhw: Nomenclature question:  We have CourseTableTask and CourseRecord, but ExtractCoursePartitionTask.  What is the role of "Extract" in this?  Do we need something to indicate what kind of course it is if "Course" by itself is not clear enough?  But it seems to me that _all_ our tasks are Extract tasks of some sort, so that's not relevant.  But that we may need an additional modifier on Course.

edx/analytics/tasks/course_enrollment_summary.py
```diff
@@ -0,0 +1,203 @@
...
+                program.program_title,
+                program.catalog_course,
+                course.start_time,
+                course.end_time,
+                course.pacing_type,
+                course.availability,
+                enrollment_end.count,
+                (enrollment_end.count - COALESCE(enrollment_start.count, 0)) as count_change_7_days
+                enrollment_end.cumulative_count,
+                enrollment_end.mode,
```
brianhw: Hmm, now I'm confused.  Why does the order of these columns not match the order in the CourseEnrollmentSummaryRecord?  I would have thought that they needed to match.

edx/analytics/tasks/course_enrollment_summary.py
```diff
@@ -0,0 +1,203 @@
...
+            FROM {enrollment_table} as enrollment_end
+            LEFT JOIN {enrollment_table} as enrollment_start
+                ON enrollment_start.course_id = enrollment_end.course_id
+                AND enrollment_start.mode = enrollment_end.mode
+            LEFT JOIN {program_table} as program
+                ON program.course_id = ce_recent.course_id
+            LEFT JOIN {course_table} as course
+                ON course.course_id = ce_recent.course_id
+            WHERE ce_recent.date = '{end_date}'
+            AND ce_early.date = '{start_date}';
```
brianhw: What is ce_recent and ce_early?

edx/analytics/tasks/course_enrollment_summary.py
```diff
@@ -0,0 +1,203 @@
...
+        }
+        yield ImportCourseSummaryEnrollmentsIntoMysql(**kwargs)
+
+
+class ImportCourseSummaryEnrollmentsIntoMysql(CourseEnrollmentSummaryDownstreamMixin,
+                                              HiveQueryToMysqlTask):
+
+    @property
+    def table(self):
+        return 'course_summary'
```
brianhw: course_summary_enrollments?   Or course_enrollment_summary?  I'm confused by the sudden inversion in naming.

brianhw (commented):
Great progress!  Mostly nits at this point.

edx/analytics/tasks/course_enrollment_summary.py
dsjen: I noticed it's used in [module_engagement](https://github.com/edx/edx-analytics-pipeline/blob/cc233ee70dfc40fd4f7d80d3825d2ccc21475acf/edx/analytics/tasks/module_engagement.py#L1169).  I could refactor the name to something more fitting--maybe too much for this PR though.

edx/analytics/tasks/course_enrollment_summary.py
dsjen: Ok, I'll plan on adding descriptions to the fields and format as needed.

edx/analytics/tasks/course_enrollment_summary.py
dsjen: Ah, thanks!  I'll reorder.

edx/analytics/tasks/course_enrollment_summary.py
dsjen: Ugh, I'm being sloppy on the naming.  I did noticed that course_enrollment_summary is used in [enrollments.py](https://github.com/edx/edx-analytics-pipeline/blob/457dde57faac923256752270ccca2e684c7a3eeb/edx/analytics/tasks/enrollments.py#L596), but inverting the summary and enrollments doesn't seem like a good solution.  What do you think if I called this `course_meta_summary_enrollment` instead?  Too weird?

edx/analytics/tasks/course_enrollment_summary.py
dsjen: I didn't realize!  I will make them match.

edx/analytics/tasks/course_enrollment_summary.py
dsjen: Left over from an earlier prototype.  ce_recent is enrollment_end and ce_early is enrollment_start.  Thanks for catching that!

edx/analytics/tasks/load_internal_reporting_course_catalog.py
dsjen: I didn't see "extract" in the other parts of the codebase either and thought that it could be renamed--maybe just removing "extract" entirely as it's assumed.  Thoughts?

I'm sort of at a loss for the "course" naming conundrum.  What does it mean to be a program course vs. just a course?  One appears to have marketing information attached to it and the other program information.  Maybe:
- MarketingCourse
- ProgramCourse

Would it make sense to have an AbstractCourseRecord?

AUTHORS
```diff
@@ -11,3 +11,5 @@ James Rowan <jrowan@mit.edu>
 Dmitry Viskov <dmitry.viskov@webenterprise.ru>
 Sanford Student <sstudent@edx.org>
 Jillian Vogel <jill@opencraft.com>
+Dennis Jen <djen@edx.org>
```
mulby, line 14: :smile:

edx/analytics/tasks/course_enrollment_summary.py
```diff
@@ -0,0 +1,203 @@
...
+    catalog_course = StringField(nullable=False, length=255)
+
+
+class CourseEnrollmentSummaryDownstreamMixin(CourseEnrollmentDownstreamMixin, LoadInternalReportingCourseCatalogMixin):
+    """Combines course enrollment and catalog parameters."""
+    pass
+
+
+class CourseEnrollmentSummaryWrapperTask(CourseEnrollmentSummaryDownstreamMixin,
+                                         luigi.WrapperTask):
```
mulby: I think you can remove this class

edx/analytics/tasks/course_enrollment_summary.py
```diff
@@ -0,0 +1,203 @@
...
+                ON program.course_id = ce_recent.course_id
+            LEFT JOIN {course_table} as course
+                ON course.course_id = ce_recent.course_id
+            WHERE ce_recent.date = '{end_date}'
+            AND ce_early.date = '{start_date}';
+        """.format(
+            database_name=hive_database_name(),
+            enrollment_table='course_enrollment_mode_daily',
+            program_table='program_course',
+            course_table='course_catalog',
```
mulby: I don't see a huge amount of value in pulling these table names out into the format string unless you were to actually dynamically fetch the table names from the requirements. We usually just hardcode table names in queries like this one (for better or worse)!

edx/analytics/tasks/load_internal_reporting_course_catalog.py
mulby: I think just removing "Extract" is sufficient IMHO.

mulby (commented):
Shaping up very nicely!

edx/analytics/tasks/course_enrollment_summary.py
dsjen: Gotcha!

edx/analytics/tasks/load_internal_reporting_course_catalog.py
dsjen: Ok, I'll plan on removing "Extract" unless there are additional thoughts on this!

edx/analytics/tasks/course_enrollment_summary.py
dsjen: I was feeling a little iffy on this class too as it's not needed.  If I remember correctly, @brianhw had suggested to me that having a wrapper class as the entry point would be helpful in case other workflow were to be added.  Is there a pattern that's typically done?  Is this preemptive optimization?

edx/analytics/tasks/course_summary_enrollment.py
```diff
@@ -0,0 +1,232 @@
...
+                                       description='Difference in enrollment counts over the past 7 days')
+    cumulative_count = IntegerField(nullable=True, description='The cumulative total of all users ever enrolled')
+
+
+class CourseSummaryEnrollmentDownstreamMixin(CourseEnrollmentDownstreamMixin, LoadInternalReportingCourseCatalogMixin):
+    """Combines course enrollment and catalog parameters."""
+    pass
+
+
+class CourseSummaryEnrollmentWrapperTask(CourseSummaryEnrollmentDownstreamMixin,
```
brianhw: We also mark entry points with a decorator:  @workflow_entry_point

edx/analytics/tasks/course_summary_enrollment.py
```diff
@@ -0,0 +1,232 @@
...
+                                         luigi.WrapperTask):
+    """Entrypoint for course summary enrollment task."""
```
brianhw: nit: I would suggest making this two words:  "Entry point"....

edx/analytics/tasks/course_summary_enrollment.py
brianhw: Also not sure about how important, but I think we generally define the classes in the order they are used.  So since this uses ImportCourseSummaryEnrollmentsIntoMysql, it would normally appear after it.  (I think it's required if it inherits, but less so if the dependency is in a function.)

edx/analytics/tasks/course_summary_enrollment.py
```diff
@@ -0,0 +1,232 @@
...
+                program_title,
+                catalog_course,
+                start_time,
+                end_time,
+                pacing_type,
+                availability,
+                enrollment_mode,
+                count,
+                count_change_7_days,
+                cumulative_count
```
brianhw: Hmm.  This is all the fields in the CourseSummaryEnrollmentRecord, right?  I wonder if there's an easy way to extract the list from there.  Something like ','.join([field.name for field in CourseSummaryEnrollmentRecord.fields()]).  Just to make it DRYer.

brianhw: Or one could just get it from the self.columns, which already gets it from the record's get_sql_schema.  Just extract the names.

edx/analytics/tasks/course_summary_enrollment.py
```diff
@@ -0,0 +1,232 @@
...
+            n_reduce_tasks=self.n_reduce_tasks,
+            source=self.source,
+            interval=self.interval,
+            pattern=self.pattern,
+            warehouse_path=self.warehouse_path,
+            overwrite_n_days=self.overwrite_n_days,
+            date=self.date,
+            api_root_url=self.api_root_url,
+            api_page_size=self.api_page_size,
+            overwrite=self.overwrite,
```
brianhw: Do you have a sense of the scope of the overwrite as opposed to the overwrite_n_days?  I'm assuming the former is for catalog, and the latter for enrollment?  (Possibly worth a comment somewhere in that class?  Or where?)

brianhw (commented):
Looks good!  A few nits.  

But I was looking for how you were going to handle the "Edge cases", as it were.  And I think I missed it.

edx/analytics/tasks/course_summary_enrollment.py
dsjen: Thanks, I'll add the decorator, and I'm happy to reorder the classes by usage.

edx/analytics/tasks/course_summary_enrollment.py
dsjen: Oh, I like that!  Will do.

edx/analytics/tasks/course_summary_enrollment.py
dsjen: Right, `overwrite_n_days` is for enrollment and `overwrite` is for the catalog.  I'll add relevant comments.  Thanks!

dsjen:
Hooray!  Acceptance tests for test_course_summary_enrollment and test_enrollments (because I changed some of the enrollment event fixtures) pass and `make coverage` passes.  I'm running the entire acceptance test suite currently.

To recap, this PR includes:
- In `load_internal_reporting_course_catalog`, wrapped Extract* classes into `HivePartionTask` and created analogous `BareHiveTableTask`
- Added `course_enrollment_summary.py` that includes the job to create a course summary sql table with course metadata and recent enrollment data

The job can be run with a parameter to turn off population of the course catalog tables for the case that the catalog service isn't available.

Please feel free to review!  I'll be sure to squash commits once approved.

@thallada @HassanJaveed84 @brianhw @mulby

edx/analytics/tasks/course_summary_enrollment.py
```diff
@@ -0,0 +1,171 @@
...
+            CoursePartitionTask(
+                date=self.date,
+                warehouse_path=self.warehouse_path,
+                api_root_url=self.api_root_url,
+                api_page_size=self.api_page_size,
+                overwrite=self.overwrite,
+            ),
+        ]
+
+        if self.disable_course_catalog:
```
brianhw: This is a nice clean solution.  It might warrant a brief comment.

edx/analytics/tasks/tests/acceptance/test_course_summary_enrollment.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+@ddt
+class CourseEnrollmentSummaryAcceptanceTest(AcceptanceTestCase):
+    """Ensure course enrollment summary is populated in the result store."""
+
+    CATALOG_DATE = '2016-09-08'
+
+    def setUp(self):
+        ''' Loads enrollment and course catalog fixtures. '''
```
brianhw: nit:  Using single quotes followed by a space looks cool, but it's different than our standard.

brianhw (approved):
LGTM!  👍

edx/analytics/tasks/course_summary_enrollment.py
dsjen: Can do!

edx/analytics/tasks/tests/acceptance/test_course_summary_enrollment.py
dsjen: Of course!  I'll update.

dsjen:
I believe all acceptance tests should be fixed now.  I'm running the entire suite now.  Please review when you have a moment.

dsjen:
FYI, acceptance tests have all passed!

edx/analytics/tasks/course_summary_enrollment.py
```diff
@@ -0,0 +1,173 @@
...
+    """Creates the course summary enrollment sql table."""
+
+    @property
+    def indexes(self):
+        return [('course_id',)]
+
+    @property
+    def query(self):
+        end_date = self.interval.date_b - datetime.timedelta(days=1)
+        start_date = self.interval.date_b - datetime.timedelta(days=7)
```
mulby: don't we want this to be `end_date - datetime.timedelta(days=7)`?

edx/analytics/tasks/course_summary_enrollment.py
```diff
@@ -0,0 +1,173 @@
...
+                enrollment_end.count,
+                (enrollment_end.count - COALESCE(enrollment_start.count, 0)) AS count_change_7_days,
+                enrollment_end.cumulative_count
+            FROM course_enrollment_mode_daily enrollment_end
+            LEFT OUTER JOIN course_enrollment_mode_daily enrollment_start
+                ON enrollment_start.course_id = enrollment_end.course_id
+                AND enrollment_start.mode = enrollment_end.mode
+                AND enrollment_start.date = '{start_date}'
+            LEFT OUTER JOIN program_course program
+                ON program.course_id = enrollment_end.course_id
```
mulby: Note that courses can belong to multiple programs, so this could result in multiple rows in the output for each course - is that the desired behavior? I suppose we could handle the display logic downstream if we wanted to consolidate in the API or UI tiers...

edx/analytics/tasks/course_summary_enrollment.py
```diff
@@ -0,0 +1,173 @@
...
+                api_root_url=self.api_root_url,
+                api_page_size=self.api_page_size,
+                overwrite=self.overwrite,
+            ),
+
+        ]
+
+        # if disable_course_catalog is set, create the hive tables w/o loading data
+        if self.disable_course_catalog:
+            catalog_tasks = [task.hive_table_task for task in catalog_tasks]
```
mulby: nit: reassignment of the `catalog_tasks` variable here was a little confusing to me... I think something like this is more clear:

``` python
if self.disable_course_catalog:
    # this Open edX installation isn't running a catalog service, so just create empty hive tables without loading any data into them
    yield [task.hive_table_task for task in catalog_tasks]
else:
    # create the hive tables and populate them with data
    yield catalog_tasks
```

edx/analytics/tasks/course_summary_enrollment.py
```diff
@@ -0,0 +1,173 @@
...
+    def table(self):
+        return 'course_meta_summary_enrollment'
+
+    @property
+    def columns(self):
+        return CourseSummaryEnrollmentRecord.get_sql_schema()
+
+    @property
+    def required_table_tasks(self):
+        yield EnrollmentByModeTask(
```
mulby: Historically we have referenced other tables like this using external URLs since this could cause some wacky race conditions if this workflow's schedule overlapped with the enrollment workflow.

We may want to carefully consider how these three workflows are related and when each one needs to run relative to one another:
1. The workflow that pulls the catalog API data into the warehouse.
2. The enrollment workflow
3. This new workflow

I see potential here for them to stomp on one another depending on how they are scheduled.

mulby (commented):
Only blockers here are the workflow scheduling and the date math.

Great work!

edx/analytics/tasks/course_summary_enrollment.py
dsjen: Makes sense!

edx/analytics/tasks/course_summary_enrollment.py
dsjen: I didn't realize that.  Good catch!  It would be good to account for here I think.  I think I'm  misunderstanding the left outer join as I thought it avoided this problem.  Is there a different join to use or ordering of the query?

mulby: It's actually a kind of hard problem, you generally need to figure out what to do when a course belongs to multiple programs - should we just use the first one alphabetically? Should the programs be presented as a comma separated list?

edx/analytics/tasks/course_summary_enrollment.py
dsjen: Good catch!  I think I was confused.  I was thinking that lets says that `date_b` is 2016-11-17.  `end_date` would then be 2016-11-16 and the `start_date` would be 2016-11-10 for a total of 7 complete days.

However, the ask was to be able to essentially compare yesterday's enrollment counts to the same day, e.g. compare Monday and the previous Monday.  I'll update this!

edx/analytics/tasks/course_summary_enrollment.py
dsjen: Oh, that's very interesting.  I guess it might be confusing to the user to see just the first program, and your suggestion of turning them into a comma separated list would help solve that.  I can change the query so that both `catalog_course_title` and `catalog_course` from from the course_catalog table instead, which then leaves a group by (right?) for the `program_id` and `program_title`.

The program data isn't needed for the MVP course listing table in Insights, so maybe we can punt on including the program data for now.  Opinions?

@katymyw - is it important that we include `program_id` and `program_title` in the course listing table in the resultstore?  It's not in the wireframes, but perhaps you had some forward thinking reason for wanting it available.

katymyw: @dsjen -- this was a little more forward-thinking (e.g., lightweight program dashboard could be just filtering to the courses in a program). In which case we would want to know all programs a course is in. But I'd say scrap it since it's not immediately needed and adds complexity here.

dsjen:
Unit tests, linter, and acceptance tests on enrollment have passed successfully.  I'm running the full acceptance test suite now.  Please leave any additional feedback ASAP!

edx/analytics/tasks/enrollments.py
```diff
@@ -859,24 +866,145 @@ def columns(self):
...
+    disable_course_catalog = luigi.BooleanParameter(
+        config_path={'section': 'course-summary-enrollment', 'name': 'disable_course_catalog'},
+        default=False,
+        description="Disables course catalog data jobs and only creates empty tables for them."
+    )
+
+
+class CourseSummaryEnrollmentRecord(Record):
+    """Recent enrollment summary and metadata for a course."""
+    course_id = StringField(length=255, nullable=False, description='A unique identifier of the course')
```
brianhw: tiny-nit:  this is the only one with length before nullable.

edx/analytics/tasks/enrollments.py
```diff
@@ -859,24 +866,145 @@ def columns(self):
         ]
 
 
+class CourseSummaryEnrollmentDownstreamMixin(CourseEnrollmentDownstreamMixin, LoadInternalReportingCourseCatalogMixin):
+    """Combines course enrollment and catalog parameters."""
+
+    disable_course_catalog = luigi.BooleanParameter(
+        config_path={'section': 'course-summary-enrollment', 'name': 'disable_course_catalog'},
+        default=False,
```
brianhw, line 874: I think this should be true by default.  Or we change the sense of it, and call it "enable_course_catalog" with a default of False.  I think most people outside of edX who will be running enrollment will not have a course catalog set up.   Once we have something better documented for helping them do the setup or set the flag, then we could change the default in the code if we really wanted to.  But I think it would be pretty disruptive to have people run enrollment and have to find out about setting this flag.

brianhw (approved):
Just one issue about a default value.  Otherwise LGTM. 👍

edx/analytics/tasks/enrollments.py
dsjen: Ok, I'll update!

edx/analytics/tasks/enrollments.py
dsjen: Done!

dsjen:
@brianhw -- reran the enrollment acceptance tests with your requested changes and they've passed.  I'm rerunning the entire acceptance suite with the change now.

dsjen:
The test_enrollment_validation acceptance test passed.  I would be shocked if the run of the entire acceptance suite doesn't pass (but I will wait until they pass before merging).  Please chime in if there's anything else!

mulby (approved):
LGTM

dsjen:
Acceptance tests have passed!  Merging soon.

PR #310, opened 2016-10-27
Discovery: Task to import a table from mysql into vertica.

HassanJaveed84:
@mulby @brianhw this is what I did as part of the discovery.

### DIFF
--- a/edx/analytics/tasks/database_imports.py
+++ b/edx/analytics/tasks/database_imports.py
@@ -4,14 +4,19 @@ Import data from external RDBMS databases into Hive.
 import datetime
 import logging
 import textwrap
+import re
 
 import luigi
 from luigi.hive import HiveQueryTask, HivePartitionTarget
 
 from edx.analytics.tasks.sqoop import SqoopImportFromMysql
-from edx.analytics.tasks.url import url_path_join
+from edx.analytics.tasks.url import url_path_join, ExternalURL, get_target_from_url
 from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
-from edx.analytics.tasks.util.hive import hive_database_name, hive_decimal_type
+from edx.analytics.tasks.util.hive import hive_database_name, hive_decimal_type, WarehouseMixin
+from edx.analytics.tasks.mysql_load import MysqlInsertTaskMixin, CredentialFileMysqlTarget
+from edx.analytics.tasks.util.record import Record, StringField, FloatField, DateTimeField, IntegerField, BooleanField
+from edx.analytics.tasks.vertica_load import VerticaCopyTask, VerticaCopyTaskMixin
+
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


HassanJaveed84:
A wrapper task would query all the table names out of the DB and exclude the ones we don't want to import.

edx/analytics/tasks/database_imports.py
```diff
@@ -213,6 +218,131 @@ def requires(self):
...
+            credentials=self.credentials,
+            num_mappers=self.num_mappers,
+            verbose=self.verbose,
+            overwrite=self.overwrite,
+            database=self.database,
+            # Hive expects NULL to be represented by the string "\N" in the data. You have to pass in "\\N" to sqoop
+            # since it uses that string directly in the generated Java code, so "\\N" actually looks like "\N" to the
+            # Java code. In order to get "\\N" onto the command line we have to use another set of escapes to tell the
+            # python code to pass through the "\" character.
+            null_string='\\\\N',
```
HassanJaveed84, line 256: We are not importing data in hive, what should we use as null_string here ? "NULL" ?

edx/analytics/tasks/sqoop.py
```diff
@@ -103,7 +103,7 @@ def output(self):
 
     def metadata_output(self):
         """Return target to which metadata about the task execution can be written."""
-        return get_target_from_url(url_path_join(self.destination, '.metadata'))
+        return get_target_from_url(url_path_join(self.destination, '_metadata'))
```
HassanJaveed84, line 106: .metadata is picked up when loading data into Vertica.

edx/analytics/tasks/database_imports.py
```diff
@@ -213,6 +218,131 @@ def requires(self):
...
+                if column_type == 'int' or column_type == 'smallint':
+                    setattr(self.record, column_name, IntegerField(nullable=nullable))
+                elif column_type == 'tinyint':
+                    setattr(self.record, column_name, BooleanField(nullable=nullable))
+                elif column_type == 'varchar':
+                    setattr(self.record, column_name, StringField(nullable=nullable, length=column_length))
+                elif column_type == 'datetime' or column_type == 'date':
+                    setattr(self.record, column_name, DateTimeField(nullable=nullable))
+                elif column_type == 'longtext':
+                    setattr(self.record, column_name, StringField(nullable=nullable, length=65000))
```
HassanJaveed84, line 306: This is not a exhaustive list.

edx/analytics/tasks/database_imports.py
```diff
@@ -213,6 +218,131 @@ def requires(self):
...
+            # This is a string that is interpreted as an octal number, so it is equivalent to the character Ctrl-A
+            # (0x01). This is the default separator for fields in Hive.
+            fields_terminated_by='\x2C',
+            # Replace delimiters with a single space if they appear in the data. This prevents the import of malformed
+            # records. Hive does not support escape characters or other reasonable workarounds to this problem.
+            delimiter_replacement=' ',
+        )
+
+    def output(self):
+        return get_target_from_url(self.hive_partition_path(self.import_table, self.import_date.isoformat()))
```
HassanJaveed84, line 268: Point to the output of the sqoop task.

edx/analytics/tasks/sqoop.py
brianhw: So did the renaming to _metadata solve your problem?  That's good information (and good news).

edx/analytics/tasks/database_imports.py
```diff
@@ -213,6 +218,131 @@ def requires(self):
...
+                column_type_info = column[1]
+                column_nullable = column[2]
+                nullable = True if column_nullable == 'YES' else False
+
+                match = re.search("(.*)\((\d*)\)", column_type_info)
+                if match:
+                    column_type = match.group(1)
+                    column_length = match.group(2)
+                else:
+                    column_type = column_type_info
```
brianhw, line 295: nit: I'm assuming this if/else could be collapsed by making the second group optional?

edx/analytics/tasks/database_imports.py
brianhw: Thinking that this logic would eventually make sense to roll into Record.  Maybe as simple as adding to each Field a new entry which is a set of equivalent types, instead of the canonical version.  So for IntegerField, it would have sql_base_type = 'INT', but sql_equivalent_type = ('INT', 'INTEGER', 'SMALLINT').  For BooleanField, sql_equivalent_type = ('TINYINT', 'BOOLEAN').  Then this matching would simply involve finding the (first) Field that has a matching entry in the field's sql_equivalent_type.

edx/analytics/tasks/database_imports.py
brianhw: Doesn't the Vertica code still look for this, however, in order to know what to treat as NULL?

edx/analytics/tasks/database_imports.py
```diff
@@ -213,6 +218,131 @@ def requires(self):
...
+    def default_columns(self):
+        return []
+
+    @property
+    def table(self):
+        return self.import_table
+
+    @property
+    def columns(self):
+        return self.insert_source_task.record.get_sql_schema()
```
brianhw, line 341: So we didn't need a different definition of columns between Mysql and Vertica after all?  Then in fact we might not need to do the Record mapping in between after all, but just use the Mysql column definitions directly?

edx/analytics/tasks/database_imports.py
brianhw: Can it just ask the import_task for its output?  Well, maybe that's not very important.  Same information being provided as destination in SqoopImportFromMysql initialization as is used here.

brianhw (commented):
Looking good!  Very encouraging in its relative simplicity.

edx/analytics/tasks/sqoop.py
HassanJaveed84: It did, thanks for the useful info.

edx/analytics/tasks/database_imports.py
```diff
@@ -213,6 +218,131 @@ def requires(self):
...
+
+
+class LoadMysqlToVerticaTableTask(WarehouseMixin, VerticaCopyTask):
+
+    import_table = luigi.Parameter()
+
+    @property
+    def copy_delimiter(self):
+        """The delimiter in the data to be copied.  Default is tab (\t)"""
+        return "','"
```
mulby, line 321: Isn't this set to `Ctrl-A` up here: https://github.com/edx/edx-analytics-pipeline/pull/310/files#diff-2c8ff49110cd4ccd29585a6dc421c3abR261 ?

edx/analytics/tasks/database_imports.py
mulby: We may want to have a mechanism by which we fail to load a single table if we discover it has a type we don't recognize, but we continue with the rest of the load. We would then need to be notified of the error so that we could patch in the appropriate mapping.

mulby (commented):
Overall, I think this spike is very illustrative. Great work!

I have the following thoughts:
1. Running sqoop for each table may take a prohibitively long amount of time since the overhead of each run is not insignificant. Not sure if there is a way to run it once and have it process all of the tables at once or something. A lot of the value Sqoop provides is in the parallel extraction of data from very large tables. Maybe we don't need it yet and could just use the MySQL connector directly? We might even be able to stream the records directly into Vertica without actually landing them to S3?
2. Most of the value of `Record` comes from our ability to use it to enforce contracts between disparate types of tasks (Hive, Hadoop, Python etc). I'm not sure it buys us a whole lot in this case since it's only used for schema mapping (no data serialization or deserialization). I wonder if we could exactly replicate the MySQL schema in Vertica without performing any mapping? Are there actually any types MySQL supports that Vertica doesn't?

HassanJaveed84:
Initially, I was hoping to use sqoop-import-all-tables, but it doesn't specify any mechanism to exclude tables. Skipped using that as it would also import the 2 very large tables from LMS. We could just import everything to s3, and skip those tables while loading into Vertica. Not sure what kind of speed benefit it would provide. 

We can possibly use mysql.connector to fetch 'n' records at a time and stream that directly into Vertica. I'm not sure what that would look like in code, but it should be possible. 

Regarding the data types mapping, I've seen 'longtext' in mysql tables, whereas I can't find that data type in Vertica documentation.

edx/analytics/tasks/database_imports.py
HassanJaveed84: Oh, I actually used a comma there(\x2C).

PR #311, opened 2016-11-02
Remove "NOT NULL" constraint on country_code.

brianhw:
We were seeing that '\N' values in a Hive table was being written incorrectly as zero-length strings in Mysql when run in production and on our acceptance test server, but being written as NULL and failing the constraint when run in acceptance tests on analyticstack.  Removing the constraint apparently results in NULLs being written instead of zero-length strings on both production and analyticstack.  

@mulby @HassanJaveed84

### DIFF
--- a/edx/analytics/tasks/location_per_course.py
+++ b/edx/analytics/tasks/location_per_course.py
@@ -347,7 +347,7 @@ class InsertToMysqlCourseEnrollByCountryTaskBase(MysqlInsertTask):  # pylint: di
         return [
             ('date', 'DATE NOT NULL'),
             ('course_id', 'VARCHAR(255) NOT NULL'),
-            ('country_code', 'VARCHAR(10) NOT NULL'),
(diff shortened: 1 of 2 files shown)


mulby (approved):
Weird, might have something to do with the python MySQL driver being too clever?

brianhw:
But it's the same driver on analyticstack and production.  And it behaved differently on each when the constraint was present (which is how this emerged).  Maybe then Mysql version was different in a significant way.   

Anyway, I'm running this on Stage, to check this before merging.

brianhw:
Works on stage, by running:

  ALTER TABLE course_enrollment_location_current MODIFY country_code varchar(10) NULL;

... and confirming that loaded data contains NULLs now.  And that insights continues to count NULLs along with other values that are not countries into the "Unknown Country" category.  

Ready for thumb, @HassanJaveed84 .

HassanJaveed84 (approved):

PR #313, opened 2016-11-09
Task to import a DB from mysql into vertica.

HassanJaveed84:
@mulby @brianhw this is ready for an initial review. 

I especially would want to get your feedback on Tasks which fetch the schema and list of tables. 

There are some outstanding issues like handling nulls, enclosing character etc.

### DIFF
--- a/edx/analytics/tasks/database_imports.py
+++ b/edx/analytics/tasks/database_imports.py
@@ -4,16 +4,24 @@ Import data from external RDBMS databases into Hive.
 import datetime
 import logging
 import textwrap
+import re
+import json
 
 import luigi
 from luigi.hive import HiveQueryTask, HivePartitionTarget
 
 from edx.analytics.tasks.sqoop import SqoopImportFromMysql
-from edx.analytics.tasks.url import url_path_join
+from edx.analytics.tasks.url import url_path_join, ExternalURL
 from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
-from edx.analytics.tasks.util.hive import hive_database_name, hive_decimal_type
+from edx.analytics.tasks.util.hive import hive_database_name, hive_decimal_type, WarehouseMixin, HivePartition
+from edx.analytics.tasks.vertica_load import VerticaCopyTask
 
 log = logging.getLogger(__name__)
+
+try:
+    import mysql.connector
+except ImportError:
+    log.warn('Unable to import mysql client libraries')
 
 
 class DatabaseImportMixin(object):
@@ -38,7 +46,7 @@ class DatabaseImportMixin(object):
         description='Path to the external access credentials file.',
     )
     database = luigi.Parameter(
-        default_from_config={'section': 'database-import', 'name': 'database'}
+        config_path={'section': 'database-import', 'name': 'database'},
     )
     import_date = luigi.DateParameter(
         default=None,
@@ -212,6 +220,201 @@ class ImportMysqlToHiveTableTask(DatabaseImportMixin, ImportIntoHiveTableTask):
             # records. Hive does not support escape characters or other reasonable workarounds to this problem.
             delimiter_replacement=' ',
         )
+
+
+class MysqlToVerticaTaskMixin(WarehouseMixin):
+    """
+    Parameters for importing a mysql database into the warehouse.
+    """
+
+    db_credentials = luigi.Parameter(
+        config_path={'section': 'database-import', 'name': 'credentials'},
+        description='Path to the external access credentials file.',
+    )
+    database = luigi.Parameter(
+        config_path={'section': 'database-import', 'name': 'database'}
+    )
+
+
+def get_mysql_query_results(credentials, database, query):
+    """
+    Executes a mysql query on the provided database and returns the results.
+    """
+
+    credentials_target = ExternalURL(url=credentials).output()
+    cred = None
+    with credentials_target.open('r') as credentials_file:
+        cred = json.load(credentials_file)
+
+    connection = mysql.connector.connect(user=cred.get('username'),
+                                         password=cred.get('password'),
+                                         host=cred.get('host'),
+                                         port=cred.get('port'),
+                                         database=database)
+
+    try:
+        cursor = connection.cursor()
+        cursor.execute(query)
+        results = cursor.fetchall()
+    finally:
+        connection.close()
+
+    return results
+
+
+class LoadMysqlToVerticaTableTask(MysqlToVerticaTaskMixin, VerticaCopyTask):
+    """
+    Task to import a table from mysql into vertica.
+    """
+
+    table_name = luigi.Parameter(
+        description='The name of the table.',
+    )
+
+    date = luigi.DateParameter(
+        default=datetime.datetime.utcnow().date(),
+    )
+
+    def __init__(self, *args, **kwargs):
+        super(LoadMysqlToVerticaTableTask, self).__init__(*args, **kwargs)
+        self.table_schema = []
+
+    def requires(self):
+        if self.required_tasks is None:
+            self.required_tasks = {
+                'credentials': ExternalURL(url=self.credentials),
+                'insert_source': self.insert_source_task,
+            }
+        return self.required_tasks
+
+    def vertica_compliant_schema(self):
+        """Transforms mysql table schema into a vertica compliant schema."""
+
+        if not self.table_schema:
+            results = get_mysql_query_results(self.db_credentials, self.database, 'describe {}'.format(self.table_name))
+            for result in results:
+                field_name = result[0].strip()
+                field_type = result[1].strip()
+                field_null = result[2].strip()
+
+                types_with_parentheses = ['tinyint', 'smallint', 'int', 'bigint', 'datetime']
+                if any(_type in field_type for _type in types_with_parentheses):
+                    field_type = field_type.rsplit('(')[0]
+                elif field_type == 'longtext':
+                    field_type = 'LONG VARCHAR'
+                elif field_type == 'double':
+                    field_type = 'DOUBLE PRECISION'
+
+                if field_null == "NO":
+                    field_type = field_type + " NOT NULL"
+
+                field_name = "\"{}\"".format(field_name)
+
+                self.table_schema.append((field_name, field_type))
+
+        return self.table_schema
+
+    @property
+    def copy_delimiter(self):
+        """The delimiter in the data to be copied."""
+        return "','"
+
+    @property
+    def copy_null_sequence(self):
+        """The null sequence in the data to be copied."""
+        return "'NULL'"
+
+    @property
+    def enclosed_by(self):
+        return "''''"
+
+    @property
+    def insert_source_task(self):
+        partition_path_spec = HivePartition('dt', self.date.isoformat()).path_spec
+        destination = url_path_join(
+            self.warehouse_path,
+            "database_import",
+            self.database,
+            self.table_name,
+            partition_path_spec
+        ) + '/'
+        return SqoopImportFromMysql(
+            table_name=self.table_name,
+            credentials=self.db_credentials,
+            database=self.database,
+            destination=destination,
+            overwrite=self.overwrite,
+            mysql_delimiters=True,
+        )
+
+    @property
+    def default_columns(self):
+        return []
+
+    @property
+    def table(self):
+        return self.table_name
+
+    @property
+    def columns(self):
+        return self.vertica_compliant_schema()
+
+    @property
+    def auto_primary_key(self):
+        return None
+
+
+class ImportMysqlToVerticaTask(MysqlToVerticaTaskMixin, luigi.WrapperTask):
+    """Provides entry point for importing a mysql database into Vertica."""
+
+    schema = luigi.Parameter(
+        config_path={'section': 'vertica-export', 'name': 'schema'},
+        description='The schema to which to write.',
+    )
+    credentials = luigi.Parameter(
+        config_path={'section': 'vertica-export', 'name': 'credentials'},
+        description='Path to the external access credentials file.',
+    )
+    date = luigi.DateParameter(
+        default=datetime.datetime.utcnow().date(),
+    )
+    overwrite = luigi.BooleanParameter(
+        default=False,
+        significant=False,
+    )
+
+    exclude = luigi.Parameter(
+        is_list=True,
+        default=(),
+    )
+
+    def __init__(self, *args, **kwargs):
+        super(ImportMysqlToVerticaTask, self).__init__(*args, **kwargs)
+        self.table_list = []
+
+    def should_exclude_table(self, table_name):
+        """Determines whether to exlude a table during the import."""
+        if any(re.match(pattern, table_name) for pattern in self.exclude):
+            return True
+        return False
+
+    def requires(self):
+        if not self.table_list:
+            results = get_mysql_query_results(self.db_credentials, self.database, 'show tables')
+            self.table_list = [result[0].strip() for result in results]
+
+        for table_name in self.table_list:
+            if not self.should_exclude_table(table_name):
+                yield LoadMysqlToVerticaTableTask(
+                    credentials=self.credentials,
+                    schema=self.schema,
+                    db_credentials=self.db_credentials,
+                    database=self.database,
+                    warehouse_path=self.warehouse_path,
+                    table_name=table_name,
+                    overwrite=self.overwrite,
+                    date=self.date,
+                )
 
 
 class ImportStudentCourseEnrollmentTask(ImportMysqlToHiveTableTask):
--- a/edx/analytics/tasks/sqoop.py
+++ b/edx/analytics/tasks/sqoop.py
@@ -17,6 +17,7 @@ from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
 
 log = logging.getLogger(__name__)
 
+METADATA_FILENAME = '_metadata'
 
 def load_sqoop_cmd():
     """Get path to sqoop command from Luigi configuration."""
@@ -103,7 +104,7 @@ class SqoopImportTask(OverwriteOutputMixin, luigi.hadoop.BaseHadoopJobTask):
 
     def metadata_output(self):
         """Return target to which metadata about the task execution can be written."""
-        return get_target_from_url(url_path_join(self.destination, '.metadata'))
+        return get_target_from_url(url_path_join(self.destination, METADATA_FILENAME))
 
     def job_runner(self):
         """Use simple runner that gets args from the job and passes through."""
--- a/edx/analytics/tasks/tests/acceptance/test_database_import.py
+++ b/edx/analytics/tasks/tests/acceptance/test_database_import.py
@@ -1,0 +1,57 @@
+"""
+End to end test for importing mysql database into Vertica task.
+"""
+
+import os
+import logging
+import datetime
+import pandas
+
+from pandas.util.testing import assert_frame_equal
+
+from edx.analytics.tasks.tests.acceptance import AcceptanceTestCase, when_vertica_available
+
+log = logging.getLogger(__name__)
+
+
+class DatabaseImportAcceptanceTest(AcceptanceTestCase):
+    """Validate workflow to load mysql database into Vertica."""
+
+    DATE = '2014-07-01'
+
+    def setUp(self):
+        super(DatabaseImportAcceptanceTest, self).setUp()
+        self.execute_sql_fixture_file('load_database_import_test_table.sql')
+
+    @when_vertica_available
+    def test_database_import(self):
+        self.task.launch([
+            'ImportMysqlToVerticaTask',
+            '--date', self.DATE,
+            '--overwrite'
+        ])
+
+        self.validate_output()
+
+    def validate_output(self):
+        with self.vertica.cursor() as cursor:
+            expected_output_csv = os.path.join(
+                self.data_dir,
+                'output',
+                'database_import',
+                'expected_database_import_test_table.csv'
+            )
+
+            def convert_date(date_string):
+                """Convert date string to a date object."""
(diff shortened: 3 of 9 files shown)


edx/analytics/tasks/database_imports.py
```diff
@@ -214,6 +216,191 @@ def requires(self):
...
+class ImportMysqlToVerticaTask(WarehouseMixin, DatabaseImportMixin, luigi.WrapperTask):
+
+    schema = luigi.Parameter()
+
+    def requires(self):
+        tables_task = GetTablesFromMysqlTask(
+            warehouse_path=self.warehouse_path,
+            database=self.database
+        )
+        luigi.build([tables_task], local_scheduler=True)
```
HassanJaveed84: This feels hackish, but without dynamic requirements I did not see another way.

edx/analytics/tasks/database_imports.py
```diff
@@ -213,6 +215,234 @@ def requires(self):
             delimiter_replacement=' ',
         )
 
+class MysqlQueryTaskMixin(WarehouseMixin):
+    db_credentials = luigi.Parameter(
+        config_path={'section': 'database-import', 'name': 'credentials'},
+        description='Path to the external access credentials file.',
+    )
+    database = luigi.Parameter(
+        default_from_config={'section': 'database-import', 'name': 'database'}
```
brianhw: default_from_config is deprecated.  You should use config_path.

edx/analytics/tasks/sqoop.py
```diff
@@ -103,7 +103,7 @@ def output(self):
 
     def metadata_output(self):
         """Return target to which metadata about the task execution can be written."""
-        return get_target_from_url(url_path_join(self.destination, '.metadata'))
+        return get_target_from_url(url_path_join(self.destination, '_metadata'))
```
brianhw: When the time comes, it would probably be worthwhile to make this a constant.  And then use it in enrollment_validation.py.  And then remove "ugly hack" in util/obfuscate_util.py.

Actually, maybe it's worthwhile just making a separate PR to fix this.

edx/analytics/tasks/database_imports.py
```diff
@@ -213,6 +215,234 @@ def requires(self):
...
+
+
+class MysqlTableSchemaTask(MysqlQueryTaskBase):
+
+    table_name = luigi.Parameter(
+        description='The name of the table.',
+    )
+
+    def output(self):
+        url_with_filename = url_path_join(self.warehouse_path, "sql_schema", "{0}.tsv".format(self.table_name))
```
brianhw: wonder if we should include database name in here as well.   We have at least two we're planning on....

edx/analytics/tasks/database_imports.py
```diff
@@ -213,6 +215,234 @@ def requires(self):
...
+            connection.rollback()
+            raise
+        finally:
+            connection.close()
+
+
+class GetTablesFromMysqlTask(OverwriteOutputMixin, MysqlQueryTaskMixin, luigi.Task):
+
+    def output(self):
+        url_with_filename = url_path_join(self.warehouse_path, "sql_schema", "table_list")
```
brianhw: same thing: indicate the database that is being listed.   And maybe this shoudn't go in sql_schema, since it's not actually a schema.  And we may also want to date them, as we do with other "pulls" of information.

edx/analytics/tasks/database_imports.py
```diff
@@ -213,6 +215,234 @@ def requires(self):
...
+
+    @property
+    def insert_source_task(self):
+        return SqoopImportFromMysql(
+            table_name=self.table_name,
+            credentials=self.db_credentials,
+            database=self.database,
+            # TODO: We may want to make the explicit passing in of columns optional as it prevents a direct transfer.
+            # Make sure delimiters and nulls etc. still work after removal.
+            destination=self.hive_partition_path(self.table_name, self.date.isoformat()),
```
brianhw: I think it would be good to have some substructure within the warehouse_path, and not just dump all table_name entries to the "root".  We already have some tables (like auth_user, etc.) that are at root, and their contents would be different.

edx/analytics/tasks/database_imports.py
```diff
@@ -213,6 +215,234 @@ def requires(self):
...
+            }
+        return self.required_tasks
+
+    def vertica_compliant_schema(self):
+        if not self.table_schema:
+            with self.input()['mysql_schema_task'].open('r') as schema_file:
+                for line in schema_file:
+                    field_name, field_type, field_null = line.split('\t')
+
+                    int_types = ['tinyint', 'smallint', 'mediumint', 'int', 'bigint', 'datetime']
```
brianhw: datetime is an int_type?   Or just that it is followed by parentheses that should be stripped?  (Assuming not all parentheses should be stripped e.g.  varchar(128).)

edx/analytics/tasks/database_imports.py
```diff
@@ -213,6 +215,234 @@ def requires(self):
...
+    )
+
+    def requires(self):
+        tables_task = GetTablesFromMysqlTask(
+            db_credentials=self.db_credentials,
+            database=self.database,
+            warehouse_path=self.warehouse_path,
+            overwrite=self.overwrite,
+        )
+        luigi.build([tables_task], local_scheduler=True)
```
brianhw: This is probably better than what I was proposing.  The thought I had was just to get the table list as a requirement, and then to do everything else in a run method rather than as requirements.

brianhw (commented):
I threw out a few other nits while going through the code, but I don't have a better suggestion about doing the dynamic scheduling.  (I also don't fully understand how this works yet.)

edx/analytics/tasks/database_imports.py
HassanJaveed84: You're correct about the types which have parentheses(but not all of them). Initially, I only came across Integer types, hence the name. I should change the name.

edx/analytics/tasks/sqoop.py
HassanJaveed84: Noted.

edx/analytics/tasks/database_imports.py
brianhw: This is definitely better.  It seems to be what Luigi supports, both in the version we're using as well as in the current release.  Good find!

mulby: Would it be cleaner to just call a python function here that fetches the table list and then cache that result? That way we wouldn't have to use this work-around.

edx/analytics/tasks/database_imports.py
```diff
@@ -214,6 +217,256 @@ def requires(self):
...
+                for pattern in self.exclude:
+                    match = re.match(pattern, table_name)
+                if not match:
+                    yield LoadMysqlToVerticaTableTask(
+                        credentials=self.credentials,
+                        schema=self.schema,
+                        db_credentials=self.db_credentials,
+                        database=self.database,
+                        warehouse_path=self.warehouse_path,
+                        table_name=line.strip('\n'),
```
mulby: nit: This could just be `table_name=table_name`

edx/analytics/tasks/vertica_load.py
```diff
@@ -387,6 +387,10 @@ def copy_null_sequence(self):
         """The null sequence in the data to be copied.  Default is Hive NULL (\\N)"""
         return "'\\N'"
 
+    @property
+    def enclosed_by(self):
+        return "''"
```
mulby, line 393: Do we need to tell sqoop to quote the data it writes to S3?

Will this impact data that does not include quotes? For example:

```
foo,bar,'baz'
```

mulby (commented):
So far so good!

edx/analytics/tasks/vertica_load.py
HassanJaveed84: Our current imports do not quote the data, I've tested it and vertica seems to treat this as no enclosing character. 

The new import specifies mysql delimiters for sqoop import, which uses a single quote as optionally-enclosed-by, and also specifies the enclosing character as `'`

```
@property
    def enclosed_by(self):
        return "''''"
```

HassanJaveed84:
@brianhw @mulby can you advise on how to improve the tests on this ? What should the acceptance test cover ?

HassanJaveed84:
Updated some parts.
Switched to using a function instead of tasks for getting table list and table schema.

edx/analytics/tasks/database_imports.py
```diff
@@ -214,6 +223,186 @@ def requires(self):
...
+    def requires(self):
+        if not self.table_list:
+            results = get_mysql_query_results(self.db_credentials, self.database, 'show tables')
+            self.table_list = [result[0].strip() for result in results]
+
+        for table_name in self.table_list:
+            match = None
+            for pattern in self.exclude:
+                match = re.match(pattern, table_name)
+            if not match:
```
mulby: Wait, won't this always be the result of the last pattern match?

I think we want something like this:

``` python
def table_matches_pattern(pattern):
    return re.match(pattern, table_name) is not None

if not any(table_matches_pattern(pattern) for pattern in self.exclude):
    # ...
```

Maybe we could move this logic to a function so that we could unit test it more easily? The logic is non-trivial...

edx/analytics/tasks/vertica_load.py
```diff
@@ -400,13 +404,24 @@ def copy_data_table_from_target(self, cursor):
 
         with self.input()['insert_source'].open('r') as insert_source_file:
             log.debug("Running stream copy from source file")
+            log.debug(
+                "COPY {schema}.{table} ({cols}) FROM STDIN ENCLOSED BY {enclosed_by} DELIMITER AS {delim} NULL AS {null} DIRECT ABORT ON ERROR NO COMMIT;".format(
```
mulby: nit: store the query in a variable that is used both by the log statement and the actual `cursor.copy()` call

edx/analytics/tasks/database_imports.py
HassanJaveed84: Ah, good catch. I'm sure I had put a if match:break in there initially, not sure why I removed it.

edx/analytics/tasks/database_imports.py
```diff
@@ -214,6 +223,186 @@ def requires(self):
...
+                                         host=cred.get('host'),
+                                         port=cred.get('port'),
+                                         database=database)
+
+    try:
+        cursor = connection.cursor()
+        cursor.execute(query)
+        results = cursor.fetchall()
+    except:
+        raise
```
brianhw: nit: does this do anything?  Do you just want a try-finally?

edx/analytics/tasks/database_imports.py
```diff
@@ -214,6 +223,186 @@ def requires(self):
         )
 
 
+class MysqlQueryTaskMixin(WarehouseMixin):
```
brianhw: While I'm fresh on the topic, do we want overwrite functionality in this as well?  Or is it hardwired to always overwrite?

edx/analytics/tasks/database_imports.py
```diff
@@ -214,6 +223,186 @@ def requires(self):
...
+
+    def vertica_compliant_schema(self):
+        if not self.table_schema:
+            results = get_mysql_query_results(self.db_credentials, self.database, 'describe {}'.format(self.table_name))
+            for result in results:
+                field_name = result[0].strip()
+                field_type = result[1].strip()
+                field_null = result[2].strip()
+
+                types_with_parantheses = ['tinyint', 'smallint', 'mediumint', 'int', 'bigint', 'datetime']
```
brianhw: nit:  parentheses, not parantheses.

edx/analytics/tasks/sqoop.py
brianhw: If it is a separate PR, I assume it needs to be done before this merges.  Or just do it in this PR.

edx/analytics/tasks/tests/acceptance/test_database_import.py
```diff
@@ -0,0 +1,51 @@
...
+    DATE = '2014-07-01'
+
+    def setUp(self):
+        super(DatabaseImportAcceptanceTest, self).setUp()
+        #self.execute_sql_fixture_file('load_auth_user_for_internal_reporting_user.sql')
+        #self.execute_sql_fixture_file('load_auth_userprofile.sql')
+        self.execute_sql_fixture_file('load_certificates_generatedcertificate.sql')
+        # self.execute_sql_fixture_file('load_course_groups_courseusergroup.sql')
+        # self.execute_sql_fixture_file('load_course_groups_courseusergroup_users.sql')
+        # self.execute_sql_fixture_file('load_student_courseenrollment.sql')
```
brianhw: Another option is if certs doesn't test every type of column type we care about, then we should make a synthetic one that does.   Then check that we get it to Vertica successfully and compare.

edx/analytics/tasks/tests/test_database_imports.py
```diff
@@ -5,11 +5,59 @@
...
+                    slug,varchar(255),YES
+                    site_id,int(11),NO
+                    parent_id,int(11),YES
+                    lft,int(10) unsigned,NO
+                    rght,int(10) unsigned,NO
+                    tree_id,int(10) unsigned,NO
+                    level,int(10) unsigned,NO
+                    article_id,int(11),NO
+                    feedback,longtext,YES
+                    value,double,NO
```
brianhw: This covers many of the types.  Are there some others that should be added?

brianhw (commented):
Looks good -- a few nits.

edx/analytics/tasks/database_imports.py
HassanJaveed84: The class name needs to be updated. 
The workflow does indeed accept 'overwrite' parameter defined in 'ImportMysqlToVerticaTask'.
'LoadMysqlToVerticaTableTask' already inherits OverwriteOutputMixin from VerticaCopyTask.

edx/analytics/tasks/database_imports.py
HassanJaveed84: Yes, it should be try-finally.

edx/analytics/tasks/database_imports.py
```diff
@@ -214,6 +223,186 @@ def requires(self):
...
+    db_credentials = luigi.Parameter(
+        config_path={'section': 'database-import', 'name': 'credentials'},
+        description='Path to the external access credentials file.',
+    )
+    database = luigi.Parameter(
+        config_path={'section': 'database-import', 'name': 'database'}
+    )
+
+
+def get_mysql_query_results(credentials, database, query):
```
HassanJaveed84, line 239: Should this method be moved somewhere ?

edx/analytics/tasks/database_imports.py
```diff
@@ -214,6 +223,186 @@ def requires(self):
...
+        return "'NULL'"
+
+    @property
+    def enclosed_by(self):
+        return "''''"
+
+    @property
+    def insert_source_task(self):
+        partition_path_spec = HivePartition('dt', self.date.isoformat()).path_spec
+        destination = url_path_join(
```
HassanJaveed84, line 334: Does this path seem reasonable ?

edx/analytics/tasks/database_imports.py
```diff
@@ -214,6 +223,186 @@ def requires(self):
...
+            table_name=self.table_name,
+            credentials=self.db_credentials,
+            database=self.database,
+            destination=destination,
+            overwrite=self.overwrite,
+            mysql_delimiters=True,
+        )
+
+    @property
+    def default_columns(self):
```
HassanJaveed84, line 351: Do we want a default column ?

edx/analytics/tasks/tests/acceptance/test_database_import.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def validate_output(self):
+        with self.vertica.cursor() as cursor:
+            expected_output_csv = os.path.join(
+                self.data_dir,
+                'output',
+                'database_import',
+                'expected_certificates_generatedcertificate.csv'
+            )
+            expected = pandas.read_csv(expected_output_csv, parse_dates=[11,12])
```
HassanJaveed84: Again, unable to get the dates parsed with `parse_dates=True`.
We can also remove and call frame.convert_objects.

edx/analytics/tasks/tests/acceptance/test_database_import.py
```diff
@@ -0,0 +1,51 @@
...
+            expected.fillna('', inplace=True)
```
HassanJaveed84: pandas filled empty values with Nan, this replaces them with empty strings.

edx/analytics/tasks/tests/acceptance/test_database_import.py
```diff
@@ -0,0 +1,51 @@
...
+
+            cursor.execute(
+                "SELECT * FROM {schema}.certificates_generatedcertificate".format(schema=self.vertica.schema_name)
+            )
+            response = cursor.fetchall()
+            certificates_generatedcertificate = pandas.DataFrame(response, columns=list(expected.columns))
+            certificates_generatedcertificate = certificates_generatedcertificate.convert_objects(convert_numeric=True)
```
HassanJaveed84: The expected frame stored grade as float, and the actual data frame used object for the grade column.

edx/analytics/tasks/tests/test_database_imports.py
HassanJaveed84: This test needs to be updated. vertica_compliant_schema method now calls get_mysql_query_results to generate the table schema. I was hoping I could somehow mock get_mysql_query_results function. 
Is there an example in the repo that mocks a function in another module ?

I will add more types.

edx/analytics/tasks/database_imports.py
brianhw: Seems reasonable to me.  Thanks for separating by database as well as table name.

edx/analytics/tasks/database_imports.py
brianhw: Main issue with adding one is what to call it, and how to ensure that it's not already defined in one of the tables.   But if we wanted something like that:  maybe "warehouse_load_timestamp".

edx/analytics/tasks/tests/test_database_imports.py
brianhw: There are several examples of patching in the unit tests.  The is from test_pathutil.py, for example:

```
@patch('edx.analytics.tasks.pathutil.boto.connect_s3')
def test_requires(self, connect_s3_mock):
    s3_conn_mock = connect_s3_mock.return_value
    bucket_mock = s3_conn_mock.get_bucket.return_value

    class FakeKey(object):
        """A test double of the structure returned by boto when listing keys in an S3 bucket."""
        def __init__(self, path):
            self.key = path
            self.size = 10

    bucket_mock.list.return_value = [FakeKey(path) for path in self.SAMPLE_KEY_PATHS]
```

edx/analytics/tasks/database_imports.py
```diff
@@ -214,6 +223,186 @@ def requires(self):
...
+        results = cursor.fetchall()
+    except:
+        raise
+    finally:
+        connection.close()
+
+    return results
+
+
+class LoadMysqlToVerticaTableTask(MysqlQueryTaskMixin, VerticaCopyTask):
```
mulby: It just occurred to me that this will break if the schema of the table changes. Perhaps we have to pivot our strategy to:
1. Load a temporary copy of the table "foo_loading_2016_11_18"
2. If it's successful swap the copy with the real one

HassanJaveed84: What if we override `def create_table` method for this task to always drop before creating ?

mulby: I think that would break any queries that users were making using the table

HassanJaveed84: That's true. I had not considered that.

mulby: That said... we could address this in a future PR given that this is somewhat of an experiment. I'm just nervous we are going to have to be fairly frequently performing manual interventions whenever a table changes.

HassanJaveed84:
Facing this really annoying issue of getting the acceptance test to pass because of object types.
When the data frame is constructed using vertica's response, the 'date' type column is stored as datetime.date within the frame.

When the expected frame is constructed, with parse_dates enabled, the 'date' is stored as pandas.tslib.Timestamp with a '00:00:00' appended to the date. This causes the test to fail.
If we don't parse the dates, the date is stored as type 'str', the test still fails. 

Have you ever dealt with something like this before ?

brianhw:
I've not really used pandas myself in doing acceptance test validation.  I'm wondering if it is causing more difficulty than help.

HassanJaveed84:

mulby (approved):
LGTM - we should file a separate ticket for addressing the schema change issue

HassanJaveed84:
Created one: https://openedx.atlassian.net/browse/AN-8126

edx/analytics/tasks/database_imports.py
brianhw: This question got me thinking.  I've had in the back of my mind for a long time to reorganize the code better.  One part of that is to separate out the warehouse-loading workflows from the insights-loading workflows.  We've generally been pretty good about this, putting the former into load_internal_reporting_xxx files.  

So I see two things here.  One is that this mysql.connector-specific function could very reasonably go into mysql_load.py.  But I'm wondering if the other code would reasonably go into a separate load_internal_reporting_ kind of file, instead of database_imports.   It seems very Vertica-specific.

brianhw:
Looks good to me!  👍 

Only question I have now is whether to take the time now to separately Vertica-specific code into a separate file rather than leaving in database_imports.   That could be done in a separate PR, but I think it would help to keep our code better organized.

HassanJaveed84:
The acceptance tests passed, merging this.
I would reorganize vertica-specific code in a separate PR.

PR #314, opened 2016-11-15
Incremental calculation of user's location

brianhw:
This moves calculation of user location into an incremental framework.  

The first part is to first calculate historical data that stores each user's IP address on the last interaction with a course on a particular day. 

LastDailyIpAddressOfUserTask --local-scheduler \
  --interval $(date +%Y-%m-%d -d "$FROM_DATE")-$(date +%Y-%m-%d -d "$TO_DATE") \
  --n-reduce-tasks $NUM_REDUCE_TASKS \
  --warehouse-path s3://edx-analytics-scratch/brian/test-incremental-location

Calculations of location per user and location counts per course are calculated with the same entry points as before (e.g. InsertToMysqlCourseEnrollByCountryWorkflow), with the addition of an --overwrite-n-days parameter that specifies the number of days for which to recalculate LastDailyIpAddressOfUserTask before doing the rest of the aggregations.

Changes mainly involve the splitting out the LastDailyIpAddressOfUserTask functionality from  LastCountryOfUser so that it can be calculated incrementally.  LastCountryOfUser continues to find the last IP address for each user independent of the course, but the course_id is in the historical data for when we want to change this.

Additional work was done to encapsulate geolocation functionality more completely into GeolocationMixin.

### DIFF
--- a/edx/analytics/tasks/location_per_course.py
+++ b/edx/analytics/tasks/location_per_course.py
@@ -1,93 +1,334 @@
 """
 Determine the number of users in each country are enrolled in each course.
 """
+import datetime
 import logging
 import textwrap
+from collections import defaultdict
 
 import luigi
 from luigi.hive import HiveQueryTask, ExternalHiveTask
+from luigi.parameter import DateIntervalParameter
 
 from edx.analytics.tasks.database_imports import ImportStudentCourseEnrollmentTask, ImportAuthUserTask
 from edx.analytics.tasks.database_imports import ImportIntoHiveTableTask
-from edx.analytics.tasks.mapreduce import MapReduceJobTask, MapReduceJobTaskMixin
+from edx.analytics.tasks.decorators import workflow_entry_point
+from edx.analytics.tasks.mapreduce import MapReduceJobTask, MapReduceJobTaskMixin, MultiOutputMapReduceJobTask
 from edx.analytics.tasks.mysql_load import MysqlInsertTask
-from edx.analytics.tasks.pathutil import EventLogSelectionMixin, EventLogSelectionDownstreamMixin
-from edx.analytics.tasks.url import ExternalURL, get_target_from_url, url_path_join
+from edx.analytics.tasks.pathutil import (
+    PathSelectionByDateIntervalTask,
+    EventLogSelectionMixin,
+    EventLogSelectionDownstreamMixin,
+)
+from edx.analytics.tasks.url import ExternalURL, get_target_from_url, url_path_join, UncheckedExternalURL
 from edx.analytics.tasks.util.geolocation import (
     GeolocationMixin, GeolocationDownstreamMixin, UNKNOWN_COUNTRY, UNKNOWN_CODE,
 )
 from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
 from edx.analytics.tasks.util import eventlog
 from edx.analytics.tasks.util.hive import WarehouseMixin, hive_database_name
-from edx.analytics.tasks.decorators import workflow_entry_point
+from edx.analytics.tasks.util.record import Record, StringField
 
 log = logging.getLogger(__name__)
 
 
-class LastCountryOfUserMixin(
+class LastIpAddressRecord(Record):
+    """
+    Store information about last IP address observed for a given user in a given course.
+
+    Values are not written to a database, so string lengths are not specified.
+    """
+    timestamp = StringField(description='Timestamp of last event by user in a course.')
+    ip_address = StringField(description='IP address recorded on last event by user in a course.')
+    username = StringField(description='Username recorded on last event by user in a course.')
+    course_id = StringField(description='Course ID recorded on last event by user in a course.')
+
+
+class LastDailyIpAddressOfUserTask(
         WarehouseMixin,
-        MapReduceJobTaskMixin,
-        EventLogSelectionDownstreamMixin,
-        GeolocationDownstreamMixin,
-        OverwriteOutputMixin):
+        OverwriteOutputMixin,
+        EventLogSelectionMixin,
+        MultiOutputMapReduceJobTask):
     """
-    Defines parameters for LastCountryOfUser task and downstream tasks that require it.
+    Task to extract IP address information from eventlogs over a given interval.
 
-    """
-    pass
-
-
-class LastCountryOfUser(LastCountryOfUserMixin, EventLogSelectionMixin, GeolocationMixin, MapReduceJobTask):
-    """
-    Identifies the country of the last IP address associated with each user.
-
-    Uses :py:class:`LastCountryOfUserMixin` to define parameters, :py:class:`EventLogSelectionMixin`
-    to define required input log files, and :py:class:`GeolocationMixin` to provide geolocation setup.
-
+    This would produce a different output file for each day within the interval
+    containing that day's last IP address for each user only.  If there are no
+    events on a given day, an empty output file is produced for that day.
     """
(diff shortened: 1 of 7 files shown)


brianhw:
@mulby @HassanJaveed84 Still planning to review and update the unit tests, but this is ready for initial review.  (Acceptance tests passed and edge data validated.)

edx/analytics/tasks/location_per_course.py
```diff
@@ -24,69 +27,285 @@
 log = logging.getLogger(__name__)
 
 
-class LastCountryOfUserMixin(
+class LastDailyAddressOfUserTask(
```
mulby: Maybe `LastDailyIpAddressOfUserTask`? At first I thought this meant mailing address...

edx/analytics/tasks/location_per_course.py
```diff
@@ -24,69 +27,285 @@
...
+        # It's okay if it isn't.  Not sure if there are particular
+        # types of course interaction we care about, but we might want
+        # to only collect the course_id off of explicit events, and
+        # ignore implicit events as not being "real" interactions with
+        # course content.  Or maybe we add a flag indicating explicit
+        # vs. implicit, so that this can be better teased apart.
+        # (E.g.  we use the latest explicit event for a course, but if
+        # there are none, then use the latest implicit event for the
+        # course, and if there are none, then use the latest overall
+        # event.
```
mulby: nit: missing closing paren

edx/analytics/tasks/location_per_course.py
```diff
@@ -24,69 +27,285 @@
...
+        last_timestamp = defaultdict()
+        for value in values:
+            (timestamp, ip_address, course_id, username) = value
+
+            # We are storing different IP addresses depending on the username
+            # *and* the course.  This anticipates a future requirement to provide
+            # different countries depending on which course.
+            last_key = (username, course_id)
+
+            last_time = last_timestamp[last_key] if last_key in last_timestamp else ""
```
mulby: nit: replace with `last_time = last_timestamp.get(last_key, "")`

edx/analytics/tasks/location_per_course.py
```diff
@@ -24,69 +27,285 @@
...
+            if timestamp > last_time:
+                last_ip[last_key] = ip_address
+                last_timestamp[last_key] = timestamp
+
+        # Now output the resulting "last" values for each key.
+        for last_key, ip_address in last_ip.iteritems():
+            timestamp = last_timestamp[last_key]
+            username, course_id = last_key
+            value = [timestamp, ip_address, username, course_id]
+            output_file.write('\t'.join([unicode(field).encode('utf8') for field in value]))
```
mulby: Perhaps initialize a record here and use it to serialize the data to the output file?

edx/analytics/tasks/location_per_course.py
```diff
@@ -24,69 +27,285 @@
...
+            ),
+        )
+
+    def downstream_input_tasks(self):
+        """
+        MultiOutputMapReduceJobTask returns marker as output.
+        This method returns the external tasks which can then be used as input in other jobs.
+        Note that this method does not verify the existence of the underlying urls. It assumes that
+        there is an output file for every date within the interval. Any MapReduce job
+        which uses this as input would fail if there is missing data for any date within the interval.
```
mulby: I would make a note here that we fill in empty files for dates that have no data

edx/analytics/tasks/location_per_course.py
```diff
@@ -24,69 +27,285 @@
...
+    def requires_local(self):
+        if not self.cached_local_requirements:
+            requirements = super(LastCountryOfUser, self).requires_local()
+            # Default is an empty list, but assume that any real data added is done
+            # so as a dict.
+            if not requirements:
+                requirements = {}
+
+            if self.overwrite_n_days > 0:
+                overwrite_interval = DateIntervalParameter().parse('{}-{}'.format(
```
mulby: nit: perhaps use `luigi.date_interval.Custom(self.overwrite_from_date, self.interval.date_b)`?

edx/analytics/tasks/location_per_course.py
```diff
@@ -24,69 +27,285 @@
...
+    # Calculate requirements once.
+    cached_local_requirements = None
+    cached_hadoop_requirements = None
+
+    def __init__(self, *args, **kwargs):
+        super(LastCountryOfUser, self).__init__(*args, **kwargs)
+
+        self.overwrite_from_date = self.interval.date_b - datetime.timedelta(days=self.overwrite_n_days)
+
+    def requires_local(self):
```
mulby, line 239: I recommend including a comment here explaining why we need this separation between hadoop and local dependencies. Particularly given the fact that the "local" dependencies aren't actually used at all by this task.

edx/analytics/tasks/location_per_course.py
```diff
@@ -24,69 +27,285 @@
...
+        # This is different from remove_output_on_overwrite()
+        # in that it also removes the target directory if
+        # the success marker file is missing.
+        if not self.complete() and output_target.exists():
+            output_target.remove()
+        super(LastCountryOfUser, self).run()
+
+    def init_local(self):
+        super(LastCountryOfUser, self).init_local()
+        # This makes sure the attempted-removal flag is set.
```
mulby: _sigh_ - what a tangled web!

edx/analytics/tasks/location_per_course.py
```diff
@@ -24,69 +27,285 @@
...
+        self.remove_output_on_overwrite()
+
+    def mapper(self, line):
+        (
+            timestamp,
+            ip_address,
+            username,
+            _course_id,
+        ) = line.split('\t')
```
mulby: Using a Record above would also clean this up a bit...

mulby (approved):
Just nits

edx/analytics/tasks/location_per_course.py
brianhw: Done.

brianhw: Done

edx/analytics/tasks/location_per_course.py
brianhw: Done.

edx/analytics/tasks/location_per_course.py
brianhw: I wasn't originally using a Record because I mostly think of them as useful for output to databases (i.e. providing schema).  But it's generally useful for communicating between any two tasks, really.

edx/analytics/tasks/location_per_course.py
brianhw: done.

edx/analytics/tasks/location_per_course.py
brianhw: done.

edx/analytics/tasks/location_per_course.py
brianhw: Yes, that looks better.

edx/analytics/tasks/location_per_course.py
brianhw: I'm moving the call to run(), like it is everywhere else.  Not sure why it was in init_local() originally, though luigi calls one and then the other, so it should make no difference.

brianhw:
This should be ready for final review.  I'll then squash and merge, after confirming that it still works (one more day).  

@HassanJaveed84 @mulby

mulby (approved):

PR #315, opened 2016-11-23
Close Vertica connection if the connection is not passed.

HassanJaveed84:
'ImportMysqlToVerticaTask' fails when it checks for completeness. Luigi calls exists() on the all dependencies, which opens a new connection per task(per table).
The workflow fails with 
"ConnectionError: Severity: FATAL, Message: New session rejected due to limit, already 50 sessions active"

This PR aims to resolve this issue by closing the connection. I'm open to suggestions.

@mulby @brianhw


mulby (approved):
LGTM

brianhw (approved):
LGTM too.  👍

PR #317, opened 2016-11-25
Remove intermediate argument from user location.

brianhw:
This includes manual testing of the ExternalLastCountryOfUserToHiveTask implementation.

@mulby @HassanJaveed84 
fyi: @tobz

### DIFF
--- a/edx/analytics/tasks/location_per_course.py
+++ b/edx/analytics/tasks/location_per_course.py
@@ -26,8 +26,13 @@ from edx.analytics.tasks.util.geolocation import (
 )
 from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
 from edx.analytics.tasks.util import eventlog
-from edx.analytics.tasks.util.hive import WarehouseMixin, hive_database_name
-from edx.analytics.tasks.util.record import Record, StringField
+from edx.analytics.tasks.util.hive import (
+    WarehouseMixin,
+    hive_database_name,
+    BareHiveTableTask,
+    HivePartitionTask,
+)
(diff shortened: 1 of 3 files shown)


edx/analytics/tasks/location_per_course.py
```diff
@@ -409,42 +424,39 @@ def requires(self):
...
-        yield ExternalURL(
-            url=url_path_join(self.warehouse_path, 'last_country_of_user', self.partition_spec()) + '/'
-        )
+    @property
+    def data_task(self):
+        url = super(ExternalLastCountryOfUserToHiveTask, self).data_task.output().path
+        return ExternalURL(url.rstrip('/') + '/')
 
 
 class InsertToMysqlLastCountryOfUserTask(LastCountryOfUserDownstreamMixin, MysqlInsertTask):
```
mulby, line 449: Not a blocker, but I think we can delete this now... we shouldn't need to load this data into the result store now that we've retired pentaho.

edx/analytics/tasks/location_per_course.py
```diff
@@ -409,42 +424,39 @@ def requires(self):
...
+    interval_start = None
+    interval_end = None
     date = luigi.DateParameter()
 
-    @property
-    def partition_date(self):
-        return self.date.isoformat()  # pylint: disable=no-member
+    def __init__(self, *args, **kwargs):
+        self.interval_start = kwargs.get('date')
+        self.interval_end = kwargs.get('date')
```
mulby, line 436: hrmmm, this assumes the param is in the kwargs and not the args...

mulby (approved):
No blockers. LGTM

PR #319, opened 2016-12-01
Address table schema changes for database to warehouse import.

HassanJaveed84:
Leaving some questions/comments for feedback.

Do we need to do any validation checks before we do the swapping ?

### DIFF
--- a/edx/analytics/tasks/load_internal_reporting_database.py
+++ b/edx/analytics/tasks/load_internal_reporting_database.py
@@ -12,6 +12,7 @@ from edx.analytics.tasks.url import url_path_join, ExternalURL
 from edx.analytics.tasks.util.hive import WarehouseMixin, HivePartition
 from edx.analytics.tasks.vertica_load import VerticaCopyTask
 from edx.analytics.tasks.mysql_load import get_mysql_query_results
+from edx.analytics.tasks.load_warehouse import SchemaManagementTask
 
 log = logging.getLogger(__name__)
 
@@ -132,6 +133,56 @@ class LoadMysqlToVerticaTableTask(MysqlToVerticaTaskMixin, VerticaCopyTask):
         return None
 
 
+class PreImportDatabaseTask(SchemaManagementTask):
+    """
+    Task needed to run before importing database into warehouse.
+    """
+    priority = 100
+
+    @property
+    def queries(self):
+        return [
+            "DROP SCHEMA IF EXISTS {schema} CASCADE;".format(schema=self.schema_loading),
+            "CREATE SCHEMA IF NOT EXISTS {schema}".format(schema=self.schema_loading),
+        ]
+
+    @property
+    def marker_name(self):
+        return 'pre_database_import_{schema}_{date}'.format(
+            schema=self.schema,
+            date=self.date.strftime('%Y-%m-%d')
+        )
+
+
+class PostImportDatabaseTask(SchemaManagementTask):
+    """
+    Task needed to run after importing database into warehouse.
+    """
+    priority = -100
+
+    @property
+    def queries(self):
+        return [
+            "DROP SCHEMA IF EXISTS {schema} CASCADE;".format(schema=self.schema),
+            "ALTER SCHEMA {schema_loading} RENAME TO {schema};".format(
+                schema_loading=self.schema_loading,
+                schema=self.schema
+            ),
+            "GRANT USAGE ON SCHEMA {schema} TO {roles};".format(schema=self.schema, roles=self.vertica_roles),
+            "GRANT SELECT ON ALL TABLES IN SCHEMA {schema} TO {roles};".format(
+                schema=self.schema,
+                roles=self.vertica_roles
+            ),
+        ]
+
+    @property
+    def marker_name(self):
+        return 'post_database_import_{schema}_{date}'.format(
+            schema=self.schema,
+            date=self.date.strftime('%Y-%m-%d')
+        )
+
+
 class ImportMysqlToVerticaTask(MysqlToVerticaTaskMixin, luigi.WrapperTask):
     """Provides entry point for importing a mysql database into Vertica."""
 
@@ -156,6 +207,10 @@ class ImportMysqlToVerticaTask(MysqlToVerticaTaskMixin, luigi.WrapperTask):
         default=(),
     )
 
+    marker_schema = luigi.Parameter(
+        description='The marker schema to which to write the marker table.'
+    )
+
     def __init__(self, *args, **kwargs):
         super(ImportMysqlToVerticaTask, self).__init__(*args, **kwargs)
         self.table_list = []
@@ -171,15 +226,33 @@ class ImportMysqlToVerticaTask(MysqlToVerticaTaskMixin, luigi.WrapperTask):
             results = get_mysql_query_results(self.db_credentials, self.database, 'show tables')
             self.table_list = [result[0].strip() for result in results]
 
+        pre_import_task = PreImportDatabaseTask(
+            date=self.date,
+            schema=self.schema,
+            credentials=self.credentials,
+            marker_schema=self.marker_schema,
+            overwrite=self.overwrite
+        )
+        yield pre_import_task
+
         for table_name in self.table_list:
             if not self.should_exclude_table(table_name):
                 yield LoadMysqlToVerticaTableTask(
                     credentials=self.credentials,
-                    schema=self.schema,
+                    schema=pre_import_task.schema_loading,
                     db_credentials=self.db_credentials,
                     database=self.database,
                     warehouse_path=self.warehouse_path,
                     table_name=table_name,
                     overwrite=self.overwrite,
                     date=self.date,
+                    marker_schema=self.marker_schema,
                 )
+
+        yield PostImportDatabaseTask(
+            date=self.date,
+            schema=self.schema,
+            credentials=self.credentials,
+            marker_schema=self.marker_schema,
+            overwrite=self.overwrite
+        )
--- a/edx/analytics/tasks/load_warehouse.py
+++ b/edx/analytics/tasks/load_warehouse.py
@@ -60,10 +60,16 @@ class SchemaManagementTask(VerticaCopyTaskMixin, luigi.Task):
 
     date = luigi.DateParameter()
 
+    roles = luigi.Parameter(
+        is_list=True,
+        config_path={'section': 'vertica-export', 'name': 'roles'},
+    )
+
     def __init__(self, *args, **kwargs):
         super(SchemaManagementTask, self).__init__(*args, **kwargs)
         self.schema_last = self.schema + '_last'
         self.schema_loading = self.schema + '_loading'
+        self.vertica_roles = ','.join(self.roles)
 
     @property
     def queries(self):
@@ -108,8 +114,11 @@ class SchemaManagementTask(VerticaCopyTaskMixin, luigi.Task):
                 log.debug(query)
                 connection.cursor().execute(query)
             except vertica_python.errors.Error as err:
-                if (type(err) is vertica_python.errors.MissingRelation) or ('Sqlstate: 42V01' in err.args[0]):
-                    # If so, then our query error failed because the table doesn't exist.
+                is_missing_relation = type(err) is vertica_python.errors.MissingRelation
+                is_missing_table = 'Sqlstate: 42V01' in err.args[0]
+                is_missing_schema = 'Sqlstate: 3F000' in err.args[0]
+                if is_missing_relation or is_missing_table or is_missing_schema:
+                    # If so, then our query error failed because the schema or table doesn't exist.
                     pass
                 else:
                     raise
@@ -251,8 +260,11 @@ class PostLoadWarehouseTask(SchemaManagementTask):
                 schema_loading=self.schema_loading,
                 schema=self.schema
             ),
-            "GRANT USAGE ON SCHEMA {schema} TO analyst;".format(schema=self.schema),
-            "GRANT SELECT ON ALL TABLES IN SCHEMA {schema} TO analyst;".format(schema=self.schema),
+            "GRANT USAGE ON SCHEMA {schema} TO {roles};".format(schema=self.schema, roles=self.vertica_roles),
+            "GRANT SELECT ON ALL TABLES IN SCHEMA {schema} TO {roles};".format(
+                schema=self.schema,
+                roles=self.vertica_roles
+            ),
         ]
 
     @property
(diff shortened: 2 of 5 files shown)


edx/analytics/tasks/load_internal_reporting_database.py
```diff
@@ -132,6 +133,49 @@ def auto_primary_key(self):
         return None
 
 
+class PreImportDatabaseTask(SchemaManagementTask):
```
HassanJaveed84, line 136: This is a duplicate of PreLoadWarehouseTask apart from the property change 'marker_name'.
Should this be refactored into a single class where marker_name can be a parameter ?
or
As we will be using a different schema for marker table for database import(we use 'production_marker' for warehouse loading), even if the marker_name is same for load_warehouse and database import, it would not matter.
Suggestions ?

edx/analytics/tasks/load_internal_reporting_database.py
```diff
@@ -132,6 +133,49 @@ def auto_primary_key(self):
...
+            "DROP SCHEMA IF EXISTS {schema} CASCADE;".format(schema=self.schema_loading),
+            "CREATE SCHEMA IF NOT EXISTS {schema}".format(schema=self.schema_loading),
+        ]
+
+    @property
+    def marker_name(self):
+        return 'pre_database_import_' + self.date.strftime('%Y-%m-%d')
+
+
+class PostImportDatabaseTask(SchemaManagementTask):
```
HassanJaveed84, line 157: Similar to what I mentioned above. 
This also includes an extra query 'DROP SCHEMA'

edx/analytics/tasks/load_warehouse.py
```diff
@@ -118,6 +118,7 @@ def run(self):
         connection = self.output().connect()
 
         try:
+            self.output().create_marker_schema(connection)
```
HassanJaveed84: We call ```init_touch``` do delete the marker table entry and then call ```self.output().touch()```(which creates the marker_schema and marker_table if they do no exist as well as the marker entry). This seems like a design issue to me.

I had trouble getting the acceptance test to pass as the marker_schema does not exist before the task tries to clear the marker table entry. Thoughts ?

edx/analytics/tasks/load_internal_reporting_database.py
```diff
@@ -132,6 +133,49 @@ def auto_primary_key(self):
...
+class PreImportDatabaseTask(SchemaManagementTask):
+    """
+    Task needed to run before importing database into warehouse.
+    """
+    priority = 100
+
+    @property
+    def queries(self):
+        return [
+            "DROP SCHEMA IF EXISTS {schema} CASCADE;".format(schema=self.schema_last),
```
brianhw: Just curious, but why do we also drop the output schema at the beginning?

edx/analytics/tasks/load_internal_reporting_database.py
```diff
@@ -132,6 +133,49 @@ def auto_primary_key(self):
...
+    def queries(self):
+        return [
+            "ALTER SCHEMA {schema} RENAME TO {schema_last};".format(schema=self.schema, schema_last=self.schema_last),
+            "ALTER SCHEMA {schema_loading} RENAME TO {schema};".format(
+                schema_loading=self.schema_loading,
+                schema=self.schema
+            ),
+            "GRANT USAGE ON SCHEMA {schema} TO analyst;".format(schema=self.schema),
+            "GRANT SELECT ON ALL TABLES IN SCHEMA {schema} TO analyst;".format(schema=self.schema),
+            "DROP SCHEMA {schema} CASCADE;".format(schema=self.schema_last),
```
brianhw: Why do we drop the last schema at the end, rather than before renaming schema to schema_last?

edx/analytics/tasks/load_internal_reporting_database.py
brianhw: I think it is small enough -- and different enough -- to allow it to be a near-duplicate for now.  But I think it would be helpful to have comments in both places pointing out the differences in dealing with marker tables.

edx/analytics/tasks/load_internal_reporting_database.py
```diff
@@ -171,15 +217,32 @@ def requires(self):
...
+            marker_schema=self.marker_schema,
+            overwrite=self.overwrite
+        )
+
         for table_name in self.table_list:
             if not self.should_exclude_table(table_name):
                 yield LoadMysqlToVerticaTableTask(
                     credentials=self.credentials,
-                    schema=self.schema,
+                    schema=self.schema + '_loading',
```
brianhw: Not sure how to handle this better, but seems like an implementational detail of SchemaManagementTask has leaked out?

edx/analytics/tasks/load_warehouse.py
brianhw: Sorry that I don't understand (though it may be the hour).  Is this a particularity of the acceptance tests?  That is, is it the testing mechanism that is screwing up the marker table?  Or is this a broader issue?

brianhw (commented):
Looks good in general!

edx/analytics/tasks/load_warehouse.py
HassanJaveed84: I think it's a broader issue and without this change we would need to manually create the marker schema.

edx/analytics/tasks/load_internal_reporting_database.py
HassanJaveed84: If we are not planning to keep the 'last' schema, I ought to remove this.

edx/analytics/tasks/load_internal_reporting_database.py
HassanJaveed84: I think we planned to 'not' keep the last schema, but only the latest one, hence the rename, then the deletion.

edx/analytics/tasks/load_internal_reporting_database.py
```diff
@@ -132,6 +133,49 @@ def auto_primary_key(self):
...
+
+    @property
+    def queries(self):
+        return [
+            "ALTER SCHEMA {schema} RENAME TO {schema_last};".format(schema=self.schema, schema_last=self.schema_last),
+            "ALTER SCHEMA {schema_loading} RENAME TO {schema};".format(
+                schema_loading=self.schema_loading,
+                schema=self.schema
+            ),
+            "GRANT USAGE ON SCHEMA {schema} TO analyst;".format(schema=self.schema),
```
mulby: Is this GRANT hard-coded elsewhere?

We should probably make this a configurable setting (a list of roles that are granted access to Vertica schemas).

edx/analytics/tasks/load_internal_reporting_database.py
mulby: If we aren't keeping foo_last, maybe we shouldn't bother creating it at all?

That is, remove all references to a "last" schema from this file.

mulby (commented):
Mostly looks good to me

edx/analytics/tasks/load_internal_reporting_database.py
HassanJaveed84: We also have it here:
https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/load_warehouse.py#L254

edx/analytics/tasks/load_internal_reporting_database.py
```diff
@@ -132,6 +133,49 @@ def auto_primary_key(self):
...
+class PostImportDatabaseTask(SchemaManagementTask):
+    """
+    Task needed to run after importing database into warehouse.
+    """
+    priority = -100
+
+    @property
+    def queries(self):
+        return [
+            "ALTER SCHEMA {schema} RENAME TO {schema_last};".format(schema=self.schema, schema_last=self.schema_last),
```
brianhw: Right.  So if we're not going to maintain _last, then this line also needs to be removed.

edx/analytics/tasks/load_internal_reporting_database.py
brianhw: Agreed.

edx/analytics/tasks/load_internal_reporting_database.py
brianhw: Do we have access here to self.schema_loading instead of trying to construct it?

edx/analytics/tasks/load_warehouse.py
brianhw: I was wondering if the alternative is to change the way the task tries to clear the marker table entry so that it's not an error if the marker table doesn't exist yet.  It doesn't seem to me that it should be a cause of failure.

edx/analytics/tasks/load_internal_reporting_database.py
```diff
@@ -156,6 +200,8 @@ class ImportMysqlToVerticaTask(MysqlToVerticaTaskMixin, luigi.WrapperTask):
         default=(),
     )
 
+    marker_schema = luigi.Parameter()
```
brianhw: Eventually add description to parameters.

edx/analytics/tasks/load_internal_reporting_database.py
brianhw: Would we have any reason to want the grants to be different from the regular warehouse loads?   Might be good to have a single configuration setting for both of these.

edx/analytics/tasks/load_internal_reporting_database.py
brianhw: And an extra rename.

edx/analytics/tasks/load_internal_reporting_database.py
brianhw: Sounds good.

brianhw (commented):
Just a few more comments.

edx/analytics/tasks/load_internal_reporting_database.py
HassanJaveed84: We don't. What about something like :
```python
        pre_import_task =  PreImportDatabaseTask(
            date=self.date,
            schema=self.schema,
            credentials=self.credentials,
            marker_schema=self.marker_schema,
            overwrite=self.overwrite
        )
       yield pre_import_task
```
We can then use ```pre_import_task.schema_loading``` for LoadMysqlToVerticaTableTask.

HassanJaveed84:
This is ready for another review.

edx/analytics/tasks/load_internal_reporting_database.py
```diff
@@ -132,6 +133,47 @@ def auto_primary_key(self):
...
+    """
+    Task needed to run after importing database into warehouse.
+    """
+    priority = -100
+
+    @property
+    def queries(self):
+        return [
+            "DROP SCHEMA IF EXISTS {schema} CASCADE;".format(schema=self.schema),
+            "ALTER SCHEMA {schema_loading} RENAME TO {schema};".format(
```
mulby, line 167: I think we can do an atomic schema swap:

https://my.vertica.com/docs/7.1.x/HTML/Content/Authoring/SQLReferenceManual/Statements/ALTERSCHEMA.htm

Not sure if it operates differently than DROP followed by ALTER... which one method did we test with?

I just want to make sure the DROP won't wait for queries to finish...

mulby (approved):
One question, but overall LGTM

edx/analytics/tasks/load_internal_reporting_database.py
HassanJaveed84: I tested drop while running a query on the schema, vertica dropped the schema and the query finished as well.

mulby: Awesome!

edx/analytics/tasks/load_internal_reporting_database.py
brianhw: That sounds like a good improvement.   Especially since we are creating and yielding that task anyway.

brianhw: That sounds like a good improvement, especially since we are creating and yielding the task anyway.

edx/analytics/tasks/load_internal_reporting_database.py
```diff
@@ -132,6 +133,47 @@ def auto_primary_key(self):
...
+                schema_loading=self.schema_loading,
+                schema=self.schema
+            ),
+            "GRANT USAGE ON SCHEMA {schema} TO {roles};".format(schema=self.schema, roles=self.vertica_roles),
+            "GRANT SELECT ON ALL TABLES IN SCHEMA {schema} TO {roles};".format(schema=self.schema, roles=self.vertica_roles),
+        ]
+
+    @property
+    def marker_name(self):
+        return 'post_database_import_' + self.date.strftime('%Y-%m-%d')
```
brianhw: This means we can do one run a day, right?  Do we need to at least include the schema name in this marker_name, so that tests to different output schemas can be run without having to use a different marker_schema?

brianhw (approved):
LGTM.  Just one change that you suggested, and one question.

edx/analytics/tasks/load_warehouse.py
```diff
@@ -108,8 +114,8 @@ def init_touch(self, connection):
                 log.debug(query)
                 connection.cursor().execute(query)
             except vertica_python.errors.Error as err:
-                if (type(err) is vertica_python.errors.MissingRelation) or ('Sqlstate: 42V01' in err.args[0]):
-                    # If so, then our query error failed because the table doesn't exist.
+                if (type(err) is vertica_python.errors.MissingRelation) or ('Sqlstate: 42V01' in err.args[0]) or ('Sqlstate: 3F000' in err.args[0]):
```
HassanJaveed84: Any suggestions on how to break this up into multiple lines ? 
I tried following https://www.python.org/dev/peps/pep-0008/#multiline-if-statements style guide but can't seem to get 'make coverage' to stop complaining.

HassanJaveed84:
Squashed, just one question before I merge this.

edx/analytics/tasks/load_warehouse.py
mulby: Sometimes I will break up the boolean conditions and assign them to variables with shorter names:

```python
is_missing_relation = type(err) is vertica_python.errors.MissingRelation
is_42V01 = 'Sqlstate: 42V01' in err.args[0]
is_3F000 = 'Sqlstate: 3F000' in err.args[0]
if is_missing_relation or is_42V01 or is_3F000:
    print('foo')
```

Ideally we could use more human-readable variable names for the 42V01 state and the 3F000 state -> what do those actually indicate?

HassanJaveed84: 42V01 is for an undefined table, 3F000 is for invalid schema name.

PR #320, opened 2016-12-01
Ability to run arbitrary SQL scripts, and to run a defined list of them.

tobz:
This work supports AN-7229, which is designed to allow analysts to run SQL scripts after a warehouse load in a self-service fashion.

We have two tasks, `RunSqlScript` and `RunSqlScript` which allow running an arbitrary SQL script against a prescribed schema, and subsequently, run a whole slew of them, in a deterministic order.

It's a little hacky in how we schedule the individual runs from the parent task because our version of Luigi is old enough that we can't yield tasks for scheduling from the `run()` method.

### DIFF
--- a/edx/analytics/tasks/run_vertica_sql_script.py
+++ b/edx/analytics/tasks/run_vertica_sql_script.py
@@ -1,0 +1,149 @@
+"""
+Support for running a SQL script against an HP Vertica database.
+"""
+import datetime
+import logging
+
+import luigi
+import luigi.configuration
+from edx.analytics.tasks.url import ExternalURL
+from edx.analytics.tasks.util.vertica_target import VerticaTarget, CredentialFileVerticaTarget
+
+log = logging.getLogger(__name__)
+
+try:
+    import vertica_python
+    vertica_client_available = True  # pylint: disable-msg=C0103
+except ImportError:
+    log.warn('Unable to import Vertica client libraries')
+    # On hadoop slave nodes we don't have Vertica client libraries installed so it is pointless to ship this package to
+    # them, instead just fail noisily if we attempt to use these libraries.
+    vertica_client_available = False  # pylint: disable-msg=C0103
+
+
+class BaseVerticaSqlScriptTaskMixin(object):
+    """
+    Parameters for running a SQL script against an HP Vertica database.
+    """
+    date = luigi.DateParameter(
+        default=datetime.datetime.utcnow().date(),
+        description='Default is today, UTC.',
+    )
+    schema = luigi.Parameter(
+        config_path={'section': 'run-vertica-sql-script', 'name': 'schema'},
+        description='Name of the schema to which to write.',
+    )
+    credentials = luigi.Parameter(
+        config_path={'section': 'run-vertica-sql-script', 'name': 'credentials'},
+        description='Path to the external access credentials file.',
+    )
+    read_timeout = luigi.IntParameter(
+        config_path={'section': 'run-vertica-sql-script', 'name': 'read_timeout'},
+        description='Timeout in seconds for reading from a Vertica database connection.'
+    )
+    marker_schema = luigi.Parameter(
+        default=None,
+        description='Name of the schema to which to write the marker table. marker_schema would '
+        'default to the schema value if the value here is None.'
+    )
+
+    def update_id(self):
+        """
+        Unique string identifying this task run, based on the input parameters.
+        """
+        return str(self)
+
+
+class RunVerticaSqlScriptTaskMixin(BaseVerticaSqlScriptTaskMixin):
+    """
+    Parameters required to run a single SQL script against an HP Vertica database.
+    """
+    source_script = luigi.Parameter(
+        description='Path to the source script to execute.'
+    )
+    script_name = luigi.Parameter(
+        description='Unique identifier for the purposes of tracking whether or not this '
+        'script ran successfully i.e. the table created by this script, or the ticket related to it.'
+    )
+
+
+class RunVerticaSqlScriptTask(RunVerticaSqlScriptTaskMixin, luigi.Task):
+    """
+    A task for running a SQL script against an HP Vertica database.
+    """
+    required_tasks = None
+    output_target = None
+    depends_on = None
+
+    def add_dependency(self, dependency):
+      """
+      Adds a custom dependency/requirement for this task.
+
+      Note: this currently *sets* a single, custom dependency.  You cannot add multiple dependencies to this task.
+      The last dependency to be added is the only one that will stick.  It will, however, not be the only dependency,
+      as this task has a "base" set of dependencies.
+      """
+      self.depends_on = dependency
+
+    def requires(self):
+        if self.required_tasks is None:
+            self.required_tasks = {
+                'credentials': ExternalURL(url=self.credentials),
+                'source_script': ExternalURL(url=self.source_script),
+            }
+
+            if self.depends_on is not None:
+              self.required_tasks['depends_on'] = self.depends_on
+
+        return self.required_tasks
+
+    def output(self):
+        """
+        Returns a VerticaTarget representing the inserted dataset.
+        """
+        if self.output_target is None:
+            self.output_target = CredentialFileVerticaTarget(
+                credentials_target=self.input()['credentials'],
+                table=self.script_name,
+                schema=self.schema,
+                update_id=self.update_id(),
+                read_timeout=self.read_timeout,
+                marker_schema=self.marker_schema,
+            )
+
+        return self.output_target
+
+    def run(self):
+        """
+        Runs the given SQL script against the Vertica target.
+        """
+        # Make sure we can connect to Vertica.
+        self.check_vertica_availability()
+        connection = self.output().connect()
+
+        try:
+            # Set up our connection to point to the specified schema so that scripts can have unqualified
+            # table references and not necessarily need to know or care about where they're running.
+            connection.cursor().execute('SET SEARCH_PATH = {schema};'.format(schema=self.schema))
+
+            with self.input()['source_script'].open('r') as script_file:
+                # Read in our script and execute it.
+                script_body = script_file.read()
+                connection.cursor().execute(script_body)
+
+                # If we're here, nothing blew up, so mark as complete.
+                self.output().touch(connection)
+
+                connection.commit()
+                log.debug("Committed transaction.")
+        except Exception as exc:
+            log.exception("Rolled back the transaction; exception raised: %s", str(exc))
+            connection.rollback()
+            raise
+        finally:
+            connection.close()
+
+    def check_vertica_availability(self):
+        """Call to ensure fast failure if this machine doesn't have the Vertica client library available."""
+        if not vertica_client_available:
+            raise ImportError('Vertica client library not available')
--- a/edx/analytics/tasks/run_vertica_sql_scripts.py
+++ b/edx/analytics/tasks/run_vertica_sql_scripts.py
@@ -1,0 +1,97 @@
+"""
+Support for running multiple SQL scripts against an HP Vertica database in a deterministic fashion.
+"""
+import yaml
+import datetime
+import logging
+from os import path
+
+import luigi
+import luigi.configuration
+from edx.analytics.tasks.url import ExternalURL
+from edx.analytics.tasks.run_vertica_sql_script import BaseVerticaSqlScriptTaskMixin, RunVerticaSqlScriptTask
+
+
+log = logging.getLogger(__name__)
+
+
+class RunVerticaSqlScriptsTaskMixin(BaseVerticaSqlScriptTaskMixin):
+    """
+    Parameters for running multiple SQL scripts against an HP Vertica database in a deterministic fashion.
+    """
+    script_configuration = luigi.Parameter(
+        description='Path to the configuration file that specifies which scripts to run.'
+    )
+    script_root = luigi.Parameter(
+        default='',
+        description='Root directory from which the script locations in the configuration '
+        'are referenced from.'
+    )
+
+
+class RunVerticaSqlScriptsTask(RunVerticaSqlScriptsTaskMixin, luigi.WrapperTask):
+    """
+    A wrapper task for running multiple SQL scripts against an HP Vertica database in a deterministic fashion.
+
+    We use a YAML file that defines a list of scripts to run.  We run the scripts in the order they are defined.
+    By using RunVerticaSqlScriptTask, each script stores its own marker, thus allowing us to idempotently run this
+    task until all required tasks (aka our scripts) have successfully run for a given date.
+    """
+    downstream_task = None
+
+    def requires(self):
+        if self.downstream_task is None:
+            self.downstream_task = self.get_downstream_task()
+
+        if self.downstream_task is not None:
+            yield self.downstream_task
+
+    def validate_script_entry(self, script):
+      # It has to be a dictionary.
+      if not isinstance(script, dict):
+        return False
+
+      # It needs to have a name and a script location.
+      for attr in ['name', 'location']:
+        if attr not in script:
+          return False
+
+      return True
+
+    def get_downstream_task(self):
+        # If no downstream task has been set, load our configuration and generate our tasks and dependency chain.
+        if self.downstream_task is None:
+            script_conf_target = ExternalURL(url=self.script_configuration).output()
+            with script_conf_target.open('r') as script_conf_file:
+                config = yaml.safe_load(script_conf_file)
+                if config is not None and isinstance(config, dict):
+                    previous_task = None
+
+                    scripts = config.get('scripts', [])
+
+                    # Iterate over the list of scripts in the configuration file in reverse order.  We also zip a list of integers,
+                    # representing the zero-based index position of the given script in the overall list.  We iterate in reverse
+                    # in order to link each task together, using requires(), to ensure that tasks run sequentially, and in the intended
+                    # order: from the top of the file, downwards.
+                    for script in scripts:
+                        if not self.validate_script_entry(script):
+                            log.warn("encountered invalid script entry!")
+                            continue
+
+                        new_task = RunVerticaSqlScriptTask(
+                            credentials=self.credentials, schema=self.schema, marker_schema=self.marker_schema,
+                            date=self.date, read_timeout=self.read_timeout, source_script=path.join(self.script_root, script['location']),
+                            script_name=script.get('name'))
+
+                        # If we previously configured a task, set it as a dependency of this one, so it runs prior to.
+                        if previous_task is not None:
+                            new_task.add_dependency(previous_task)
+
+                        # Mark this as the previously-created task.
+                        previous_task = new_task
+
+                    self.downstream_task = previous_task
+
+        # If a downstream task has been set, yield it, triggering Luigi to schedule our scripts.
+        if self.downstream_task is not None:
+            yield self.downstream_task
--- a/edx/analytics/tasks/tests/test_run_vertica_sql_script.py
+++ b/edx/analytics/tasks/tests/test_run_vertica_sql_script.py
@@ -1,0 +1,92 @@
+"""
+Ensure we can write to Vertica data sources.
+"""
+from __future__ import absolute_import
+
+import textwrap
+
+import luigi
+import luigi.task
+
+from mock import call
+from mock import MagicMock
+from mock import patch
+from mock import sentinel
+
+from edx.analytics.tasks.run_vertica_sql_script import RunVerticaSqlScriptTask
+from edx.analytics.tasks.tests import unittest
+from edx.analytics.tasks.tests.target import FakeTarget
+from edx.analytics.tasks.tests.config import with_luigi_config
+
+
+class RunVerticaSqlScriptTaskTest(unittest.TestCase):
+    """
+    Ensure we can connect to and write data to Vertica data sources.
+    """
+
+    def setUp(self):
+        patcher = patch('edx.analytics.tasks.run_vertica_sql_script.vertica_python.vertica')
+        self.mock_vertica_connector = patcher.start()
+        self.addCleanup(patcher.stop)
+
+    def create_task(self, credentials=None, source_script=None):
+        """
+        Emulate execution of a generic RunVerticaSqlScriptTask.
+        """
+        # Make sure to flush the instance cache so we create a new task object.
+        luigi.task.Register.clear_instance_cache()
+        task = RunVerticaSqlScriptTask(
+            credentials=sentinel.ignored,
+            script_name='my simple script',
+            source_script=sentinel.ignored,
+        )
+
+        if not credentials:
+            credentials = '''\
+                {
+                    "host": "db.example.com",
+                    "port": 5433,
+                    "user": "exampleuser",
+                    "password": "example password"
+                }'''
+
+        # This SQL doesn't actually run, but I've used real SQL to provide context. :)
+        source = '''
+        DELETE TABLE my_schema.my_table;
+        CREATE TABLE my_schema.my_table AS SELECT foo, bar, baz FROM my_schema.another_table;
+        '''
+
+        fake_input = {
+            'credentials': FakeTarget(value=textwrap.dedent(credentials)),
+            'source_script': FakeTarget(value=textwrap.dedent(source))
+        }
+
+        fake_output = MagicMock(return_value=self.mock_vertica_connector)
+        self.mock_vertica_connector.marker_schema = "name_of_marker_schema"
+        self.mock_vertica_connector.marker_table = "name_of_marker_table"
+
+        task.input = MagicMock(return_value=fake_input)
+        task.output = fake_output
+        return task
+
+    def test_run_with_default_credentials(self):
+        self.create_task(credentials='{}').run()
+
+    def test_run(self):
+        self.create_task().run()
+        mock_conn = self.mock_vertica_connector.connect()
+        self.assertTrue(mock_conn.cursor().execute.called)
+        self.assertFalse(mock_conn.rollback.called)
+        self.assertTrue(mock_conn.commit.called)
+        self.assertTrue(mock_conn.close.called)
+
+    def test_run_with_failure(self):
+        task = self.create_task()
+        task.output().touch = MagicMock(side_effect=Exception("Failed to update marker"))
+        with self.assertRaises(Exception):
+            task.run()
+        mock_conn = self.mock_vertica_connector.connect()
+        self.assertTrue(mock_conn.cursor().execute.called)
+        self.assertTrue(mock_conn.rollback.called)
+        self.assertFalse(mock_conn.commit.called)
+        self.assertTrue(mock_conn.close.called)
--- a/edx/analytics/tasks/tests/test_run_vertica_sql_scripts.py
+++ b/edx/analytics/tasks/tests/test_run_vertica_sql_scripts.py
@@ -1,0 +1,102 @@
+"""
+Ensure we can write to Vertica data sources.
+"""
+from __future__ import absolute_import
+
+import textwrap
+from os import path
+
+import luigi
+import luigi.task
+
+from mock import sentinel
+
+from edx.analytics.tasks.run_vertica_sql_scripts import RunVerticaSqlScriptsTask
+from edx.analytics.tasks.tests import unittest
+
+
+class RunVerticaSqlScriptsTaskTest(unittest.TestCase):
+    """
+    Ensure we can run SQL scripts that read and write data to Vertica data sources.
+    """
+
+    def create_task(self, credentials='', script_config=''):
+        """
+        Emulate execution of a generic RunVerticaSqlScriptsTask.
+        """
+        # Make sure to flush the instance cache so we create a new task object.
+        luigi.task.Register.clear_instance_cache()
+        task = RunVerticaSqlScriptsTask(
+            credentials=sentinel.ignored,
+            script_configuration=script_config,
+            script_root='',
+        )
+
+        return task
+
+    def get_configured_task_chain_for_config(self, config_path):
+        """
+        Creates the task based on a given YAML configuration file.
+        """
+        return self.create_task(script_config=path.join('edx/analytics/tasks/tests/fixtures/sql_scripts', config_path)).get_downstream_task()
+
+    def get_tasks_in_chain(self, chain):
+        # Enumerate the chain, revealing whether or not anything was actually generated.
+        chain_items = list(chain)
+
+        items = []
+        if len(chain_items) == 0:
+          return items
+
+        # If we're returning more than one item from the requirements, that means we're risking concurrent
+        # task execution, which breaks our desire to serially execute tasks one after the other.
+        self.assertEqual(len(chain_items), 1)
+
+        # Pull out the chain of dependent tasks.
+        current_item = chain_items[0]
+        while current_item is not None:
+          items.append(current_item)
+          current_item = current_item.depends_on
+
+        return list(reversed(items))
+
+    def test_run_with_empty_configuration(self):
+        # It's ... valid YAML, but there just won't be anything there.  No generated tasks.
+        chain = self.get_configured_task_chain_for_config('empty_config.yaml')
+        tasks = self.get_tasks_in_chain(chain)
+
+        self.assertEqual(len(tasks), 0)
+
+    def test_run_with_jacked_up_configuration(self):
+        # Again, this is more like "weird, unexpected content that is still valid."  No generated tasks.
+        chain = self.get_configured_task_chain_for_config('nonsense_config.yaml')
+        tasks = self.get_tasks_in_chain(chain)
+
+        self.assertEqual(len(tasks), 0)
+
+    def test_run_with_single_script(self):
+        # A real configuration.  Should be a single generated task.
+        chain = self.get_configured_task_chain_for_config('single_script.yaml')
+        tasks = self.get_tasks_in_chain(chain)
+
+        self.assertEqual(len(tasks), 1)
+
+    def test_run_with_two_scripts(self):
+        # A real configuration.  Should be two generated tasks: script_two, then script_one.
+        chain = self.get_configured_task_chain_for_config('two_scripts.yaml')
+        tasks = self.get_tasks_in_chain(chain)
+
+        self.assertEqual(len(tasks), 2)
+        self.assertEqual(tasks[0].script_name, "script one")
+        self.assertEqual(tasks[1].script_name, "script two")
+
+    def test_run_with_four_scripts(self):
+        # A real configuration.  Should be four generated tasks: script_four, then script_three, then script_two, then script_one.
+        chain = self.get_configured_task_chain_for_config('four_scripts.yaml')
+        tasks = self.get_tasks_in_chain(chain)
+
+        self.assertEqual(len(tasks), 4)
+        self.assertEqual(tasks[0].script_name, "script one")
+        self.assertEqual(tasks[1].script_name, "script two")
+        self.assertEqual(tasks[2].script_name, "script three")
+        self.assertEqual(tasks[3].script_name, "script four")
--- a/AUTHORS
+++ b/AUTHORS
@@ -12,4 +12,4 @@ Dmitry Viskov <dmitry.viskov@webenterprise.ru>
 Sanford Student <sstudent@edx.org>
 Jillian Vogel <jill@opencraft.com>
 Dennis Jen <djen@edx.org>
-
+Toby Lawrence <tlawrence@edx.org>
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ test-requirements: requirements
 test-local:
 	# TODO: when we have better coverage, modify this to actually fail when coverage is too low.
 	rm -rf .coverage
-	LUIGI_CONFIG_PATH='config/test.cfg' python -m coverage run --rcfile=./.coveragerc -m nose --with-xunit --xunit-file=unittests.xml -A 'not acceptance'
+	LUIGI_CONFIG_PATH='config/test.cfg' python -m coverage run --rcfile=./.coveragerc -m nose --with-xunit --xunit-file=unittests.xml -A 'not acceptance' -s
 
 test: test-requirements develop test-local
 
--- a/config/devstack.cfg
+++ b/config/devstack.cfg
@@ -87,3 +87,7 @@ number_of_shards = 5
 
 [ccx]
 enabled = false
+
+[run-vertica-sql-script]
+schema = testing
+read_timeout = 5
--- a/config/test.cfg
+++ b/config/test.cfg
@@ -134,3 +134,7 @@ api_root_url = http://example.com/api/v1/
 
 [ccx]
 enabled = false
+
+[run-vertica-sql-script]
+schema = testing
+read_timeout = 5
--- a/edx/analytics/tasks/load_internal_reporting_user_activity.py
+++ b/edx/analytics/tasks/load_internal_reporting_user_activity.py
@@ -10,10 +10,11 @@ import luigi
 from edx.analytics.tasks.pathutil import PathSetTask
 from edx.analytics.tasks.url import ExternalURL, url_path_join
 from edx.analytics.tasks.user_activity import UserActivityTableTask
-from edx.analytics.tasks.vertica_load import VerticaCopyTask, VerticaCopyTaskMixin, CredentialFileVerticaTarget
+from edx.analytics.tasks.vertica_load import VerticaCopyTask, VerticaCopyTaskMixin
 from edx.analytics.tasks.database_imports import ImportAuthUserTask
 from edx.analytics.tasks.util.hive import HiveTableFromQueryTask, WarehouseMixin, HivePartition
 from edx.analytics.tasks.util.weekly_interval import WeeklyIntervalMixin
+from edx.analytics.tasks.util.vertica_target import CredentialFileVerticaTarget
 from edx.analytics.tasks.user_activity import CourseActivityWeeklyTask
 
 log = logging.getLogger(__name__)
--- a/edx/analytics/tasks/load_warehouse.py
+++ b/edx/analytics/tasks/load_warehouse.py
@@ -12,8 +12,9 @@ from edx.analytics.tasks.load_internal_reporting_user_activity import LoadIntern
 from edx.analytics.tasks.load_internal_reporting_user_course import LoadUserCourseSummary
 from edx.analytics.tasks.load_internal_reporting_user import LoadInternalReportingUserToWarehouse
 from edx.analytics.tasks.course_catalog import DailyLoadSubjectsToVerticaTask
-from edx.analytics.tasks.vertica_load import VerticaCopyTaskMixin, CredentialFileVerticaTarget
+from edx.analytics.tasks.vertica_load import VerticaCopyTaskMixin
 
+from edx.analytics.tasks.util.vertica_target import CredentialFileVerticaTarget
 from edx.analytics.tasks.util.hive import WarehouseMixin
 from edx.analytics.tasks.url import ExternalURL
 
--- a/edx/analytics/tasks/tests/fixtures/sql_scripts/four_scripts.yaml
+++ b/edx/analytics/tasks/tests/fixtures/sql_scripts/four_scripts.yaml
@@ -1,0 +1,9 @@
+scripts:
+  - name: script one
+    location: script_one.sql
+  - name: script two
+    location: script_two.sql
+  - name: script three
+    location: script_three.sql
+  - name: script four
+    location: script_four.sql
--- a/edx/analytics/tasks/tests/fixtures/sql_scripts/nonsense_config.yaml
+++ b/edx/analytics/tasks/tests/fixtures/sql_scripts/nonsense_config.yaml
@@ -1,0 +1,1 @@
+asdasdaikubdnaiksudb
--- a/edx/analytics/tasks/tests/fixtures/sql_scripts/single_script.yaml
+++ b/edx/analytics/tasks/tests/fixtures/sql_scripts/single_script.yaml
@@ -1,0 +1,4 @@
+scripts:
+  - name: Simple Testing Script
+    script_id: my_lil_test_script
+    location: foobar.sql
--- a/edx/analytics/tasks/tests/fixtures/sql_scripts/two_scripts.yaml
+++ b/edx/analytics/tasks/tests/fixtures/sql_scripts/two_scripts.yaml
@@ -1,0 +1,5 @@
+scripts:
+  - name: script one
+    location: script_one.sql
+  - name: script two
+    location: script_two.sql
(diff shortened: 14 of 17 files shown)


mulby:
@tobz - great start! You should also add yourself to the AUTHORS file. I'll make other comments inline.

edx/analytics/tasks/run_sql_script.py
```diff
@@ -0,0 +1,137 @@
...
+        config_path={'section': 'run-sql-script', 'name': 'credentials'},
+        description='Path to the external access credentials file.',
+    )
+    read_timeout = luigi.IntParameter(
+        config_path={'section': 'run-sql-script', 'name': 'read_timeout'}
+    )
+    source_script = luigi.Parameter(
+        description='The path to the source script to execute'
+    )
+    table = luigi.Parameter(
```
mulby: Wondering aloud: could we just use the file path for this purpose?

edx/analytics/tasks/run_sql_scripts.py
```diff
@@ -0,0 +1,97 @@
...
+        """
+        return str(self)
+
+    def output(self):
+        """
+        Returns a SentinelTarget which represents a task that generates no output
+        and also is not intended to be marked for complete.
+        """
+        if self.output_target is None:
+            self.output_target = SentinelTarget()
```
mulby: If this is never complete, I think Luigi will complain about it not running successfully even if it actually did. I would recommend using some kind of marker here. Much like we do in the SchemaMangementTarget:

https://github.com/edx/edx-analytics-pipeline/blob/add0ddcb1786a7132d8a8050aa20579516820dad/edx/analytics/tasks/load_warehouse.py#L136

edx/analytics/tasks/run_sql_scripts.py
```diff
@@ -0,0 +1,97 @@
...
+                # Fire off a task for each script. Make sure raise_on_error is false because
+                # we shouldn't stop the job if one subtask happens to fail.  This is hacky, because
+                # the version of Luigi we're on is so old that it doesn't support the yield-tasks-from-run
+                # functionality.  Alas.
+                current_task = RunSqlScriptTask(
+                    credentials=self.credentials, schema=self.schema, marker_schema=self.marker_schema,
+                    date=self.date, read_timeout=self.read_timeout, source_script=script['location'],
+                    table=script['table'], raise_on_error=False)
+                if not current_task.output().exists():
+                  current_task.run()
```
mulby: If we do go this route, I think we want to use `luigi.build()`?

I'm a bit wary of this pattern, since it would be the first place we do something like this. That said, we could just make a note about it and come back to clean it up when we upgrade to the latest version of luigi.

mulby (commented):
Still thinking about the dynamic dependency problem... I'll post more comments here if I have more thoughts. Curious what Hassan and Brian think as well.

edx/analytics/tasks/run_sql_script.py
tobz: So, I guess it depends what we end up using this for beyond the current task.  I copied this all from the vertica load stuff, and let it in because if someone did want to run the task as a way to generate an output used downstream, you'd want to know the table.

I could just make it default to nothing, which removes it from needing to be passed in except for cases where you care.

edx/analytics/tasks/run_sql_scripts.py
tobz: Let me doing a little testing.  I implicitly assumed that if I ever raised an error back to the parent (so `RunSqlScripts` getting an error from an individual `RunSqlScript` run) that it would nuke the parent run, thereby not satisfying our criterion of "don't let one bad script nuke the whole run".

edx/analytics/tasks/run_sql_scripts.py
tobz: Ah, yeah, that would definitely be nicer.

tobz:
Deep thought: should we bake the schema into the credentials file or have it be part of the arguments into the task?

In this case, the credentials and the schema sort of go hand-in-hand, because it's intended to be a sandbox, so it makes me think that we should, in fact, put them together.  This would remove a necessary parameter from invoking the task.

edx/analytics/tasks/run_sql_scripts.py
mulby: @HassanJaveed84 was looking at this stuff most recently - any suggestions / ideas here?

Also curious what @brianhw thinks...

mulby:
Yeah, we initially had the schema in the credential file. However, we find it very useful to be able to easily override the schema at the command line for RC builds and other testing purposes, so I would advocate for keeping the parameter.

tobz:
Fair enough.

edx/analytics/tasks/run_sql_script.py
```diff
@@ -0,0 +1,137 @@
...
+        config_path={'section': 'run-sql-script', 'name': 'schema'},
+        description='The schema to which to write.',
+    )
+    credentials = luigi.Parameter(
+        config_path={'section': 'run-sql-script', 'name': 'credentials'},
+        description='Path to the external access credentials file.',
+    )
+    read_timeout = luigi.IntParameter(
+        config_path={'section': 'run-sql-script', 'name': 'read_timeout'}
+    )
```
brianhw: This could use a description, indicating what the units are.  (For that matter, so can the parameter in VerticaCopyTaskMixin, from whence this was cloned.)

edx/analytics/tasks/run_sql_script.py
```diff
@@ -0,0 +1,137 @@
+"""
+Support for running a SQL script against an HP Vertica database.
+"""
+from collections import namedtuple
```
brianhw: is this used?

edx/analytics/tasks/run_sql_script.py
```diff
@@ -0,0 +1,137 @@
...
+    marker_schema = luigi.Parameter(
+        default=None,
+        description='The marker schema to which to write the marker table. marker_schema would '
+        'default to the schema value if the value here is None.'
+    )
+
+
+class RunSqlScriptTask(RunSqlScriptTaskMixin, luigi.Task):
+    """
+    A task for running a SQL script against an HP Vertica database.
```
brianhw: You mentioned that you tried this out with Postgres instead of Vertica.  There's nothing in the Mixin that is Vertica-specific, but the code below is Vertica-specific.  So I think the name of the task should reflect this.  And then we can make similar tasks for Mysql, Postgres, and so forth.   I.e. RunVerticaSqlScriptTask.

edx/analytics/tasks/run_sql_script.py
```diff
@@ -0,0 +1,137 @@
...
+except ImportError:
+    log.warn('Unable to import Vertica client libraries')
+    # On hadoop slave nodes we don't have Vertica client libraries installed so it is pointless to ship this package to
+    # them, instead just fail noisily if we attempt to use these libraries.
+    vertica_client_available = False  # pylint: disable-msg=C0103
+
+
+class RunSqlScriptTaskMixin(object):
+    """
+    Parameters for running a SQL script against an HP Vertica database.
```
brianhw: Hmm.  In looking at these, I thought they were generic, and not specifically related to Vertica.  If they were vertica-specific, then I would like to see the class name changed to reflect that dependency.  The overall values are generic, but I realize in looking at the overlap here with VerticaCopyTaskMixin that the problem is the scope of the default config-path values.  Most of these are the same parameter names as used in VerticaCopyTaskMixin, but are put in a "run-sql-script" section instead of "vertica-export".   So I think the name of the task should be Vertica-specific, and the name of the section should also.   (After all, we need to use a different credentials file and schema, at the very least.)  "vertica-sql-script", as defined in RunVerticaSqlScriptTaskMixin?

edx/analytics/tasks/run_sql_script.py
```diff
@@ -0,0 +1,137 @@
...
+
+        # Make sure we can connect to Vertica.
+        self.check_vertica_availability()
+        connection = self.output().connect()
+
+        try:
+            # Set up our connection to point to the specified schema so that scripts can have unqualified
+            # table references and not necessarily need to know or care about where they're running.
+            connection.cursor().execute('SET SEARCH_PATH = {schema};'.format(schema=self.schema))
+
```
brianhw: The vertica copy also calls "connection.cursor().execute("SET TIMEZONE TO 'GMT';")".  Is that needed here too, or is that only needed on import?

edx/analytics/tasks/run_sql_scripts.py
```diff
@@ -0,0 +1,97 @@
...
+    schema = luigi.Parameter(
+        config_path={'section': 'run-sql-scripts', 'name': 'schema'},
+        description='The schema to which to write.',
+    )
+    credentials = luigi.Parameter(
+        config_path={'section': 'run-sql-scripts', 'name': 'credentials'},
+        description='Path to the external access credentials file.',
+    )
+    read_timeout = luigi.IntParameter(
+        config_path={'section': 'run-sql-scripts', 'name': 'read_timeout'}
```
brianhw: I'm not clear why these are redefined here, rather than just using RunVerticaSqlScriptTaskMixin or a base mixin class for the two.  We shouldn't have to need to define these values to have defaults that are different between run-script and run-scripts.

edx/analytics/tasks/run_sql_scripts.py
```diff
@@ -0,0 +1,97 @@
...
+
+import luigi
+import luigi.configuration
+from edx.analytics.tasks.url import ExternalURL
+from edx.analytics.tasks.run_sql_script import RunSqlScriptTask
+from edx.analytics.tasks.util.sentinel_target import SentinelTarget
+
+log = logging.getLogger(__name__)
+
+class RunSqlScriptsTaskMixin(object):
```
brianhw: same thing about vertica-specific as elsewhere....

edx/analytics/tasks/run_sql_scripts.py
brianhw: Heh, if we wanted to, we could treat the script config file as a makefile: allow each script listed in the script config to provide an identifier, and also provide a list of the identifiers on which it depends.  (It's left as an exercise to the reader to calculate a script ordering. : )  I think that "table" is not so clear a name for the "identifier", because it's possible for some scripts to not create a table, or for multiple scripts to insert to the same table.  So I'd go for something more generic, such as ... "identifier". 

It's possible to just roll our own dependence stuff here, with a check of a script's dependent scripts before running the script, then storing its status for checking future scripts' dependencies.  

But I think the same thing could be done with luigi.build() by injecting tasks into other tasks using an add_requirement() method that allows them to be added to the credentials and source_script dependencies already included.   Then Luigi deals with calculating dependency trees and the like, and we just run luigi.build here with an array of the constructed tasks.

edx/analytics/tasks/tests/test_run_sql_script.py
```diff
@@ -0,0 +1,106 @@
...
+                {
+                    "host": "db.example.com",
+                    "port": 5433,
+                    "user": "exampleuser",
+                    "password": "example password"
+                }'''
+
+        source = '''
+        DELETE TABLE my_schema.my_table;
+        CREATE TABLE my_schema.my_table AS SELECT foo, bar, quux FROM my_schema.another_table;
```
brianhw: what is "quux"?

edx/analytics/tasks/tests/test_run_sql_script.py
```diff
@@ -0,0 +1,106 @@
...
+        mock_conn = self.mock_vertica_connector.connect()
+        self.assertTrue(mock_conn.cursor().execute.called)
+        self.assertFalse(mock_conn.rollback.called)
+        self.assertTrue(mock_conn.commit.called)
+        self.assertTrue(mock_conn.close.called)
+
+    def test_run_with_failure(self):
+        task = self.create_task()
+        task.output().touch = MagicMock(side_effect=Exception("Failed to update marker"))
+        with self.assertRaises(Exception):
```
brianhw: Might be stronger to assert the exception string appears as well.  I forget the specific name, but something like assertRaisesRegexp(Exception, 'Failed to update marker').

mulby:
Idea for the dynamic requirements problem: What if we checked out the repo and read the config file in the `requires()` method? Once it was read in we could build the entire requirements graph?

brianhw (commented):
Looks very good overall.  I guess I'm most concerned about how this fits with the other RDBMS and Vertica-specific code:  how much DRY to attempt, and how to scope.

edx/analytics/tasks/run_sql_script.py
```diff
@@ -0,0 +1,137 @@
...
+                'credentials': ExternalURL(url=self.credentials),
+                'source_script': ExternalURL(url=self.source_script),
+            }
+        return self.required_tasks
+
+    def update_id(self):
+        """This update id will be a unique identifier for this SQL script execution."""
+        return str(self)
+
+    def output(self):
```
brianhw: This method and update_id above it are cut-and-paste from VerticaCopyTask.  Is it worthwhile to extract them into a Vertica base class?

brianhw:
Just saw Gabe's comment, and that's about what we're doing with the ImportMysqlToVerticaTask that Hassan wrote.  We looked at having requires() call luigi.build() to run a task or two that are needed to build the list of tables to export, but then we figured there's no need here for that logic to be in a separate task.  See load_internal_reporting_database.py.

edx/analytics/tasks/run_sql_script.py
tobz: Shouldn't be, no; a hold-over from copying `vertica_load.py`.  Will remove.

edx/analytics/tasks/run_sql_script.py
tobz: Do we think this is worth generalizing?

I'm assuming that we have an output target class that can map to MySQL, and I'm not sure if `SET SEARCH_PATH` would blow up on MySQL, but we could just try/catch it.  We'd also have to modify the check that sees if the Vertica client module could be loaded.

All in all, it doesn't seem too onerous, but if we really only see ourselves doing this sort of stuff against the warehouse, then it might be just fine and dandy and be explicit about this code/module/file being Vertica-only.

edx/analytics/tasks/run_sql_script.py
tobz: I'm not explicitly sure if it's required, but it seems like a good invariant to provide for analyst tasks.  I'll add it.

edx/analytics/tasks/tests/test_run_sql_script.py
tobz: Just another dummy value along with "foo" and "bar".

edx/analytics/tasks/tests/test_run_sql_script.py
tobz: Yep, that's a good point.

edx/analytics/tasks/run_sql_script.py
tobz: There's probably more to be extracted into a common base class/mixin, yeah.  I started doing so by moving `CredentialFileVerticaTarget` out of `vertica_load.py`, but didn't really consider any other commonalities.

I'll keep this in mind to address once my other question (re: generalize this for MySQL/others vs make it Vertica-specific) has a good answer.

edx/analytics/tasks/run_sql_scripts.py
tobz: At first, I was ready to agree with your point, but it struck me: if we want these tasks to be separate for the purpose of having another workflow simply run a one-off SQL script, then it's conceivable we do, in fact, want a cascading value here.

What if I created, pursuant to one of your comments below, a base mixin that represented the basics for a Vertica-related task -- schema, marker schema, read timeout, etc -- with a more general name?  One configuration section, one mixin, used to define these values centrally, in a way that could be enforced across all tasks that used it.

I would still mix this in to both `RunSqlScriptTask` and `RunSqlScriptsTask`, to allow the aforementioned configurability, but at least they would both derive from the same place, and not explicitly redefine things.

tobz:
Yeah, I could totally have this happen in the `requires`.  My initial reluctance was that it would mean we couldn't pull in the configuration file as an `ExternalURL` resource.  I could obviously create an instance of the class, call the methods myself to get the file stream, etc... it just wouldn't be happening in Luigi itself, it wouldn't be shortcircuiting the scheduling/execution if the file was unavailable, etc.

In essence, it would simply be a more contained instance of doing things manually instead of it happening natively in Luigi.

edx/analytics/tasks/run_sql_script.py
brianhw: I think the set search would indeed blow up on Mysql, but more to the point, the connection stuff would all be database-specific.  My interest is more in making the names reflect the database-specific nature.  Right now, the names hide that, at the config level, the class level, and even the filename level.

edx/analytics/tasks/run_sql_script.py
brianhw: Cool.  This is definitely a lower-priority refactor, and may not be worth the complexity.

edx/analytics/tasks/run_sql_script.py
brianhw: @HassanJaveed84 I think you encountered the need for this with the copy.  Do you think it's useful in general?

edx/analytics/tasks/run_sql_scripts.py
brianhw: Yup, that's what I meant by a base mixin class.  But moreso, the base mixin would define a single config section, rather than one for script and one for scripts.   And to me that seems fine -- while one might want to run a one-off SQL script, I would argue that we only need to define one default value in the config file.  One-offs would define the schema and credentials on the command line and propagate it down that way. 

Note that we don't want a single mixin for credentials or schema to be shared with the VerticaCopy code, because it will generally have a different default value, since it's run as a different user and writes to a different schema by default.

edx/analytics/tasks/tests/test_run_sql_script.py
brianhw: Yes, I see that. : )   It's just I know the story behind foo and bar, and wondered if there were one related to quux.  (Not a technical question, just curiosity.)  Lisp had "foo", "bar", "baz".  All I found was that 'quux' is commonly used after 'baz', and that it was attributed to Guy Steele, who wrote the book on Common Lisp.   Anyway....

edx/analytics/tasks/util/sentinel_target.py
```diff
@@ -0,0 +1,22 @@
+"""
+Luigi target that marks the completion of a task which has no output.
+"""
+import luigi
+
+
+class SentinelTarget(luigi.Target):
+    """
+    A target that is a sentinel value for tasks that have no output.  This target will never exist.
```
brianhw: To echo Gabe's point earlier, I don't see what utility this provides.  When would one ever want the target to never exist, except for testing.  In which case, this should move to a testing directory.

tobz:
Alright: I've reworked my approach to use the dynamic requirements technique, and now `RunVerticaSqlScriptsTask` is much simpler, and works as intended.

edx/analytics/tasks/run_vertica_sql_scripts.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def requires(self):
+        script_configuration_target = ExternalURL(url=self.script_configuration).output()
+        with script_configuration_target.open('r') as script_conf_file:
+            # Read in our script configuration, and iterate the scripts in reverse order.  The way that Luigi seems to schedule tasks
+            # is that it runs them in the reverse order that you yield them.  By having users specify their scripts in top-first order
+            # and then reversing that order here, they run under Luigi in the intended order.
+            config = yaml.safe_load(script_conf_file)
+            for script in reversed(config['scripts']):
+                yield RunVerticaSqlScriptTask(
```
mulby: I agree, I think this is much cleaner...

That said, the tasks will be scheduled in parallel (regardless of the order of the `yield` statements). We need to somehow pass the previous script task into the current one so that it can return it from *its* requires function.

edx/analytics/tasks/run_vertica_sql_script.py
```diff
@@ -0,0 +1,135 @@
...
+    )
+
+    def update_id(self):
+        """
+        A unique string identifying this task run, based on the input parameters.
+        """
+        return str(self)
+
+
+class RunVerticaSqlScriptTaskMixin(object):
```
brianhw: Why not have this extend BaseVerticaSqlScriptTaskMixin?

edx/analytics/tasks/run_vertica_sql_script.py
```diff
@@ -0,0 +1,135 @@
...
+    source_script = luigi.Parameter(
+        description='The path to the source script to execute'
+    )
+    table = luigi.Parameter(
```
brianhw: Maybe just call this script_id then?

edx/analytics/tasks/run_vertica_sql_scripts.py
```diff
@@ -0,0 +1,48 @@
...
+import datetime
+from os import path
+
+import luigi
+import luigi.configuration
+from edx.analytics.tasks.url import ExternalURL
+from edx.analytics.tasks.run_vertica_sql_script import BaseVerticaSqlScriptTaskMixin, RunVerticaSqlScriptTask
+
+
+class RunVerticaSqlScriptsTaskMixin(object):
```
brianhw: Again, you should probably extend BaseVerticaSqlScriptTaskMixin.

edx/analytics/tasks/run_vertica_sql_scripts.py
brianhw: And I believe the assumption that tasks run in a particular order separate from explicit dependencies breaks down when we upgrade to Luigi 2.   (That being one of the reasons why only a small subset of our workflows actually work there, I suspect.)

So we could try a few things:
* Try setting priorities, in the way that pre-load and post-load do so for load_warehouse.  
* Try making every script's task depend on the previous script's task, so the whole thing runs sequentially.
* Implement a dependency setting that user's would have to specify if a script depends on a previous script, and translate that into requirements.  

In any case, one would probably need to inject the requirement or priority into the task before yielding it.

edx/analytics/tasks/run_vertica_sql_script.py
tobz: Good point.  Will do.

edx/analytics/tasks/run_vertica_sql_script.py
tobz: You're right.  Will do.

edx/analytics/tasks/run_vertica_sql_scripts.py
tobz: You're right.  Will do.

edx/analytics/tasks/run_vertica_sql_scripts.py
tobz: Yeah, that's fair.  It seemed to order them distinctly, but I was only using two tasks.

I'll take a crack at lashing them together to form a sequential dependency chain.

edx/analytics/tasks/run_vertica_sql_scripts.py
```diff
@@ -0,0 +1,64 @@
...
+        with script_configuration_target.open('r') as script_conf_file:
+            # Pull out our configuration and link them up, so that the last task depends on the one before it to run, and on and on.
+            config = yaml.safe_load(script_conf_file)
+            first_task = None
+            previous_task = None
+            for script in reversed(config['scripts']):
+                new_task = RunVerticaSqlScriptTask(
+                    credentials=self.credentials, schema=self.schema, marker_schema=self.marker_schema,
+                    date=self.date, read_timeout=self.read_timeout, source_script=path.join(self.script_root, script['location']),
+                    table=script['table'])
```
mulby: change to script_id, maybe this defaults to the ordering in the file "script_0", "script_1" etc...

```python
script_id=script.get('script_id', 'script_{}'.format(script_num))
```

edx/analytics/tasks/run_vertica_sql_script.py
```diff
@@ -0,0 +1,138 @@
...
+            )
+
+        return self.output_target
+
+    def run(self):
+        """
+        Runs the given SQL script against the Vertica target.
+        """
+        if not self.script_id:
+            raise Exception("table needs to be specified")
```
mulby: Parameters are required by default, so trying to instantiate this class without specifying this parameter would raise an error.

They can only be `None` if it's explicitly passed in that way, or if you provide a default value in the declaration of the parameter.

edx/analytics/tasks/tests/test_run_vertica_sql_script.py
```diff
@@ -0,0 +1,92 @@
...
+from mock import patch
+from mock import sentinel
+
+from edx.analytics.tasks.run_vertica_sql_script import RunVerticaSqlScriptTask
+from edx.analytics.tasks.tests import unittest
+from edx.analytics.tasks.tests.target import FakeTarget
+from edx.analytics.tasks.tests.config import with_luigi_config
+
+
+class RunVerticaSqlScriptTaskTest(unittest.TestCase):
```
mulby, line 22: We may want to add a few more clearly fixture our business requirements / rules:

* A test that ensures tasks are run sequentially.
* A test that ensures that if one task fails the others don't run.
* Tests that exercise the various formats of the input config file (no "scripts" section for example).

mulby (commented):
Overall this is looking great.

We still need to implement the bits that actually checks out the scripts repo from `git`. We could either do it at deploy time in the ansible, or do it in the requires method. We could also do this as part of a separate PR.

edx/analytics/tasks/run_vertica_sql_scripts.py
tobz: Ah, woops.  Missed that one.

edx/analytics/tasks/run_vertica_sql_script.py
tobz: Gotcha.  Will remove.

edx/analytics/tasks/tests/test_run_vertica_sql_script.py
tobz: Ah, you're totally right.  I don't have tests for the RunVerticaSqlScript**s**Task portion here. :)

tobz:
The `git` work happens in the Jenkins task already.

edx/analytics/tasks/run_vertica_sql_script.py
```diff
@@ -0,0 +1,138 @@
...
+    )
+
+    def update_id(self):
+        """
+        A unique string identifying this task run, based on the input parameters.
+        """
+        return str(self)
+
+
+class RunVerticaSqlScriptTaskMixin(BaseVerticaSqlScriptTaskMixin):
```
brianhw, line 57: nit: docstring.

edx/analytics/tasks/run_vertica_sql_script.py
```diff
@@ -0,0 +1,138 @@
...
+        config_path={'section': 'run-vertica-sql-script', 'name': 'schema'},
+        description='The schema to which to write.',
+    )
+    credentials = luigi.Parameter(
+        config_path={'section': 'run-vertica-sql-script', 'name': 'credentials'},
+        description='Path to the external access credentials file.',
+    )
+    read_timeout = luigi.IntParameter(
+        config_path={'section': 'run-vertica-sql-script', 'name': 'read_timeout'},
+        description='Timeout in seconds for reading from a Vertica database connection'
```
brianhw: nit: end with period, like others.

edx/analytics/tasks/run_vertica_sql_script.py
```diff
@@ -0,0 +1,138 @@
...
+        return str(self)
+
+
+class RunVerticaSqlScriptTaskMixin(BaseVerticaSqlScriptTaskMixin):
+    source_script = luigi.Parameter(
+        description='The path to the source script to execute'
+    )
+    script_id = luigi.Parameter(
+        description='A unique identifier for the purposes of tracking whether or not this '
+        'script ran successfully i.e. the table created by this script, or the ticket related to it'
```
brianhw: nit: end with period, like others.

edx/analytics/tasks/run_vertica_sql_script.py
```diff
@@ -0,0 +1,135 @@
...
+    def update_id(self):
+        """
+        A unique string identifying this task run, based on the input parameters.
+        """
+        return str(self)
+
+
+class RunVerticaSqlScriptTaskMixin(BaseVerticaSqlScriptTaskMixin):
+    source_script = luigi.Parameter(
+        description='The path to the source script to execute'
```
brianhw: nit: end with period, like others.

edx/analytics/tasks/run_vertica_sql_scripts.py
```diff
@@ -0,0 +1,71 @@
...
+from edx.analytics.tasks.url import ExternalURL
+from edx.analytics.tasks.run_vertica_sql_script import BaseVerticaSqlScriptTaskMixin, RunVerticaSqlScriptTask
+
+
+class RunVerticaSqlScriptsTaskMixin(BaseVerticaSqlScriptTaskMixin):
+    """
+    Parameters for running multiple SQL scripts against an HP Vertica database in a deterministic fashion.
+    """
+    script_configuration = luigi.Parameter(
+        description='The path to the configuration file that specifies which scripts to run'
```
brianhw: nit: end with period, like others.

edx/analytics/tasks/run_vertica_sql_scripts.py
```diff
@@ -0,0 +1,71 @@
...
+    )
+
+
+class RunVerticaSqlScriptsTask(RunVerticaSqlScriptsTaskMixin, luigi.WrapperTask):
+    """
+    A wrapper task for running multiple SQL scripts against an HP Vertica database in a deterministic fashion.
+
+    We use a YAML file that defines a list of scripts to run.  We run the scripts in the order they are defined.
+    By using RunVerticaSqlScriptTask, each script stores its own marker, thus allowing us to idempotently run this
+    task until all require tasks (aka our scripts) have successfully run for a given date.
```
brianhw: nit:  "all required tasks"

edx/analytics/tasks/run_vertica_sql_scripts.py
```diff
@@ -0,0 +1,71 @@
...
+        script_configuration_target = ExternalURL(url=self.script_configuration).output()
+        with script_configuration_target.open('r') as script_conf_file:
+            # Pull out our configuration and link them up, so that the last task depends on the one before it to run, and on and on.
+            config = yaml.safe_load(script_conf_file)
+
+            script_index = 0
+            first_task = None
+            previous_task = None
+
+            for script in reversed(config['scripts']):
```
brianhw: could just use enumerate() here to get script_index, instead of manually managing a counter.

edx/analytics/tasks/run_vertica_sql_scripts.py
```diff
@@ -0,0 +1,71 @@
...
+                # Generate a sequential script ID if one wasn't passed in.
+                script_id = script.get('script_id', 'sql_script_{}'.format(script_index))
+
+                new_task = RunVerticaSqlScriptTask(
+                    credentials=self.credentials, schema=self.schema, marker_schema=self.marker_schema,
+                    date=self.date, read_timeout=self.read_timeout, source_script=path.join(self.script_root, script['location']),
```
brianhw: we've tended to use url_path_join instead of path.join.

edx/analytics/tasks/run_vertica_sql_scripts.py
```diff
@@ -0,0 +1,71 @@
...
+    A wrapper task for running multiple SQL scripts against an HP Vertica database in a deterministic fashion.
+
+    We use a YAML file that defines a list of scripts to run.  We run the scripts in the order they are defined.
+    By using RunVerticaSqlScriptTask, each script stores its own marker, thus allowing us to idempotently run this
+    task until all require tasks (aka our scripts) have successfully run for a given date.
+    """
+
+    def requires(self):
+        script_configuration_target = ExternalURL(url=self.script_configuration).output()
+        with script_configuration_target.open('r') as script_conf_file:
```
brianhw: requires() can get called several times, so it might be good to pull this into a helper and only call once.  Probably only need to cache first_task (as e.g. self.first_task), and then have the yield at the bottom.

edx/analytics/tasks/run_vertica_sql_scripts.py
```diff
@@ -0,0 +1,71 @@
...
+                    credentials=self.credentials, schema=self.schema, marker_schema=self.marker_schema,
+                    date=self.date, read_timeout=self.read_timeout, source_script=path.join(self.script_root, script['location']),
+                    script_id=script_id)
+
+                # Is this the first task we've configured? If so, mark it, so we know where to fire off execution from.
+                if first_task is None:
+                    first_task = new_task
+
+                # If we previously configured a task, set this new one as a requirement of the previous, so it runs prior to.
+                if previous_task is not None:
```
brianhw: nit:  I got confused at first by "previous".   It's previous in definition, but not in execution order.  Maybe call it "downstream_task"?

edx/analytics/tasks/run_vertica_sql_script.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def requires(self):
+        if self.required_tasks is None:
+            self.required_tasks = {
+                'credentials': ExternalURL(url=self.credentials),
+                'source_script': ExternalURL(url=self.source_script),
+            }
+
+            if self.depends_on is not None:
+              self.required_tasks['depends_on'] = self.depends_on
```
brianhw, line 96: nit: We don't need it for the current functionality, but I'd probably prefer the ability to have multiple dependencies be set, so that depends_on would be a list instead of a single task.  And I would probably prefer to have a self.add_dependency() that adds stuff to the internal self.depends_on, instead of setting it directly externally, to make it clearer what the task's interface is.

edx/analytics/tasks/run_vertica_sql_scripts.py
```diff
@@ -0,0 +1,71 @@
...
+            # Pull out our configuration and link them up, so that the last task depends on the one before it to run, and on and on.
+            config = yaml.safe_load(script_conf_file)
+
+            script_index = 0
+            first_task = None
+            previous_task = None
+
+            for script in reversed(config['scripts']):
+                # Generate a sequential script ID if one wasn't passed in.
+                script_id = script.get('script_id', 'sql_script_{}'.format(script_index))
```
brianhw: nit: Do we want the ids to go in reverse order?    This would label the end script to be 0 and the first one to have the highest id.   Having it work this way seems to expose the implementation.

edx/analytics/tasks/run_vertica_sql_scripts.py
```diff
@@ -0,0 +1,71 @@
...
+
+                new_task = RunVerticaSqlScriptTask(
+                    credentials=self.credentials, schema=self.schema, marker_schema=self.marker_schema,
+                    date=self.date, read_timeout=self.read_timeout, source_script=path.join(self.script_root, script['location']),
+                    script_id=script_id)
+
+                # Is this the first task we've configured? If so, mark it, so we know where to fire off execution from.
+                if first_task is None:
+                    first_task = new_task
```
brianhw: nit: after getting confused by "previous", I realized I had the same issue with "first".  I thought at first it meant the first that would be run, but it's the last to run.   But harder to come up with a name:  "most_downstream_task"?  "last_task_to_run"?

brianhw (commented):
Real progress.  Mostly down to naming things now.

edx/analytics/tasks/run_vertica_sql_scripts.py
mulby: Also, it just occurred to me, do we want to reverse this list? I feel like we don't:

Given this config:
```
scripts:
    - foo:
        location: sql/foo.sql
    - bar:
        location: sql/bar.sql
```

I would expect `foo` to run first, and `bar` to depend on `foo` (and run second).

edx/analytics/tasks/run_vertica_sql_script.py
tobz: Talked with Gabe about this: going to leave the single dependency bit as-is.  If people need multiple dependencies, they probably want to be writing their own Luigi task because this would otherwise imply that we have a way, in the Yaml file, to specify multiple dependencies... which we currently don't.

I'll add `add_dependency` which will abstract the internal machinery, and ultimately, that will let us changes things down the line if we so choose.

tobz: Scratch that.  I'm going to call it `set_dependency`, which accurately represents what is happening.

edx/analytics/tasks/run_vertica_sql_scripts.py
tobz: A ha, you're absolutely right!  Good catch.  Even aside from exposing the implementation, it doesn't jive with the intended visible behavior of things running top to bottom, and would potentially confuse users when "script #0" is actually running last, etc.

edx/analytics/tasks/run_vertica_sql_script.py
brianhw: Heh.  Actually, it's adding a dependency, because the other things in requires() are also dependencies.  This is just the one script dependency.

tobz: You're right, but I'm torn, because "add" implies it can be run multiple times, when this is actually a "set" under the hood.  Aiiiieeeeee, I don't know what to do. :P

tobz: Okay, I'm going with `add_dependency`, but with a note in the docstring to make it crystal clear what's happening.

edx/analytics/tasks/run_vertica_sql_script.py
```diff
@@ -0,0 +1,149 @@
...
+    required_tasks = None
+    output_target = None
+    depends_on = None
+
+    def add_dependency(self, dependency):
+      """
+      Adds a custom dependency/requirement for this task.
+
+      Note: this currently *sets* a single, custom dependency.  You cannot add multiple dependencies to this task.
+      The last dependency to be added is the only one that will stick.  It will, however, not be the only dependnency,
```
brianhw: nit: dependnency -> dependency

edx/analytics/tasks/run_vertica_sql_scripts.py
```diff
@@ -0,0 +1,76 @@
...
+
+                previous_task = None
+
+                scripts = config.get('scripts', [])
+                scripts_indexes = range(0, len(scripts))
+
+                # Iterate over the list of scripts in the configuration file in reverse order.  We also zip a list of integers,
+                # representing the zero-based index position of the given script in the overall list.  We iterate in reverse
+                # in order to link each task together, using requires(), to ensure that tasks run sequentially, and in the intended
+                # order: from the top of the file, downwards.
```
brianhw: Please remind me again why we create the scripts in reverse order.  Why not go through them in forward order, which each subsequent task adding the upstream_task (that was created in the previous iteration) as its dependency?    Then you could call it either previous_task or upstream_task, and then just take the value of new_task at the end of the for loop.

brianhw (commented):
a nit and a proposed simplification....

edx/analytics/tasks/run_vertica_sql_scripts.py
tobz: Yeah, that's totally true.  I forget why I originally chose to reverse the enumeration.  You're right, though, it works in either direction.

edx/analytics/tasks/tests/test_run_vertica_sql_scripts.py
```diff
@@ -0,0 +1,109 @@
...
+        # A real configuration.  Should be two generated tasks: script_two, then script_one.
+        chain = self.get_configured_task_chain_for_config('two_scripts.yaml')
+        tasks = self.get_tasks_in_chain(chain)
+
+        self.assertEqual(len(tasks), 2)
+        self.assertEqual(tasks[0].script_name, "script two")
+        self.assertEqual(tasks[1].script_name, "script one")
+
+    def test_run_with_four_scripts(self):
+      # A real configuration.  Should be four generated tasks: script_four, then script_three, then script_two, then script_one.
```
brianhw: nit: comment is not aligned.

edx/analytics/tasks/tests/test_run_vertica_sql_scripts.py
```diff
@@ -0,0 +1,109 @@
...
+        self.mock_vertica_connector.marker_schema = "name_of_marker_schema"
+        self.mock_vertica_connector.marker_table = "name_of_marker_table"
+
+        return task
+
+    def get_configured_task_chain_for_config(self, config_path):
+        return self.create_task(script_config=path.join('edx/analytics/tasks/tests/fixtures/sql_scripts', config_path)).get_downstream_task()
+
+    def get_tasks_in_chain(self, chain):
+        chain_items = [x for x in chain]
```
brianhw: I'm somewhat confused by the multiple chain items in the chain.  Only one item is pulled from it, as current_item.  All others are ignored.  Are we always expecting there to be zero or one?  If so, we should make that more apparent (e.g. by asserting).  

Also, I think this should return a list that has been reversed.

edx/analytics/tasks/tests/test_run_vertica_sql_scripts.py
```diff
@@ -0,0 +1,109 @@
...
+        self.assertEqual(len(tasks), 1)
+
+    def test_run_with_two_scripts(self):
+        # A real configuration.  Should be two generated tasks: script_two, then script_one.
+        chain = self.get_configured_task_chain_for_config('two_scripts.yaml')
+        tasks = self.get_tasks_in_chain(chain)
+
+        self.assertEqual(len(tasks), 2)
+        self.assertEqual(tasks[0].script_name, "script two")
+        self.assertEqual(tasks[1].script_name, "script one")
```
brianhw: As mentioned above, I think that tasks should be in order of execution, so should be reversed.  And so the list returned would be "script one" and then "script two".

edx/analytics/tasks/tests/test_run_vertica_sql_scripts.py
```diff
@@ -0,0 +1,109 @@
...
+    def test_run_with_four_scripts(self):
+      # A real configuration.  Should be four generated tasks: script_four, then script_three, then script_two, then script_one.
+        chain = self.get_configured_task_chain_for_config('four_scripts.yaml')
+        tasks = self.get_tasks_in_chain(chain)
+
+        self.assertEqual(len(tasks), 4)
+        self.assertEqual(tasks[0].script_name, "script four")
+        self.assertEqual(tasks[1].script_name, "script three")
+        self.assertEqual(tasks[2].script_name, "script two")
+        self.assertEqual(tasks[3].script_name, "script one")
```
brianhw: likewise:  one, two, three, four.

edx/analytics/tasks/tests/test_run_vertica_sql_scripts.py
```diff
@@ -0,0 +1,109 @@
...
+        )
+
+        fake_output = MagicMock(return_value=self.mock_vertica_connector)
+        self.mock_vertica_connector.marker_schema = "name_of_marker_schema"
+        self.mock_vertica_connector.marker_table = "name_of_marker_table"
+
+        return task
+
+    def get_configured_task_chain_for_config(self, config_path):
+        return self.create_task(script_config=path.join('edx/analytics/tasks/tests/fixtures/sql_scripts', config_path)).get_downstream_task()
```
brianhw, line 41: Just curious if there is a way to implement this without creating actual fixture files.  We've been able to keep down the proliferation as well as keeping the data closer to the test code by inlining the content of files as multi-line strings in the code.  A possible example is the construction and reading of a config file for event exports.  (See test_event_exports.py.)   One can provide a dict at test time, convert to yaml, and have it mocked to return it as the appropriate input file.  

Then again, maybe the issue that this will run into is that it assumes that the scripts are local filesystem paths rather than URLs.  In that case, one would have to create temp files in a temp directory to avoid the fixture files.    For that, see test_course_subjects.py for sample code.

mulby: Toby and I discussed this in person. I told him I was fine with fixture files (which I am).

mulby (approved):
LGTM

edx/analytics/tasks/tests/test_run_vertica_sql_scripts.py
brianhw: Are fixture files your preference? I can live with it in the short term, but I consider them in unit tests to be tech debt.

brianhw (approved):
Please add the missing docstring, and reverse the chain values returned in the test.  Then squash and merge. 👍

edx/analytics/tasks/tests/test_run_vertica_sql_scripts.py
mulby: I don't have a strong preference. The chunk of information needs to live somewhere... I don't really care whether it lives in a separate file or is inline in the test. Do you feel strongly about this?

tobz: There's also a technical reason here: because of the dynamic requires approach, there's no (good) way to mock in a `FakeTarget` - it expects a file path string which is taken directly and fed to an `ExternalURL`.

One way or another, we're putting something on disk, although I do grant that having it be a temporary file written during the test would leave the YAML itself closer to the test code.

edx/analytics/tasks/tests/test_run_vertica_sql_scripts.py
tobz: Right.  It's only ever going to be one item or `None`, as we only care about returning the most downstream task.  The list comprehension is just a simple way, IMO, to pull out that one item, or nothing.  Just came to me more naturally.  I could wrap it in `list(...)`, which presumably would do the same thing?

edx/analytics/tasks/tests/test_run_vertica_sql_scripts.py
tobz: Sure, I'm OK with this.

edx/analytics/tasks/run_vertica_sql_scripts.py
```diff
@@ -0,0 +1,97 @@
...
+                            new_task.add_dependency(previous_task)
+
+                        # Mark this as the previously-created task.
+                        previous_task = new_task
+
+                    self.downstream_task = previous_task
+
+        # If a downstream task has been set, yield it, triggering Luigi to schedule our scripts.
+        if self.downstream_task is not None:
+            yield self.downstream_task
```
mulby, line 97: @tobz I think this line is causing problems... needs to be `return self.downstream_task`

On [line 44](https://github.com/edx/edx-analytics-pipeline/pull/320/files#diff-1578710f7ae81bd490dbf8b7bccb2f60R44) we call `self.downstream_task = self.get_downstream_task()`. However, since `get_downstream_task()` is a generator it returns *immediately* without actually executing the body of the function (since something needs to attempt to consume the generator first). However, when luigi goes to actually consume the generator at that point `self.downstream_task` is no longer `None` since it is a pointer to the generator, so the actually body of the function is never executed.

Kind of confusing, but I'm happy to step through it with you if that would help clarify things...

PR #323, opened 2016-12-06
Ability to save org_id in the tags_distribution results table (besides course_id)

dmitry-viskov:


openedx-webhooks:
Thanks for the pull request, @dmitry-viskov! I've created [OSPR-1616](https://openedx.atlassian.net/browse/OSPR-1616) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

dmitry-viskov:
@baldwin47 can you please add more detailed description - why this feature is needed and what is the final goal?

baldwin47:
We use microsites and different organizations heavily in our edX
instance. We have a use case for being able to review aggregated data
across one or more courses. Our customers (colleges and universities)
want to be able to see performance across courses. With the stock edX
Analytics performance reports, it can be problematic to compare courses
with differing grade configurations. Using the Learning Outcomes report
that we created (already merged to edx master), we can compare aggregate
by Learning Outcome tags. These tags are applied to questions and are an
alternative dimension to course structure. This allows us to roll up
courses to a single learning outcome.

The most important part of this update relates to the analytics pipeline
and api. This data view rolls up statistics across courses, and could be
utilized by edX Analytics Dashboard, or by a third party data
visualization tool.

Our goal is to build an analytics dashboard report off of this data
view. This report would allow selecting one or more courses to view in
aggregate, and would be able to be scoped by Organization.

edx/analytics/tasks/tags_dist.py
```diff
@@ -61,6 +61,8 @@ def mapper(self, line):
         if not course_id:
             return
 
+        org_id = event.get('context').get('org_id')
```
brianhw: Or you can use opaque_key_util.get_org_id_for_course(course_id), and always just pull the org out of the course_id.

edx/analytics/tasks/tags_dist.py
```diff
@@ -73,19 +75,19 @@ def mapper(self, line):
 
         saved_tags = event.get('context').get('asides', {}).get('tagging_aside', {}).get('saved_tags', {})
 
-        yield (course_id, problem_id), (timestamp, saved_tags, is_correct)
+        yield (course_id, org_id, problem_id), (timestamp, saved_tags, is_correct)
```
brianhw, line 79: I was initially confused why you put this in the key rather than the values, as it doesn't add any differentiation.  But I see it's a shortcut to having to take the same value out of every entry in "values".

brianhw (approved):
LGTM, and a small enough change.  Are there no unit tests?  Just surprised that nothing breaks with this change.

edx/analytics/tasks/tags_dist.py
dmitry-viskov: thanks @brianhw ! I've changed this line

dmitry-viskov:
@brianhw thanks for the review. I've updated code and tests.

dmitry-viskov:
@brianhw any news according this PR?

brianhw (approved):
LGTM.  👍  

Do you need me to merge it?

brianhw:
Does merging this need to be coordinated with a parallel change to the data-api?

dmitry-viskov:
@brianhw 

>Do you need me to merge it?

yes. I don't have such permissions

>Does merging this need to be coordinated with a parallel change to the data-api?

This change should not affect existing functionality because it only adds one new column into existed MySQL table. But after this PR will be merged I'm going to open the new PR to the API project with the new API endpoint to return tags by org

PR #324, opened 2016-12-06
Spike: Pull course subject data from Course Catalog API

HassanJaveed84:
@mulby @brianhw

### DIFF
--- a/edx/analytics/tasks/load_internal_reporting_course_catalog.py
+++ b/edx/analytics/tasks/load_internal_reporting_course_catalog.py
@@ -136,7 +136,7 @@ class CourseSubjectTask(LoadInternalReportingCourseCatalogMixin, luigi.Task):
                     course = json.loads(course_str)
                     for subject in course.get('subjects', []):
                         record = CourseSubjectRecord(
-                            course_id=course['key']
+                            course_id=course['key'],
                             date=self.date,
                             subject_title=subject.get('name')
                         )
--- a/edx/analytics/tasks/load_internal_reporting_course_catalog.py
+++ b/edx/analytics/tasks/load_internal_reporting_course_catalog.py
@@ -12,7 +12,7 @@ from edx.analytics.tasks.url import get_target_from_url
 from edx.analytics.tasks.url import url_path_join
 from edx.analytics.tasks.util.edx_api_client import EdxApiClient
 from edx.analytics.tasks.util.hive import BareHiveTableTask, HivePartitionTask, WarehouseMixin
-from edx.analytics.tasks.util.record import Record, StringField, FloatField, DateTimeField, IntegerField
+from edx.analytics.tasks.util.record import Record, StringField, FloatField, DateTimeField, IntegerField, DateField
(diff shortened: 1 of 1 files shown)


edx/analytics/tasks/load_internal_reporting_course_catalog.py
```diff
@@ -50,6 +50,31 @@ class LoadInternalReportingCourseCatalogMixin(WarehouseMixin, OverwriteOutputMix
     )
 
 
+class PullCoursesAPIData(LoadInternalReportingCourseCatalogMixin, luigi.Task):
```
mulby, line 53: No need to do anything now, but when we go to productize this, I'd prefer to use the output from this task as the input for *all* of the other tasks. I think we could get rid of `PullCourseCatalogAPIData` completely and use the data from this set of API calls instead.

HassanJaveed84: /courses endpoint does not accept a 'partner' parameter. Right now, we call /course_runs for each partner and manually add 'partner_short_code' to the output. 
If we were to switch to /courses as input to all other tasks, we would lose the 'partner_short_code' as it is not included in the response.

mulby: Ah, right... maybe we should patch the discovery service to expose this data in the response body... this is great info from the spike. We can evaluate our options now that we know this landmine is here.

edx/analytics/tasks/load_internal_reporting_course_catalog.py
```diff
@@ -84,6 +109,80 @@ def output(self):
...
+        )
+
+    @property
+    def table(self):
+        return "d_course_subjects"
+
+    @property
+    def auto_primary_key(self):
+        """Overridden since the database schema specifies a different name for the auto incrementing primary key."""
+        return ('row_number', 'AUTO_INCREMENT')
```
brianhw, line 174: How many tables are we using the default in, and how many are we using this "row_number" in?  Seems potentially rather fragmented.  And I don't know that we really use it for anything anyway.

edx/analytics/tasks/load_internal_reporting_course_catalog.py
```diff
@@ -50,6 +50,31 @@ class LoadInternalReportingCourseCatalogMixin(WarehouseMixin, OverwriteOutputMix
...
+            for response in client.paginated_get(url, params=params):
+                parsed_response = response.json()
+                for course in parsed_response.get('results', []):
+                    output_file.write(json.dumps(course))
+                    output_file.write('\n')
+
+    def output(self):
+        return get_target_from_url(
+            url_path_join(
+                self.hive_partition_path('course_catalog_raw', partition_value=self.date), 'courses.json'
```
brianhw, line 73: I wasn't sure if this should go in the same "course_catalog_raw" as the course_catalog.json (below), but I think it's okay.

brianhw (commented):
LGTM so far!

brianhw:
Superceded by PR #457 .

PR #325, opened 2016-12-15
support extra repos and packages

mulby:
This allows us to pass in two new parameters into the remote-task script.

"--package":
This parameter allows additional python packages to be specified at deploy time. This allows tasks to be gathered from a variety of different repos etc and all installed into the same virtualenv.

For example: `--package flask --package git@github.com:edx/pipeline-tasks.git`

"--extra-repo":
A URL to a git repository that is checked out on the cluster alongside edx-analytics-pipeline. Internally we actually use the same mechanism to checkout the pipeline repo, the secure config and whatever other repos the user has specified. This allows us to gather static files from different repos and make them available to the data pipeline at runtime. Specifically, we are looking to land SQL scripts to disk so that they can be run by #320.

For example: `--extra-repo git+ssh://git@github.com:edx/some-private-repo.git?dir_name=private&branch=origin%2Frelease --extra-repo https://github.com:edx/edx-public-repo.git?dir_name=public_repo&branch=master`

On the cluster these repos will be checked out alongside the pipeline, so you will see:

<pipeline root dir>/
  config/ (optional, this is the secure config repo)
  repo/ (this is edx-analytics-pipeline)
  private/ (this is git@github.com:edx/some-private-repo.git in the above example)
  public/ (this is https://github.com:edx/edx-public-repo.git in the above example)

When the pipeline runs it can reference files in these other locations:

`launch-task SomeTask --sql-script-root ../private/sql-scripts/`

### DIFF
--- a/edx/analytics/tasks/launchers/remote.py
+++ b/edx/analytics/tasks/launchers/remote.py
@@ -7,6 +7,7 @@ import os
 import pipes
 from subprocess import Popen, PIPE
 import sys
+from urlparse import urlparse, parse_qsl
 import uuid
 
 
@@ -42,6 +43,8 @@ def main():
     parser.add_argument('--wheel-url', help='url of the wheelhouse', default=None)
     parser.add_argument('--virtualenv-extra-args', help='additional arguments passed to virtualenv command when creating the virtual environment', default=None)
     parser.add_argument('--skip-setup', action='store_true', help='assumes the environment has already been configured and you can simply run the task')
+    parser.add_argument('--package', action='append', help='pip install these packages in the pipeline virtual environment')
+    parser.add_argument('--extra-repo', action='append', help="""additional git repositories to checkout on the cluster during the deployment""")
     arguments, extra_args = parser.parse_known_args()
     arguments.launch_task_arguments = extra_args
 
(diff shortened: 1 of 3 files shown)


share/task.yml
```diff
@@ -158,26 +166,34 @@
...
-      command: git fetch --all chdir={{ secure_config_repo_dir }}
-      when: secure_config is defined
-
-    - name: secure config branch checked out
-      command: git checkout {{ secure_config_branch }} chdir={{ secure_config_repo_dir }}
-      when: secure_config is defined
-
     - name: secure config installed
-      command: cp {{ secure_config_repo_dir }}/{{ secure_config }} {{ working_repo_dir }}/override.cfg
+      command: cp {{ working_dir }}/{{ secure_config }} {{ working_repo_dir }}/override.cfg
```
brianhw, line 199: I missed something -- did this file move?  Isn't working_dir the root?  Do we have to change all our secure_config paths to include the secure repo's dirname now?

brianhw:
Looks good.  Just one question.

share/task.yml
mulby: I removed the concept of "secure_config_repo_dir". Instead the "secure_config" var needs to be the complete path to the config file including the directory that the secure config repo was checked out to.

Note that we shouldn't need to change anything since I also updated remote.py to include this information in the path, so you pass in the same information to that script and it prepends the appropriate directory name before passing the info along to ansible.

tobz (approved):
Yeah.  This looks good to me. 👍

edx/analytics/tasks/launchers/remote.py
```diff
@@ -148,27 +151,63 @@ def convert_args_to_extra_vars(arguments, uid):
...
+            'branch': 'origin/release',
+            'dir_name': 'config'
+        }
+        if arguments.secure_config_branch:
+            repos['secure']['branch'] = arguments.secure_config_branch
+
+    if arguments.override_config:
+        extra_vars['override_config'] = arguments.override_config
+    elif arguments.secure_config:
+        secure_repo_dir_name = repos.get('secure', {}).get('dir_name', 'config')
```
brianhw: How can 'dir_name' be anything other than config?  Does it make sense to only call this if 'secure' in repos?  Otherwise, if it's a repo-independent path, they would use override_config.   Not a big deal -- just checking my understanding.

share/task.yml
```diff
@@ -158,26 +166,34 @@
         WHEEL_PYVER: "{{ wheel_pyver }}"
       when: ansible_distribution in common_redhat_variants
 
+    - name: additional packages installed on Debian
+      shell: >
+        . {{ working_venv_dir }}/bin/activate && pip install {{ item }}
+        chdir={{ working_repo_dir }}
+      environment:
+        WHEEL_URL: "{{ wheel_url | default('http://edx-wheelhouse.s3-website-us-east-1.amazonaws.com/Debian/squeeze') }}"
```
brianhw: nit: DRY -- should these wheel_url values (debian and emr4) be pulled out so they're not repeated twice?

share/task.yml
brianhw: Okay, I found that code now.

brianhw: Is there a reason that this uses a 'cp' command rather than a copy: directive like the two steps below?  Including setting of mode?

brianhw (approved):
Just a few nits for your consideration, and then 👍

share/task.yml
mulby: Yes - `copy` is used to copy a file from the machine that is running ansible to the *remote* machine. `cp` allows me copy a file that is already on the remote machine to another location on the remote machine.

edx/analytics/tasks/launchers/remote.py
mulby: Good suggestion. Addressed in cf2fb41.

share/task.yml
mulby: Fixed in cf2fb41.

share/task.yml
brianhw: Thanks.

brianhw (approved):
LGTM.  👍

PR #328, opened 2016-12-15
Avoid getting url from path of another target.

brianhw:
This is a problem if a target is an HdfsTarget, because get_target_from_url(url).path is not the same as url for HDFS paths.


brianhw:
Please review:  @mulby, @tobz, @HassanJaveed84 .

mulby (approved):
LGTM

PR #329, opened 2016-12-15
Fix requires() bug in run_vertica_sql_scripts.

tobz:
Previously, we were storing a generator and then trying to yield that generator once we lazily generated it.  Here, we're still letting `get_downstream_task()` do its own internal caching/lazy loading, but we're returning the generator it produces directly in `requires()`.


mulby (approved):
I also tested this on a live cluster. Works like a charm.

brianhw:
👍

PR #330, opened 2016-12-19
WIP: Incremental User activity

HassanJaveed84:
@mulby @brianhw this is ready for an initial review. I still need to update the unit/acceptance tests.

UserActivity is rather simple, however, I could not find a way to use HivePartitionTask over data returned from the UserActivityIntervalTask and PathSelectionByDateIntervalTask. Suggestions ?

### DIFF
--- a/edx/analytics/tasks/user_activity.py
+++ b/edx/analytics/tasks/user_activity.py
@@ -207,6 +207,7 @@ class UserActivityPartitionTask(UserActivityDownstreamMixin, HivePartitionTask):
             pattern=self.pattern,
             n_reduce_tasks=self.n_reduce_tasks,
             warehouse_path=self.warehouse_path,
+            output_root=self.partition_location,
             overwrite=True,
         )
 
--- a/edx/analytics/tasks/user_activity.py
+++ b/edx/analytics/tasks/user_activity.py
@@ -129,7 +129,7 @@ class UserActivityIntervalTask(
     def output_path_for_key(self, key):
         date_string = key
         return url_path_join(
-            self.hive_partition_path('user_activity', date_string),
+            self.hive_partition_path('user_activity_daily', date_string),
             'user_activity_{date}'.format(
                 date=date_string,
             ),
@@ -216,7 +216,7 @@ class UserActivityPartitionTask(UserActivityDownstreamMixin, HivePartitionTask):
             self.interval.date_a,
             self.overwrite_from_date,
         ))
-        user_activity_root = url_path_join(self.warehouse_path, 'user_activity')
+        user_activity_root = url_path_join(self.warehouse_path, 'user_activity_daily')
         path_selection_task = PathSelectionByDateIntervalTask(
             source=[user_activity_root],
             interval=path_selection_interval,
--- a/edx/analytics/tasks/user_activity.py
+++ b/edx/analytics/tasks/user_activity.py
@@ -11,9 +11,10 @@ from edx.analytics.tasks.mapreduce import MapReduceJobTask, MapReduceJobTaskMixi
 from edx.analytics.tasks.pathutil import EventLogSelectionMixin, EventLogSelectionDownstreamMixin
 from edx.analytics.tasks.url import get_target_from_url
 import edx.analytics.tasks.util.eventlog as eventlog
-from edx.analytics.tasks.util.hive import WarehouseMixin, HiveTableTask, HivePartition, HiveQueryToMysqlTask
+from edx.analytics.tasks.util.hive import WarehouseMixin, HiveTableTask, HivePartition, HiveQueryToMysqlTask, BareHiveTableTask, HivePartitionTask, HiveQueryTask
 from edx.analytics.tasks.util.weekly_interval import WeeklyIntervalMixin
 from edx.analytics.tasks.decorators import workflow_entry_point
+from edx.analytics.tasks.mysql_load import MysqlInsertTask
 
 log = logging.getLogger(__name__)
 
@@ -126,12 +127,15 @@ class UserActivityDownstreamMixin(WarehouseMixin, EventLogSelectionDownstreamMix
     pass
 
 
-class UserActivityTableTask(UserActivityDownstreamMixin, HiveTableTask):
-    """Hive table that stores the set of users active in each course over time."""
+class UserActivityTableTask(UserActivityDownstreamMixin, BareHiveTableTask):
 
     @property
     def table(self):
         return 'user_activity_daily'
+
+    @property
+    def partition_by(self):
+        return 'dt'
 
     @property
     def columns(self):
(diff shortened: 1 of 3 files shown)


edx/analytics/tasks/user_activity.py
```diff
@@ -111,29 +110,159 @@ def _encode_tuple(self, values):
...
 
 
 class UserActivityDownstreamMixin(WarehouseMixin, EventLogSelectionDownstreamMixin, MapReduceJobTaskMixin):
     """All parameters needed to run the UserActivityTableTask task."""
     pass
 
 
-class UserActivityTableTask(UserActivityDownstreamMixin, HiveTableTask):
-    """Hive table that stores the set of users active in each course over time."""
+class UserActivityTask(UserActivityDownstreamMixin, OverwriteOutputMixin, MapReduceJobTask):
```
HassanJaveed84: Can this class go away ? All it's doing is combining the outputs and storing under 'user_activity_daily' path, which is then used in HivePartitionTask

edx/analytics/tasks/user_activity.py
```diff
@@ -23,23 +28,17 @@
...
+    EventLogSelectionMixin,
+    MultiOutputMapReduceJobTask):
 
-    Analyze the history of user actions and categorize their activity. Note that categories are not mutually exclusive.
-    A single event may belong to multiple categories. For example, we define a generic "ACTIVE" category that refers
-    to any event that has a course_id associated with it, but is not an enrollment event. Other events, such as a
-    video play event, will also belong to other categories.
 
-    The output from this job is a table that represents the number of events seen for each user in each course in each
-    category on each day.
```
brianhw, line 36: nit: Are these comments no longer valid?

edx/analytics/tasks/user_activity.py
brianhw: But it's also storing it in a single date under HivePartitionTask, best I can tell.  So that all the dates get copied under user_activity_path/dt=(today).  That seems to be at the heart of this:  that HivePartitionTask specifies a particular day and you are feeding it an interval.  So eitehr we need this class to convert from individual days to a single day (i.e. today), or you need to create a separate HivePartitionTask for each day in the interval.  The latter is what module_engagement does -- see ModuleEngagementPartitionTask.

edx/analytics/tasks/user_activity.py
```diff
@@ -143,86 +272,84 @@ def columns(self):
...
-        return 'course_activity'
+        query_format = textwrap.dedent("""
+            USE {database_name};
+            DROP TABLE IF EXISTS {table_name};
+            CREATE EXTERNAL TABLE {table_name} (
+                course_id STRING,
+                interval_start TIMESTAMP,
+                interval_end TIMESTAMP,
+                label STRING,
+                count INT
```
brianhw, line 407: At some point, pull these out into a Record.

edx/analytics/tasks/user_activity.py
```diff
@@ -338,13 +459,22 @@ def interval(self):
...
+        query_format = textwrap.dedent("""
+            USE {database_name};
+            DROP TABLE IF EXISTS {table_name};
+            CREATE EXTERNAL TABLE {table_name} (
+                course_id STRING,
+                year INT,
+                month INT,
+                label STRING,
+                count INT
+            )
```
brianhw, line 533: Record here too....

brianhw (commented):
Overall, it looks like we have two patterns for incremental:  module_engagement and user_location.  The former uses individual partitions with a day of data, and the latter has each partition with a complete set.  One might thing that module_engagement and user_activity would be more similar, since they are both calculating much the same thing.

edx/analytics/tasks/user_activity.py
```diff
@@ -111,29 +110,159 @@ def _encode_tuple(self, values):
...
-    def reducer(self, key, values):
-        """Cumulate number of events per key."""
-        num_events = sum(values)
-        if num_events > 0:
-            yield key, num_events
 
-    def output(self):
-        return get_target_from_url(self.output_root)
+    def multi_output_reducer(self, _date_string, values, output_file):
+        frequency = defaultdict(int)
```
mulby, line 115: I think we might have a scale issue here... if we are only grouping by date and then building a dict with each key being a distinct (course_id, username, label) tuple, that could get pretty big.

Not sure if it would get big enough to warrant a different strategy or not.

edx/analytics/tasks/user_activity.py
```diff
@@ -111,29 +110,159 @@ def _encode_tuple(self, values):
...
+
+        for value in values:
+            frequency[value] += 1
+
+        for key in frequency.keys():
```
mulby, line 120: nit: `for key, num_events in frequency.iteritems():`

edx/analytics/tasks/user_activity.py
```diff
@@ -111,29 +110,159 @@ def _encode_tuple(self, values):
...
+            num_events = frequency[key]
+            course_id, username, date_string, label = key
+            value = [course_id, username, date_string, label, num_events]
+            output_file.write('\t'.join([str(field) for field in value]))
```
mulby, line 124: nit: we could use a `Record` here!

edx/analytics/tasks/user_activity.py
mulby: I also think there is probably a way to make this class go away. Could we just make it a wrapper class that executes the overwrite logic but doesn't actually process any data? I'd rather not copy the data around for no reason (no transform is happening).

I think this is the first time we've need to have hive partitions for the downstream stuff to work - I would look into hive's [partition recovery feature](https://cwiki.apache.org/confluence/display/Hive/LanguageManual+DDL#LanguageManualDDL-RecoverPartitions(MSCKREPAIRTABLE)). We could use that until we get our act together and start supporting a persistent hive metastore.

Ideally we would have a "user_activity" table that has one partition per day. When the cluster spins up it overwrites N days of partitions in that table and then generates Hive partitions for the data that will be analyzed by queries later. We then run queries, which, in turn, generate partitions that are appended to their respective table. So the weekly table would have one partition per week, etc.

Similarly, when we load into MySQL and Vertica we should just be overwriting N-1 partitions and appending 1 partition.

mulby (commented):
Some initial thoughts

edx/analytics/tasks/user_activity.py
HassanJaveed84: I didn't quite understand what you meant by: ' We then run queries, which, in turn, generate partitions that are appended to their respective table. So the weekly table would have one partition per week, etc.' 
Are we talking about partitioning 'course_activity_daily', 'course_activity' and 'course_activity_monthly' tables ?

mulby: Yeah.

Although, now that I'm looking at this I'm wondering if we shouldn't just remove the daily and monthly tables since they aren't used.

PR #331, opened 2016-12-20
Hassan/inc activity

HassanJaveed84:
@mulby @brianhw 

This is an attempt using one partition per day for user activity. Currently it rebuilds the partitions, I haven't had a chance to look into recovering partitions yet, presumably it should be straight forward. 
Wanted to share this before our discussion regarding the design.


brianhw:
This looks like a very clean implementation.  Very clear what the intent of it all is, and logically consistent.  Downside is that loading history is very slow.  But that could be addressed by writing a special history loader to speed that process up.

edx/analytics/tasks/user_activity.py
```diff
@@ -143,230 +146,167 @@ def columns(self):
...
-        default=datetime.datetime.utcnow().date(),
-        description='A date within the month that will be the upper bound of the closed-open interval. '
-        'Default is today, UTC.',
-    )
-    months = luigi.IntParameter(
-        default=6,
-        description='The number of months to include in the analysis, counting back from the month that contains '
-        'the end_date.',
-    )
+class CourseActivityPartitionTask(WeeklyIntervalMixin, MapReduceJobTaskMixin, HivePartitionTask):
```
HassanJaveed84, line 287: This never gets scheduled.

HassanJaveed84:
@mulby @brianhw I'd been working on the partitioning and this is what I came up with. 

MysqlInsertTask requires an 'insert_source_task' property that needs to return a output target that supports .open. In order to achieve that I defined ```CourseActivityDataTask``` class which defines the output with ```get_target_from_url```. I had also defined ```CourseActivityPartitionTask``` which now looks redundant. 

I've removed a bunch of code for ease of making a sense of things.

edx/analytics/tasks/user_activity.py
```diff
@@ -375,5 +315,17 @@ def columns(self):
...
+    @property
+    def insert_source_task(self):
+        partition_task = CourseActivityPartitionTask(
+            end_date=self.interval.date_b,
+            weeks=1,
+            n_reduce_tasks=self.n_reduce_tasks,
+            warehouse_path=self.warehouse_path,
+            overwrite=True,
+        )
+        return partition_task.data_task
```
HassanJaveed84: We could use ```CourseActivityDataTask``` here directly..

edx/analytics/tasks/user_activity.py
mulby: Ah, I see what you mean... we probably don't need it given that we don't have any downstream Hive tasks that would query this table.

mulby:
@HassanJaveed84 This is awesome! Great work. Still needs the Vertica insert task for the user-course granularity data. Let me know if that isn't working the way we expect it to...

mulby:
We also have to figure out the "1 year problem" - that is, the MySQL database should only contain 1 year worth of weekly aggregate data.

PR #332, opened 2016-12-21
Speed up the enrollment validation acceptance test

mulby:
Previously this test was taking roughly 3.5 hours to execute because it was running S3 queries for all dates from 2014-08-01 up until the time when the test was actually executed (currently 2016-12-21). As test executions have gotten further and further away from august 2014, the test has been taking longer and longer to run.

This change simply introduces the concept of a templated fixture file so that we can adjust event times in the fixture before executing the test.

FYI @tobz


mulby:
Runtime was reduced to ~18 minutes from ~3.5 hours

brianhw (approved):
Looks great.   I didn't check the test failure, but once it passes on Jenkins and also runs on a local analyticstack, then 👍

mulby:
@HassanJaveed84 can you review this when you get a chance?

HassanJaveed84:
LGTM 👍

HassanJaveed84 (approved):

PR #336, opened 2016-12-28
Reorganize directory structure of edx-analytics-pipeline repository

brianhw:
Files have been moved, and the unit tests have been run locally.  

Still need to run acceptance tests, but ready for first-pass review.

@mulby @HassanJaveed84 @tobz


edx/analytics/tasks/launchers/local.py
```diff
@@ -52,12 +54,14 @@ def main():
...
 
     # Tell luigi what dependencies to pass to the Hadoop nodes
+    # - edx.analytics.tasks is used to load the pipeline code, since we cannot trust all will be loaded automatically.
     # - boto is used for all direct interactions with s3.
     # - cjson is used for all parsing event logs.
     # - filechunkio is used for multipart uploads of large files to s3.
     # - opaque_keys is used to interpret serialized course_ids
     #   - opaque_keys extensions:  ccx_keys
     #   - dependencies of opaque_keys:  bson, stevedore
+    luigi.hadoop.attach(edx.analytics.tasks)
```
brianhw, line 64: Here is the biggest change in this PR.  I added this here because it was by far the most convenient way to get the entire edx.analytics.tasks codebase loaded onto worker instances for running Hadoop jobs.  The existing method for loading edx.analytics.tasks is through Luigi's hadoop code loading the "current module" into the packaging jar.  With the reorg, the current model is edx.analytics.tasks.insights (or similar), and the other packages that are needed are no longer included.

brianhw:
Acceptance tests now pass, so this is now ready for final review.  I would like to get it in sooner than later since it touches a lot of stuff and I really want to avoid rebasing with all the renamed files.

While most of the PR is mechanical, there is one fundamental thing I changed.  I added "edx.analytics.tasks" as an extra-module in launch-task (local.py).  I made that change because it was by far the most convenient way to get the entire edx.analytics.tasks codebase loaded onto worker instances for running Hadoop jobs. The existing method for loading edx.analytics.tasks is through Luigi's hadoop code loading the "current module" into the packaging jar. With the reorg, the current model is edx.analytics.tasks.insights (or similar), and the other packages that are needed are no longer included.

tobz (approved):
This looks good to me, so long as local tests and acceptance tests run cleanly.

mulby (approved):
LGTM

PR #338, opened 2016-12-28
Make pipeline tasks easier to debug.

HassanJaveed84:
@mulby @brianhw I added a script that pulls the emr logs and displays Traceback info from the downloaded logs.

### DIFF
--- a/edx/analytics/tasks/tools/debug_emr_logs.py
+++ b/edx/analytics/tasks/tools/debug_emr_logs.py
@@ -1,0 +1,113 @@
+import argparse
+import gzip
+import os
+import re
+import shutil
+import boto.emr
+
+from subprocess import Popen, PIPE
+
+DEFAULT_REGION = 'us-east-1'
+
+def main():
+    arg_parser = argparse.ArgumentParser(description='Get Traceback information from emr-logs.')
+    arg_parser.add_argument(
+        '-b', '--bucket-name',
+        help='Bucket name',
+        default=os.environ.get('EMR_LOGS_BUCKET_NAME', None)
+    )
+    arg_parser.add_argument(
+        '-c', '--cluster-name',
+        help='Cluster name',
+    )
+    arg_parser.add_argument(
+        '-j', '--job-flow-name',
+        help='Job flow name.'
+    )
+    arg_parser.add_argument(
+        '-a', '--application-id',
+        default='',
+        help='Application ID.'
+    )
+    arg_parser.add_argument(
+        '-o', '--output-path',
+        default='./',
+        help='Output path for downloaded logs.'
+    )
+    arg_parser.add_argument(
+        '--cleanup',
+        action='store_true',
+        help='Cleanup logs.'
+    )
+
+    args = arg_parser.parse_args()
+
+    if not args.bucket_name:
+        exit(arg_parser.print_usage())
+
+    if not args.job_flow_name:
+        connection = boto.emr.connect_to_region(DEFAULT_REGION)
+        for cluster in connection.list_clusters().clusters:
+            if cluster.name == args.cluster_name:
+                args.job_flow_name = cluster.id
+                break
+
+    s3_emr_logs_url = "s3://{bucket_name}/{job_flow}/containers/{application_id}".format(
+        bucket_name=args.bucket_name,
+        job_flow=args.job_flow_name,
+        application_id=args.application_id,
+    )
+    return_code = download_emr_logs(s3_emr_logs_url, args.output_path)
+    if return_code != 0:
+       raise RuntimeError('Unable to download logs from S3.')
+
+    extract_files(args.output_path)
+
+    display_errors(args.output_path)
+
+    if args.cleanup:
+        shutil.rmtree(args.output_path)
+
+def download_emr_logs(s3_emr_logs_url, output_path):
+    print("Downloading logs.")
+    if not os.path.exists(output_path):
+        os.makedirs(output_path)
+
+    proc = Popen(
+        ['aws','s3','sync',s3_emr_logs_url, output_path, '--exclude', '*', '--include', '*stderr.gz'],
+        stdout=PIPE,
+    )
+    stdout = proc.communicate()[0]
+    return proc.returncode
+
+def extract_files(root_path):
+    print("Unzipping files.")
+
+    for path, dirs, files in os.walk(root_path):
+        for filename in files:
+            if filename.endswith('.gz'):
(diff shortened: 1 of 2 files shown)


brianhw:
Thanks for doing this, Hassan!  

I would be interested in knowing how performant this approach is.  What Gabe described is very similar to what I used to do, though I used my s3_util script to only pull down files ending with "stderr" to avoid pulling "stdlog" (or something) files that were big but not useful.  

With the switch to Hadoop 2 and the non-reuse of containers, I found that some of the nightly jobs generated so many stderr files (one per container) that it took a lot of time (and disk space) to do the download.  Your approach might be faster by using aws s3 sync instead, but I'd be curious. 

Another approach that would minimize local disk space use would be to get a listing of the remote files and then stream them one by one, and implement the grep either in code or on a temp file. 

A small thing:  it might be nice to occasionally override the grep parameter for after context, in those cases where 30 is not adequate.   (Seems like a fine default.)

edx/analytics/tasks/launchers/debug_emr_logs.py
```diff
@@ -0,0 +1,83 @@
...
+    )
+    arg_parser.add_argument(
+        '-o', '--output-path',
+        default='./',
+        help='Output path for downloaded logs.'
+    )
+
+    args = arg_parser.parse_args()
+
+    s3_emr_logs_url = "s3://{bucket_name}/{job_flow}/containers/{application_id}".format(
```
mulby: job_flow here has to be the cluster ID (for example: j-123457AC83754). I often want to just give this script the human-readable cluster name, like "Enrollment".

When the cluster is deployed we store this human readable name on it... ideally this script could use boto to find the most recently deployed cluster with that name and grab it's cluster ID to build this S3 URL.

See https://github.com/edx/edx-analytics-configuration/blob/master/batch/library/emr#L40

edx/analytics/tasks/launchers/debug_emr_logs.py
```diff
@@ -0,0 +1,83 @@
...
+       raise RuntimeError('Unable to download logs from S3.')
+
+    extract_files(args.output_path)
+
+    extract_errors(args.output_path)
+
+def download_emr_logs(s3_emr_logs_url, output_path):
+    print("Downloading logs.")
+    proc = Popen(
+        ['aws','s3','sync',s3_emr_logs_url, output_path],
```
mulby: Ideally, I'd prefer not to actually persist the logs to my local disk... but I'd be OK with this if needed!

We also might be able to use the "--include" or "--exclude" parameters to limit the set of files returned to just the stderr ones.

edx/analytics/tasks/launchers/debug_emr_logs.py
```diff
@@ -0,0 +1,83 @@
...
+            if filename.endswith('.gz'):
+                filename_with_path = os.path.join(path, filename)
+                basename = os.path.basename(filename_with_path)
+                destination_path = os.path.join(path, basename[:-3])
+                with gzip.open(filename_with_path, 'rb') as input_file:
+                    with open(destination_path, 'wb') as output_file:
+                        for line in input_file:
+                            output_file.write(line)
+
+def extract_errors(root_path):
```
mulby: We could probably borrow the logic that luigi uses here: https://github.com/edx/luigi/blob/edx/release/luigi/hadoop.py#L350

I think the stack trace is hex encoded by mrrunner.py here: https://github.com/edx/luigi/blob/edx/release/luigi/mrrunner.py#L63

So we could just take advantage of that fact to find them all without having a fixed window of N lines of output.

edx/analytics/tasks/launchers/debug_emr_logs.py
```diff
@@ -0,0 +1,83 @@
+import argparse
+import gzip
+import os
+
+from subprocess import Popen, PIPE
+
+def main():
+    arg_parser = argparse.ArgumentParser(description='Get Traceback information from emr-logs.')
+    arg_parser.add_argument(
```
mulby: It would be great if this defaulted to a value from some kind of environment variable so I could throw it into my .bashrc and never think about it again.

mulby (approved):
This is awesome! Already so much better than my current process.

I put in a few suggestions to make it even better, but I'm fine with this as is.

edx/analytics/tasks/launchers/debug_emr_logs.py
tobz: Yep.

edx/analytics/tasks/launchers/debug_emr_logs.py
tobz: I think it makes sense to allow both.  Give the user an option to specify the cluster name, which finds the latest job flow ID which matches the name, or let them specify the job flow ID specifically.

If the cluster name is present, it just looks it up, and the rest of the code is the same, as if they passed in a job flow ID.

The worst thing would be that we actually start to use this task because it's helpful, and then we want to go look at the error logs for the 2nd to last run of the workflow, and it can only show us the latest one. :P

brianhw: Amen to the last point -- it may often be that we do the analysis of why a job failed after it has been started a second time.  

I'm imagining down the road adding a build step that runs to do this dumping if a job fails as part of the definition on Jenkins.  For that we might still have the cluster running, so asking for the latest that had completed might not quite be the right one.  But we would hopefully get access to the current job flow id.

edx/analytics/tasks/launchers/debug_emr_logs.py
brianhw: It would also be good to know what file a given traceback was found from.

edx/analytics/tasks/launchers/debug_emr_logs.py
brianhw: Or again, use an s3util-like mechanism to provide an outer loop for processing a single file at a time.

HassanJaveed84:
- Added option to specify bucket name via env. variable. 
- Added option to cleanup downloaded logs.
- Pull cluster id from latest cluster if cluster id is not passed.
- Used luigi's error lookup.

@mulby @brianhw @tobz

mulby (approved):
Awesome!

LGTM - we can improve over time as we use it more.

tobz (approved):
Yep, LGTM. :shipit:

tobz:
We can ignore code coverage check results for now, because the thresholds haven't been set at all.

brianhw (approved):
LGTM 👍

PR #339, opened 2016-12-28
Add TravisCI support.

tobz:
TravisCI is already enabled on the repository, but it won't build branches without a `.travis.yml` present, so this will add that.

Right now, we only build against Python 2.7 since some of the Ansible stuff seems to break on Python 3. x and it would just red light every single build.

Builds are somewhere between 5m and 6m30s.  There's packages that just don't have wheels (like `psycopg2`) and some that seem to need OS-based libraries (like ATLAS, BLAS, LAPACK, etc) which I've yet to figure out a clean way to install precompiled.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,17 @@
+language: python
+python:
+  - "2.7"
+cache: pip
+
+addons:
+  apt:
+    packages:
+      - libblas-dev
+      - liblapack-dev
+      - libatlas-base-dev
+
+script:
+  - make test
+
+after_success:
+  - bash <(curl -s https://codecov.io/bash)
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,6 @@ test-local:
 
 test: test-requirements develop test-local
 
-
 test-acceptance: test-requirements
 	LUIGI_CONFIG_PATH='config/test.cfg' python -m coverage run --rcfile=./.coveragerc -m nose --nocapture --with-xunit -A acceptance $(ONLY_TESTS)
 


brianhw:
Looks good.  I have a vague memory of others adding a requirements.txt file at root for some reason, and I thought it was for Travis.  You didn't need to do that?

tobz:
Nope.  Since we're invoking `make test` directly, which will make sure dependencies are installed, we're covered.

.travis.yml
```diff
@@ -0,0 +1,15 @@
...
+addons:
+  apt:
+    packages:
+      - libblas-dev
+      - liblapack-dev
+      - libatlas-base-dev
+
+script:
+  - env
+  - make test
```
mulby, line 14: I usually use "make coverage" to also generate coverage metrics and run quality checks. Did you explicitly want to handle those separately? codecov etc?

mulby (approved):
LGTM

One question about the make target - not a blocker.

.travis.yml
tobz: So, with the way Travis CI is integrated, all checks would generate a single pass/fail result.

So, if you invoke multiple commands -- tests, coverage, quality, etc -- your PR would fail entirely if any one of them failed individually.  I'm OK with doing that, but it's a departure from how we plug our Jenkins infrastructure into edx-platform.

I think it's probably OK to do it this way, and just bite the bullet and deal with failures as they come, so long as all of us are good with that.

brianhw: Is there an alternative?  That is, it would be great to get the quality and coverage artifacts built and accessible to PRs.  It sounds like it's not possible to have Travis do separate things in separate builds, because there is one yaml file with one scripts directive.

brianhw:
Oh, and you mentioned Python 3.  We've not pushed to support that because AWS EMR doesn't support it.  Only python 2.7 now (though at first they were on 2.6 when we first started, so there's some small remnants of that around).

.travis.yml
tobz: The alternative, if you want your PR checks to look like the ones we have in edx-platform, is to do something like what we do with edx-platform (multiple Jenkins webhooks or whatever) or to use separate integrations -- like push coverage metrics from `make coverage` to a coverage service which pushes THAT as a separate check into the PR.

To be explicit, we can do all of these things in a single TravisCI build.. you'd just have to go into the build results if there was a failure so you could figure out precisely *what* failed.

mulby:
It looks like in the configuration repo they have the build split up into several "jobs" which are each separate make targets... not sure how to do this... but maybe that's an option here?

Example:
https://travis-ci.org/edx/configuration/builds/187281075

tobz:
Right, but notice how it's still a single line item in the list of checks for the PR?  That's specifically what I was referring to.

brianhw:
I think I'm fine with having a single check in the PR, as long as I can get to an appropriate artifact or artifacts to tell me what went wrong.  Generally I expect unit tests to pass, and at the moment we don't fail on coverage or pep8/pylint changes.  So I see it partly as a way to see what the pep8/pylint/coverage results are for a PR.

mulby:
Oh, I'm fine with a single line in the PR, particularly given that if I drill into it it's broken out by job.

PR #341, opened 2016-12-29
Upgrade pip to latest version.

brianhw:
Also avoid installing virtualenv onto itself.  See AN-4322 for description of the problem.  Jenkins has an old version of virtualenv (among other places), which corrupts an existing virtualenv that contains a more recent version of pip.  Instead of trying to hunt down the problems with virtualenv and upgrade them all, this approach just assumes we avoid running virtualenv when the virtual env already exists. 

This also required a change on Jenkins job configuration to only run virtualenv as part of the configuration bash script if the virtualenv didn't already exist.  The same one-line change can be made to the other jobs that run acceptance tests.


brianhw:
Running acceptance tests now.

mulby (approved):
LGTM

Does this mean we can get rid of all the "wheel URL" stuff?

brianhw:
It does look like it is finding a lot of wheels -- not for everything, but for some of the more significant stuff anyway.

tobz (approved):
LGTM.  Love that we can upgrade the crypto library now. ❤️

HassanJaveed84 (approved):
I've come across this frustrating issue several times, great to have this fix. 👍

brianhw:
The module_engagement acceptance test failed due to the upgrade in numpy.  I reverse engineering the failure, and found the problem was a change in the behavior of numpy.percentile() when presented with a single 'inf' value.  This happens when a single learner has tried problems but completed none of them.  The old behavior for 15th and 85th percentile of [inf] was (inf, inf).  The new (broken?) behavior is (nan, nan).  So instead of { low: (0, inf), normal: (inf, inf) }, the ranges were { normal: (nan, nan), high: (nan, inf) }.   When written to Mysql, the 'inf' values are replaced with NULL, and the 'nan' values are replaced with 0.   I'm not sure what the real behavior here should be.

I went to update the unit tests for ModuleEngagementSummaryMetricRangesDataTask and found that none were actually run.   The test input had zero for days, so all was being skipped.   I fixed that and a few other details.  However, there was one test that failed, and I would appreciate reviewing whether the code is wrong or the test is wrong.

Finally, I bisected numpy versions (which is much faster with wheels!), and found that the change in numpy.percentile() behavior was introduced between 1.8.2 and 1.9.0 -- soon after the old version we've been using (1.8.0).  The current version is 1.11.3.  So I've rolled this PR back to 1.8.2 to see if that is adequate to work with Travis. 

I think it would be helpful to have some review (again) of what the values are that should come out of this.  I don't see much documentation that addresses the boundary cases that we discussed ad nauseum during the original development, but maybe I'm not looking in the right place.

brianhw:
Nope.  Numpy 1.8.2 is not good enough to get Travis to work.

mulby:
Great catch @brianhw. The intent was for the tests to be the documentation, but since they weren't being run... they weren't in sync with the code (they mostly were). I changed the logic so that zero values are included in the calculation and must have forgotten to update the test since it wasn't being run. Oops.

As for this numpy upgrade. The difference between `float('NaN')` and `float('inf')` is not consequential here, so I just coerced 'nan' into 'inf'. Hopefully this works.

mulby:
Acceptance tests now pass on this branch!

http://jenkins-ci.analytics.edx.org/view/ad-hoc/job/edx-analytics-pipeline-acceptance-test-manual/1241/

PR #342, opened 2016-12-29
Run coverage and tests seperately.

tobz:
As stated in the subject.  This will run TravisCI in a build matrix mode, running both unit tests and coverage/quality separately and concurrently.

I had to do some annoying git reclone action here because Travis defaults to cloning in a detached state, which prevents `diff-cover` and `diff-quality` from comparing the branch to `master`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ env:
 install:
   - git clone https://github.com/$TRAVIS_REPO_SLUG.git $TRAVIS_REPO_SLUG
   - cd $TRAVIS_REPO_SLUG
-  - git checkout -qf $TRAVIS_PULL_REQUEST_SHA || git checkout -qf $TRAVIS_BRANCH
+  - git checkout -f ${TRAVIS_PULL_REQUEST_BRANCH:-$TRAVIS_BRANCH}
 
 script:
   - make $MAKE_TARGET
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,17 @@ addons:
       - liblapack-dev
       - libatlas-base-dev
 
+env:
+  - MAKE_TARGET=test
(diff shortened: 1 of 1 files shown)


mulby:
it looks this checkout method doesn't work?

Approach looks good to me though.

tobz:
Yeah, Travis checks out in detached mode normally, so that was supposed to
be a way to re-checkout in not-detached mode so we could reference origin,
etc, for diff-cover and diff-quality.  It seems to choke, though, with
rebased force-pushes.

I'm going to try and have it checkout the branch rather than the commit
once I figure out the syntax to make it work for both normal pushes and PRs.

tobz:
I'm going to close this PR for now.

The gist is basically: because Travis CI offers no way to *not* shallow clone a repository, we `diff-cover` and `-diff-quality` won't be able to refer back to the base branch to check against.  We could, potentially, check out a 2nd time before tests run, not detached, but this poses issues when Travis tries to run not only against the branch, but the PR merge result.

It was hacky enough trying to get the former to work, and there seems to be no way to detect that you're "in" one or another.

If we do want to do this, I think we need to run it against the entire codebase, not just the diff... or we'd need to use another CI service.

mulby:
I'm fine with just running it over the whole code base.

Coveralls provides the incremental coverage difference... so all we are really losing is the "diff-quality"... maybe we could find another SaaS vendor that provides a coveralls-like service for pep8/pylint that could fill that gap?

I'd also be OK with just having it run over the entire code-base. Particularly if we did one big lift to get it 100% clean.

tobz:
Right, it would just be that things would always be failing until it happens.  I worry about having it be noisy as it's brand new.

I could also just enable it in a PR where I also do many/all of the fixes to make sure that it does pass.

PR #344, opened 2017-01-03
Add support needed for loading segment events.

brianhw:
Include unit tests for these, so they can go to master.  Functionality includes:

* Support for sparse Record objects, where values are specified only for some fields.
* Support for "timestamp" pattern when finding files in an interval. 
* Support for defining partitions when loading data to Vertica.


mulby (approved):
LGTM

tobz (approved):
LGTM.

PR #347, opened 2017-01-09
Remove support for pip using wheels

brianhw:


Makefile
```diff
@@ -1,10 +1,5 @@
 
-# If a wheel repository is defined, then have pip use that.  But don't require the use of wheel.
-ifneq ($(strip $(WHEEL_URL)),)
-	PIP_INSTALL = pip install --use-wheel --find-links=$$WHEEL_URL/Python-$$WHEEL_PYVER
-else
-	PIP_INSTALL = pip install
-endif
+PIP_INSTALL = pip install
```
mulby: Maybe we should remove this variable? It doesn't appear to actually be providing any value at this point...

mulby (commented):
Also need to clean up the acceptance tests and remote-task:

https://github.com/edx/edx-analytics-pipeline/search?utf8=%E2%9C%93&q=wheel

And this spot in edx/configuration: https://github.com/edx/configuration/blob/b10da53da28358ee82f56307ec30eaf9c329e292/playbooks/roles/analytics_pipeline/files/acceptance.json#L16

Makefile
brianhw: Seems reasonable.  Done.

brianhw:
Do we have to update configuration in order to not have things break?  It's one thing to set an environment variable that is no longer used, and another to pass an argument that is no longer supported.

mulby:
We shouldn't need to update configuration. Just something to do eventually (and not lose track of).

brianhw:
Enrollment acceptance tests pass on analyticstack and on Jenkins.  Running all acceptance tests now...

tobz (approved):
LGTM. :shipit:

PR #348, opened 2017-01-10
Fix to acceptance test on analyticstack. 

brianhw:
For problem response acceptance test, update the pattern used for finding output files.

### DIFF
--- a/edx/analytics/tasks/tests/acceptance/test_problem_response.py
+++ b/edx/analytics/tasks/tests/acceptance/test_problem_response.py
@@ -70,10 +70,13 @@ class ProblemResponseReportWorkflowAcceptanceTest(AcceptanceTestCase):
 
     def validate_reports(self):
         """Check the generated reports against the expected output files."""
+        actual_output_targets = self.get_targets_from_remote_path(self.report_output_root)
+        self.assertEqual(len(actual_output_targets), 2)
+
         for course_id in ('OpenCraft_PRDemo1_2016', 'OpenCraft_PRDemo2_2016'):
             report_file_name = '{}_problem_response.csv'.format(course_id)
-            actual_output_targets = self.get_targets_from_remote_path(self.report_output_root, report_file_name)
-            self.assertEqual(len(actual_output_targets), 1, '{} not created'.format(report_file_name))
+            actual_output_targets = self.get_targets_from_remote_path(self.report_output_root, "*{}".format(report_file_name))
+            self.assertEqual(len(actual_output_targets), 1, '{} not created in {}'.format(report_file_name, self.report_output_root))
             actual_output = actual_output_targets[0].open('r').read()
 
             expected_output_file = os.path.join(self.data_dir, 'output', 'problem_response', report_file_name)


edx/analytics/tasks/tests/acceptance/test_problem_response.py
```diff
@@ -70,10 +70,13 @@ def validate_hive(self):
...
+        actual_output_targets = self.get_targets_from_remote_path(self.report_output_root)
+        self.assertEqual(len(actual_output_targets), 2)
+
         for course_id in ('OpenCraft_PRDemo1_2016', 'OpenCraft_PRDemo2_2016'):
             report_file_name = '{}_problem_response.csv'.format(course_id)
-            actual_output_targets = self.get_targets_from_remote_path(self.report_output_root, report_file_name)
-            self.assertEqual(len(actual_output_targets), 1, '{} not created'.format(report_file_name))
+            actual_output_targets = self.get_targets_from_remote_path(self.report_output_root, "*{}".format(report_file_name))
+            self.assertEqual(len(actual_output_targets), 1, '{} not created in {}'.format(report_file_name, self.report_output_root))
             actual_output = actual_output_targets[0].open('r').read()
```
tobz, line 80: Can you add an assertion before this line that the size of `actual_output_targets` is 1?

edx/analytics/tasks/tests/acceptance/test_problem_response.py
```diff
@@ -70,10 +70,13 @@ def validate_hive(self):
...
         """Check the generated reports against the expected output files."""
+        actual_output_targets = self.get_targets_from_remote_path(self.report_output_root)
+        self.assertEqual(len(actual_output_targets), 2)
+
         for course_id in ('OpenCraft_PRDemo1_2016', 'OpenCraft_PRDemo2_2016'):
             report_file_name = '{}_problem_response.csv'.format(course_id)
-            actual_output_targets = self.get_targets_from_remote_path(self.report_output_root, report_file_name)
-            self.assertEqual(len(actual_output_targets), 1, '{} not created'.format(report_file_name))
+            actual_output_targets = self.get_targets_from_remote_path(self.report_output_root, "*{}".format(report_file_name))
+            self.assertEqual(len(actual_output_targets), 1, '{} not created in {}'.format(report_file_name, self.report_output_root))
```
tobz, line 79: Yay for more detailed logging!

edx/analytics/tasks/tests/acceptance/test_problem_response.py
```diff
@@ -70,10 +70,13 @@ def validate_hive(self):
...
     def validate_reports(self):
         """Check the generated reports against the expected output files."""
+        actual_output_targets = self.get_targets_from_remote_path(self.report_output_root)
+        self.assertEqual(len(actual_output_targets), 2)
+
         for course_id in ('OpenCraft_PRDemo1_2016', 'OpenCraft_PRDemo2_2016'):
             report_file_name = '{}_problem_response.csv'.format(course_id)
-            actual_output_targets = self.get_targets_from_remote_path(self.report_output_root, report_file_name)
-            self.assertEqual(len(actual_output_targets), 1, '{} not created'.format(report_file_name))
+            actual_output_targets = self.get_targets_from_remote_path(self.report_output_root, "*{}".format(report_file_name))
```
tobz, line 78: Why do we now want to match anything at the front of the filename?

brianhw:
Acceptance tests passed.  Awaiting thumbs.

tobz (changes requested):

edx/analytics/tasks/tests/acceptance/test_problem_response.py
brianhw: Because it's actually comparing against the pattern against the entire path for HDFS paths, whereas for S3 paths it only matches against the part that is below the output_root.   And it turns out that the latter is the filename.  As a result, this only worked on Jenkins but not analyticstack.

edx/analytics/tasks/tests/acceptance/test_problem_response.py
brianhw: I'm confused.  That is what the previous line I added was doing, no?

tobz: Ha.  Sorry.  You're right. :)

tobz (approved):
Alright.  This all looks good to me after your responses.

mulby (approved):

PR #349, opened 2017-01-12
Hassan/characterize events

HassanJaveed84:
@mulby @brianhw @tobz 

These are the tasks that I ran as part of my story.

### DIFF
--- a/edx/analytics/tasks/monitor/characterize_events.py
+++ b/edx/analytics/tasks/monitor/characterize_events.py
@@ -38,8 +38,9 @@ class ListS3FilesWithDateTask(OverwriteOutputMixin, luigi.Task):
                 last_modified = key.last_modified
                 values = [key_url, last_modified]
                 metadata = key.metadata
+
                 if metadata:
-                    match = re.match('.*?mtime:(?P<mtime>\d{10}).*?ctime:(?P<ctime>\d{10})', metadata)
+                    match = re.match('.*?mtime:(?P<mtime>\d{10}).*?ctime:(?P<ctime>\d{10})', metadata.get('s3cmd-attrs', ''))
                     if match:
                         mtime = match.group('mtime')
                         values.append(mtime)
--- a/edx/analytics/tasks/monitor/characterize_events.py
+++ b/edx/analytics/tasks/monitor/characterize_events.py
@@ -2,8 +2,10 @@ import luigi
 import os
 import logging
 import re
+import datetime
 
 from boto.s3.connection import S3Connection
+from dateutil import parser
 
 from edx.analytics.tasks.common.mapreduce import MapReduceJobTask
 from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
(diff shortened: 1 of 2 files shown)


edx/analytics/tasks/monitor/characterize_events.py
```diff
@@ -0,0 +1,98 @@
...
+
+    bucket = luigi.Parameter()
+
+    key_prefix = luigi.Parameter(default='logs/tracking/')
+
+
+    url_to_timestamp_map = {}
+
+    def requires_local(self):
+        return ListS3FilesWithDateTask(
```
brianhw, line 52: Is there any reason to call super here?   I'm looking at events_obfuscate.py as a model of something that works.

brianhw: And then you could add the target explicitly to the requirements as a dict entry.  If super doesn't return a dict, then create an empty one and add the super's return to that new dict.  Then add this requirement as a particular key value in the returned dict.

edx/analytics/tasks/monitor/characterize_events.py
```diff
@@ -0,0 +1,98 @@
...
+            key_prefix=self.key_prefix,
+            output_root=self.output_root,
+        )
+
+    def init_local(self):
+        super(CharacterizeEventsTask, self).init_local()
+        with self.input_local().open() as f_in:
+            lines = f_in.readlines()
+            for line in lines:
+                file_url, timestamp = line.split()
```
brianhw, line 64: Does it make sense to split on \t instead?  That wouldn't explain your issue, but...

edx/analytics/tasks/monitor/characterize_events.py
```diff
@@ -0,0 +1,98 @@
...
+            bucket=self.bucket,
+            key_prefix=self.key_prefix,
+            output_root=self.output_root,
+        )
+
+    def init_local(self):
+        super(CharacterizeEventsTask, self).init_local()
+        with self.input_local().open() as f_in:
+            lines = f_in.readlines()
+            for line in lines:
```
brianhw: nit: you could do 'for line in f_in',  I believe.

edx/analytics/tasks/monitor/characterize_events.py
```diff
@@ -0,0 +1,98 @@
...
+                file_url, timestamp = line.split()
+                self.url_to_timestamp_map[file_url] = timestamp
+
+    def mapper(self, line):
+        input_file = os.environ['mapreduce_map_input_file']
```
brianhw: nit:  use self.get_map_input_file()

edx/analytics/tasks/monitor/characterize_events.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def requires_local(self):
+        return ListS3FilesWithDateTask(
+            bucket=self.bucket,
+            key_prefix=self.key_prefix,
+            output_root=self.output_root,
+        )
+
+    def init_local(self):
+        super(CharacterizeEventsTask, self).init_local()
```
brianhw: I would probably try adding log statements in here as well, to confirm that it's being read.  That it's being called, and that the dict is getting populated.

brianhw (commented):
I don't see anything wrong here that would explain why you are getting zero entries.  Again, it looks close enough to the filtering done in events_obfuscation, which (hopefully) works.

edx/analytics/tasks/monitor/characterize_events.py
```diff
@@ -0,0 +1,98 @@
...
+class CharacterizeEventsTask(OverwriteOutputMixin, EventLogSelectionMixin, MapReduceJobTask):
+
+    output_root = luigi.Parameter()
+
+    bucket = luigi.Parameter()
+
+    key_prefix = luigi.Parameter(default='logs/tracking/')
+
+
+    url_to_timestamp_map = {}
```
HassanJaveed84: We need to set the data structure on the object for it to be available on hadoop nodes.

PR #350, opened 2017-01-18
Use default value for overwrite throughout.

brianhw:
So in ImportCourseSummaryEnrollmentsIntoMysql, it would be true for
writing to Mysql.


brianhw:
Enrollment acceptance test passed.

mulby (approved):
LGTM

HassanJaveed84 (approved):
👍

PR #351, opened 2017-01-30
support multiple roles

mulby:
Use different roles for different schemas loaded into Vertica. This creates three categories of roles: standard, restricted, and business_intelligence_team. It also adds the concept of "restricted columns" which are only visible to the users with the `restricted` role.

FYI @HassanJaveed84


mulby:
OK - my manual tests passed:

* Ran a warehouse load into production_mulby
* Ran an event load
* Ran an otto database import

This is just waiting on review to merge.

edx/analytics/tasks/common/vertica_load.py
```diff
@@ -428,6 +428,30 @@ def copy_data_table_from_target(self, cursor):
...
             )
 
+    @property
+    def restricted_columns(self):
+        return []
+
+    def create_access_policies(self, connection):
+        cursor = connection.cursor()
+        for column in self.restricted_columns:
+            restricted_roles_param = luigi.Parameter(is_list=True, config_path={'section': 'vertica-export', 'name': 'restricted_roles'})
```
brianhw: This looks like it is required to be found in the config file.  Any reason for that?  I guess that's okay if one is declaring a table as having a restricted column.  Alternative is to have restricted roles be empty, so only dbadmin will be enabled for seeing the column.

edx/analytics/tasks/common/vertica_load.py
```diff
@@ -428,6 +428,30 @@ def copy_data_table_from_target(self, cursor):
...
+CREATE ACCESS POLICY ON {schema}.{table} FOR COLUMN {column}
+CASE WHEN {expression} THEN {column}
+ELSE 'Restricted'
+END
+ENABLE;""".format(schema=self.schema, table=self.table, column=column, expression=expression)
+            try:
+                cursor.execute(statement)
+            except Exception:
+                # This is expected to fail if the access policy already exists.
+                # TODO: check if this is an "expected" failure, or an unexpected one.
```
brianhw: Shouldn't this be figured out before merging to master?

edx/analytics/tasks/warehouse/load_internal_reporting_events.py
```diff
@@ -1219,6 +1219,11 @@ class PruneEventPartitionsInVertica(EventRecordLoadDownstreamMixin, SchemaManage
     # create queries to delete the excess partitions.
     earliest_date = None
 
+    roles = luigi.Parameter(
```
brianhw: nit: I had to look at code elsewhere to realize this was overwriting the standard definition of roles, and that vertica_roles derives from roles.  Might help to have a brief comment saying that we are overwriting the
parameter in order to restrict access.

brianhw (commented):
LGTM overall.  One question about error handling, however (and nits on doc).

edx/analytics/tasks/common/vertica_load.py
mulby: Good point, I've made the parameter optional!

edx/analytics/tasks/common/vertica_load.py
mulby: Yeah, I'll run a test now.

edx/analytics/tasks/warehouse/load_internal_reporting_events.py
mulby: Good idea. Done.

edx/analytics/tasks/common/vertica_load.py
```diff
@@ -428,6 +429,37 @@ def copy_data_table_from_target(self, cursor):
...
+            log.debug(statement)
+            try:
+                cursor.execute(statement)
+            except QueryError as query_error:
+                # This is expected to fail if the access policy already exists.
+                expected_error = 'Access policy for COLUMN "{column}" already exists on "{table}"'.format(
+                    column=column,
+                    table=self.table
+                )
+                if query_error.error_response.mssage == expected_error:
```
brianhw: Is this a typo? (mssage?)

brianhw (approved):
Fix the typo, otherwise LGTM.  👍

tobz (approved):
LGTM. 👍

PR #352, opened 2017-01-31
[AN-8382] Add script to save Hadoop counters after workflow completion

tobz:
This script grabs all available job counters from the HistoryServer once a workflow finishes, and transforms them according to configuration, making them suitable to then ship to a local statsd endpoint.

The script will collect all available counters -- default and custom -- and convert their Java-esque names to a simpler format.  It also allows multiple "templates" to be defined, which allows generating multiple output metrics (which allow dimensions to be added, removed, for aggregation, etc) from a single input metric.

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -3,6 +3,7 @@
 # These dependencies are explicitly included in code.
 
 argparse==1.2.1 	# Python Software Foundation License
+boto3==1.4.4 # Apache 2.0
 ciso8601==1.0.3         # MIT
 edx-opaque-keys==0.4    # AGPL
 edx-ccx-keys==0.2.1     # AGPL
@@ -21,11 +22,13 @@ python-gnupg==0.3.9	# BSD
 pytz==2016.10		# ZPL
 requests==2.12.4        # Apache 2.0
 six==1.10.0		# MIT
+statsd==3.2.1  # MIT
 stevedore==1.19.1 	# Apache 2.0
 ua-parser==0.3.6 	# Apache
 urllib3==1.19.1         # MIT
 user-agents==0.3.2	# MIT
 vertica-python==0.6.11  # MIT
+yarn-api-client==0.2.3 # BSD
 
 git+https://github.com/edx/luigi.git@a73700ca51685974220ef6069d2f078312055444#egg=luigi 		# Apache License 2.0
 git+https://github.com/edx/pyinstrument.git@a35ff76df4c3d5ff9a2876d859303e33d895e78f#egg=pyinstrument     # BSD
--- a/scripts/collect-hadoop-metrics.py
+++ b/scripts/collect-hadoop-metrics.py
@@ -1,0 +1,289 @@
+import json
+import sys
+import boto3
+import requests
+import statsd
+import yaml
+from yarn_api_client import HistoryServer
+
+
+def query_metadata_service(path=''):
+    """
+    Gets the value at the specified path by querying the instance metadata service.
+
+    Returns the textual response and does not attempt to deserialize the response body.  The instance metadata service[1]
+    provides basic information about the given EC2 instance the endpoint is called from, providing things like instance ID,
+    security group IDs, availability zone, block device mapping, and more.
+
+    [1] - http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html
+    """
+    try:
+        r = requests.get('http://169.254.169.254/latest/meta-data/' + path)
+        return r.text
+    except Exception as e:
+        print "[collect-hadoop-metrics] exception when querying instance metadata service for path '{}': {}".format(path, e)
+        return None
+
+
+def get_cluster_region():
+    """
+    Gets the AWS region this cluster is running in.
+    """
+    node_az = query_metadata_service('placement/availability-zone')
+    # node_az is the full AZ identifier i.e. us-east-1d so we want to take everything
+    # but the last character, which gives us the region.  Historically, we have no reason
+    # to believe that AZs will start to look like us-east-1tp or anything, so I think
+    # this is a reasonably safe bet.
+    if node_az is not None:
+        return node_az[:-1]
+
+    return None
+
+
+def get_local_address_on_emr():
+    """
+    Gets the local/private IPv4 address of this instance, assuming we're on AWS.
+    """
+    return query_metadata_service('local-ipv4')
+
+
+def get_instance_id():
+    """
+    Gets the instance ID of this instance.
+    """
+    return query_metadata_service('instance-id')
+
+
+def get_job_flow_id():
+    """
+    Gets the job flow ID that this instance belongs to.
+
+    We utilize the job flow information[1] put on disk when running an EMR cluster
+
+    [1] - http://docs.aws.amazon.com/emr/latest/DeveloperGuide/Config_JSON.html
+    """
+    job_flow_id = None
+    try:
+        with open('/mnt/var/lib/info/job-flow.json', 'r') as job_flow_file:
+            parsed = json.loads(job_flow_file.read())
+            job_flow_id = parsed.get('jobFlowId', None)
+    except:
+        print "[collect-hadoop-metrics] Error reading job flow information.  Not on AWS/EMR?"
+
+    return job_flow_id
+
+
+def get_cluster_name():
+    """
+    Gets the name of the EMR cluster.
+    """
+    cluster_id = get_job_flow_id()
+    if cluster_id is not None:
+        try:
+            emr = boto3.client('emr', region_name=get_cluster_region())
+            cluster = emr.describe_cluster(ClusterId=cluster_id)
+
+            return cluster['Cluster']['Name']
+        except:
+            print "[collect-hadoop-metrics] Error querying EMR API for cluster name"
+            return None
+
+    return None
+
+
+def get_context():
+    """
+    Gets contextual information about where the script is being run.
+
+    This includes things like the instance ID, and in the cases of special environments like EMR, the job flow ID.
+    """
+    return {
+        'instance_id': get_instance_id(),
+        'job_flow_id': get_job_flow_id(),
+        'cluster_name': get_cluster_name(),
+    }
+
+
+def get_job_name_from_luigi_task_id(task_id):
+    """
+    Gets the name of the job based on the task ID that Luigi uses.
+
+    This is usually in the form of TaskClassName(param=value, param=value, ...) and so we take the simple approach
+    of chopping off whatever is in front of the parameters tuple.
+    """
+    return task_id.split('(')[0]
+
+
+def get_prefix_from_counter_group_name(group_name):
+    """
+    Gets the simplified version of a counter group name, suitable for use with Graphite.
+
+    Normally, Hadoop counter groups have names which correspond to the fully qualified Java class name
+    which emitted the counters e.g. org.apache.hadoop.mapreduce.lib.input.FileInputFormatCounter.
+
+    We're simply grabbing the last component in the full string, and then running it through a mapping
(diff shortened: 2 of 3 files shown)


requirements/default.txt
```diff
@@ -26,6 +27,7 @@ ua-parser==0.3.6 	# Apache
 urllib3==1.19.1         # MIT
 user-agents==0.3.2	# MIT
 vertica-python==0.6.11  # MIT
+yarn-api-client==0.2.3 # BSD
```
brianhw, line 31: Does this pull in any other dependencies?  If so, they should be pinned below....

scripts/collect-hadoop-metrics.py
```diff
@@ -0,0 +1,287 @@
+import json
+import requests
+import boto3
+import statsd
```
brianhw: should this be pinned?  I don't see it in the requirements...

brianhw: Also, we've been using pylint settings that call for separating python libs from third-party.  So json and sys would be listed first, then boto3, requests, statsd, yarn_api_client.

scripts/collect-hadoop-metrics.py
```diff
@@ -0,0 +1,287 @@
...
+import sys
+from yarn_api_client import HistoryServer
+
+
+def query_metadata_service(path=''):
+  """
+  Gets the value at the specified path by querying the instance metadata serice.
```
brianhw: nit:  serice => service

scripts/collect-hadoop-metrics.py
```diff
@@ -0,0 +1,287 @@
...
+
+  Returns the textual response and does not attempt to deserialize the response body.
+  """
+  try:
+    r = requests.get('http://169.254.169.254/latest/meta-data/' + path)
```
brianhw: This is a special URL?  Would be good to document what it is.  In particular, what the host is.

scripts/collect-hadoop-metrics.py
```diff
@@ -0,0 +1,287 @@
...
+
+def get_cluster_region():
+  """
+  Gets the AWS region this cluster is running in.
+  """
+  node_az = query_metadata_service('placement/availability-zone')
+  # node_az is the full AZ identifier i.e. us-east-1d so we want to take everything
+  # but the last character, which gives us the region.  Historically, we have no reason
+  # to believe that AZs will start to look like us-east-1tp or anything, so I think
+  # this is a reasonably safe bet.
```
brianhw: We can't just ask for the region, instead of pulling it out of the AZ?  That's pretty annoying...

scripts/collect-hadoop-metrics.py
```diff
@@ -0,0 +1,287 @@
...
+  return query_metadata_service('instance-id')
+
+
+def get_job_flow_id():
+  """
+  Gets the job flow ID that this instance belongs to.
+  """
+  job_flow_id = None
+  try:
+    with open('/mnt/var/lib/info/job-flow.json', 'r') as job_flow_file:
```
brianhw: It might be good to add a comment here pointing to http://docs.aws.amazon.com/emr/latest/DeveloperGuide/Config_JSON.html

requirements/default.txt
mulby, line 31: Do either of these new deps have new transitive dependencies that we also need to pin?

scripts/collect-hadoop-metrics.py
```diff
@@ -0,0 +1,287 @@
...
+    config = yaml.load(f)
+
+  input_config = config.get('input', {})
+  output_config = config.get('output', {})
+
+  metric_templates = input_config.get('templates', [])
+
+  # Check the HistoryServer address specifically, because we can't do anything without it.
+  hs_address = input_config.get('hs_address', None)
+  hs_address = input_config.get('hs_address', None)
```
mulby: duplicated line!

scripts/collect-hadoop-metrics.py
```diff
@@ -0,0 +1,287 @@
...
+  # Grab all jobs from the history server.
+  hs = HistoryServer(hs_address)
+  response = hs.jobs()
+
+  jobs = response.data.get('jobs', {})
+  if jobs is None:
+    print "[collect-hadoop-metrics] HistoryServer indicates no jobs have run.  Exiting."
+    sys.exit(0)
+
+  job_visits = {}
```
brianhw: nit: maybe call this "job_instances"?   Or job_counts?   It's not really a "visit"....

scripts/collect-hadoop-metrics.py
mulby: do we need to add anything to our requirements for this?

scripts/collect-hadoop-metrics.py
```diff
@@ -0,0 +1,287 @@
...
+        # Pull out the counter value for both the map and reduce stages, as well as the total.
+        metric_name = metric['name'].lower()
+        for metric_type in ['total', 'map', 'reduce']:
+          metric_value = metric[metric_type + 'CounterValue']
+          full_metric_path = '.'.join([metric_prefix, metric_name, metric_type])
+
+          formatted_metrics[full_metric_path] = {
+            'value': metric_value,
+            'context': {
+              'job_id': job['id'],
```
brianhw: nit: you could use job_id here, since you defined it on L#205.

scripts/collect-hadoop-metrics.py
```diff
@@ -0,0 +1,287 @@
...
+    print "[collect-hadoop-metrics] HistoryServer indicates no jobs have run.  Exiting."
+    sys.exit(0)
+
+  job_visits = {}
+  jobs = jobs.get('job', [])
+  for job in jobs:
+    # Grab details about the job so we get the full name.
+    job_id = job['id']
+    job_info = hs.job(job_id)
+    job_info = job_info.data['job']
```
brianhw: nit: can this be put on one line?  I find it confusing to have the same variable defined twice.

scripts/collect-hadoop-metrics.py
```diff
@@ -0,0 +1,287 @@
...
+
+          formatted_metrics[full_metric_path] = {
+            'value': metric_value,
+            'context': {
+              'job_id': job['id'],
+              'job_name': job_name,
+              'job_index': job_index
+            }
+          }
+
```
brianhw: nit: extra space

scripts/collect-hadoop-metrics.py
```diff
@@ -0,0 +1,287 @@
...
+    job_id = job['id']
+    job_info = hs.job(job_id)
+    job_info = job_info.data['job']
+
+    job_name = get_job_name_from_luigi_task_id(job_info['name'])
+
+    # Adjust job name to include a zero-based counter if we've seen it before.  This is kind of weird because
+    # it only makes "sense" in the context of a given application i.e. to differentiate Sqoop import tasks
+    # from one another, and it may change over time (size of list and thus the indexes) and so be not AS useful
+    # for over-time analysis, but it suffices because the indexing will be accurate within a single job flow.
```
brianhw: I guess this will also work for the hive jobs-within-a-task as well.  I'll be curious to see what they look like once we run them, and whether they are interpretable.

scripts/collect-hadoop-metrics.py
brianhw: any reason this is called twice?

mulby (approved):
Overall looks good. Ready to go once we remove the duplicated line assuming it works and all dependencies are pinned appropriately.

scripts/collect-hadoop-metrics.py
```diff
@@ -0,0 +1,287 @@
+import json
+import requests
+import boto3
+import statsd
+import sys
+from yarn_api_client import HistoryServer
+
+
+def query_metadata_service(path=''):
+  """
```
brianhw: Any chance that we can standardize on the same number of spaces for indents?

scripts/collect-hadoop-metrics.py
```diff
@@ -0,0 +1,287 @@
...
+
+  print "[collect-hadoop-metrics] Targeting HistoryServer at '{}'.".format(hs_address)
+
+  statsd_config = output_config.get('statsd', {})
+  statsd_host = statsd_config.get('host', 'localhost')
+  statsd_port = statsd_config.get('port', 8125)
+  statsd_prefix = statsd_config.get('prefix', 'edx.analytics.emr')
+
+  # Actually collect the metrics.
+  metrics = collect_metrics(local_address, metric_templates)
```
brianhw: What is local_address?  Should this be hs_address?

brianhw (commented):
First pass.

requirements/default.txt
tobz: `yarn-api-client` is entirely self-contained / only uses stdlib.

scripts/collect-hadoop-metrics.py
tobz: Yes, this is totally a missing import.  I'll also adjust the order.

scripts/collect-hadoop-metrics.py
tobz: This is a special host, yes.  It's the address for the [EC2 instance metadata service](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html).

I'll add a comment explaining as much.

scripts/collect-hadoop-metrics.py
tobz: Fixed.

scripts/collect-hadoop-metrics.py
tobz: Yep. :(

scripts/collect-hadoop-metrics.py
tobz: Done.

scripts/collect-hadoop-metrics.py
tobz: Sure.  Done.

scripts/collect-hadoop-metrics.py
tobz: Sure.  Done.

scripts/collect-hadoop-metrics.py
tobz: Right, I *think* it will work for that purpose.  Just have to test it.

scripts/collect-hadoop-metrics.py
tobz: Yep.  Done.

scripts/collect-hadoop-metrics.py
tobz: Fixed.

scripts/collect-hadoop-metrics.py
tobz: Nope, that's an error!  Fixed.

scripts/collect-hadoop-metrics.py
tobz: Yep, it should!  Fixed.

requirements/default.txt
tobz: `yarn-api-client` has no transitive dependencies.  I missed adding in a requirement for `statsd`, though.

`statsd` itself has no runtime requirements; it's requirements are all about testing (nose, mock, flake8).  Do we care about pinning those, so long as tests cleanly pass?  That feels good enough, to me, for possibly changed testing dependencies.

scripts/collect-hadoop-metrics.py
tobz: Yep, I missed that one. :)

scripts/collect-hadoop-metrics.py
tobz: Fixed.

brianhw (approved):
LGTM. 👍

PR #353, opened 2017-02-02
Revert "support multiple roles"

mulby:
Reverts edx/edx-analytics-pipeline#351


mulby:
FYI @tobz @brianhw @HassanJaveed84 - reverting until I can fix the outstanding issues

PR #357, opened 2017-02-16
Use a retry library to retry calls to the HistoryServer API.

tobz:
It seems like the HistoryServer can get a little choked up on bigger jobs, or by calling it too quickly after the jobs finish.

Now we have a little exponential backoff going on.  We'll try a total of six times to collect the metrics, with an exponential backoff of 2^n*1000ms (n is number of retries) with a maximum of 10000ms delay.

So... 0 seconds delay, then 1s, then 4s, then 10s, and 10s until the retries are up.

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -21,7 +21,6 @@ python-dateutil==2.6.0 	# BSD
 python-gnupg==0.3.9	# BSD
 pytz==2016.10		# ZPL
 requests==2.12.4        # Apache 2.0
-retrying==1.3.3
 six==1.10.0		# MIT
 statsd==3.2.1  # MIT
 stevedore==1.19.1 	# Apache 2.0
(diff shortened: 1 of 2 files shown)


brianhw (approved):
LGTM.  We had written our own retry utility code, but this seems easier.  👍

mulby (approved):
Heh - I didn't think this type of simple thing was worth an external dependency. Particularly given all of the difficulties we have with managing versions of `six` (which is a transitive dependency of retrying).

I guess I have a slight preference toward using our utility method given that we already have it.

That said - I don't feel strongly about this and am OK committing it as is.

tobz:
Hmm, I didn't originally realize we had a function for doing backoff.  I also didn't realize that this had a transitive dependency.

Let me take a quick look at what we have, and maybe other existing libraries w/o transitive dependencies.  It'd be nice to standardize things one way or another.

mulby:
here's our existing decorator: https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/util/retry.py#L21

mulby:
example usage: https://github.com/edx/edx-analytics-pipeline/blob/ccdd27967547b1bfb15fa9a9f30860236bc90134/edx/analytics/tasks/util/edx_api_client.py#L197

mulby (approved):
LGTM

PR #358, opened 2017-02-16
Remove traceback info.

HassanJaveed84:
Remove traceback info(this caused Jenkins marking the build as failed).

### DIFF
--- a/edx/analytics/tasks/export/data_obfuscation.py
+++ b/edx/analytics/tasks/export/data_obfuscation.py
@@ -191,9 +191,12 @@ class ObfuscateCoursewareStudentModule(ObfuscateSqlDumpTask):
         # but merely transform double backslashes.
         state_str = row[4].replace('\\\\', '\\')
         try:
-            state_dict = cjson.decode(state_str, all_unicode=True)
-            # Traverse the dictionary, looking for entries that need to be scrubbed.
-            updated_state_dict = self.obfuscator.obfuscate_structure(state_dict, u"state", user_info)
+            if state_str == 'NULL':
+                updated_state_dict = {}
+            else:
+                state_dict = cjson.decode(state_str, all_unicode=True)
+                # Traverse the dictionary, looking for entries that need to be scrubbed.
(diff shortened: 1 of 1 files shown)


tobz (approved):
LGTM. 👍

edx/analytics/tasks/export/data_obfuscation.py
```diff
@@ -196,7 +196,7 @@ def filter_row(self, row):
             # Traverse the dictionary, looking for entries that need to be scrubbed.
             updated_state_dict = self.obfuscator.obfuscate_structure(state_dict, u"state", user_info)
         except Exception:   # pylint:  disable=broad-except
-            log.exception(u"Unable to parse state as JSON for record %s: type = %s, state = %r",
+            log.warning(u"Unable to parse state as JSON for record %s: type = %s, state = %r",
```
brianhw: I was really imagining just avoiding the error in the first place, by not trying to parse a null entry.  But this is easy enough way to avoid the tracebacks.   The downside is that we lose information in those cases where we really didn't expect an error.

brianhw (commented):
Is it reasonable to just skip the decode when the string is 'null' (or whatever is the failing value)?

edx/analytics/tasks/export/data_obfuscation.py
HassanJaveed84: That makes sense. Would we still want a try/except block to handle other un-parseable state strings(other than 'NULL') ?

brianhw: Sorry for dropping the ball on this.   Do we know what string value we're looking for here?  Is it 'NULL'?  I would add an if statement around the decode/obfuscate inside the try, and set updated_state_dict to {} otherwise.   And I would continue to throw an exception, so when we rerun this again in future, we will know if we're getting an unexpected error.

HassanJaveed84:
@brianhw @mulby should we merge this in if this looks good ?

mulby (approved):

PR #361, opened 2017-03-03
Spike: Added task to load data from S3 into Redshift

HassanJaveed84:


edx/analytics/tasks/common/redshift_load.py
```diff
@@ -0,0 +1,173 @@
...
+                coldefs=coldefs,
+                table_attributes=self.table_attributes)
+
+            connection.cursor().execute(query)
+        else:
+            raise ValueError("create_table() found no columns for %r"
+                             % self.table)
+
+
+    def truncate_table(self, connection):
```
HassanJaveed84, line 103: TRUNCATE command commits the transaction, I think it probably makes sense to use DELETE here(like we do for Vertica). Also, Redshift recommends using VACUUM and ANALYZE after DELETE operations.

edx/analytics/tasks/common/redshift_load.py
```diff
@@ -0,0 +1,173 @@
...
+
+        path = self.s3_load_path()
+        output = self.output()
+        connection = output.connect()
+        cursor = connection.cursor()
+
+        self.init_copy(connection)
+        self.copy(cursor, path)
+
+        self.init_touch(connection)
```
brianhw, line 140: So the main difference of this run() from the original is that it adds this call, and removes the create-table retry.

edx/analytics/tasks/common/redshift_load.py
```diff
@@ -0,0 +1,173 @@
...
+        output.touch(connection)
+        connection.commit()
+
+        # commit and clean up
+        connection.close()
+
+    def copy(self, cursor, f):
+        cursor.execute("""
+         COPY {table} from '{source}'
+         IAM_ROLE 'arn:aws:iam::{id}:role/{role}'
```
brianhw, line 150: ... and this differs in the use of IAM_ROLE instead of CREDENTIALS mainly, but also in the rolling of column_separator into copy_options.

edx/analytics/tasks/common/redshift_load.py
```diff
@@ -0,0 +1,173 @@
...
+from luigi import configuration
+from luigi.contrib.redshift import S3CopyToTable
+from edx.analytics.tasks.insights.enrollments import EnrollmentSummaryRecord
+from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
+
+log = logging.getLogger(__name__)
+
+
+class RedshiftS3CopyToTable(OverwriteOutputMixin, S3CopyToTable):
+
```
brianhw, line 13: Seems that the only thing really inherited is output(), which returns a RedshiftTarget.  The base code seems to provide (or need to provide) little else.  Somehow I thought there would be more there there.

brianhw (commented):
This looks good to me.  I really had only one tiny  question, and the rest were annotations to myself that may or may not be useful to have as comments.  I was curious what was different from the original support, and I guess I'm still a little curious how different it is from Redshift support in current Luigi, were we to eventually upgrade to the current version.

mulby:
As a spike - this looks great to me!

PR #362, opened 2017-03-07
Adds counter calls to event-loading, enrollment, and video workflows

brianhw:
The main goal is to provide easier access to event volumes and error rates in certain map-reduce tasks than is provided by logging.   A secondary goal was to provide some degree of completeness, so that if problems are identified at a coarse level, some finer-grained counts can provide further information.  Thirdly, the naming of counters is hopefully more self-documenting by following some conventions. 

Counters for events being removed from processing are prefixed with "Discard".   Counters for events that are subsetted by a certain property are prefixed by "Subset".    Counters for events that contain some data problem but are not removed from processing are prefixed with "Quality". 

This does not provide a measure of "total events per day", because counters only apply to a single map-reduce job, and the processing of segment and tracking log events are done in separate tasks. 

* Number of segment events per day: SegmentEventRecordDataTask:Segment Event Exports:Output From Mapper
* Number of tracking log events per day:  TrackingEventRecordDataTask:Tracking Event Exports:Output From Mapper
* Malformed events per day:
** Each workflow provides central output of :Event:Discard Parseable Event, while other forms of malformedness are addressed on a per-workflow basis.
** No errors currently found from Segment
** Tracking-loading task returns counts for Tracking Event Exports:Discard Missing Event Data
** Enrollment workflow checks for several things, but not currently encountering any problems.
** Video workflow checks for several things, but currently encountering only missing username, course_id, and video time offsets.

Enrollment and video workflows also have state machines which this instruments.  For Video, three main categories exist:  the video events themselves, then "viewing starts" and "viewing ends".   Events discarded from each are counted with more general categories as well as specific labels.   Categories include "Discard Video Missing Something", "Discard Viewing Start", and "Discard Viewing End".  Counts for what was not discarded are also maintained, as well as some subsets by event type.   

For Enrollment, there are three tasks that perform map-reduce.  CourseEnrollmentEventsTask, CourseEnrollmentTask, and CourseEnrollmentSummaryTask.  The first provides counts related to checking and discarding enrollment events.  The second and third process nearly-identical state machines, but differ only in aggregation.  So issues with state transitions are counted up similarly for both.

### DIFF
--- a/edx/analytics/tasks/insights/enrollments.py
+++ b/edx/analytics/tasks/insights/enrollments.py
@@ -50,47 +50,64 @@ class CourseEnrollmentEventsTask(
     # We use warehouse_path to generate the output path, so we make this a non-param.
     output_root = None
 
+    counter_category_name = 'Enrollment Events'
+
     def mapper(self, line):
         value = self.get_event_and_date_string(line)
         if value is None:
             return
         event, date_string = value
+        self.incr_counter(self.counter_category_name, 'Inputs with Dates', 1)
 
         event_type = event.get('event_type')
         if event_type is None:
             log.error("encountered event with no event_type: %s", event)
+            self.incr_counter(self.counter_category_name, 'Discard Missing Event Type', 1)
             return
 
         if event_type not in (DEACTIVATED, ACTIVATED, MODE_CHANGED):
+            self.incr_counter(self.counter_category_name, 'Discard Non-Enrollment Event Type', 1)
             return
 
         timestamp = eventlog.get_event_time_string(event)
         if timestamp is None:
             log.error("encountered event with bad timestamp: %s", event)
+            self.incr_counter(self.counter_category_name, 'Discard Enroll Missing Timestamp', 1)
+            self.incr_counter(self.counter_category_name, 'Discard Enroll Missing Something', 1)
             return
 
         event_data = eventlog.get_event_data(event)
         if event_data is None:
+            self.incr_counter(self.counter_category_name, 'Discard Enroll Missing Event Data', 1)
+            self.incr_counter(self.counter_category_name, 'Discard Enroll Missing Something', 1)
             return
 
         course_id = opaque_key_util.normalize_course_id(event_data.get('course_id'))
         if course_id is None or not opaque_key_util.is_valid_course_id(course_id):
             log.error("encountered explicit enrollment event with invalid course_id: %s", event)
+            self.incr_counter(self.counter_category_name, 'Discard Enroll Missing course_id', 1)
+            self.incr_counter(self.counter_category_name, 'Discard Enroll Missing Something', 1)
             return
 
         user_id = event_data.get('user_id')
         if user_id is None:
             log.error("encountered explicit enrollment event with no user_id: %s", event)
+            self.incr_counter(self.counter_category_name, 'Discard Enroll Missing user_id', 1)
+            self.incr_counter(self.counter_category_name, 'Discard Enroll Missing Something', 1)
             return
 
         mode = event_data.get('mode')
         if mode is None:
             log.error("encountered explicit enrollment event with no mode: %s", event)
+            self.incr_counter(self.counter_category_name, 'Discard Enroll Missing mode', 1)
+            self.incr_counter(self.counter_category_name, 'Discard Enroll Missing Something', 1)
             return
 
+        self.incr_counter(self.counter_category_name, 'Output From Mapper', 1)
         yield date_string, (course_id, user_id, timestamp, event_type, mode)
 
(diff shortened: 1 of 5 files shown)


edx/analytics/tasks/insights/enrollments.py
```diff
@@ -50,47 +50,66 @@ class CourseEnrollmentEventsTask(
...
             return
 
         mode = event_data.get('mode')
         if mode is None:
             log.error("encountered explicit enrollment event with no mode: %s", event)
+            self.incr_counter(self.counter_category_name, 'Discard Enroll Missing mode', 1)
+            self.incr_counter(self.counter_category_name, 'Discard Enroll Missing Something', 1)
             return
 
+        self.incr_counter(self.counter_category_name, 'Output From Mapper', 1)
```
mulby, line 106: This is automatically tracked by the mapreduce framework IIRC.

edx/analytics/tasks/insights/enrollments.py
```diff
@@ -50,47 +50,66 @@ class CourseEnrollmentEventsTask(
...
         yield date_string, (course_id, user_id, timestamp, event_type, mode)
 
     def multi_output_reducer(self, _date_string, values, output_file):
+        self.incr_counter(self.counter_category_name, 'Output Dates with Events', 1)
```
mulby, line 110: Also tracked natively by hadoop IIRC. Something like reducer input records.

edx/analytics/tasks/insights/enrollments.py
```diff
@@ -50,47 +50,66 @@ class CourseEnrollmentEventsTask(
     # We use warehouse_path to generate the output path, so we make this a non-param.
     output_root = None
 
+    @property
+    def counter_category_name(self):
+        return 'Enrollment Events'
```
mulby: nit: Given that this value is static, perhaps we could just do:

```python
counter_category_name = 'Enrollment Events'
```

tobz:
This, in general, looks fine from a quality / correctness standpoint.

Some of these metric names are going to be veryyyyyyy long, though.  As an example, for "Enrollment Summary -> Discard User_Course With Missing Enrollment Event", the end result will be something like:

    edx.analytics.emr.Enrollment.job-flow-id.CourseEnrollment.0.hadoop.counters.enrollment_summary.discard_user_course_with_missing_enrollment_event

Not terribly unreasonable to work with, since you'll only set up the dashboard once, but, just wanted to point it out. :)

tobz (approved):

edx/analytics/tasks/insights/video.py
```diff
@@ -59,48 +59,68 @@ def init_local(self):
...
+        self.incr_counter(self.counter_category_name, 'Inputs with Dates', 1)
 
         event_type = event.get('event_type')
         if event_type is None:
             log.error("encountered event with no event_type: %s", event)
+            self.incr_counter(self.counter_category_name, 'Discard Missing Event Type', 1)
             return
 
         if event_type not in VIDEO_EVENT_TYPES:
+            self.incr_counter(self.counter_category_name, 'Discard Non-Video Event Type', 1)
```
mulby, line 83: Given how often this counter will be hit, it might actually cause a non-trivial performance degradation since all counters have to be written out to stderr of the process...

mulby (approved):
Overall looks good to me.

My primary concern is that the volume of metrics tracked will actually have a consequential effect on performance since each increment call results in a line being written out to stderr. In some of these cases I suspect that would result in millions of extra I/O operations to write all of this data out to stderr.

Definitely worth running with, though, to see the impact - might be negligible.

brianhw:
Timing results for video-edge look surprisingly flat.  See view/All/job/video-timeline-edge/buildTimeTrend on Jenkins, and also /view/Edge/job/enrollment-edge/buildTimeTrend.  I'm not as surprised with Enrollment, as its set up to run the events job incrementally, but the video result was surprising, given the heavy amount of instrumentation.

edx/analytics/tasks/insights/enrollments.py
brianhw: Yes, this is redundant, but I added it because I wanted everything needed to be in the same category.

edx/analytics/tasks/insights/enrollments.py
brianhw: Yes, this is the same as "Map-Reduce Framework":"Reduce input groups".  But I wanted it to be more self-documenting, so that someone working in Grafana wouldn't have to know as much about implementation details to know what a counter meant.

brianhw: By the way, the value for this counter is 3.  So it's not called many times.  (And presumably it matches the ovewrite_n_days parameter.)

PR #364, opened 2017-03-08
AN-8398

azafty468:
This is a WIP pull request encompassing the work in [AN-8398](https://openedx.atlassian.net/browse/AN-8398).  The new process has been documented in a temporary home here: https://openedx.atlassian.net/wiki/display/~azafft/Data+Analytics+Pipeline+-+Video+Processing until the Data Engineering space is built out.  

Currently the code is **not** passing acceptance tests (within the Data Tasks) but is in a good enough state for a review.

### DIFF
--- a/edx/analytics/tasks/insights/video.py
+++ b/edx/analytics/tasks/insights/video.py
@@ -1,25 +1,29 @@
-"""Tasks for aggregating statisics about video viewing."""
-
+"""Tasks for aggregating statistics about video viewing."""
 from collections import namedtuple
 import json
 import logging
 import math
 import re
+import textwrap
 import urllib
 
 import ciso8601
 import luigi
 from luigi import configuration
+from luigi.hive import HiveQueryTask
 
 from edx.analytics.tasks.common.mapreduce import MapReduceJobTask, MapReduceJobTaskMixin
+from edx.analytics.tasks.common.mysql_load import MysqlInsertTask
 from edx.analytics.tasks.common.pathutil import EventLogSelectionMixin, EventLogSelectionDownstreamMixin
 from edx.analytics.tasks.util.decorators import workflow_entry_point
 from edx.analytics.tasks.util import eventlog
-from edx.analytics.tasks.util.hive import WarehouseMixin, HivePartition, HiveTableTask, HiveQueryToMysqlTask
-from edx.analytics.tasks.util.url import get_target_from_url, url_path_join
+from edx.analytics.tasks.util.hive import (WarehouseMixin, HivePartition, HiveTableTask, BareHiveTableTask,
+                                           HivePartitionTask, hive_database_name)
+from edx.analytics.tasks.util.url import url_path_join, get_target_from_url
+from edx.analytics.tasks.util.record import Record, StringField, IntegerField
+
 
 log = logging.getLogger(__name__)
-
 
 VIDEO_PLAYED = 'play_video'
 VIDEO_PAUSED = 'pause_video'
@@ -41,6 +45,89 @@ VIDEO_VIEWING_MINIMUM_LENGTH = 0.25  # seconds
 
 VideoViewing = namedtuple('VideoViewing', [   # pylint: disable=invalid-name
     'start_timestamp', 'course_id', 'encoded_module_id', 'start_offset', 'video_duration'])
+
+
+class VideoTimelineRecord(Record):
+    """
+    Video Segment Information used to populate the video_timeline table
+    """
+
+    pipeline_video_id = StringField(length=255,
+                                    nullable=False,
+                                    description='A concatenation of the course_id and the HTML encoded ID of the video.'
+                                                ' Intended to uniquely identify an instance of a video in a particular '
+                                                'course. Note that ideally we would use an XBlock usage_id here, but '
+                                                'it isn\'t present on the legacy events.')
+    segment = IntegerField(description='An integer representing the rank of the segment within the video. 0 is the '
+                                       'first segment, 1 is the second etc. Note that this does not specify the length '
+                                       'of the segment.')
+    num_users = IntegerField(description='The number of unique users who watched any part of this segment of the '
+                                         'video.')
+    num_views = IntegerField(description='The total number of times any part of the segment was viewed, regardless of '
+                                         'who was watching it.')
+
+
+class VideoSegmentSummaryRecord(Record):
+    """
+    Video Segment Summary Information used to populate the video table
+    """
+
+    pipeline_video_id = StringField(length=255,
+                                    nullable=False,
+                                    description='A concatenation of the course_id and the HTML encoded ID of the video.'
+                                                ' Intended to uniquely identify an instance of a video in a particular '
+                                                'course. Note that ideally we would use an XBlock usage_id here, but '
+                                                'it isn\'t present on the legacy events.')
+    course_id = StringField(length=255,
+                            nullable=False,
+                            description='Course the video was displayed in. This is an opaque key serialized to '
+                                        'a string.')
+    encoded_module_id = StringField(length=255,
+                                    nullable=False,
+                                    description='This is the HTML encoded module ID for the video. Ideally this would '
+                                                'be an XBlock usage_id, but that data is not present on legacy events.')
+    duration = IntegerField(description='The video length in seconds. This can be inferred for some videos. We don\'t '
+                                        'have reliable metadata for the length of videos in the source data.')
+    segment_length = IntegerField(description='The length of each segment, in seconds.')
+    users_at_start = IntegerField(description='The number of users who watched the first segment of the video.')
+    users_at_end = IntegerField(description='The number of users who watched the end of the video. Note that this is '
+                                            'not the number of users who watched the last segment of the video.')
+    total_viewed_seconds = IntegerField(description='The total number of seconds viewed by all users across all '
+                                                    'segments.')
+
+
+class VideoSegmentDetailRecord(Record):
+    """
+    Video Segment Usage Detail
+    """
+
+    pipeline_video_id = StringField(length=255,
+                                    nullable=False,
+                                    description='A concatenation of the course_id and the HTML encoded ID of the video.'
+                                                ' Intended to uniquely identify an instance of a video in a particular '
+                                                'course. Note that ideally we would use an XBlock usage_id here, but '
+                                                'it isn\'t present on the legacy events.')
+    course_id = StringField(length=255,
+                            nullable=False,
+                            description='Course the video was displayed in. This is an opaque key serialized to '
+                                        'a string.')
+    encoded_module_id = StringField(length=255,
+                                    nullable=False,
+                                    description='This is the HTML encoded module ID for the video. Ideally this would '
+                                                'be an XBlock usage_id, but that data is not present on legacy events.')
+    duration = IntegerField(description='The video length in seconds. This can be inferred for some videos. We don\'t '
+                                        'have reliable metadata for the length of videos in the source data.')
+    segment_length = IntegerField(description='The length of each segment, in seconds.')
+    users_at_start = IntegerField(description='The number of users who watched the first segment of the video.')
+    users_at_end = IntegerField(description='The number of users who watched the end of the video. Note that this is '
+                                            'not the number of users who watched the last segment of the video.')
+    segment = IntegerField(description='An integer representing the rank of the segment within the video. 0 is the '
+                                       'first segment, 1 is the second etc. Note that this does not specify the length '
+                                       'of the segment.')
+    num_users = IntegerField(description='The number of unique users who watched any part of this segment of the '
+                                         'video.')
+    num_views = IntegerField(description='The total number of times any part of the segment was viewed, regardless of '
+                                         'who was watching it.')
 
 
 class UserVideoViewingTask(EventLogSelectionMixin, MapReduceJobTask):
@@ -420,18 +507,18 @@ class VideoUsageTask(VideoTableDownstreamMixin, MapReduceJobTask):
         users_at_end = len(usage_map.get(self.complete_end_segment(video_duration), {}).get('users', []))
         for segment in sorted(usage_map.keys()):
             stats = usage_map[segment]
-            yield (
-                pipeline_video_id,
-                course_id,
-                encoded_module_id,
-                int(video_duration),
-                VIDEO_VIEWING_SECONDS_PER_SEGMENT,
-                users_at_start,
-                users_at_end,
-                segment,
-                len(stats.get('users', [])),
-                stats.get('views', 0),
-            )
+            yield VideoSegmentDetailRecord(
+                pipeline_video_id=pipeline_video_id,
+                course_id=course_id,
+                encoded_module_id=encoded_module_id,
+                duration=int(video_duration),
+                segment_length=VIDEO_VIEWING_SECONDS_PER_SEGMENT,
+                users_at_start=users_at_start,
+                users_at_end=users_at_end,
+                segment=segment,
+                num_users=len(stats.get('users', [])),
+                num_views=stats.get('views', 0)
+            ).to_string_tuple()
             if segment == final_segment:
                 break
 
@@ -477,25 +564,14 @@ class VideoUsageTableTask(VideoTableDownstreamMixin, HiveTableTask):
         return 'video_usage'
 
     @property
-    def columns(self):
-        return [
-            ('pipeline_video_id', 'STRING'),
-            ('course_id', 'STRING'),
-            ('encoded_module_id', 'STRING'),
-            ('duration', 'INT'),
-            ('segment_length', 'INT'),
-            ('users_at_start', 'INT'),
-            ('users_at_end', 'INT'),
-            ('segment', 'INT'),
-            ('num_users', 'INT'),
-            ('num_views', 'INT'),
-        ]
+    def columns(self):  # pragma: no cover
+        return VideoSegmentDetailRecord.get_hive_schema()
 
     @property
-    def partition(self):
+    def partition(self):  # pragma: no cover
         return HivePartition('dt', self.interval.date_b.isoformat())  # pylint: disable=no-member
 
-    def requires(self):
+    def requires(self):  # pragma: no cover
         return VideoUsageTask(
             mapreduce_engine=self.mapreduce_engine,
             n_reduce_tasks=self.n_reduce_tasks,
@@ -506,19 +582,48 @@ class VideoUsageTableTask(VideoTableDownstreamMixin, HiveTableTask):
             output_root=self.partition_location
         )
 
-    def output(self):
+    def output(self):  # pragma: no cover
         return self.requires().output()
 
 
-class InsertToMysqlVideoTimelineTask(VideoTableDownstreamMixin, HiveQueryToMysqlTask):
-    """Insert information about video segments from a Hive table into MySQL."""
+class VideoTimelineTableTask(BareHiveTableTask):
+    """Creates the Hive storage table used to hold video_timeline data."""
 
     @property
-    def table(self):
-        return "video_timeline"
+    def partition_by(self):  # pragma: no cover
+        return 'dt'
 
     @property
-    def query(self):
+    def table(self):  # pragma: no cover
+        return 'video_timeline'
+
+    @property
+    def columns(self):  # pragma: no cover
+        return VideoTimelineRecord.get_hive_schema()
+
+
+class VideoTimelinePartitionTask(VideoTableDownstreamMixin, HivePartitionTask):
+    """Creates the Hive storage partition used to hold video_timeline data."""
+
+    @property
+    def hive_table_task(self):  # pragma: no cover
+        return VideoTimelineTableTask(
+            warehouse_path=self.warehouse_path,
+        )
+
+    @property
+    def partition_value(self):  # pragma: no cover
+        """Use a dynamic partition value based on the date parameter."""
+        return self.interval.date_b.isoformat()  # pylint: disable=no-member
+
+
+class VideoTimelineDataTask(VideoTableDownstreamMixin, HiveQueryTask):
+    """Execute the query on video_usage and persist the results into video_timeline."""
+
+    @property
+    def insert_query(self):  # pragma: no cover
+        """The insert query that specifies the fields from the source table."""
+
         return """
             SELECT
                 pipeline_video_id,
@@ -528,18 +633,24 @@ class InsertToMysqlVideoTimelineTask(VideoTableDownstreamMixin, HiveQueryToMysql
             FROM video_usage
         """
 
-    @property
-    def columns(self):
-        return [
-            ('pipeline_video_id', 'VARCHAR(255)'),
-            ('segment', 'INTEGER'),
-            ('num_users', 'INTEGER'),
-            ('num_views', 'INTEGER'),
-        ]
+    def query(self):  # pragma: no cover
+        full_insert_query = """
+                    USE {database_name};
+                    INSERT INTO TABLE {table}
+                    PARTITION ({partition.query_spec})
+                    {insert_query};
+                    """.format(database_name=hive_database_name(),
+                               table=self.partition_task.hive_table_task.table,
+                               partition=self.partition,
+                               insert_query=self.insert_query.strip(),  # pylint: disable=no-member
+                              )
+
+        return textwrap.dedent(full_insert_query)
 
     @property
-    def required_table_tasks(self):
-        return VideoUsageTableTask(
+    def partition_task(self):  # pragma: no cover
+        """The task that creates the partition used by this job."""
+        return VideoTimelinePartitionTask(
             mapreduce_engine=self.mapreduce_engine,
             n_reduce_tasks=self.n_reduce_tasks,
             source=self.source,
@@ -549,25 +660,102 @@ class InsertToMysqlVideoTimelineTask(VideoTableDownstreamMixin, HiveQueryToMysql
         )
 
     @property
-    def indexes(self):
+    def partition(self):  # pragma: no cover
+        """A shorthand for the partition information on the upstream partition task."""
+        return self.partition_task.partition  # pylint: disable=no-member
+
+    def requires(self):  # pragma: no cover
+        for requirement in super(VideoTimelineDataTask, self).requires():
+            yield requirement
+        yield self.partition_task
+
+        yield VideoUsageTableTask(
+            mapreduce_engine=self.mapreduce_engine,
+            n_reduce_tasks=self.n_reduce_tasks,
+            source=self.source,
+            interval=self.interval,
+            pattern=self.pattern,
+            warehouse_path=self.warehouse_path,
+        )
+
+    def output(self):  # pragma: no cover
+        output_root = url_path_join(self.warehouse_path,
+                                    self.partition_task.hive_table_task.table,
+                                    self.partition.path_spec + '/')
+        return get_target_from_url(output_root, marker=True)
+
+    def on_success(self):  # pragma: no cover
+        """Overload the success method to touch the _SUCCESS file.  Any class that uses a separate Marker file from the
+        data file will need to override the base on_success() call to create this marker."""
+        self.output().touch_marker()
+
+
+class InsertToMysqlVideoTimelineTask(VideoTableDownstreamMixin, MysqlInsertTask):
+    """Insert information about video timelines from a Hive table into MySQL."""
+
+    @property
+    def table(self):  # pragma: no cover
+        return 'video_timeline'
+
+    @property
+    def insert_source_task(self):  # pragma: no cover
+        return VideoTimelineDataTask(
+            mapreduce_engine=self.mapreduce_engine,
+            n_reduce_tasks=self.n_reduce_tasks,
+            source=self.source,
+            interval=self.interval,
+            pattern=self.pattern,
+            warehouse_path=self.warehouse_path,
+        )
+
+    @property
+    def columns(self):  # pragma: no cover
+        return VideoTimelineRecord.get_sql_schema()
+
+    @property
+    def indexes(self):  # pragma: no cover
         return [
             ('pipeline_video_id',),
         ]
 
-    @property
-    def partition(self):
-        return HivePartition('dt', self.interval.date_b.isoformat())  # pylint: disable=no-member
 
+class VideoTableTask(BareHiveTableTask):
+    """Creates the Hive storage table used to hold video data."""
 
-class InsertToMysqlVideoTask(VideoTableDownstreamMixin, HiveQueryToMysqlTask):
-    """Insert non-segment information about videos from a Hive table into MySQL."""
-
-    @property
-    def table(self):
-        return "video"
+    @property  # pragma: no cover
+    def partition_by(self):
+        return 'dt'
 
     @property
-    def query(self):
+    def table(self):  # pragma: no cover
+        return 'video'
+
+    @property
+    def columns(self):  # pragma: no cover
+        return VideoSegmentSummaryRecord.get_hive_schema()
+
+
+class VideoPartitionTask(VideoTableDownstreamMixin, HivePartitionTask):
+    """Creates the Hive storage partition used to hold video data."""
+
+    @property
+    def hive_table_task(self):  # pragma: no cover
+        return VideoTableTask(
+            warehouse_path=self.warehouse_path,
+        )
+
+    @property
+    def partition_value(self):  # pragma: no cover
+        """ Use a dynamic partition value based on the date parameter. """
+        return self.interval.date_b.isoformat()  # pylint: disable=no-member
+
+
+class VideoDataTask(VideoTableDownstreamMixin, HiveQueryTask):
+    """Execute the query on video_usage and persist the results into video."""
+
+    @property
+    def insert_query(self):  # pragma: no cover
+        """The fields used in the source table."""
         return """
             SELECT
                 pipeline_video_id,
@@ -589,22 +777,72 @@ class InsertToMysqlVideoTask(VideoTableDownstreamMixin, HiveQueryToMysqlTask):
                 users_at_end
         """
 
-    @property
-    def columns(self):
-        return [
-            ('pipeline_video_id', 'VARCHAR(255)'),
-            ('course_id', 'VARCHAR(255)'),
-            ('encoded_module_id', 'VARCHAR(255)'),
-            ('duration', 'INTEGER'),
-            ('segment_length', 'INTEGER'),
-            ('users_at_start', 'INTEGER'),
-            ('users_at_end', 'INTEGER'),
-            ('total_viewed_seconds', 'INTEGER'),
-        ]
+    def query(self):  # pragma: no cover
+        full_insert_query = """
+                    USE {database_name};
+                    INSERT INTO TABLE {table}
+                    PARTITION ({partition.query_spec})
+                    {insert_query};
+                """.format(database_name=hive_database_name(),
+                           table=self.partition_task.hive_table_task.table,
+                           partition=self.partition,
+                           insert_query=self.insert_query.strip(),  # pylint: disable=no-member
+                          )
+        return textwrap.dedent(full_insert_query)
 
     @property
-    def required_table_tasks(self):
-        return VideoUsageTableTask(
+    def partition(self):  # pragma: no cover
+        """A shorthand for the partition object on the upstream partition task."""
+        return self.partition_task.partition  # pylint: disable=no-member
+
+    @property
+    def partition_task(self):  # pragma: no cover
+        """Returns the task representing the work to create the partition."""
+        return VideoPartitionTask(
+            mapreduce_engine=self.mapreduce_engine,
+            n_reduce_tasks=self.n_reduce_tasks,
+            source=self.source,
+            interval=self.interval,
+            pattern=self.pattern,
+            warehouse_path=self.warehouse_path,
+        )
+
+    def requires(self):  # pragma: no cover
+        for requirement in super(VideoDataTask, self).requires():
+            yield requirement
+        yield self.partition_task
+
+        yield VideoUsageTableTask(
+            mapreduce_engine=self.mapreduce_engine,
+            n_reduce_tasks=self.n_reduce_tasks,
+            source=self.source,
+            interval=self.interval,
+            pattern=self.pattern,
+            warehouse_path=self.warehouse_path,
+        )
+
+    def output(self):  # pragma: no cover
+        output_root = url_path_join(self.warehouse_path,
+                                    self.partition_task.hive_table_task.table,
+                                    self.partition.path_spec + '/')
+        return get_target_from_url(output_root, marker=True)
+
+    def on_success(self):  # pragma: no cover
+        """Overload the success method to touch the _SUCCESS file.  Any class that uses a separate Marker file from the
+        data file will need to override the base on_success() call to create this marker."""
+        self.output().touch_marker()
+
+
+class InsertToMysqlVideoTask(VideoTableDownstreamMixin, MysqlInsertTask):
+    """Insert summary information into the video table in MySQL."""
+
+    @property
+    def table(self):  # pragma: no cover
+        return 'video'
+
+    @property
+    def insert_source_task(self):  # pragma: no cover
+        return VideoDataTask(
             mapreduce_engine=self.mapreduce_engine,
             n_reduce_tasks=self.n_reduce_tasks,
             source=self.source,
@@ -614,14 +852,14 @@ class InsertToMysqlVideoTask(VideoTableDownstreamMixin, HiveQueryToMysqlTask):
         )
 
     @property
-    def indexes(self):
+    def columns(self):  # pragma: no cover
+        return VideoSegmentSummaryRecord.get_sql_schema()
+
+    @property
+    def indexes(self):  # pragma: no cover
         return [
             ('course_id', 'encoded_module_id'),
         ]
-
-    @property
-    def partition(self):
-        return HivePartition('dt', self.interval.date_b.isoformat())  # pylint: disable=no-member
 
 
 @workflow_entry_point
--- a/edx/analytics/tasks/util/url.py
+++ b/edx/analytics/tasks/util/url.py
@@ -18,8 +18,40 @@ import luigi.configuration
 import luigi.format
 import luigi.hdfs
 import luigi.s3
+from luigi.hdfs import HdfsTarget
+from luigi.s3 import S3Target
 
 from edx.analytics.tasks.util.s3_util import ScalableS3Client, S3HdfsTarget
+
+
+class MarkerMixin(object):
+    """This mixin handles Targets that cannot accurately be measured by the existence of data files, and instead need
+    another positive marker to indicate Task success."""
+
+    def exists(self):  # pragma: no cover
+        """Completion of this target is based solely on the existence of the marker file."""
+        return self.fs.exists(self.path + "/_SUCCESS")
+
+    def touch_marker(self):  # pragma: no cover
+        """Generate the marker file using file system native to the parent Target."""
+        marker = self.__class__(path=self.path + "/_SUCCESS")
+        marker.open("w").close()
+
+
+class S3MarkerTarget(MarkerMixin, S3Target):
+    """An S3 Target that uses a marker file to indicate success."""
+
+
+class HdfsMarkerTarget(MarkerMixin, HdfsTarget):
+    """An HDFS Target that uses a marker file to indicate success."""
+
+
+class LocalMarkerTarget(MarkerMixin, luigi.LocalTarget):
+    """A Local Target that uses a marker file to indicate success."""
+
+
+class S3HdfsMarkerTarget(MarkerMixin, S3HdfsTarget):
+    """An S3 HDFS Target that uses a marker file to indicate success."""
 
 
 class ExternalURL(luigi.ExternalTask):
@@ -58,11 +90,25 @@ URL_SCHEME_TO_TARGET_CLASS = {
     's3+https': luigi.s3.S3Target,
 }
 
+DEFAULT_MARKER_TARGET_CLASS = LocalMarkerTarget
+URL_SCHEME_TO_MARKER_TARGET_CLASS = {
+    'hdfs': HdfsMarkerTarget,
+    's3': S3HdfsMarkerTarget,
+    's3n': S3HdfsMarkerTarget,
+    'file': LocalMarkerTarget,
+    's3+https': S3MarkerTarget,
+}
 
-def get_target_class_from_url(url):
+
+def get_target_class_from_url(url, marker=False):
     """Returns a luigi target class based on the url scheme"""
     parsed_url = urlparse.urlparse(url)
-    target_class = URL_SCHEME_TO_TARGET_CLASS.get(parsed_url.scheme, DEFAULT_TARGET_CLASS)
+
+    if marker:
+        target_class = URL_SCHEME_TO_MARKER_TARGET_CLASS.get(parsed_url.scheme, DEFAULT_MARKER_TARGET_CLASS)
+    else:
+        target_class = URL_SCHEME_TO_TARGET_CLASS.get(parsed_url.scheme, DEFAULT_TARGET_CLASS)
+
     kwargs = {}
     if issubclass(target_class, luigi.hdfs.HdfsTarget) and url.endswith('/'):
         kwargs['format'] = luigi.hdfs.PlainDir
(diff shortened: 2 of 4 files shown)


edx/analytics/tasks/insights/video.py
```diff
@@ -171,7 +219,6 @@ def _check_time_offset(self, time_value, line):
     def reducer(self, key, events):
         """
         Constructs "viewing" records for each user in a course video module.
-
```
brianhw: nit: I actually think it's the normal style for docstrings to have a summary line as the first line, and additional content separated from it by a blank line.

edx/analytics/tasks/insights/video.py
```diff
@@ -1,22 +1,27 @@
...
 from luigi import configuration
+from luigi.hive import HiveQueryTask
 
 from edx.analytics.tasks.common.mapreduce import MapReduceJobTask, MapReduceJobTaskMixin
+from edx.analytics.tasks.common.mysql_load import MysqlInsertTask
 from edx.analytics.tasks.common.pathutil import EventLogSelectionMixin, EventLogSelectionDownstreamMixin
 from edx.analytics.tasks.util.decorators import workflow_entry_point
 from edx.analytics.tasks.util import eventlog
-from edx.analytics.tasks.util.hive import WarehouseMixin, HivePartition, HiveTableTask, HiveQueryToMysqlTask
+from edx.analytics.tasks.util.hive import WarehouseMixin, HivePartition, HiveTableTask, BareHiveTableTask, \
```
brianhw: nit: our code style guidelines prefer parentheses over backslashes for multi-line statements.

edx/analytics/tasks/insights/video.py
```diff
@@ -210,7 +257,7 @@ def start_viewing():
                 else:
                     start_offset = current_time
 
-                return VideoViewing(
+                return VideoViewing( #TODO Should I promote this into a first class object?
```
brianhw: I think it is a first class object already.  Do you mean should it be defined as a class?  Is there anything else you would add to it?  Otherwise I think it's okay for it to be a namedtuple-defined object.

edx/analytics/tasks/insights/video.py
```diff
@@ -239,7 +286,7 @@ def end_viewing(end_time):
                               viewing, event, key)
                     return None
 
-                return (
+                return ( #TODO Is this a new object?
```
brianhw: It's returning a tuple here.  That's an okay return value.  If we were doing much more with it (i.e. it matched some kind of record), then we could create that instead.  But if not, there seems little value in formalizing this further.

edx/analytics/tasks/insights/video.py
```diff
@@ -284,10 +331,10 @@ def end_viewing(end_time):
     def output(self):
         return get_target_from_url(self.output_root)
 
+
```
brianhw: nit: PEP8 should complain about this:  should be one blank line between methods in a class, and two blank lines between stuff not in a class (like another class).

edx/analytics/tasks/insights/video.py
```diff
@@ -284,10 +331,10 @@ def end_viewing(end_time):
...
     def get_video_duration(self, youtube_id):
         """
         For youtube videos, queries Google API for video duration information.
-
```
brianhw: nit: again, we probably want these, here and below.

edx/analytics/tasks/insights/video.py
```diff
@@ -421,16 +464,18 @@ def reducer(self, key, viewings):
...
+                    course_id=course_id,
+                    encoded_module_id=encoded_module_id,
+                    duration=int(video_duration),
+                    segment_length=VIDEO_VIEWING_SECONDS_PER_SEGMENT,
+                    users_at_start=users_at_start,
+                    users_at_end=users_at_end,
+                    segment=segment,
+                    num_users=len(stats.get('users', [])),
+                    num_views=stats.get('views', 0)
+                ).to_string_tuple()   # TODO This is a signature change!  Instead of being mixes of types now they are all strings!
```
brianhw: Interesting!  Since we don't have history for this task, the fact that this might not work with existing data will likely not matter.  It should just recalculate from scratch.

edx/analytics/tasks/insights/video.py
```diff
@@ -510,15 +544,54 @@ def output(self):
         return self.requires().output()
 
 
-class InsertToMysqlVideoTimelineTask(VideoTableDownstreamMixin, HiveQueryToMysqlTask):
-    """Insert information about video segments from a Hive table into MySQL."""
+class VideoTimelineCreateTableTask(BareHiveTableTask):
```
brianhw: nit:  add a docstring, when you get the chance....

edx/analytics/tasks/insights/video.py
```diff
@@ -510,15 +544,54 @@ def output(self):
...
 
     @property
     def table(self):
-        return "video_timeline"
+        return 'video_timeline'
 
     @property
-    def query(self):
+    def columns(self):
+        return VideoSegmentSummaryRecord.get_hive_schema()
```
brianhw: Just curious if the name of this record, the class name and the table name should all be more in agreement?

edx/analytics/tasks/insights/video.py
```diff
@@ -510,15 +544,54 @@ def output(self):
...
+        return self.hive_table_task.partition_by
+
+    @property
+    def partition_value(self):
+        """Use a dynamic partition value based on the date parameter."""
+        return self.interval.date_b.isoformat()  # pylint: disable=no-member
+
+    @property
+    def data_task(self):
+        """A bit of an abuse here as data should be for populating this table instead of an upstream table"""
```
brianhw: Is there a reason it is here then?  It is really a requirement of the Hive query, since it defines the table being queried, and should probably be added as a requirement there.

edx/analytics/tasks/insights/video.py
```diff
@@ -528,18 +601,55 @@ def query(self):
...
-            ('pipeline_video_id', 'VARCHAR(255)'),
-            ('segment', 'INTEGER'),
-            ('num_users', 'INTEGER'),
-            ('num_views', 'INTEGER'),
-        ]
+    def partition(self):
+        return self.requires_local().partition  # pylint: disable=no-member
+        # return HivePartition('dt', self.interval.date_b.isoformat())  # pylint: disable=no-member
+
+# Is this appropriate?  Requires local feels like it's too low level and instead I should pass these as parameters
```
brianhw: Requirements get flattened, so they can be defined as dicts.  We will often do that if there are particular requirements that are needed elsewhere.  So I would create a dict here, and move the VideoUsageTableTask here as well.  Then the query above can get the table name from self.requires()['destination'].hive_table_task.table, and the partition from self.requires()['destination'].partition.   Actually, I would use the latter in the partition(self) definition, then use that in the full_insert_query.

edx/analytics/tasks/insights/video.py
```diff
@@ -528,18 +601,55 @@ def query(self):
...
+            n_reduce_tasks=self.n_reduce_tasks,
+            source=self.source,
+            interval=self.interval,
+            pattern=self.pattern,
+            warehouse_path=self.warehouse_path,
+        )
+
+# TODO correct this to be the proper location.  Also how do I touch this file?
+    # def output(self):
+    #     marker_url = url_path_join(self.marker, str(hash(self)))
```
brianhw: I'm assuming the output() would be a url pointing to the partition.   It's then a different matter to figure out how to implement complete() on the task.  That might need a marker....

edx/analytics/tasks/insights/video.py
```diff
@@ -549,25 +659,64 @@ def required_table_tasks(self):
         )
 
     @property
+    def columns(self):
+        return VideoSegmentSummaryRecord.get_sql_schema()
```
brianhw: Maybe this should be renamed VideoTimelineRecord?

edx/analytics/tasks/insights/video.py
```diff
@@ -549,25 +659,64 @@ def required_table_tasks(self):
...
+class VideoCreateTableTask(BareHiveTableTask):
     @property
-    def partition(self):
-        return HivePartition('dt', self.interval.date_b.isoformat())  # pylint: disable=no-member
+    def partition_by(self):
+        return 'dt'
 
+    @property
+    def table(self):
+        return 'video'
```
brianhw: Actually, since this is not an incremental workflow, we could probably go ahead and change the name of this, if we wanted to.  And all the classes too (except perhaps the entry point).

edx/analytics/tasks/insights/video.py
```diff
@@ -613,16 +768,35 @@ def required_table_tasks(self):
             warehouse_path=self.warehouse_path,
         )
 
+
+class InsertToMysqlVideoTask(VideoTableDownstreamMixin, MysqlInsertTask):
+    """Insert non-segment information about videos from a Hive table into MySQL."""
+
+    @property
+    def table(self):
+        return "video"
```
brianhw: Note that the names of tables in Mysql should not change, since the analytics data api is looking for them.

brianhw (commented):
Looks good!  A bunch of nits related to naming and doc and stuff, and the only real issue is the definition of output() in one task, and moving another task from data_task() to requires().

azafty468:
Thanks @brianhw!  I'll take a look at these changes and incorporate them

edx/analytics/tasks/insights/video.py
azafty468: Yup, so the class name had been taken from the comments in of each of the jobs.  You are totally right about alignment here however, so I've changed each of these data objects to be VideoTimelineRecord and VideoSummarySegmentRecord

edx/analytics/tasks/insights/video.py
azafty468: Right so data_task is driven by the HivePartitionTask which runs primarily after an MR file has been created.  My creation pattern is slightly different (partition created before data insert).  So the intent was that this data task represents data to be loaded into this partition, and I'm using it as a chain for the previous upstream task.  I'm probably going to swap this out to the requires_local as it is a bit more in line with how the underlying BaseHadoopJobTask is run.  But I'm also open to suggestions incase I'm misinterpreting this.

edx/analytics/tasks/insights/video.py
azafty468: Alright, sounds good!  I had drilled down to BaseHadoopJobTask and noted a few comments differentiating requires(), requires_local(), and requires_hadoop() and it looked like there were differences in what variables would be available when.  But if they are just mashed down then I can put them in a dict and make the whole process more readable.   I'll look into doing likewise with your previous comment to see if it makes sense.

edx/analytics/tasks/insights/video.py
azafty468: Hahah standards evolution!  Probably a handful of other ones out there.  Does someone have a link to the code style guidelines?  Some are triggered on a build but I do not believe a review of this issue was one of them.  

image

edx/analytics/tasks/insights/video.py
```diff
@@ -43,6 +48,49 @@
     'start_timestamp', 'course_id', 'encoded_module_id', 'start_offset', 'video_duration'])
 
 
+class VideoSegmentSummaryRecord(Record):
+    """
+    Video Segment Information
+    """
+
+    pipeline_video_id = StringField(length=255, nullable=False, description='Pipeline Video Id.')
```
mulby: nit: we should probably provide a more detailed description of what a "pipeline video id" is here. IIRC it's actually a compound identifier formed by concatenating two other identifiers, which is probably worth calling out here.

Now that I write this, @brianhw or I should probably just crank through these column descriptions and provide you with more detailed information about what they represent.

edx/analytics/tasks/insights/video.py
mulby: I suspect that these will actually result in the same output since luigi is effectively calling `str()` on each element of the tuple anyway: https://github.com/edx/luigi/blob/edx/release/luigi/hadoop.py#L697

Which is roughly equivalent to `Record` which does something like `unicode(foo).encode('utf8')`.

edx/analytics/tasks/insights/video.py
```diff
@@ -510,15 +544,54 @@ def output(self):
...
-        return "video_timeline"
+        return 'video_timeline'
 
     @property
-    def query(self):
+    def columns(self):
+        return VideoSegmentSummaryRecord.get_hive_schema()
+
+
+class VideoTimelineCreatePartitionTask(VideoTableDownstreamMixin, HivePartitionTask):
```
mulby: Nit: we usually call this just "FooPartitionTask" instead of "FooCreatePartitionTask" - that is we describe the end state instead of the transformation that results in that end state.

This hasn't always been the case... more of a "modern" convention.

edx/analytics/tasks/insights/video.py
mulby: nit: I tend to prefer the style of "FooTableTask" as opposed to "FooCreateTableTask" - one describes the desired end state, the second describes one of the possible actions that might be taken.

edx/analytics/tasks/insights/video.py
```diff
@@ -43,6 +47,49 @@
     'start_timestamp', 'course_id', 'encoded_module_id', 'start_offset', 'video_duration'])
 
 
+class VideoTimelineRecord(Record):
+    """
+    Video Segment Information used to populate the video_timeline table
+    """
+
+    pipeline_video_id = StringField(length=255, nullable=False, description='Pipeline Video Id.')
```
mulby: @brianhw and I should probably go through all of these columns and provide more useful descriptions

edx/analytics/tasks/insights/video.py
```diff
@@ -510,15 +543,62 @@ def output(self):
...
+    @property
+    def partition_by(self):
+        return self.hive_table_task.partition_by
+
+    @property
+    def partition_value(self):
+        """Use a dynamic partition value based on the date parameter."""
+        return self.interval.date_b.isoformat()  # pylint: disable=no-member
+
+    # def requires(self):
```
mulby: You could do something like this:

```python
for requirement in super(VideoTimelineCreatePartitionTask, self).requires():
    yield requirement

yield VideoUsageTableTask(
            source=self.source,
            interval=self.interval,
            pattern=self.pattern,
        )
```

You could also do something like this:
```python
requirements = list(super(VideoTimelineCreatePartitionTask, self).requires())
requirements.append(VideoUsageTableTask())
```

Or even use luigi's `flatten` helper:
```python
requirements = flatten(super(VideoTimelineCreatePartitionTask, self).requires())
requirements.append(VideoUsageTableTask())
```

This is a python generator... which is actually cleaned up quite a bit in python 3. For now you have to do something like the above. In python 3 you can do this:

```python
yield from super(VideoTimelineCreatePartitionTask, self).requires()
yield VideoUsageTableTask()
```

No matter what you choose here, I recommend returning the `VideoUsageTableTask` from `requires()` instead of `requires_local()`.

mulby (commented):
First batch of comments, haven't gotten all of the way through yet.

mulby:
Overall this is looking great so far!

edx/analytics/tasks/insights/video.py
azafty468: So after a short implementation I'm going to suggest we do not do this.  The reason being  because HivePartitionTask is expecting the table task to be return by the hive_table_task() call and is used in some shared functionality (such as when determining the storage location).  I would have to override base functionality to retarget it to a dict object and it feels like a bit too much mucking around in here in exchange for simplification of requirements.

edx/analytics/tasks/insights/video.py
brianhw: I agree... otherwise all our tasks will have "Create" in them.

edx/analytics/tasks/insights/video.py
```diff
@@ -528,18 +605,61 @@ def query(self):
...
+        return VideoTimelineCreatePartitionTask(
+            mapreduce_engine=self.mapreduce_engine,
+            n_reduce_tasks=self.n_reduce_tasks,
+            source=self.source,
+            interval=self.interval,
+            pattern=self.pattern,
+            warehouse_path=self.warehouse_path,
+        )
+
+    # TODO I need to define what the output is here and what complete means.  This job is finishing but the downstream task cannot confirm this is complete
```
brianhw: Is this still the case?  I would think the downstream task is saying it's complete because the partition exists.  Not sufficient, but necessary, and should get happy-path acceptance tests working.

edx/analytics/tasks/insights/video.py
```diff
@@ -528,18 +605,61 @@ def query(self):
...
+    def output(self):
+        return HivePartitionTarget(
+            self.requires_local().requires_local()["table_task"].table,
```
brianhw: This stuff could use a comment, when you get a chance.

edx/analytics/tasks/insights/video.py
```diff
@@ -549,25 +669,69 @@ def required_table_tasks(self):
...
+    @property
+    def partition_by(self):
+        return self.requires_local()["table_task"].partition_by
+
+    @property
+    def partition_value(self):
+        # Use a dynamic partition value based on the date parameter.
+        return self.interval.date_b.isoformat()  # pylint: disable=no-member
+
+    def requires_local(self):
```
brianhw: Just curious why using requires_local instead of just requires()?  I thought I saw a PR comment somewhere addressing this, but it would be good to get that in a comment in the code.  (Because I'm sure to ask it again.   And again. : )

edx/analytics/tasks/insights/video.py
```diff
@@ -589,22 +753,26 @@ def query(self):
...
+    def query(self):
+        full_insert_query = """
+                    USE {database_name};
+                    INSERT INTO TABLE {table}
+                    PARTITION ({partition.query_spec})
+                    {insert_query};
+                """.format(
+                    database_name=hive_database_name(),
+                    table=self.requires_local().requires_local()["table_task"].table,
+                    partition=self.requires_local().partition,
```
brianhw: This should just call self.partition

edx/analytics/tasks/insights/video.py
```diff
@@ -613,16 +781,45 @@ def required_table_tasks(self):
             warehouse_path=self.warehouse_path,
         )
 
+    # TODO I need to define what the output is here and what complete means.  This job is finishing but the downstream task cannot confirm this is complete
+    def output(self):
+        return HivePartitionTarget(
+            self.requires_local().requires_local()["table_task"].table,
+            self.requires_local().partition,
```
brianhw: This should just call self.partition

edx/analytics/tasks/insights/video.py
```diff
@@ -613,16 +781,45 @@ def required_table_tasks(self):
             warehouse_path=self.warehouse_path,
         )
 
+    # TODO I need to define what the output is here and what complete means.  This job is finishing but the downstream task cannot confirm this is complete
+    def output(self):
+        return HivePartitionTarget(
+            self.requires_local().requires_local()["table_task"].table,
```
brianhw: Since this is used twice now, I would suggest pulling it out into a local method, and having a docstring on it to explain what it's doing.

edx/analytics/tasks/insights/video.py
```diff
@@ -613,16 +781,45 @@ def required_table_tasks(self):
...
+            self.requires_local().partition,
+            database=hive_database_name(),
+            fail_missing_table=True
+        )
+
```
brianhw: nit: remove extra line.

brianhw (commented):
A few more minor comments.

edx/analytics/tasks/insights/video.py
azafty468: Ooooh, neat.  I had misread that part of the luigi docs.  I had interpreted requires_local as the set of requirements that were passed down to the runner and available during execution time, and therefore any variables that would be accessed needed to be present on the local requirements.  I'll update it here and throughout the code as I used it fairly prevalently.

azafty468: Ooooh, neat.  I had misread that part of the luigi docs.  I had interpreted requires_local as the set of requirements that were passed down to the runner and available during execution time, and therefore any variables that would be accessed needed to be present on the local requirements.  I'll update it here and throughout the code as I used it fairly prevalently.

azafty468:
Alright I've made this guy pep8 compliant, incorporated the changes above and confirmed that the acceptance tests run.  I have to add some unit testing but this guy should be ready for review

edx/analytics/tasks/insights/video.py
```diff
@@ -510,15 +544,62 @@ def output(self):
...
+    def partition_value(self):
+        """Use a dynamic partition value based on the date parameter."""
+        return self.interval.date_b.isoformat()  # pylint: disable=no-member
+
+    def requires(self):
+        # There is some glue in requires that links in the data and table tasks that needs to survive unscathed
+        for requirement in super(VideoTimelinePartitionTask, self).requires():
+            yield requirement
+
+        yield VideoUsageTableTask(
```
mulby: nit: given that this isn't used here, but actually in `VideoTimelineDataTask`, I would consider adding it to the requirements of that task instead of this one.

edx/analytics/tasks/insights/video.py
```diff
@@ -528,18 +609,59 @@ def query(self):
...
+    def partition(self):
+        """ A shorthand for the partition information on the upstream partition task """
+        return self.partition_task.partition  # pylint: disable=no-member
+
+    def requires(self):
+        for requirement in super(VideoTimelineDataTask, self).requires():
+            yield requirement
+        yield self.partition_task
+
+    def output(self):
```
mulby: Interesting side effect of this output condition is that if we run the task once on say dt=2017-03-15 it will write the data to S3 and create the partition on the EMR cluster that ran the workflow. Once that cluster finishes its work, we teardown the cluster (thus deleting the partition metadata). If we run the workflow *again* for that day, Luigi will see the data in S3 and won't run this task (since the output already exists). Instead it will just run any downstream tasks.

In this particular case, there aren't any downstream tasks that would want to use this data in a hive query... but if there were, they would fail, complaining that the table didn't exist.

Not something to solve now, but this might be worth a comment here. This only works if the downstream consumers of this data are just interested in the data in S3 and don't care about the actual hive partition.

edx/analytics/tasks/insights/video.py
```diff
@@ -549,25 +671,67 @@ def required_table_tasks(self):
...
+    def partition_value(self):
+        # Use a dynamic partition value based on the date parameter.
+        return self.interval.date_b.isoformat()  # pylint: disable=no-member
+
+    def requires(self):
+        # There is some glue in requires that links in the data and table tasks that needs to survive unscathed
+        for requirement in super(VideoPartitionTask, self).requires():
+            yield requirement
+
+        yield VideoUsageTableTask(
```
mulby: nit: similar to above, I would recommend moving this to the task that actually requires it.

edx/analytics/tasks/insights/video.py
```diff
@@ -589,22 +753,29 @@ def query(self):
                 users_at_end
         """
 
+    def query(self):
+        full_insert_query = """
+                    USE {database_name};
+                    INSERT INTO TABLE {table}
+                    PARTITION ({partition.query_spec})
+                    {insert_query};
+                """.format(
```
iloveagent57: This feels like a lot of left-padding. I usually like to line up the closing `"""` with the beginning of the variable name, then indent the the body of the string by 4 spaces (or whatever # of spaces). Of course, I have no idea if that's PEP-8 compliant.

edx/analytics/tasks/insights/video.py
```diff
@@ -421,16 +464,18 @@ def reducer(self, key, viewings):
...
-                course_id,
-                encoded_module_id,
-                int(video_duration),
-                VIDEO_VIEWING_SECONDS_PER_SEGMENT,
-                users_at_start,
-                users_at_end,
-                segment,
-                len(stats.get('users', [])),
-                stats.get('views', 0),
+                VideoSegmentDetailRecord(
```
iloveagent57: I don't think you need the enclosing parentheses around the result of `to_string_tuple()`, since that function presumably returns a tuple.

iloveagent57 (commented):
I only have these two minor points to add about style, since I don't know what I'm doing.

edx/analytics/tasks/insights/video.py
```diff
@@ -613,20 +784,50 @@ def required_table_tasks(self):
...
 
-    @property
-    def partition(self):
-        return HivePartition('dt', self.interval.date_b.isoformat())  # pylint: disable=no-member
-
 
 @workflow_entry_point
 class InsertToMysqlAllVideoTask(VideoTableDownstreamMixin, luigi.WrapperTask):
-    """Insert all video data into MySQL."""
+    """Insert all video data into MySQL. """
```
brianhw: nit:  extra space?

edx/analytics/tasks/insights/video.py
```diff
@@ -549,25 +671,67 @@ def required_table_tasks(self):
...
+            warehouse_path=self.warehouse_path,
+        )
+
+    @property
+    def partition_value(self):
+        # Use a dynamic partition value based on the date parameter.
+        return self.interval.date_b.isoformat()  # pylint: disable=no-member
+
+    def requires(self):
+        # There is some glue in requires that links in the data and table tasks that needs to survive unscathed
```
brianhw: I don't understand the comment.   Oh, if you are talking about the method, I would backquote or just say requires().  I was trying to parse it as a word.  But what you mean is super().requires() or HivePartitionTask.requires().

nit: use complete sentences, so end with a period.

edx/analytics/tasks/insights/video.py
mulby: Here's some docs!

```
pipeline_video_id = A concatenation of the course_id and the HTML encoded ID of the video. Intended to uniquely identify an instance of a video in a particular course. Note that ideally we would use an XBlock usage_id here, but it isn't present on the legacy events.
segment = An integer representing the rank of the segment within the video. 0 is the first segment, 1 is the second etc. Note that this does not specify the length of the segment.
num_users = The number of unique users who watched any part of this segment of the video.
num_views = The total number of times any part of the segment was viewed, regardless of who was watching it.

course_id = Course the video was displayed in. This is an opaque key serialized to a string.
encoded_module_id = This is the HTML encoded module ID for the video. Ideally this would be an XBlock usage_id, but that data is not present on legacy events.
duration = The video length in seconds. This can be inferred for some videos. We don't have reliable metadata for the length of videos in the source data.
segment_length = The length of each segment, in seconds.
users_at_start = The number of users who watched the first segment of the video.
users_at_end = The number of users who watched the end of the video. Note that this is not the number of users who watched the last segment of the video.
total_viewed_seconds = The total number of seconds viewed by all users across all segments.
```

edx/analytics/tasks/insights/video.py
azafty468: Hmm, I do believe you are correct.  I'll update that now

edx/analytics/tasks/insights/video.py
```diff
@@ -510,15 +544,62 @@ def output(self):
...
+
+    @property
+    def hive_table_task(self):
+        return VideoTimelineTableTask(
+            warehouse_path=self.warehouse_path,
+        )
+
+    # @property
+    # def partition_by(self):
+    #     return self.hive_table_task.partition_by
```
mulby: nit: remove this commented out code!

edx/analytics/tasks/insights/video.py
```diff
@@ -613,20 +784,50 @@ def required_table_tasks(self):
...
+    """Insert summary information into the video table in MySQL."""
+
+    @property
+    def table(self):
+        return "video"
+
+    @property
+    def insert_source_task(self):
+        return VideoDataTask(
+            mapreduce_engine=self.mapreduce_engine,  # questionable about this one
```
brianhw: I don't understand the comment....

mulby (approved):
Assuming this doesn't break things - LGTM!

edx/analytics/tasks/insights/video.py
```diff
@@ -589,22 +753,29 @@ def query(self):
...
-            ('course_id', 'VARCHAR(255)'),
-            ('encoded_module_id', 'VARCHAR(255)'),
-            ('duration', 'INTEGER'),
-            ('segment_length', 'INTEGER'),
-            ('users_at_start', 'INTEGER'),
-            ('users_at_end', 'INTEGER'),
-            ('total_viewed_seconds', 'INTEGER'),
-        ]
+    def partition(self):
+        """ A shorthand for the partition object on the upstream partition task. """
```
brianhw: nit: we don't generally pad docstrings with leading and trailing whitespace....

edx/analytics/tasks/insights/video.py
```diff
@@ -549,25 +671,67 @@ def required_table_tasks(self):
...
+            n_reduce_tasks=self.n_reduce_tasks,
+            source=self.source,
+            interval=self.interval,
+            pattern=self.pattern,
+            warehouse_path=self.warehouse_path,
+        )
+
+
+class VideoDataTask(VideoTableDownstreamMixin, HiveQueryTask):
+    """Execute the query on video_usage and persist the results into video"""
```
brianhw: nit: trailing period.

edx/analytics/tasks/insights/video.py
```diff
@@ -549,25 +671,67 @@ def required_table_tasks(self):
...
-class InsertToMysqlVideoTask(VideoTableDownstreamMixin, HiveQueryToMysqlTask):
-    """Insert non-segment information about videos from a Hive table into MySQL."""
+    @property
+    def partition_by(self):
+        return 'dt'
 
     @property
     def table(self):
-        return "video"
+        return 'video'
```
brianhw: nit:  curious about this change, when video_timeline was not changed....

edx/analytics/tasks/insights/video.py
azafty468: Yup, I'm game.  Doing it now and will test

edx/analytics/tasks/insights/video.py
brianhw: Remove comment.

edx/analytics/tasks/insights/video.py
```diff
@@ -367,19 +414,15 @@ def mapper(self, line):
     def reducer(self, key, viewings):
         """
         Outputs a record for each watched segment in each video module in each course.
-
```
brianhw: nit: please restore blank lines.

edx/analytics/tasks/insights/video.py
```diff
@@ -528,18 +609,59 @@ def query(self):
...
+    def requires(self):
+        for requirement in super(VideoTimelineDataTask, self).requires():
+            yield requirement
+        yield self.partition_task
+
+    def output(self):
+        output_root = url_path_join(self.warehouse_path,
+                                    self.partition_task.hive_table_task.table,
+                                    self.partition.path_spec + '/')
+        return get_target_from_url(output_root)
```
brianhw: I'm not clear how this output() differs from the partition?  So if this task were to fail for some reason, such as the query failed or the machine crashed in the middle of executing the query, wouldn't this just not run again because the partition exists?   I thought there would need to be some kind of marker here to distinguish between the partition being created and the query being executed.

edx/analytics/tasks/insights/video.py
azafty468: Done!  Added a note:
`Warning this creates data potentially in a shared data source (i.e. S3).  This output will not naturally be wiped out between runs.  Re-running on the same partition date will short circuit the creation of the partition + table metadata in Hive.  This is ok for this job as no downstream tasks interact with the Hive metadata.  However if tasks were added in the future this could be a source of error, and any shared data locations would need to be manually removed to force a regeneration of the partition + table metadata.`

brianhw (commented):
My main concern is with the output of the Hive query tasks being the same as the output of the Hive partition tasks that they require.  (I also added a number of minor nits as well, but they're much less important.)

edx/analytics/tasks/insights/video.py
azafty468: That is really weird, I did restore these breaks between the summary and the detail

edx/analytics/tasks/insights/video.py
azafty468: Ah that was a stray comment from when I first read through this doc.  I'm wiping it now

edx/analytics/tasks/insights/video.py
brianhw: I commented on this independently below, but I think there are two problems here, if I'm understanding things.  One is as you say, whether data gets loaded into a Hive partition in the metastore, but the second is whether the query that populates that partition has been run.  I think we want the latter to work properly, even if we choose not to figure out how to get the former to work.    I think the former can be made to work by having the tasks that need the data to specify both this task and the partition task as its requirements, and not rely on this to have created the partition as well.  So it's only the latter -- separating the output() of the Hive Partition and the Hive Query tasks -- that needs to be addressed.

azafty468:
Alright this last one should address everything.  I've moved the logic into the Target objects, had to re-hook a few areas that generated the targets and I still need to override the on_success() call at the end of the Hive jobs to generate the marker.  But this implementation should be cleaner and does not require any overrides to input() or output() calls anymore.

edx/analytics/tasks/insights/video.py
```diff
@@ -1,25 +1,35 @@
...
+"""Tasks for aggregating statistics about video viewing."""
 from collections import namedtuple
 import json
 import logging
 import math
 import re
 import urllib
+import textwrap
+import urlparse
+import os
```
brianhw: nit:  imports should be grouped by type (e.g. system, external, our own), but then be alphabetic.  These three are not in order.

edx/analytics/tasks/insights/video.py
```diff
@@ -43,6 +53,178 @@
...
+                                                'course. Note that ideally we would use an XBlock usage_id here, but '
+                                                'it isn\'t present on the legacy events.')
+    course_id = StringField(length=255,
+                            nullable=False,
+                            description='Course the video was displayed in. This is an opaque key serialized to '
+                                        'a string.')
+    encoded_module_id = StringField(length=255,
+                                    nullable=False,
+                                    description='This is the HTML encoded module ID for the video. Ideally this would '
+                                                'be an XBlock usage_id, but that data is not present on legacy events.')
```
brianhw, line 88: This is perhaps more a question for @mulby, but is this intended to be like the pipeline_video_id (per video per course) or to represent the video independent of the course?   A usage_id provides the former, but not the latter.

edx/analytics/tasks/insights/video.py
```diff
@@ -43,6 +53,178 @@
...
+    duration = IntegerField(description='The video length in seconds. This can be inferred for some videos. We don\'t '
+                                        'have reliable metadata for the length of videos in the source data.')
+    segment_length = IntegerField(description='The length of each segment, in seconds.')
+    users_at_start = IntegerField(description='The number of users who watched the first segment of the video.')
+    users_at_end = IntegerField(description='The number of users who watched the end of the video. Note that this is '
+                                            'not the number of users who watched the last segment of the video.')
```
brianhw, line 94: This doesn't say what it is, though, so it's very confusing.   Perhaps "... who watched close to the end of the video."?    Without defining here what is meant by "close" but at least making clear why it isn't (necessarily) the last segment.

edx/analytics/tasks/insights/video.py
```diff
@@ -43,6 +53,178 @@
...
+                                       'of the segment.')
+    num_users = IntegerField(description='The number of unique users who watched any part of this segment of the '
+                                         'video.')
+    num_views = IntegerField(description='The total number of times any part of the segment was viewed, regardless of '
+                                         'who was watching it.')
+
+
+class S3MarkerTarget(S3Target):
+    """An S3 Target that uses a marker file to indicate success.  In reality these should be shared functionality but
+    for now they are localized in the video process."""
```
brianhw: So is this to be captured in a follow-up story?  Is the question hindering doing this just to decide which file in ../util to put this marker functionality?  And should I hold off on commenting on this until such time?

edx/analytics/tasks/insights/video.py
```diff
@@ -43,6 +53,178 @@
...
+    'hdfs': HdfsMarkerTarget,
+    's3': S3HdfsTarget,
+    's3n': S3HdfsTarget,
+    'file': LocalMarkerTarget,
+    's3+https': S3MarkerTarget,
+}
+
+
+# Lifted from url.py and modified to use the above listed classes
+def get_tag_target_class_from_url(url):
```
brianhw: I don't understand the use of "_tag_" here.   How does that relate to "marker", for example?

edx/analytics/tasks/insights/video.py
```diff
@@ -43,6 +53,178 @@
...
+        if parentfolder and not os.path.exists(parentfolder):
+            os.makedirs(parentfolder)
+
+        if self.format:
+            self.format.pipe_writer(atomic_file(self.path + "/_SUCCESS")).close()
+        else:
+            atomic_file(self.path + "/_SUCCESS").close()
+
+
+# Lifted from url.py and modified to use the above listed classes
```
brianhw: Is there a reason not to just put all this into url.py?  I guess there's a concern that others might start to use it before it's fully baked?

edx/analytics/tasks/insights/video.py
```diff
@@ -43,6 +53,178 @@
...
+DEFAULT_TARGET_CLASS = LocalMarkerTarget
+URL_SCHEME_TO_TARGET_CLASS = {
```
brianhw: URL_SCHEME_TO_MARKER_TARGET_CLASS ?   Or is the fact that S3HdfsTarget is here somehow give it a different meaning, because it doesn't implement marker logic?  Also, does S3HdfsTarget provide a touch_marker method?  That seems to be needed in order for on_success() to call it on the output of get_tag_target_from_url().

edx/analytics/tasks/insights/video.py
```diff
@@ -528,18 +728,73 @@ def query(self):
...
+
+    def output(self):  # pragma: no cover
+        output_root = url_path_join(self.warehouse_path,
+                                    self.partition_task.hive_table_task.table,
+                                    self.partition.path_spec + '/')
+        return get_tag_target_from_url(output_root)
+
+    def on_success(self):  # pragma: no cover
+        """Overload the success method to touch the _SUCCESS file.  Any class that uses a separate Marker file from the
+        data file will need to override the base on_success() call to create this marker."""
```
brianhw: Is this an override, or an add-on?  Just wondering if this should also call super().on_success()?

edx/analytics/tasks/insights/video.py
brianhw: sounds reasonable to me.

brianhw (commented):
Great progress!  I think the use of targets greatly simplifies the downstream logic -- an excellent solution.  I'm just concerned that it's not fully implemented for S3, and in particular, that it won't work in production.

edx/analytics/tasks/insights/video.py
mulby: IIRC it's the "hash" at the end of the usage_id which is *not* course specific. We want something that is course specific (ideally), but don't have it available, which is what forced the creation of the pipeline_video_id.

edx/analytics/tasks/insights/video.py
mulby: +1

edx/analytics/tasks/insights/video.py
mulby: I support moving this to util

edx/analytics/tasks/insights/video.py
mulby: I think we could get away without duplicating all of this logic by modifying the stuff in url.py to be something more like this:

```python
DEFAULT_TARGET_CLASS = luigi.LocalTarget
URL_SCHEME_TO_TARGET_CLASS = {
    'hdfs': luigi.hdfs.HdfsTarget,
    's3': S3HdfsTarget,
    's3n': S3HdfsTarget,
    'file': luigi.LocalTarget,
    's3+https': luigi.s3.S3Target,
}
DEFAULT_MARKER_TARGET_CLASS = LocalMarkerTarget
MARKER_TARGET_URL_SCHEME_TO_TARGET_CLASS = {
    'hdfs': HdfsMarkerTarget,
    's3': S3HdfsTarget,
    's3n': S3HdfsTarget,
    'file': LocalMarkerTarget,
    's3+https': S3MarkerTarget,
}

def get_target_class_from_url(url, marker=False):
    """Returns a luigi target class based on the url scheme"""
    parsed_url = urlparse.urlparse(url)
    if marker:
        target_class = URL_SCHEME_TO_TARGET_CLASS.get(parsed_url.scheme, DEFAULT_MARKER_ TARGET_CLASS)
    else:
        target_class = URL_SCHEME_TO_TARGET_CLASS.get(parsed_url.scheme, DEFAULT_TARGET_CLASS)
    kwargs = {}
    if issubclass(target_class, luigi.hdfs.HdfsTarget) and url.endswith('/'):
        kwargs['format'] = luigi.hdfs.PlainDir
    if issubclass(target_class, luigi.LocalTarget) or parsed_url.scheme == 'hdfs':
        # LocalTarget and HdfsTarget both expect paths without any scheme, netloc etc, just bare paths. So strip
        # everything else off the url and pass that in to the target.
        url = parsed_url.path
    if issubclass(target_class, luigi.s3.S3Target):
        kwargs['client'] = ScalableS3Client()

    url = url.rstrip('/')
    args = (url,)

    return target_class, args, kwargs

def get_target_from_url(url, marker=False):
    """Returns a luigi target based on the url scheme"""
    cls, args, kwargs = get_target_class_from_url(url, marker=marker)
    return cls(*args, **kwargs)
```

edx/analytics/tasks/insights/video.py
```diff
@@ -506,19 +677,48 @@ def requires(self):
...
+    @property
+    def columns(self):  # pragma: no cover
+        return VideoTimelineRecord.get_hive_schema()
+
+
+class VideoTimelinePartitionTask(VideoTableDownstreamMixin, HivePartitionTask):
+    """Creates the Hive storage partition used to hold video_timeline data."""
+
+    @property
+    def hive_table_task(self):  # pragma: no cover
```
mulby, line 609: nit (no need to do this now, just for future reference): you can move this declaration to the class level in cases like this

mulby (commented):
Overall - looking great! I love the marker target idea, but would like to move it to a more central location.

edx/analytics/tasks/insights/video.py
brianhw: Right.  So my question is whether saying that the encoded_module_id should ideally be a usage_id makes sense.  It does in the comment for the pipeline_video_id.    Maybe the comment here should mention that it is independent of the course.

edx/analytics/tasks/insights/video.py
brianhw: Heh.  So my further comments were about combining the marker into get_target_from_url.  Good.

edx/analytics/tasks/insights/video.py
brianhw: Exactly.  I was going to propose the same.    (Just note that the "if marker" clause should use the marker-related map.  And I would suggest "URL_SCHEME_TO_MARKER_TARGET_CLASS".

edx/analytics/tasks/insights/video.py
```diff
@@ -43,6 +53,178 @@
...
+                                         'who was watching it.')
+
+
+class S3MarkerTarget(S3Target):
+    """An S3 Target that uses a marker file to indicate success.  In reality these should be shared functionality but
+    for now they are localized in the video process."""
+
+    def exists(self):  # pragma: no cover
+        """Completion of this target is based solely on the existence of the marker file."""
+        return self.fs.exists(self.path + "/_SUCCESS")
```
brianhw: Also, I'm not clear whether it would be possible to implement the marker functionality as a single implementation in a common mixin.  That is, the exists() implementations I see here are all identical.  Is it possible for the touch_marker() implementations to also be implemented using existing methods of the underlying target, so we don't have to cut-and-paste such internal operations as AtomicS3File and format.pipe_writer and such?

edx/analytics/tasks/insights/video.py
azafty468: Yup!  Fully anticipated when I copied the original util.py code out so simple enough to re-incorporate back into util.py.  The one area I'm a bit concerned about however is the S3HdfsTarget file.  I'm not sure if this operation is covered in an acceptance test or if I need to write one.  I don't mind extending out the exist() and touch() calls into a new S3HdfsMarkerTarget class, but I didn't want to do that without some basic tests

brianhw: Do the acceptance tests work on Jenkins?  I assume the existing video acceptance test uses S3, while running on analyticstack uses HDFS.   And so I would expect the test to break on Jenkins.

edx/analytics/tasks/insights/video.py
azafty468: Tough request on the touch_marker() implementation.  3 of the 4 base classes for the markers are defined in the Luigi packages so I won't have access to modify them.  I could write custom mixins but each would be unique to the specific filesystem, leaving us with no saved lines of code and method inheritance obscured behind mixins.  

The exists() function can be extricated.  My usual inclination would be to not do this as we are replacing a 1 line function definition with a mixin inclusion.  However I would make this change if it is more consistent with the rest of the library.  So if normally mixins would be utilized here then I will update it.

brianhw: I also don't think a mixin is worth it for the exists() functionality alone.  It was more a question about whether the touch_marker() implementation could be generalized.

azafty468: Ahhhh, gotcha gotcha gotcha.  Let me take a second look at it and see if there is something that can be done.

brianhw: One suggestion:  in touch_marker(), create a second Target object of the same type, using the original URL (self.path?) with _SUCCESS appended.  Then just call .open() and .close() on that.

edx/analytics/tasks/util/url.py
```diff
@@ -18,10 +18,56 @@
...
+    another positive marker to indicate Task success."""
+
+    def exists(self):  # pragma: no cover
+        """Completion of this target is based solely on the existence of the marker file."""
+        return self.fs.exists(self.path + "/_SUCCESS")
+
+    def touch_marker(self):  # pragma: no cover
+        """Generate the marker file using file system native to the parent Target."""
+
+        tmp = self.__copy__()
```
brianhw: Just wondering if there is a simpler way to do this, that involves just getting the class from self.class or something similar, and calling a constructor on the class with named arguments to generate a new instance of the class with the new path argument.  I'm concerned that the format may not quite be the right thing, for example, if the target here is a directory and the marker target should be a plain file.

edx/analytics/tasks/insights/video.py
```diff
@@ -548,26 +678,84 @@ def required_table_tasks(self):
             warehouse_path=self.warehouse_path,
         )
 
+    def output(self):  # pragma: no cover
+        output_root = url_path_join(self.warehouse_path,
+                                    self.partition_task.hive_table_task.table,
+                                    self.partition.path_spec + '/')
+        return get_target_from_url(output_root, True)
```
mulby: nit: for clarity I would prefer you explicitly state `marker=True` in these calls to make it more clear what `True` represents

edx/analytics/tasks/util/url.py
mulby: I feel like this might be more clever than it's worth - taking DRY to a level that's actually unhelpful. I liked it better before when you a different `touch_marker` method in each class.

Brian and I disagree on this point periodically...

mulby (approved):
LGTM once this passes tests and a clean RC build

edx/analytics/tasks/util/url.py
brianhw: Right, we periodically do.   But I was hoping that we wouldn't need special copy methods per class.  I think originally I was hoping we could just wrap or monkey-patch the classes returned by get_target_from_url() with a new exists() and touch_marker().  Since we need to define marker classes as derived classes, if only so that the issubclass logic works, that probably rules out the wrapper approach.  But I really would like to have as little target-specific logic copied from Luigi innards as possible, whether we use a copy method or a  constructor or whatever.

brianhw (approved):
LGTM. 👍   Thanks for your patience with working through the marker file implementation.    One remaining nit is the comment on the users-at-end, but I don't need for it to block.

azafty468:
Squashed down and ready to go

mulby:
Great work! Congrats on your first merged PR! :confetti_ball: :tada: :champagne:

PR #365, opened 2017-03-21
Comment out video counters

brianhw:
It seems that having too many counter calls causes the App Manager to die.  Twice.
Rather than try to tune the AM, we are just commenting out the calls in the short term.

Calls to increment counters that are commented out with a single hash happen more frequently.  Calls commented out with a double hash are less frequent, and are candidates to be restored first.


tobz (approved):
LGTM.

mulby (approved):
LGTM

HassanJaveed84 (approved):

PR #366, opened 2017-03-21
AN-8554 | Load course_meta_program table in export DB

iloveagent57:
This is mostly me mimicking existing code.  A piece of the pipeline already existed to pluck course-program data our of the course catalog API and into a Hive Table.  This change loads the data in the hive table into Mysql.


brianhw:
Ah, this is rather unfortunate in terms of the existing code being mimicked.  We are actually trying to rid ourselves of HiveQueryToMysqlTask.   It is currently limited to student_engagement.py, enrollments.py, user_activity.py, and video.py.  Andrew (@azafty468) is currently nearing completion on work to remove it from video.py, which will also hopefully provide us with a template for how to do the same with the others.

iloveagent57:
That's fine - the hard part of this was learning my way around the codebase and wrapping my head around the fundamental concepts of this pipeline, so I don't think we're losing any value if all of this code just gets thrown away.

brianhw:
But not all has to be thrown away.  It's a good start, and the acceptance test in particular should not need to change.

iloveagent57:
I had to merge instead of rebase to get the changes in from master (after previously rebasing off of Andrew's branch), but the diff between this branch and master is at least correct. Also, it now follows the better pattern set forth in Andrew's changes and eschews the `HiveQueryInsertToMysql` approach to creating the `course_meta_program` table.

edx/analytics/tasks/insights/enrollments.py
```diff
@@ -1068,3 +1091,134 @@ def requires(self):
...
+        return get_target_from_url(output_root, marker=True)
+
+    def on_success(self):
+        """Override the success method to touch the _SUCCESS file.  Any class
+        that uses a separate Marker file from the data file will need to
+        override the base on_success() call to create this marker."""
+        self.output().touch_marker()
+
+
+class CourseMetaProgramInsertToMysqlTask(CourseSummaryEnrollmentDownstreamMixin,
```
azafty468: Question should there be an @workflow_entry_point annotation at the point?  It will run fine without it as there already is one on ImportEnrollmentsIntoMysql but if that were to be removed...I'm not sure what would happen.

azafty468:
I'll let Gabe & Brian speak closer to the style and formatting.  The general workflow and structure of the replacement HiveQueryToMysqlTask looks good to me

edx/analytics/tasks/insights/enrollments.py
HassanJaveed84: I thought we used @workflow_entry_point for generating docs(I might be wrong) and put the entry points in setup.cfg

edx/analytics/tasks/insights/enrollments.py
```diff
@@ -1068,3 +1091,134 @@ def requires(self):
...
+    def insert_query(self):
+        query = """
+        SELECT program_id,
+               program_type,
+               program_title,
+               catalog_course,
+               catalog_course_title,
+               course_id,
+               org_id,
+               partner_short_code
```
mulby: According to the spec, I think we just need course_id, program_id, program_type, and program_title.

edx/analytics/tasks/insights/enrollments.py
```diff
@@ -1068,3 +1091,134 @@ def requires(self):
...
+    def table(self):
+        return 'course_meta_program'
+
+    @property
+    def columns(self):
+        return ProgramCourseRecord.get_sql_schema()
+
+    @property
+    def indexes(self):
+        return [('program_id',), ('course_id',)]
```
mulby: We should check with @dsjen and @thallada about the expected access pattern here. I think this will work for their use case?

edx/analytics/tasks/insights/enrollments.py
```diff
@@ -1068,3 +1091,134 @@ def requires(self):
             EnrollmentDailyTask(**enrollment_kwargs),
             ImportCourseSummaryEnrollmentsIntoMysql(**course_summary_kwargs),
```
mulby: You'll want to add your task here to ensure that it's run as part of the enrollment workflow: `CourseMetaProgramInsertToMysqlTask(**course_summary_kwargs)`. You could also insert it into the requirements chain of the course summary stuff (that actually might make more sense).

mulby (commented):
Overall - looking great!

The one thing we need to figure out is where this fits into the existing course summary-related DAG.

edx/analytics/tasks/insights/enrollments.py
mulby: It is just used to generate docs!

edx/analytics/tasks/tests/acceptance/fixtures/input/program_course_catalog.json
```diff
@@ -0,0 +1,3 @@
...   "description": null,     "height": null,     "width": null   },   "video": null,   "seats": [     {       "type": "honor",       "price": "0.00",       "currency": "USD",       "upgrade_deadline": "2017-04-01T00:00:00Z",       "credit_provider": "edx",       "credit_hours": 4     },     {       "type": "verified",       "price": "50.00",       "currency": "USD",       "upgrade_deadline": "2016-10-01T08:00:00Z",       "credit_provider": "edx",       "credit_hours": 3     }   ],   "content_language": "en-us",   "transcript_languages": [],   "instructors": [],   "staff": [],   "pacing_type": "instructor_paced",   "min_effort": 2,   "max_effort": 4,   "modified": "2016-09-07T18:38:32.207371Z",   "marketing_url": "https://www.edx.org/course/all-about-acceptance-testing-edx-testing101-3t2016",   "level_type": "Intermediate",   "availability": "Current",   "programs": [     {       "uuid": "acb243a0-1234-5abe-099e-ffcae2a340d4",       "title": "Testing",       "type": "XSeries",       "marketing_slug": "tests",       "marketing_url": "https://example.com/xseries/testing"     }   ],   "partner_short_code": "openedx",   "announcement": "2016-08-02T00:00:00Z",   "reporting_type": "test" }
```
HassanJaveed84: I think we can just use 'course_catalog.json'. We already load them for enrollments and course_catalog acceptance tests.

edx/analytics/tasks/insights/enrollments.py
dsjen: LGTM!  We're likely just going to select by course ID.

HassanJaveed84:
Looks great!
Question for @mulby and @brianhw: Does it make sense to move this new code out of enrollments.py ?

edx/analytics/tasks/insights/enrollments.py
iloveagent57: Good point.

edx/analytics/tasks/insights/enrollments.py
iloveagent57: Noted.

edx/analytics/tasks/tests/acceptance/fixtures/input/program_course_catalog.json
iloveagent57: Good point, scratching this.

edx/analytics/tasks/insights/enrollments.py
```diff
@@ -2,27 +2,50 @@
...
     LoadInternalReportingCourseCatalogMixin,
+    ProgramCoursePartitionTask,
+    ProgramCourseRecord,
 )
-from edx.analytics.tasks.common.mapreduce import MapReduceJobTaskMixin, MapReduceJobTask, MultiOutputMapReduceJobTask
+from edx.analytics.tasks.common.mapreduce import (
+    MapReduceJobTaskMixin,
+    MapReduceJobTask,
+    MultiOutputMapReduceJobTask,
+)
```
brianhw: nit: is there really a reason to split these onto separate lines?  I personally find it harder to read than single-line imports, when the latter are short enough to fit on the line (i.e. less than 120 characters).

edx/analytics/tasks/insights/enrollments.py
```diff
@@ -2,27 +2,50 @@
...
 from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
-from edx.analytics.tasks.util.record import Record, StringField, IntegerField, BooleanField, DateTimeField
+from edx.analytics.tasks.util.record import (
+    BooleanField,
+    DateTimeField,
+    IntegerField,
+    StringField,
+    Record,
+)
```
brianhw: nit: ditto with this.

iloveagent57:
Thanks for the input, all, I've made changes to address all of these points (except the one about splitting this out from `enrollments.py`, although I agree that the name of the file doesn't line up precisely with the code).

edx/analytics/tasks/insights/enrollments.py
brianhw: Choosing a class to put into the entry points in setup.cfg is arbitrary.  I have not known us to actually use any of those entry points.  They serve only to trigger the loading of the files that are referenced, so that all the tasks in the file are then available.  Since this enrollments.py file already has a "registered class" in setup.cfg, we don't need to add another.

edx/analytics/tasks/insights/enrollments.py
brianhw: There's an issue here.  We want the enrollment workflow to continue to work for installations that do not have a course catalog available to them.  And, in fact, we want it to work this way by default, so that we have to opt-in to getting functionality that does make use of course catalog.   This is presently implemented in ImportCourseSummaryEnrollmentsIntoMysql, where an "enable_course_catalog" parameter is checked before requiring CoursePartitionTask.  If it is not set, then only the table is created and is empty.  This is fine for this Hive query, because the course_catalog is only left joined against, rather than being the only table being queried.    Adding the ImportCourseSummaryEnrollmentsIntoMysql requirement here is problematic, in that if there is an empty course_catalog (using the same trick), then the Mysql will be trying to insert zero rows, and it will return an error.  So it might be necessary to make inclusion of this be entirely controlled by a parameter.    That parameter, in turn, is defined in CourseSummaryEnrollmentDownstreamMixin, and gets its value from the 'course-summary-enrollment' section of the config file.  Should we use the same value?  It would avoid needing to make another config edit....

edx/analytics/tasks/insights/enrollments.py
```diff
@@ -1067,4 +1090,136 @@ def requires(self):
             EnrollmentByEducationLevelTask(**enrollment_kwargs),
             EnrollmentDailyTask(**enrollment_kwargs),
             ImportCourseSummaryEnrollmentsIntoMysql(**course_summary_kwargs),
+            CourseMetaProgramInsertToMysqlTask(**course_summary_kwargs),
```
brianhw: Well, since we're passing "course_summary_kwargs" to it, maybe it's enough to change its run() method to do a no-op if 'enable_course_catalog' is not set.

edx/analytics/tasks/insights/enrollments.py
```diff
@@ -1067,4 +1090,136 @@ def requires(self):
...
+    def partition_by(self):
+        return 'dt'
+
+    @property
+    def table(self):
+        return 'course_meta_program'
+
+    @property
+    def columns(self):
+        column_types_by_name = dict(ProgramCourseRecord.get_hive_schema())
```
brianhw: I'm just curious what this dict() call does.  What does get_hive_schema() return?  Is this to make a copy here?  (It doesn't look like it's being modified, but just good to be safe?)

edx/analytics/tasks/insights/enrollments.py
```diff
@@ -1067,4 +1090,136 @@ def requires(self):
...
+class CourseMetaProgramTableTask(BareHiveTableTask):  # pragma: no cover
+    """Creates the Hive `course_meta_program` storage table."""
+
+    @property
+    def partition_by(self):
+        return 'dt'
+
+    @property
+    def table(self):
+        return 'course_meta_program'
```
brianhw: By the way, what does the "meta" mean here?  Not meta-level in general, but metadata?  Would it be reasonable to make that more explicit  e.g. table name 'course_program_metadata', classnames CourseProgramMetadataXXXTask?

brianhw (commented):
Partial comments, but just to get something out there sooner...

edx/analytics/tasks/insights/enrollments.py
```diff
@@ -1067,4 +1090,136 @@ def requires(self):
...
+        return CourseMetaProgramTableTask(warehouse_path=self.warehouse_path)
+
+    @property
+    def partition_value(self):
+        return self.date.isoformat()
+
+
+class CourseMetaProgramDataTask(CourseSummaryEnrollmentDownstreamMixin,
+                                HiveQueryTask):  # pragma: no cover
+    """Execute a select on the `program_course` Hive table and persist the
```
brianhw: super-nit: summary sentence is supposed to be on a single line, and therefore short enough to fit.  The details go further down in the docstring, after a blank line.  (There are a bunch below as well.)

edx/analytics/tasks/insights/enrollments.py
```diff
@@ -1067,4 +1090,136 @@ def requires(self):
             EnrollmentByEducationLevelTask(**enrollment_kwargs),
             EnrollmentDailyTask(**enrollment_kwargs),
             ImportCourseSummaryEnrollmentsIntoMysql(**course_summary_kwargs),
+            CourseMetaProgramInsertToMysqlTask(**course_summary_kwargs),
+        )
+
+
+class CourseMetaProgramTableTask(BareHiveTableTask):  # pragma: no cover
```
brianhw: nit:  It might be good to have all of this above the workflow endpoint above, just to be consistent in our tendency to put the upstream tasks nearer the top and the downstream tasks near the bottom.

edx/analytics/tasks/tests/acceptance/test_internal_reporting_program_courses.py
```diff
@@ -0,0 +1,53 @@
...
+                          '--date', self.DATE])
+
+        columns = ['course_id', 'program_id', 'program_type', 'program_title']
+
+        with self.export_db.cursor() as cursor:
+            cursor.execute(
+                '''
+                  SELECT {columns}
+                  FROM   course_meta_program
+                '''.format(columns=','.join(columns))
```
brianhw: Would a more thorough test of this to not be limited to a hardcoded list of columns here?  If you were to add another column in the code, this wouldn't fail.  Can you do a 'SELECT *' instead?

brianhw (commented):
Good progress!  A few more minor comments this time.

edx/analytics/tasks/insights/enrollments.py
iloveagent57: Ah, I set a vertical bar at 80 characters in my editor, so maybe that's why this seems more aggressive to your. But I also typically go to multi-line imports when importing more than 2 things from a module.  Lastly, multi-line imports make it easier for me to do `M-x sort-lines` (I guess I also find it easier to read multi-line imports).

edx/analytics/tasks/insights/enrollments.py
iloveagent57: I certainly like that name better, I think I just blindly copied a table name from the ticket. Assuming there's no hard requirement on the table name (@dsjen ?), I'll change it.

dsjen: Go for it!  FYI, @thallada

edx/analytics/tasks/insights/enrollments.py
iloveagent57: `get_hive_schema()` returns a list of 2-tuples, e.g. `[('course_id', 'VARCHAR(255)'), ('program_id', 'VARCHAR(16)')]`, which translates naturally to a dict (and affords me easy lookup in the list comprehension beneath).

edx/analytics/tasks/insights/enrollments.py
iloveagent57: Ah, good to know.  I'm used to writing classes like a story in order of dependency:
```python
class BuildTension():
    pass

class Climax():
    pass

class Resolution():
    pass
```

iloveagent57: (but I will adhere to previously established convention here, obviously).

edx/analytics/tasks/tests/acceptance/test_internal_reporting_program_courses.py
iloveagent57: Good point.

edx/analytics/tasks/insights/enrollments.py
iloveagent57: I guess I could also conditionally include it in the tuple of things to yield in this `requires()` method, based on the value of `self.enable_course_catalog`.  Although modifying `run()` feels somewhat safer to me, maybe?

mulby: We typically do this type of thing my modifying the `requires()` method

edx/analytics/tasks/insights/enrollments.py
```diff
@@ -1106,7 +1107,9 @@ def table(self):
 
     @property
     def columns(self):
-        return ProgramCourseRecord.get_hive_schema()
+        column_types_by_name = dict(ProgramCourseRecord.get_hive_schema())
+        return [(column_name, column_types_by_name[column_name])
+                for column_name in CourseMetaProgramDataTask.COLUMN_NAMES]
```
mulby: Given that you're doing this type of thing in a couple places, I would recommend just creating a new `Record` type. I think that would help clarify things. Not quite as DRY - but clearer, IMHO.

edx/analytics/tasks/insights/enrollments.py
mulby: Good point - I would recommend only including this task in the requirements chain if that config var is set (I don't think this warrants a separate config var).

This could result in the table not being present in the result store at all - which is fine as long as the API and Insights are able to handle that edge case.

FYI - @dsjen @thallada

edx/analytics/tasks/insights/enrollments.py
iloveagent57: On the one hand, I really dislike proliferating classes everywhere, but on the other, it is clearer. So, fine - recommendation accepted.

edx/analytics/tasks/insights/enrollments.py
brianhw: I don't understand the sort-lines comment.  In particular, I don't see how it would be very helpful when ordering imports.  It only orders the plain import statements where they are not "from", and it only orders single-line entries at that.  And it orders multi-line imports if you limit it only to one multi-line at a time, but that can't be used as a reason for using multi-line imports.

edx/analytics/tasks/insights/enrollments.py
brianhw: Yes, we can do it this way -- it's different than the ImportCourseSummaryEnrollmentsIntoMysql case, which is initially jarring, but it does make clearer that there will be no table in the result store.  The alternative to there being no table is to create the table with no rows, and override something to permit it to be created with no rows.  (I'm not sure if it's hardcoded, but it could be made to be controlled with a flag that we set here.)

edx/analytics/tasks/insights/enrollments.py
brianhw: I guess I can go either way -- I didn't find this to be too unclear, except maybe for adding a comment about the tuple-to-dict conversion.   COLUMN_NAMES is used in several places, but that seems pretty clear too.

edx/analytics/tasks/insights/enrollments.py
brianhw: I think we are agreeing?  I was suggesting writing in order of dependency, and you are agreeing.  Isn't that the previously-established convention?   I'm not clear on what is obvious..... : )

edx/analytics/tasks/insights/enrollments.py
brianhw: (Heh -- we don't typically do this type of thing. : )

brianhw:
Hmm.  @HassanJaveed84 asked about splitting some stuff out of this file, which is a good question.   I'm currently dissatisfied with the dependence of "insights" workflows on warehouse tasks, because I generally think of the former as "public" and the latter as "private" (for now).  The course catalog was the first, and this is the second to blur that boundary.  This is awkward because we also expect the course-catalog stuff to be somewhat private -- that we are the only ones with access to a course catalog API.   But we could move the guts of the course-catalog stuff into insights/course_catalog.py, including these Mysql-loaders, and leave the vertica-specific stuff in the warehouse/load-internal-...-course-catalog.py file.   And we should do that if we thought that some openedx installation might actually set up and access their course catalog API.

edx/analytics/tasks/insights/enrollments.py
```diff
@@ -1036,6 +1044,138 @@ def required_table_tasks(self):
...
+    def table(self):
+        return 'course_program_metadata'
+
+    @property
+    def columns(self):
+        return CourseProgramMetadataRecord.get_hive_schema()
+
+
+class CourseProgramMetadataPartitionTask(CourseSummaryEnrollmentDownstreamMixin,
+                                     HivePartitionTask):  # pragma: no cover
```
brianhw: nit: this line doesn't align.

edx/analytics/tasks/insights/enrollments.py
```diff
@@ -1036,6 +1044,138 @@ def required_table_tasks(self):
...
+    @property
+    def partition_value(self):
+        return self.date.isoformat()
+
+
+class CourseProgramMetadataDataTask(CourseSummaryEnrollmentDownstreamMixin,
+                                    HiveQueryTask):  # pragma: no cover
+    """Selects from `program_course` and persists results into `course_program_metadata` Hive table."""
+
+    COLUMN_NAMES = ['course_id', 'program_id', 'program_type', 'program_title']
```
brianhw: Shouldn't this go away?

edx/analytics/tasks/insights/enrollments.py
```diff
@@ -1036,6 +1044,138 @@ def required_table_tasks(self):
...
+
+    @property
+    def insert_query(self):
+        query = """
+        SELECT {columns}
+        FROM   program_course;
+        """.format(columns=','.join(self.COLUMN_NAMES))
+        return query
```
brianhw: And this should use CourseProgramMetadataRecord as the names of columns?

edx/analytics/tasks/insights/enrollments.py
```diff
@@ -1060,11 +1200,14 @@ def requires(self):
...
-        yield (
+        requirements = [
             CourseEnrollmentSummaryTableTask(**enrollment_kwargs),
             EnrollmentByGenderTask(**enrollment_kwargs),
             EnrollmentByBirthYearTask(**enrollment_kwargs),
             EnrollmentByEducationLevelTask(**enrollment_kwargs),
             EnrollmentDailyTask(**enrollment_kwargs),
-            ImportCourseSummaryEnrollmentsIntoMysql(**course_summary_kwargs),
-        )
+            ImportCourseSummaryEnrollmentsIntoMysql(**course_summary_kwargs)
```
brianhw: nit: We customarily leave trailing commas on list elements, so that we don't have to add it back on if we  add something on the following line.

edx/analytics/tasks/insights/enrollments.py
```diff
@@ -1060,11 +1200,14 @@ def requires(self):
...
+        ]
+        if self.enable_course_catalog:
+            requirements.append(CourseProgramMetadataInsertToMysqlTask(**course_summary_kwargs))
+        yield requirements
```
brianhw: I think you could also have made separate yield calls, no?  Just conditionally call yield on this last task if the flag is set.  But this is fine too.

brianhw (approved):
Looks good.  Just a few remaining nits.  (We can deal with splitting out code into a different file in a separate PR, if needed.)

edx/analytics/tasks/insights/enrollments.py
```diff
@@ -1036,9 +1044,137 @@ def required_table_tasks(self):
...
+    def partition_value(self):
+        return self.date.isoformat()
+
+
+class CourseProgramMetadataDataTask(CourseSummaryEnrollmentDownstreamMixin, HiveQueryTask):  # pragma: no cover
+    """Selects from `program_course` and persists results into `course_program_metadata` Hive table."""
+
+    @property
+    def insert_query(self):
+        column_names = [column[0] for column in CourseProgramMetadataRecord.get_sql_schema()]
```
mulby: nit: slightly clearer

```python
column_names = CourseProgramMetadataRecord.get_fields().keys()
```

edx/analytics/tasks/insights/enrollments.py
```diff
@@ -1060,11 +1196,13 @@ def requires(self):
...
             CourseEnrollmentSummaryTableTask(**enrollment_kwargs),
             EnrollmentByGenderTask(**enrollment_kwargs),
             EnrollmentByBirthYearTask(**enrollment_kwargs),
             EnrollmentByEducationLevelTask(**enrollment_kwargs),
             EnrollmentDailyTask(**enrollment_kwargs),
             ImportCourseSummaryEnrollmentsIntoMysql(**course_summary_kwargs),
-        )
+        ]
+        if self.enable_course_catalog:
+            yield [CourseProgramMetadataInsertToMysqlTask(**course_summary_kwargs)]
```
mulby: nit: you don't need to yield an array here, you should be able to just yield the task

mulby (approved):

PR #367, opened 2017-03-22
Truncate entity_id.

HassanJaveed84:
Should I kick off a build with this branch ?


brianhw (approved):
LGTM.  👍

brianhw:
Yes, I would kick off a build.  Is there cleanup necessary?  I assume that today's output needs to be deleted first so that this will rerun from the right point.

tobz (approved):
LGTM.

PR #368, opened 2017-03-28
Added letter_grade and percent fields.

HassanJaveed84:
AN-8644: Mapped new fields for the event_records table.
We need the elif clause as the column name 'percent' differs from the event payload attribute 'percent_grade'. 

The stuff we need to figure out is how to proceed with these columns. Do we run a alter table command on the existing event_records table(adding the new columns) ?

### DIFF
--- a/edx/analytics/tasks/warehouse/load_internal_reporting_events.py
+++ b/edx/analytics/tasks/warehouse/load_internal_reporting_events.py
@@ -33,7 +33,7 @@ from edx.analytics.tasks.util.url import ExternalURL, url_path_join
 
 log = logging.getLogger(__name__)
 
-VERSION = '0.2.2'
+VERSION = '0.2.3'
 
 EVENT_TABLE_NAME = 'event_records'
 
@@ -207,6 +207,7 @@ class EventRecord(SparseRecord):
     is_correct = StringField(length=255, nullable=True, description='')  # Boolean
     is_correct_location = StringField(length=255, nullable=True, description='')  # Boolean
     item_id = StringField(length=255, nullable=True, description='')  # integer, string
+    letter_grade = StringField(length=64, nullable=True)
     list_type = StringField(length=255, nullable=True, description='')  # string
     location = StringField(length=255, nullable=True, description='')  # library
     manually = StringField(length=255, nullable=True, description='')  # Boolean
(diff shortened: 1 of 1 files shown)


tobz (approved):
LGTM.

edx/analytics/tasks/warehouse/load_internal_reporting_events.py
```diff
@@ -328,6 +328,8 @@ class EventRecord(SparseRecord):
     view = StringField(length=255, nullable=True, description='')  # string
     vote_value = StringField(length=255, nullable=True, description='')  # string
     widget_placement = StringField(length=255, nullable=True, description='')  # string
+    letter_grade = StringField(length=64, nullable=True)
+    percent = FloatField(nullable=True)
```
brianhw: Since these are effectively being inserted into the middle of the record anyway, I would prefer that they be added in alphabetical order.   Also, with such a format change, the version number should be incremented.   

This will be the first actual use of FloatField.  Hopefully it works.  (I have encountered failures with some types when the actual data couldn't be coerced to the right value.  But hopefully this will be "clean".)  Those values that are strings but "should" be numbers are indicated as such with a comment at the end of the line.   

Also, I know that the original story mentioned "percent", but I generally believe that the fewer differences there are between the names of event key values and the names of corresponding columns, the easier it is for people to figure out what to include in their queries based on event documentation.  So I would strongly suggest that this be kept as "percent_grade" instead of mapping to "percent".  All other remappings are motivated by need.   And if we're really going to remap "percent_grade" to "percent", that should be reflected in a comment.

brianhw:
If we are going to use the existing data that is loaded into run14 on Vertica, and alter the table as it exists on Vertica, we will have to make sure that the data loaded from S3 matches.  I would suggest that we write to a different S3 location than the current run14 location, perhaps calling it run14a.  There are currently two jobs that are strung together -- one to generate the S3 data, and one to load it into Vertica.  If we manually change the table definition on Vertica, we will no longer be able to read the existing data from run14.

brianhw:
Also, my understanding of ALTER TABLE ADD COLUMN is that you cannot control where the column gets added.  So I would guess new columns would be added after the "Created" column.   Will the Vertica load from S3 then do the right thing in terms of mapping columns on disk to column values in Vertica?   I had assumed that the loading was done positionally, rather than by column name.

iloveagent57:
You can usually do `ALTER TABLE ADD COLUMN bar int AFTER foo;` to control the ordinal position of a new column.

brianhw:
That would make sense, but I didn't find that listed in the documentation:  https://my.vertica.com/docs/7.1.x/HTML/Content/Authoring/SQLReferenceManual/Statements/ALTERTABLE.htm.

edx/analytics/tasks/warehouse/load_internal_reporting_events.py
HassanJaveed84: So the FloatField introduction worked(I ran over one day worth of events and loaded into Vertica).

I initially made it 'percent_grade', but reverted to 'percent' according to the story. I'll move back to 'percent_grade', I think it makes more sense.

HassanJaveed84:
@brianhw I was planning on running a job to populate events, does the order of the new fields okay ?

brianhw (approved):
Approve this, once the version number is bumped.

mulby (approved):

PR #369, opened 2017-03-28
Dump out metrics as we send them.

tobz:
This is part of debugging metrics collection issues, so we can see what metrics we send out and figure out which ones are getting dropped.


iloveagent57 (approved):
👍

PR #370, opened 2017-03-29
Ship metrics to Graphite directly.

tobz:
Instead of first shipping metrics to statsd, and then onwards to Graphite, we should just ship them directly to Graphite.

As far as I can tell, we were seeing issues where metrics were being shipped from EMR master nodes to the Graphite machine we had running statsd, over UDP, and things were being mysteriously dropped.

We'll be shipping via TCP now, which eliminates at least one variable.  We're also dropping the baggage of statsd defaults such as namespacing metrics by type (like gauge vs counter) and automatic rate calculations, etc.

### DIFF
--- a/scripts/collect-hadoop-metrics.py
+++ b/scripts/collect-hadoop-metrics.py
@@ -257,7 +257,7 @@ if __name__ == "__main__":
     # Load our configuration.
     if len(sys.argv) < 2:
         print "[collect-hadoop-metrics] You must specify the configuration file to load!  Exiting."
-        sys.exit(1)
+        sys.exit(0)
 
     conf_file = sys.argv[1]
     config = {}
@@ -267,7 +267,7 @@ if __name__ == "__main__":
             config = yaml.load(f)
     except IOError:
         print "[collect-hadoop-metrics] Error reading configuration file or configuration does not exist!  Exiting."
-        sys.exit(1)
+        sys.exit(0)
 
     input_config = config.get('input', {})
     output_config = config.get('output', {})
@@ -280,7 +280,7 @@ if __name__ == "__main__":
         hs_address = get_local_address_on_emr()
     if hs_address is None:
         print "[collect-hadoop-metrics] No HistoryServer address specified and unable to query instance metadata!  Exiting."
-        sys.exit(1)
+        sys.exit(0)
 
     print "[collect-hadoop-metrics] Targeting HistoryServer at '{}'.".format(hs_address)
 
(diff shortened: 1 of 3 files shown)


azafty468:
Looks pretty straight forward.  Question though, are we really calling our yaml file: scripts/collect-hadoop-metrics.yaml.example ?

👍

tobz:
It's an example configuration, for people who aren't us (me, really) and are running this out in the wild. :)

azafty468:
Ah, spot on then!

azafty468 (approved):
looks good!

brianhw:
Your earlier comment about how this is being run by openedx installations was that UDP meant they shouldn't be affected.  (It would still require them to configure it, nonetheless, which seems weird).  Now that this is TCP, I'm assuming that argument no longer holds.  Is there a way to make this functionality dependent somewhere on it being explicitly configured?   Here or in run-automated-task.sh?

scripts/collect-hadoop-metrics.py
```diff
@@ -292,10 +293,9 @@ def collect_metrics(hs_address, metric_templates):
...
 
     # Ship them to the local statsd endpoint.
-    statsd_client = statsd.StatsClient(statsd_host, statsd_port, prefix=statsd_prefix)
+    emission_ts = int(time.time())
+    stats_client = graphitesend.init(graphite_server=graphite_host, graphite_port=graphite_port, prefix=graphite_prefix, system_name='')
     for (metric_name, metric_value) in metrics:
-        print "[collect-hadoop-metrics] shipping metrics {} -> {}".format(metric_name, metric_value)
-        statsd_client.gauge(metric_name, metric_value)
-        statsd_client.incr(metric_name, metric_value)
+        stats_client.send(metric_name, metric_value, emission_ts)
```
mulby: We should probably wrap all of this new stuff in a try-except so that it doesn't break for users that aren't running a graphite server.

mulby: Optional, but I think we could do something like this?

```python
stats_client.send_list(metrics)
```
It looks like graphitesend will use the current time if no timestamp is specified.

scripts/collect-hadoop-metrics.py
```diff
@@ -279,10 +280,10 @@ def collect_metrics(hs_address, metric_templates):
 
     print "[collect-hadoop-metrics] Targeting HistoryServer at '{}'.".format(hs_address)
 
-    statsd_config = output_config.get('statsd', {})
-    statsd_host = statsd_config.get('host', 'localhost')
-    statsd_port = statsd_config.get('port', 8125)
-    statsd_prefix = statsd_config.get('prefix', 'edx.analytics.emr')
+    graphite_config = output_config.get('graphite', {})
```
mulby, line 287: If this isn't defined, maybe just return without even attempting to send the metrics over?

That way open source users won't have to wait 2 seconds for the TCP connection to timeout.

mulby (commented):
Overall looks good, I would like to see the open source user case addressed before we merge, though.

tobz:
I think this sufficiently addresses the concerns now.

There will still be error messages that get spit out, but it won't exit with 1, and so things should not gum up on the script failing to run.

If the configuration file doesn't exist, it will simply stop running.

mulby (approved):
LGTM

brianhw (approved):
LGTM as well.  I personally would prefer to see the open source case handled in edx-analytics-configuration and not here.  I don't think this should be called in the first place if it's not requested, which is not an issue for this PR.

PR #371, opened 2017-03-31
Remove libatlas3gf-base on Debian.

brianhw:
It is not needed, and doesn't exist by that name on Xenial.

AWS EMR uses yum, and is not affected by this change.


mulby (approved):

tobz (approved):
LGTM.

PR #373, opened 2017-04-03
AN-8707 fix 

azafty468:
Found that the default overwrite handler was previously set to true, but was dropped from the last production release.  This re-instates the overwrite handler's value back to true


brianhw (approved):
LGTM.  👍 

We can test the branch before merging to master, just to be sure.

PR #374, opened 2017-04-05
DE-45 | Add hls video code.

iloveagent57:
https://openedx.atlassian.net/browse/DE-45

I think this is it...right?


brianhw (approved):
LGTM.   👍

PR #375, opened 2017-04-07
Remove output() from VideoUsageTableTask.

brianhw:
This method points to the usage data on disk, but the downstream tasks don't actually use this, but rather need to query it in Hive.  So the right output() is the one provided by HiveTableTask.


iloveagent57 (approved):
👍

tobz (approved):
LGTM.

azafty468 (approved):
👍 LGTM.

PR #376, opened 2017-04-09
Make changes necessary for Hive 2.1 upgrade.

brianhw:
* Backquote column names that use reserved keywords.

* Backquote table names as well.

* We use "interval" in the calendar table as the partition,
  so use date_interval for calendar partition instead.
  "Interval" is a reserved SQL word.

* Backquote `date` in explicit Hive queries.

In order for Hive 2.1 to work with mapreduce tasks that read the output of Hive union queries, the hadoop parameter "mapreduce.input.fileinputformat.input.dir.recursive" needs to be set to "true".  This is the case for ReconcileOrdersAndTransactionsTask, which reads its input from OrderTableTask.    (See http://grokbase.com/t/hive/user/156aaawwfa/empty-table-in-mr-with-union-all-created-in-tez.)

Acceptance tests pass now pass when run against emr-5.4.0 cluster.

### DIFF
--- a/config/devstack.cfg
+++ b/config/devstack.cfg
@@ -1,6 +1,6 @@
 [hive]
 release = apache
-version = 0.11
+version = 1.0
 database = default
 warehouse_path = hdfs://localhost:9000/edx-analytics-pipeline/warehouse/
 
--- a/config/test.cfg
+++ b/config/test.cfg
@@ -2,7 +2,7 @@
 
 [hive]
 release = apache
-version = 0.11
+version = 1.0
 database = default
 warehouse_path = s3://fake/warehouse/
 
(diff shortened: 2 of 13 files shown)


tobz (approved):
LGTM.

Random thought: does it behoove us to convert all column name usages to be backtick-quoted?  Set the precedent, and build up muscle memory that we should be backtick-quoting all the time.

Is that a feasible jump?  Is it technically possible without breaking things?  Should we also backtick-quote tables ans aliases as well?

brianhw:
The branch ran cleanly for acceptance tests running on 4.7.2 as well as 5.4.0.  The only thing left to check is if it would break on analyticstack, which is running Hive 0.11.   And the solution there may simply be to upgrade Hive on Analyticstack to something more respectable anyway.

mulby:
If this runs cleanly on analyticstack, I'm good with merging it!

brianhw:
This did not run on analyticstack with Hive 0.11, as expected.   So now analyticstack in master (but not Ficus) has been updated to Hive 1.0.  I have now changed Hive defaults in config files and code from 0.11 to 1.0, so that this now runs successfully on a new analyticstack.    But it looks like I need to do a rebase to deal with conflicts.

brianhw:
Rebased branch still passes on analyticstack with Hive 1.0.0.

mulby (approved):

PR #378, opened 2017-04-13
Remove libatlas3gf-base.

brianhw:
It is not needed, and doesn't exist by that name on Xenial.


brianhw:
This change is already in master, and is cherry-picked here so that the Ficus pipeline will work on Xenial.

mulby (approved):

PR #379, opened 2017-04-19
Update dependencies for Ficus release on xenial.

brianhw:
This has been tested against a Ficus box generated from https://github.com/edx/configuration/pull/3808.  In particular, all acceptance tests now pass.   

Ficus.master was originally branched from release 0.5.21 of edx-analytics-pipeline, but several patches were necessary for running on Xenial.  In particular, the cryptography library needed updating, which in turn required a pip upgrade.  The pip upgrade exposes a problem with ansible.  There were issues with tests, so all other dependency updates were pulled in for consistency.


brianhw:
@nedbat FYI.  Once merged, ficus.master should be ready for ficus.3.

tobz (approved):
LGTM so long as tests pass and we squash it. 👍

brianhw:
@tobz Actually, I wasn't planning on squashing this, since almost all the changes are cherry-picked (all-but-one).

requirements/base.txt
```diff
@@ -1,10 +1,25 @@
...
-MarkupSafe==0.23	# BSD
-PyYAML==3.10		# MIT License
-ecdsa==0.11 		# MIT
-httplib2==0.9 		# MIT
-paramiko==1.14.0 	# LGPL
-pycrypto==2.6.1 	# public domain
-ansible==1.4.5          # GPL v3 License
+# These packages are required for bootstrapping.
+
+ansible==1.4.5 --no-binary ansible          # GPL v3 License
```
nedbat, line 3: I'm curious why `--no-binary` was needed?

requirements/test.txt
```diff
@@ -1,11 +1,33 @@
-coverage==3.7.1
-ddt==1.0.0
+coverage==4.3.1
```
nedbat, line 1: 4.3.1 is an odd version to pick.  Why not 4.3.4?

tobz:
@brianhw Ah, right.  Makes sense. :)

requirements/base.txt
brianhw: See issue described in ansible/ansible#8875.  Without the flag, it would find an incomplete version of ansible, with files missing.  The error message was:    "ERROR: file is not a legal parameter in an Ansible task or handler".  This emerged first for us when we upgraded pip, which exposed the problem.

requirements/test.txt
brianhw: This is what was available in January, when we last upgraded our dependencies.  This branch involves upgrading dependencies to match what we have on master (and have thus some amount of testing).

requirements/base.txt
nedbat: Hmm, I wonder why we don't have this issue with the configuration repo?

nedbat:
:+1:

mulby (approved):

requirements/base.txt
brianhw: Speculation:  perhaps because configuration is running with a much more recent version of Ansible (e.g. 2.2.0.0 versus 1.4.5), and perhaps this issue had been addressed along the way.

PR #380, opened 2017-04-19
Spike: Task to load data into BigQuery.

HassanJaveed84:
Added Luigi task to load data from S3 into BigQuery.

- Supports Loading from the result of an input task to BigQuery
- Supports overwrite

Issues:
If you do a streaming insert into a BigQuery table(I do that for 'table_updates'), you lose the ability to run DML statements on that table(specifically DELETE statements in our case) until the streaming buffer is unset( 

@mulby @brianhw @azafty468 @tobz

### DIFF
--- a/edx/analytics/tasks/common/bigquery_load.py
+++ b/edx/analytics/tasks/common/bigquery_load.py
@@ -1,0 +1,180 @@
+import logging
+
+import luigi
+import json
+import time
+from luigi import configuration
+
+from google.cloud import bigquery
+from google.oauth2 import service_account
+from google.cloud.exceptions import NotFound
+
+from edx.analytics.tasks.util.url import ExternalURL
+from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
+
+log = logging.getLogger(__name__)
+
+
+class BigQueryTarget(luigi.Target):
+
+    def __init__(self, credentials_target, dataset_id, table, update_id):
+        self.dataset_id = dataset_id
+        self.table = table
+        self.update_id = update_id
+        with credentials_target.open('r') as credentials_file:
+            json_creds = json.load(credentials_file)
+            self.project_id = json_creds['project_id']
+            credentials = service_account.Credentials.from_service_account_info(json_creds)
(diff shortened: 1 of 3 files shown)


mulby:
@HassanJaveed84 I was hoping we could use their "load from file" HTTP based loading method like this: https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/bigquery/cloud-client/load_data_from_file.py

More info here: https://cloud.google.com/bigquery/loading-data-post-request#bigquery-import-file-python

I'm worried about us having to clean up the files on GCS after we are done loading. Not terribly difficult I suppose, but would like to avoid the complexity if possible...

HassanJaveed84:
Initially I was thinking of using the post method but avoided trying it out as there are a bunch of requirements for the file object passed. Going to try to see if it works.

mulby:
Yeah it would be awesome if we could pass it a luigi target:

```
with self.input().open('r') as input_file:
    job = table.upload_from_file(
            input_file, source_format='text/csv')
```

HassanJaveed84:
I've updated and tested it with an ExternalUrl type as the insert source task. Other input tasks should behave the same way.  
Great suggestion!

azafty468 (approved):
👍 I'm not familiar with the ins-and-outs of Big Query but the general structure looks good to me

edx/analytics/tasks/common/bigquery_load.py
```diff
@@ -0,0 +1,193 @@
...
+        client = self.output().client
+        self.create_dataset(client)
+        self.init_copy(client)
+        self.create_table(client)
+
+        dataset = client.dataset(self.dataset_id)
+        table = dataset.table(self.table, self.schema)
+
+        with self.input()['source'].open('r') as source_file:
+            job = table.upload_from_file(source_file, source_format='text/csv', field_delimiter=self.field_delimiter)
```
brianhw, line 136: Just out of curiosity, does this load our TSV data from S3?  I'm assuming we would need a way to specify the field_delimiter, to allow changing the value from a comma to a tab.

PR #381, opened 2017-04-21
Import the grades_persistentcoursegrade table

iloveagent57:
In the interest of smaller, more frequent PRs, here's the first piece of https://openedx.atlassian.net/browse/TNL-6871. The new task isn't wired up to anything yet, that will happen in the next PR for this story. Also included here is some refactoring to reduce Luigi parameter repetition between mixin classes.

### DIFF
--- a/edx/analytics/tasks/common/sqoop.py
+++ b/edx/analytics/tasks/common/sqoop.py
@@ -23,10 +23,8 @@ def load_sqoop_cmd():
     return luigi.configuration.get_config().get('sqoop', 'command', 'sqoop')
 
 
-class SqoopImportTask(OverwriteOutputMixin, luigi.hadoop.BaseHadoopJobTask):
-    """
-    An abstract task that uses Sqoop to read data out of a database and
-    writes it to a file in CSV format.
+class SqoopImportMixin(object):
+    """Mixin to expose useful parameters when importing from a database using Sqoop.
 
     In order to protect the database access credentials they are
     loaded from an external file which can be secured appropriately.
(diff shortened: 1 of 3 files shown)


mulby:
I'm still nervous about the size of this table growing significantly over time... I wonder if we couldn't come up with a simple and clean incremental solution using the events?

edx/analytics/tasks/common/sqoop.py
```diff
@@ -23,57 +23,66 @@ def load_sqoop_cmd():
...
+        "host": "db.example.com",
+        "port": "3306",
+        "username": "exampleuser",
+        "password": "example password"
+    }
     """
     destination = luigi.Parameter(
         config_path={'section': 'database-import', 'name': 'destination'},
-        description='The directory to write the output files to.',
+        description='The directory to write the output files to.'
```
brianhw: nit:  interesting that this removes a comma, when the database description below adds a comma.  I think the comma is good, e.g. when adding a new line like 'significant'.

edx/analytics/tasks/common/sqoop.py
```diff
@@ -23,57 +23,66 @@ def load_sqoop_cmd():
...
         default=False,
+        significant=False,
         description='Print more information while working.',
     )
-    table_name = luigi.Parameter(
-        description='The name of the table to import.',
-    )
     where = luigi.Parameter(
         default=None,
+        significant=False,
```
brianhw: Why is this not significant?   Adding such a clause should change the result, where num_mappers and verbose settings do not.

edx/analytics/tasks/common/sqoop.py
```diff
@@ -23,57 +23,66 @@ def load_sqoop_cmd():
...
         description='A "where" clause to be passed to Sqoop.  Note that '
         'no spaces should be embedded and special characters should '
         'be escaped.  For example:  --where "id\<50". ',
     )
+
+
+class SqoopImportTask(OverwriteOutputMixin, luigi.hadoop.BaseHadoopJobTask, SqoopImportMixin):
```
brianhw: Is there a basic ordering here?  Generally we have put the mixins before the base class.

brianhw (approved):
LGTM.   Just a few nits.

edx/analytics/tasks/common/sqoop.py
iloveagent57: I think that was copypasta.

edx/analytics/tasks/common/sqoop.py
iloveagent57: Good point.

PR #382, opened 2017-04-21
Document tasks to run to update insights

mulby:
As part of our efforts to centralize our documentation, this is a port of this wiki page: https://github.com/edx/edx-analytics-pipeline/wiki/Tasks-to-Run-to-Update-Insights

FYI - @iloveagent57 @azafty468 @tobz @HassanJaveed84

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -2,18 +2,21 @@
 
 .. edX analytics pipeline documentation master file
 
-###################
-Edx Analytics Tasks
-###################
+######################
+edX Analytics Pipeline
+######################
 
-This guide provides information about the tasks in the edX analytics pipeline.
-
-.. include:: toctree.rst
+This guide provides information about the edX data pipeline.
 
 .. toctree::
    :maxdepth: 1
 
+   running_tasks.rst
    faq.rst	     
+
+The following documentation is generated dynamically from comments in the code. There is detailed documentation for every Luigi task.
+
+.. include:: toctree.rst
 
 ##############
 Indexes
--- a/docs/source/running_tasks.rst
+++ b/docs/source/running_tasks.rst
@@ -10,9 +10,9 @@ General Notes
 
 #. These tasks are intended to be kicked off by some scheduler (Jenkins, cron etc)
 #. You can use a script to automatically deploy a cluster on EMR, run the task and then shut it down. Here is an example: `run-automated-task.sh <https://github.com/edx/edx-analytics-configuration/blob/master/automation/run-automated-task.sh>`_.
-#. Tweak ``NUM_REDUCE_TASKS`` based on the size of your cluster. If the cluster is not being used for anything else a good rule of thumb is to make ``NUM_REDUCE_TASKS`` equal the number of available reduce slots on your cluster.
-#. There are a bunch of ``s3://`` paths listed below, but those could easily be replaced with ``hdfs://`` paths.
-#. "credentials" files are json files should be stored somewhere secure and have the following format
+#. Tweak ``NUM_REDUCE_TASKS`` based on the size of your cluster. If the cluster is not being used for anything else a good rule of thumb is to make ``NUM_REDUCE_TASKS`` equal the number of available reduce slots on your cluster. See hadoop docs to determine the number of reduce slots available on your cluster.
+#. Luigi, the underlying workflow engine, has support for both S3 and HDFS when specifying input and output paths. ``s3://`` can be replaced with ``hdfs://`` in all examples below.
+#. "credentials" files are json files should be stored somewhere secure and have the following format. They are often stored in S3 or HDFS but can also be stored on the local filesystem of the machine running the data pipeline.
 
   ::
 
@@ -25,7 +25,6 @@ General Notes
           "password": "passwordforsomeuser",
         }
 
-#. I've started to put comments next to some parameters, please feel free to add in your own comments as well. Note that bash will get unhappy if you try to run these commands with those comments in place, be sure to strip them out of your commands before running. Also, ensure that there is no space following the trailing backslash on line endings.
 
 Performance (Graded and Ungraded)
 ---------------------------------
@@ -44,17 +43,32 @@ Task
 ::
 
     AnswerDistributionWorkflow --local-scheduler \
-      --src s3://path/to/tracking/logs/ \ [This should be the HDFS/S3 path to your tracking logs]
-      --dest s3://folder/where/intermediate/files/go/ \ [This can be any location in HDFS/S3 that doesn't exist yet]
-      --name unique_name \ [This can be any alphanumeric string, using the same string will attempt to use the same intermediate outputs etc]
-      --output-root s3://final/output/path/ \ [This can be any location in HDFS/S3 that doesn't exist yet]
-      --include '*tracking.log*.gz' \ [This glob pattern should match all of your tracking log files]
-      --manifest "s3://scratch/path/to/manifest.txt" \ [This can be any path in HDFS/S3 that doesn't exist yet, a file will be written here]
-      --base-input-format "org.edx.hadoop.input.ManifestTextInputFormat" \ [This is the name of the class within the jar to use to process the manifest]
-      --lib-jar "hdfs://localhost:9000/edx-analytics-pipeline/packages/edx-analytics-hadoop-util.jar" \ [This is the path to the jar containing the above class, note that it should be an HDFS/S3 path]
+      --src s3://path/to/tracking/logs/ \
+      --dest s3://folder/where/intermediate/files/go/ \
+      --name unique_name \
+      --output-root s3://final/output/path/ \
+      --include '*tracking.log*.gz' \
+      --manifest "s3://scratch/path/to/manifest.txt" \
+      --base-input-format "org.edx.hadoop.input.ManifestTextInputFormat" \
+      --lib-jar "hdfs://localhost:9000/edx-analytics-pipeline/packages/edx-analytics-hadoop-util.jar" \
       --n-reduce-tasks $NUM_REDUCE_TASKS \
-      --marker $dest/marker \ [This should be an HDFS/S3 path that doesn't exist yet. If this marker exists, the job will think it has already run.]
-      --credentials s3://secure/path/to/result_store_credentials.json [See discussion of credential files above, these should be the credentials for the result store database to write the result to]
+      --marker $dest/marker \
+      --credentials s3://secure/path/to/result_store_credentials.json
+
+Parameter Descriptions
+~~~~~~~~~~~~~~~~~~~~~~
+
+* ``--src``: This should be the HDFS/S3 path to your tracking logs.
+* ``--dest``: This can be any location in HDFS/S3 that doesn't exist yet.
+* ``--name``: This can be any alphanumeric string, using the same string will attempt to use the same intermediate outputs etc.
+* ``--output-root``: This can be any location in HDFS/S3 that doesn't exist yet.
+* ``--include``: This glob pattern should match all of your tracking log files.
+* ``--manifest``: This can be any path in HDFS/S3 that doesn't exist yet, a file will be written here.
+* ``--base-input-format``: This is the name of the class within the jar to use to process the manifest.
+* ``--lib-jar``: This is the path to the jar containing the above class, note that it should be an HDFS/S3 path.
+* ``--n-reduce-tasks``: Number of reduce tasks to schedule.
+* ``--marker``: This should be an HDFS/S3 path that doesn't exist yet. If this marker exists, the job will think it has already run.
+* ``--credentials``: See discussion of credential files above. These should be the credentials for the result store database to write the result to.
 
 Functional example:
 ~~~~~~~~~~~~~~~~~~~
@@ -75,13 +89,6 @@ Functional example:
       --marker hdfs://localhost:9000/tmp/pipeline-task-scheduler/AnswerDistributionWorkflow/1449177792/marker  \
       --credentials /edx/etc/edx-analytics-pipeline/output.json
 
-The contents of a credentials file look something like this:
-
-::
-
-    {"username": "pipeline001", "host": "localhost", "password": "password", "port": 3306}
-
-
 Enrollment
 ----------
 
@@ -90,10 +97,10 @@ Notes
 
 * Intended to run nightly.
 * This populates most of the data needed by the "Enrollment" lens in insights, including the demographic breakdowns by age, gender, and level of education.
-* This uses more up-to-date patterns.
 * Requires the following sections in config files: hive, database-export, database-import, map-reduce, event-logs, manifest, enrollments. The course-summary-enrollment and course-catalog-api sections are optional.
 * It *does not* require the "enrollment-reports" section. That section is used to generate static CSV reports.
 * The interval here, should be the beginning of time essentially. It computes enrollment by observing state changes from the beginning of time.
+* ``$FROM_DATE`` can be any string that is accepted by the unix utility ``date``. Here are a few examples: "today", "yesterday", and "2016-05-01".
 
 Task
 ~~~~
@@ -135,7 +142,6 @@ Notes
 * This is also one of our older tasks.
 * Finds the most recent event for every user and geolocates the IP address on the event.
 * This currently uses the student_courseenrollment table to figure out which users are enrolled in which courses. It should really be using the "course_enrollment" table computed by the enrollment and demographics related tasks.
-* This no longer supports a separate ``user-country-output`` parameter for intermediate data.  This is now written to a dated partition under ``(warehouse_path)/last_country_of_user/``.
 * Requires a maxmind data file (country granularity) to be uploaded to HDFS or S3 (see the ``geolocation`` section of the config file).  Getting a data file could look like this:
 
 ::
@@ -181,7 +187,7 @@ On December 5, 2016 the ``--course-country-output`` parameter was removed.  That
 History task
 ~~~~~~~~~~~~
 
-To load the historical enrollment events, you would need to first run:
+To load the historical location data, you would need to first run:
 
 ::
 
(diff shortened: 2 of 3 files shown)


docs/source/running_tasks.rst
```diff
@@ -0,0 +1,276 @@
...
+
+Tasks to Run to Update Insights
+===============================
+
+General Notes
+-------------
+
+#. These tasks are intended to be kicked off by some scheduler (Jenkins, cron etc)
+#. You can use a script to automatically deploy a cluster on EMR, run the task and then shut it down. Here is an example: `run-automated-task.sh <https://github.com/edx/edx-analytics-configuration/blob/master/automation/run-automated-task.sh>`_.
+#. Tweak ``NUM_REDUCE_TASKS`` based on the size of your cluster. If the cluster is not being used for anything else a good rule of thumb is to make ``NUM_REDUCE_TASKS`` equal the number of available reduce slots on your cluster.
```
tobz: This strikes me as a black magic-y sort of explanation, and even I'm not entirely familiar with what "reduce slots" are or how to definitively determine how many I have.

docs/source/running_tasks.rst
```diff
@@ -0,0 +1,276 @@
...
+Performance (Graded and Ungraded)
+---------------------------------
+
+Notes
+~~~~~
+
+* Intended to run nightly (or more frequently).
+* This was one of the first tasks we wrote so it uses some deprecated patterns.
+* You can tweak the event log pattern to restrict the amount of data this runs on, it will grab the most recent answer for each part of each problem for each student.
+* You can find the source for building edx-analytics-hadoop-util.jar at `https://github.com/edx/edx-analytics-hadoop-util <https://github.com/edx/edx-analytics-hadoop-util>`_.
```
tobz, line 38: Why is this JAR file relevant to this workflow?

docs/source/running_tasks.rst
```diff
@@ -0,0 +1,276 @@
...
+Tasks to Run to Update Insights
+===============================
+
+General Notes
+-------------
+
+#. These tasks are intended to be kicked off by some scheduler (Jenkins, cron etc)
+#. You can use a script to automatically deploy a cluster on EMR, run the task and then shut it down. Here is an example: `run-automated-task.sh <https://github.com/edx/edx-analytics-configuration/blob/master/automation/run-automated-task.sh>`_.
+#. Tweak ``NUM_REDUCE_TASKS`` based on the size of your cluster. If the cluster is not being used for anything else a good rule of thumb is to make ``NUM_REDUCE_TASKS`` equal the number of available reduce slots on your cluster.
+#. There are a bunch of ``s3://`` paths listed below, but those could easily be replaced with ``hdfs://`` paths.
```
tobz: Can you reword this?  Something like...

"Luigi, the underlying scheduler framework, has support for both S3 and HDFS when specifying input and output paths.  All examples where `s3://` is mentioned can be replaced with equivalent `hdfs://` versions."

docs/source/running_tasks.rst
```diff
@@ -0,0 +1,276 @@
...
+#. "credentials" files are json files should be stored somewhere secure and have the following format
```
tobz: Worth mentioning here that these can (and should?) be in S3 or HDFS via Luigi's built-in `s3://` or `hdfs://` support.

docs/source/running_tasks.rst
```diff
@@ -0,0 +1,276 @@
...
+      --output-root hdfs://localhost:9000/tmp/pipeline-task-scheduler/AnswerDistributionWorkflow/1449177792/course \
+      --include "*tracking.log*.gz" \
+      --manifest hdfs://localhost:9000/tmp/pipeline-task-scheduler/AnswerDistributionWorkflow/1449177792/manifest.txt \
+      --base-input-format "org.edx.hadoop.input.ManifestTextInputFormat"  \
+      --lib-jar hdfs://localhost:9000/edx-analytics-pipeline/site-packages/edx-analytics-hadoop-util.jar  \
+      --n-reduce-tasks 1 \
+      --marker hdfs://localhost:9000/tmp/pipeline-task-scheduler/AnswerDistributionWorkflow/1449177792/marker  \
+      --credentials /edx/etc/edx-analytics-pipeline/output.json
+
+The contents of a credentials file look something like this:
```
tobz: This is sort of a duplicate of the aforementioned note in the General Notes section.

docs/source/running_tasks.rst
```diff
@@ -0,0 +1,276 @@
...
+* Requires the following sections in config files: hive, database-export, database-import, map-reduce, event-logs, manifest, enrollments. The course-summary-enrollment and course-catalog-api sections are optional.
+* It *does not* require the "enrollment-reports" section. That section is used to generate static CSV reports.
+* The interval here, should be the beginning of time essentially. It computes enrollment by observing state changes from the beginning of time.
+
+Task
+~~~~
+
+::
+
+    ImportEnrollmentsIntoMysql --local-scheduler \
```
tobz, line 110: Give an example of what `$FROM_DATE` and `$TO_DATE` should look like?

docs/source/running_tasks.rst
```diff
@@ -0,0 +1,276 @@
...
+      --interval $(date +%Y-%m-%d -d "$FROM_DATE")-$(date +%Y-%m-%d -d "$TO_DATE") \
+      --n-reduce-tasks $NUM_REDUCE_TASKS
+
+Incremental implementation
```
tobz, line 114: Can we rewrite this section to:
- give a simpler explanation of incremental processing
- avoid the minutiae of implementation history

docs/source/running_tasks.rst
```diff
@@ -0,0 +1,276 @@
...
+     --course-country-output $INTERMEDIATE_OUTPUT_ROOT/$(date +%Y-%m-%d -d "$TO_DATE")/country_course \
+     --n-reduce-tasks $NUM_REDUCE_TASKS \
+     --overwrite
+
+On December 5, 2016 the ``--course-country-output`` parameter was removed.  That data is instead written to the warehouse_path.
+
+History task
+~~~~~~~~~~~~
+
+To load the historical enrollment events, you would need to first run:
```
tobz: Copypasta mishap: "enrollment"

tobz (changes requested):

pomegranited:
@mulby Sorry for my delayed reply! I'll submit a review this week.

docs/source/index.rst
```diff
@@ -2,19 +2,22 @@
...
-.. include:: toctree.rst
+This guide provides information about the edX data pipeline.
 
 .. toctree::
    :maxdepth: 1
 
+   running_tasks.rst
    faq.rst	     
 
+The following documentation is generated dynamically from comments in the code. There is detailed documentation for every Luigi task.
```
brianhw, line 17: nit: I don't know if it matters, but I believe that such documentation will be wrapped on display, and that we generally aim to have the source lines be less than 80 characters long.

docs/source/running_tasks.rst
```diff
@@ -0,0 +1,276 @@
...
+    lms-creds.json
+
+        {
+          "host": "your.mysql.host.com",
+          "port": "3306",
+          "username": "someuser",
+          "password": "passwordforsomeuser",
+        }
+
+#. I've started to put comments next to some parameters, please feel free to add in your own comments as well. Note that bash will get unhappy if you try to run these commands with those comments in place, be sure to strip them out of your commands before running. Also, ensure that there is no space following the trailing backslash on line endings.
```
brianhw: The first-person comment here sounds weird.   And now that this is no longer a wiki, we're not exactly expecting others to add their own comments anymore.

docs/source/running_tasks.rst
brianhw: This jar is needed for any workflow that reads event logs using a manifest file.  Which is pretty much every workflow that is used for insights.  So perhaps this should go in a more general setup section.

brianhw: Ah, but it's explicitly mentioned here because it's an explicit argument to the answer distribution workflow.  Our first insights-related workflow.

docs/source/running_tasks.rst
```diff
@@ -0,0 +1,276 @@
...
+      --output-root s3://final/output/path/ \ [This can be any location in HDFS/S3 that doesn't exist yet]
+      --include '*tracking.log*.gz' \ [This glob pattern should match all of your tracking log files]
+      --manifest "s3://scratch/path/to/manifest.txt" \ [This can be any path in HDFS/S3 that doesn't exist yet, a file will be written here]
+      --base-input-format "org.edx.hadoop.input.ManifestTextInputFormat" \ [This is the name of the class within the jar to use to process the manifest]
+      --lib-jar "hdfs://localhost:9000/edx-analytics-pipeline/packages/edx-analytics-hadoop-util.jar" \ [This is the path to the jar containing the above class, note that it should be an HDFS/S3 path]
+      --n-reduce-tasks $NUM_REDUCE_TASKS \
+      --marker $dest/marker \ [This should be an HDFS/S3 path that doesn't exist yet. If this marker exists, the job will think it has already run.]
+      --credentials s3://secure/path/to/result_store_credentials.json [See discussion of credential files above, these should be the credentials for the result store database to write the result to]
```
brianhw: nit: end with a period.  And period after "above" and before "These".

docs/source/running_tasks.rst
brianhw: Maybe make it a separate sub-section up there, and just link to it here.

docs/source/running_tasks.rst
```diff
@@ -0,0 +1,276 @@
...
+
+Enrollment
+----------
+
+Notes
+~~~~~
+
+* Intended to run nightly.
+* This populates most of the data needed by the "Enrollment" lens in insights, including the demographic breakdowns by age, gender, and level of education.
+* This uses more up-to-date patterns.
```
brianhw: ... compared to answer distribution.    (Certainly not our most up-to-date.)

docs/source/running_tasks.rst
```diff
@@ -0,0 +1,276 @@
...
+* Requires the following sections in config files: hive, database-export, database-import, map-reduce, event-logs, manifest, enrollments. The course-summary-enrollment and course-catalog-api sections are optional.
+* It *does not* require the "enrollment-reports" section. That section is used to generate static CSV reports.
```
brianhw: (and is deprecated!)

docs/source/running_tasks.rst
brianhw: In order for documentation to work for people who are using different versions of the code.  This was definitely the case when the section was written.   Perhaps now we would expect that people are on the same version, but were we to add incremental processing to something else, we should indicate what it relates to.   Dating them is one way to deal with that.    Maybe when they get old enough, they can be modified to release names (e.g. ficus), and then when old enough, they can just be left alone.

docs/source/running_tasks.rst
```diff
@@ -0,0 +1,276 @@
...
+
+Notes
+~~~~~
+
+* Intended to run nightly.
+* This populates the map view in insights.
+* This is also one of our older tasks.
+* Finds the most recent event for every user and geolocates the IP address on the event.
+* This currently uses the student_courseenrollment table to figure out which users are enrolled in which courses. It should really be using the "course_enrollment" table computed by the enrollment and demographics related tasks.
+* This no longer supports a separate ``user-country-output`` parameter for intermediate data.  This is now written to a dated partition under ``(warehouse_path)/last_country_of_user/``.
```
brianhw: Yeah, so maybe this should also be pulled.   Or say between eucalyptus and ficus.

docs/source/running_tasks.rst
```diff
@@ -0,0 +1,276 @@
...
+      --n-reduce-tasks $NUM_REDUCE_TASKS
+
+Video
+~~~~~
+
+Notes
+~~~~~
+
+* Intended to be run daily.
+* Still a work in progress - erroneous events can make videos appear to be much longer than they actually are.
```
brianhw: I find this to be rather cryptic....    what work is in progress?  what kinds of errors?     But it reminds me of the line about the side view mirror....

docs/source/running_tasks.rst
```diff
@@ -0,0 +1,276 @@
...
+* It is an incremental implementation, so it requires persistent storage of previous runs.  It also requires an initial load of historical data.
+* Requires the availability of a separate ElasticSearch instance running 1.5.2.  This is different from the version that the LMS uses, which is still on 0.90.
+
+History task
+~~~~~~~~~~~~
+
+The workflow assumes that new Hive-ready data has been written persistently to the ``module_engagement`` directory under warehouse_path by ModuleEngagementIntervalTask.  The workflow uses the ``overwrite_n_days`` to determine how many days back to repopulate this data. The idea is that before this point, events are not expected to change, but perhaps there might be new events that have arrived in the last few days.  We are currently running with a value of 3, and this can be overridden on the command-line or defined as a ``[module-engagement]`` parameter in the override.cfg file.  This means for us that only the last three days of raw events get scanned nightly.  It is assumed that the previous days' data has been loaded by previous runs, or by performing a historical load.
```
brianhw, line 252: At the moment, the code requires that overwrite_n_days be defined in [enrollments] as well, though the value is actually ignored.   (Part of the external dependency on enrollment data, as mentioned below.]

docs/source/running_tasks.rst
```diff
@@ -0,0 +1,276 @@
...
+      --overwrite-mysql
+
+Since module engagement in Insights only looks at the last two weeks of activity, you only need ``FROM_DATE`` to be two weeks ago.  The ``TO_DATE`` need only be within N days of today (as specified by ``--overwrite-n-days``).  Setting ``--overwrite-mysql`` will ensure that all the historical data is also written to the Mysql Result Store.  Using ``--overwrite-from-date`` is important when "fixing" data (for some reason): setting it earlier (i.e. to ``FROM_DATE``) will cause the Hive data to also be overwritten for those earlier days.
+
+Another prerequisite before running the module engagement workflow below is to have run enrollment first.  It is assumed that the ``course_enrollment`` directory under warehouse_path has been populated by running enrollment with a ``TO_DATE`` matching that used for the module engagement workflow (i.e. today).
+
+Task
+~~~~
+
+We run the module engagement job nightly, which adds the most recent day to this while it is overwriting the last N days (as set by the ``--overwrite-n-days`` parameter).  This calculates aggregates and loads them into ES and Mysql.
```
brianhw, line 271: I don't know where (or whether) we want to mention this, but open source users are currently trying out two approaches to dealing with an HDFS limitation.  On HDFS, it is not possible to move files to new names that already exist.  The history jobs do this.  To get around this, folks either avoid running history or they avoid ovewriting it.  That means they either manually delete previous data, and set --overwrite-n-days to a large value to regenerate it all from scratch, or they run the regular job with --overwrite-n-days set to zero.

brianhw (approved):
As good, or better, than the original.  Probably better to get it done and revise it later than to pick it apart for too long.   👍

docs/source/running_tasks.rst
pomegranited: +1

docs/source/running_tasks.rst
pomegranited: Since you're providing a working example of the commands below the annotated one, you can just move the commented parameters to a bulleted list or something, and then you could remove this comment entirely.

The inline comments are also difficult to read, since there's no syntax highlighting in the generated docs.

pomegranited (approved):
:+1: Major step forwards to making this information consolidated and visible  Thank you!

I've left some comments, but none are showstoppers.

- [x] I tested this by running `make docs` and viewing the generated files in my browser.
- [x] I read through the code
- [x] I checked for accessibility issues - Voiceover can read them :)
- [x] Includes documentation - heaps!

docs/source/running_tasks.rst
mulby: Fair point, I've added a note here letting them know that's a "hadoop" concept that they should consult Hadoop docs to get more info.

docs/source/running_tasks.rst
mulby: Good call. Done.

docs/source/running_tasks.rst
mulby: Great suggestion, done!

docs/source/running_tasks.rst
mulby: I like the idea of removing this line and moving the comments out into a bulleted list. I'm also trying to scrub out all first person from the doc. Great suggestions!

docs/source/running_tasks.rst
mulby: Done

docs/source/running_tasks.rst
mulby: Ou - good call. Just removed the duplicate...

docs/source/running_tasks.rst
mulby: Heh - good call. I just removed this comment...

docs/source/running_tasks.rst
mulby: Great idea, I added something like this to the notes.

docs/source/running_tasks.rst
mulby: @brianhw I'm hoping that since this doc is now committed with the code it can always reflect the code it is committed with. So as we make changes the docs evolve with the code. If someone wanted to know how it worked a while ago, they could simply pull the doc for the older version of the code by checking it out.

I think readthedocs actually supports this more elegantly by providing you with a menu at the bottom where you can switch between versions of the doc to see older stuff (for example: ficus or eucalyptus).

mulby: All of that said, the amount of re-writing here is beyond the scope of this ticket (simply migrating the doc from the github wiki) so I'll defer that to the next revision!

docs/source/running_tasks.rst
mulby: Deleted this.

docs/source/running_tasks.rst
mulby: Good catch, fixed!

docs/source/running_tasks.rst
mulby: Heh - removed!

PR #384, opened 2017-04-23
move the debug script to the scripts directory

mulby:
How do you guys feel about moving this script into the scripts directory in the top of the repo? That seems like the current place we are storing one-off scripts like this.

FYI - @azafty468 @iloveagent57 @tobz


tobz (approved):
I like this idea.  👍

HassanJaveed84 (approved):

PR #389, opened 2017-05-02
Strip id.

HassanJaveed84:
Some play_video events are appearing with a newline character at the start of the id field:

\"id\": \"\\n52968fb55cd14c428c58069812ca28b1\"

This breaks module-engagement. This fix just strips out any new line characters.


azafty468 (approved):
👍 Do we also want a unit test to cover this guy for fun?

mulby (approved):
(sigh)

brianhw:
LGTM.  👍 

Oh, I would add an inline comment after the .strip() to say that we have seen id values with leading newlines.

mulby:
Do we also need to make this change to our video timeline workflow?

HassanJaveed84:
Didn't realize we use the 'id' in the video timeline workflow. I'll create a PR for it.

PR #390, opened 2017-05-02
Spike - Use single mapreduce for event records export.

azafty468:
PR for code to be used to build out the event export.  This was cherry picked from @HassanJaveed84 's hassan/events-to-s3-single-mr branch, and when the code is merged into master his branch should be the one that is merged up.  This PR is here just for the review

### DIFF
--- a/edx/analytics/tasks/common/pathutil.py
+++ b/edx/analytics/tasks/common/pathutil.py
@@ -300,7 +300,10 @@ class EventLogSelectionMixin(EventLogSelectionDownstreamMixin):
         try:
             return event['time']
         except KeyError:
-            return None
+            try:
+                return event['timestamp']
+            except KeyError:
+                return None
 
     def get_map_input_file(self):
         """Get the name of the input file from Hadoop."""
--- a/edx/analytics/tasks/common/pathutil.py
+++ b/edx/analytics/tasks/common/pathutil.py
@@ -297,13 +297,16 @@ class EventLogSelectionMixin(EventLogSelectionDownstreamMixin):
 
     def get_event_time(self, event):
         """Returns time information from event if present, else returns None."""
-        try:
-            return event['time']
-        except KeyError:
-            try:
-                return event['timestamp']
-            except KeyError:
-                return None
+        if 'time' in event:
+            return event["time"]
+
+        if 'timestamp' in event:
+            return event['timestamp']
+
+        if 'requestTime' in event:
+            return event['requestTime']
+
+        return None
 
     def get_map_input_file(self):
         """Get the name of the input file from Hadoop."""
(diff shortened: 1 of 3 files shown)


brianhw:
LGTM.   I would suggest changing "Cherry Pick" in the title to something like "Spike" or "Do Not Merge", so we don't accidentally merge it.

tobz:
Yep, LGTM.  Generate dem events! :D

edx/analytics/tasks/warehouse/load_internal_reporting_events.py
```diff
@@ -958,6 +975,9 @@ def mapper(self, line):
         self.incr_counter(self.counter_category_name, 'Inputs with Dates', 1)
 
         segment_type = event.get('type')
+        if segment_type is None:
+            segment_type = event.get('action').lower()
```
brianhw: Just out of curiosity, is there any chance that 'action' won't be present if 'type' is not present?   If so, the call to .lower() will fail.

edx/analytics/tasks/warehouse/load_internal_reporting_events.py
```diff
@@ -862,7 +864,22 @@ def _get_time_from_segment_event(self, event, key):
...
+
+            if 'requestTime' in event:
+                self.incr_counter(self.counter_category_name, 'Supplementing requestTime for receivedAt', 1)
+                return self._get_time_from_segment_event(event, 'requestTime')
+
+            self.incr_counter(self.counter_category_name, 'Neither receivedAt nor requestTime present', 1)
+            log.error("Missing event arrival time in event '%r'", event)
+
+        except KeyError:
+            return None
```
brianhw: I presume this isn't going to happen (and that the try is not really necessary), since you check if the keys are present before calling _get_time_from_segment_event.  But if you do leave this here, it might be good to add a counter, to see if it's really getting called.   (Or does something else fail if no event-arrival-time is found?)

edx/analytics/tasks/common/pathutil.py
```diff
@@ -297,10 +297,18 @@ def get_event_and_date_string(self, line):
...
+
+        if 'timestamp' in event:
+            return event['timestamp']
+
+        if 'requestTime' in event:
+            return event['requestTime']
+
+        log.info("Error found no usable time value for event " + event)
+
+        return None
```
brianhw, line 311: This method is only getting called for tracking logs.   It has been overridden for both the tracking log and segment tasks, but it happens that this is called by TrackingEventRecordDataTask.get_event_emission_time() as a super() call.   But I assume this logic should really be going into the SegmentEventRecordDataTask.get_event_time() implementation.   See lines starting around 887.

brianhw (commented):
Just a few comments.

azafty468:
Er, you probably don't want to make too many comments in here.  This is a working branch I'm tweaking as I encounter issues.  And thanks!

brianhw:
Superceded by PR #436.

PR #393, opened 2017-05-04
Confirm s3 paths returned by S3 are valid in S3.

brianhw:
It turns out that they are not always valid, which later causes FileNotFoundExceptions from manifest processing of such paths.


azafty468 (approved):
👍 I'm fine with this but also let's test it to see if it adds just pushes an unbearable amount of overhead up the process to the manifest generation

brianhw:
Manifest generation takes place during scheduling, rather than during running.  It may or may not be faster, but it may fail more quickly and reliably.

PR #394, opened 2017-05-08
Add InsertToMysqlVideo*.allow_empty_insert

pomegranited:
Let `allow_empty_insert` be configurable for the `InsertToMysqlVideo*` tasks, because some deployments don't use video in their courses, or have quiet days when no one watches a video.

`allow_empty_insert` behaves like its equivalent on [`ModuleEngagementMysqlTask`](https://github.com/edx/edx-analytics-pipeline/blob/2da1413eebc6d0aa1a1fe3ba293a0e92e96405e7/edx/analytics/tasks/insights/module_engagement.py#L267):
* a boolean parameter
* default False
* configurable from the `[videos]` section in override.cfg

### DIFF
--- a/edx/analytics/tasks/insights/video.py
+++ b/edx/analytics/tasks/insights/video.py
@@ -778,6 +778,12 @@ class InsertToMysqlVideoTimelineTask(VideoTableDownstreamMixin, MysqlInsertTask)
                     'want.',
         significant=False
     )
+    allow_empty_insert = luigi.BooleanParameter(
+        default=False,
+        description='Allow the video table to be empty (e.g. if no video activity has occurred)',
+        config_path={'section': 'videos', 'name': 'allow_empty_insert'},
+        significant=False,
+    )
 
     @property
     def table(self):  # pragma: no cover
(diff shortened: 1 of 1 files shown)


brianhw:
Just curious:  these parameters are identical, and are defined on the two Mysql-related tasks.  But if one were to run video from the entry point (InsertToMysqlAllVideoTask), one cannot define this parameter.  So then the only way to enable it is through the configuration file.  Is that your intention?  Another possibility is to define a Mysql-related mixin that extends VideoTableDownstreamMixin with this parameter, and is used in the two mysql-related tasks and the entry point.

pomegranited:
Hey @brianhw!
> So then the only way to enable it is through the configuration file. Is that your intention?

Yep, it is.. I really just need to be able to set the parameter and forget about it, so config-only is fine.

But I'd totally be cool with a Mysql-related mixin that does the same thing, and could be shared by all the Mysql tasks..  though we'd still have to override the config section location, so maybe that doesn't really help much?

brianhw (approved):
I think this is fine enough as-is.

pomegranited:
Thanks, @brianhw !  Do I need a second approval from @edx/analytics before merging?

brianhw:
I don't think so, as it is a pretty small change.

PR #395, opened 2017-05-11
Remove ancient enrollment report

brianhw:
This was coded in warehouse/enrollment sub-directory.  It was
our first job.  There are better ways....


mulby (approved):
:confetti_ball:

PR #396, opened 2017-05-12
Fixed acceptance tests validation.

HassanJaveed84:


edx/analytics/tasks/tests/acceptance/fixtures/output/expected_financial_report.csv
```diff
@@ -1,14 +1,13 @@
...
-ORDER_BALANCED,NO_COST,NO_TRANSACTION,OCX,OCX-100021,15,N52HI334N5ZGIZLSL5UWI7BRGU======,2016-03-22 21:02:09,,,,,,,,,,,,,15,N52HI334NRUW4ZK7NF2GK3K7NFSHYMJV,15,0,1000,1,22,1000,16,VFF5A4MVV5KMNSQE,0,0,38,test_user,,seat,professional,course-v1:AccTest+Otto+Professional,AccTest,otto
-ORDER_BALANCED,PURCHASED_BALANCE_MATCHING,PURCHASE_ONE,OCX,OCX-100023,17,N52HI334N5ZGIZLSL5UWI7BRG4======,2016-03-22 21:09:45,2016-03-22,8XC042089G853230R,OBQXS4DBNR6HI4TBNZZWCY3UNFXW4X3JMR6DQWCDGA2DEMBYHFDTQNJTGIZTAUQ=,paypal,test,sale,instant_transfer,60,USD,2.04,60,2.04,17,N52HI334NRUW4ZK7NF2GK3K7NFSHYMJX,19,60,100,1,23,40,26,OTTO_CRED_40_PCT_OFF,0,0,38,test_user,,seat,credit,course-v1:AccTest+Otto+Credit,AccTest,otto
\ No newline at end of file
+ERROR_NO_ORDER_NONZERO_BALANCE,NO_ORDERITEM,PURCHASE,edx,EDX-1112,,,,2015-09-02,1234567890123456789014,MN4WEZLSONXXK4TDMV6HI4TBNZZWCY3UNFXW4X3JMR6DCMRTGQ2TMNZYHEYDCMRTGQ2TMNZYHEYDCNA=,cybersource,edx_org,sale,credit_card,50.00,USD,,50.00,,,,,,,,,,,,,,,,,,,,,
```
HassanJaveed84: Changed floating point numbers to include trailing 0's 100 -> 100.00

edx/analytics/tasks/tests/acceptance/test_financial_reports.py
```diff
@@ -58,27 +58,44 @@ def test_end_to_end(self):
 
         with self.vertica.cursor() as cursor:
             expected_output_csv = os.path.join(self.data_dir, 'output', 'expected_financial_report.csv')
-            expected = pandas.read_csv(expected_output_csv, parse_dates=True)
+            expected_output_data = []
```
HassanJaveed84: Construct a dataframe like we do for vertica response.

edx/analytics/tasks/tests/acceptance/__init__.py
```diff
@@ -112,6 +113,20 @@ def modify_target_for_local_server(target):
         return target
 
 
+def coerce_columns_to_string(row):
+    # Vertica response includes datatypes in some columns i-e. datetime, Decimal etc. so convert
+    # them into string before comparison with expected output.
+    return [str(x) for x in row]
+
+
+def read_fixture_as_list(fixture_file_path):
```
brianhw: nit:  might suggest "read_csv_fixture_as_list", so it's clear what kind of fixtures this applies to.

edx/analytics/tasks/tests/acceptance/test_course_catalog.py
```diff
@@ -52,14 +53,20 @@ def validate_output(self):
...
+            def convert_date(date_string):
+                """Convert date string to a date object."""
+                return datetime.datetime.strptime(date_string, '%Y-%m-%d').date()
+
+            expected = pandas.read_csv(expected_output_csv, converters={'date': convert_date})
 
             cursor.execute("SELECT * FROM {schema}.d_course_subjects;".format(schema=self.vertica.schema_name))
             database_subjects = cursor.fetchall()
             subjects = pandas.DataFrame(database_subjects, columns=['row_number', 'course_id', 'date', 'subject_uri',
                                                                     'subject_title', 'subject_language'])
```
brianhw: nit:  I liked the way you pulled columns out as a separate variable elsewhere.  Here it's not used in two places, but it might just look better anyway.

edx/analytics/tasks/tests/acceptance/test_course_catalog.py
```diff
@@ -4,10 +4,11 @@
 
 import logging
 import os
+import datetime
 
 import pandas
 
-from edx.analytics.tasks.tests.acceptance import AcceptanceTestCase, when_vertica_available
+from edx.analytics.tasks.tests.acceptance import AcceptanceTestCase, when_vertica_available, coerce_columns_to_string
```
brianhw: You add coerce_columns_to_string here, but don't end up actually using it.   Is there a reason that it didn't work?

edx/analytics/tasks/tests/acceptance/test_financial_reports.py
```diff
@@ -58,27 +61,27 @@ def test_end_to_end(self):
...
-                buf.seek(0)
-                expected.to_csv(buf)
-                print 'Expected:'
-                print buf.getvalue()
-                self.fail("Expected and returned data frames have different shapes or labels.")
+
+            f_orderitem_transactions = pandas.DataFrame(map(coerce_columns_to_string, response), columns=columns)
+
+            # Replace None with empty strings for validation
+            f_orderitem_transactions = f_orderitem_transactions.replace('None', '', regex=True)
```
brianhw: This confuses me.  Is there a reason that the 'None' strings can't just be put into the expected output?   And if we need to do this, does it make sense to just do it as part of coerce_columns_to_string() in the first place?

brianhw (commented):
An additional goal of this work is to also get tests to run on EMR 5.5.0.   At present,  TestImportPersistentCourseGradeTask (in test_database_import.py) fails on EMR 5.5 (and Hive 2.1 in general) because it does a string comparison of grade values.  It expects a grade equal to 0.7 to be output as the string '0.7', but Hive 2.1 outputs them with two decimal places of accuracy (e.g. 0.70).   I'm concerned that the general convert-to-string may run into this same problem on EMR 5.5.  So it would be good to check this branch against the existing 5.5 acceptance test server.  And if it works, to update TestImportPersistentCourseGradeTask to use this new stuff rather than the brittle string-compare it currently employs.

edx/analytics/tasks/tests/acceptance/test_financial_reports.py
HassanJaveed84: We would replace the empty values in our fixture files with None. Then, we don't need to do this step here.

edx/analytics/tasks/tests/acceptance/test_course_catalog.py
HassanJaveed84: I used coerce_columns_to_string here as well before, but realized that aren't any columns here that need to be coerced. Except the 'date' for which I use a converter when calling pandas.read_csv.

edx/analytics/tasks/tests/acceptance/test_financial_reports.py
HassanJaveed84: @brianhw do you think it's better if we just put 'None' in the expected out csv ?

brianhw: Offhand, yes.   Is there a reason not to?   I'm unclear why we would want the expected data to be different than what we expect, offhand....

edx/analytics/tasks/tests/acceptance/test_financial_reports.py
```diff
@@ -58,27 +61,27 @@ def test_end_to_end(self):
 
         with self.vertica.cursor() as cursor:
             expected_output_csv = os.path.join(self.data_dir, 'output', 'expected_financial_report.csv')
-            expected = pandas.read_csv(expected_output_csv, parse_dates=True)
+
+            expected_output_data = read_csv_fixture_as_list(expected_output_csv)
```
mulby, line 65: Out of curiosity - what was wrong with "pandas.read_csv"?

mulby (approved):
LGTM

edx/analytics/tasks/tests/acceptance/test_financial_reports.py
azafty468: I suspect their general lack of literacy and understanding of human languages...
...
...

azafty468 (approved):
👍

edx/analytics/tasks/tests/acceptance/test_financial_reports.py
HassanJaveed84: pandas.read_csv returns a frame where it stores values such as '2014-06-27 16:02:38' as pandas.tslib.Timestamp and float values from the csv as numpy.float64. This causes assert_frame_equal to fail giving a dtypes don't match error. You can pass in check_dtypes=False to assert_frame_equal, but then it starts complaining that values don't match.

HassanJaveed84: This behavior of read_csv also differs from when we construct a frame using pandas.DataFrame(). The latter returns different dtypes than read_csv..

HassanJaveed84:
I've started a build for all acceptance tests run, once it passes I will merge into master.
I plan to fix hive version issue with TestImportPersistentCourseGradeTask in a separate PR.

HassanJaveed84:
All of the acceptance tests passed on 4.7.2. 
Only one test TestImportPersistentCourseGradeTask failed on EMR 5.5.0(which was expected).

Okay to merge this in ?

mulby:
:+1:

PR #397, opened 2017-05-15
Partial incrementalization of AnswerDistribution via Hive

iloveagent57:
Does the answer distribution job using hive (so it can be configured to run incrementally). A lot of this work is from Brian.  The stuff I added relates to (a) the one acceptance test and (b) the added SQL to get the `first_response_count` column in there.  (b) could use some review here.

### DIFF
--- a/edx/analytics/tasks/insights/answer_hive.py
+++ b/edx/analytics/tasks/insights/answer_hive.py
@@ -624,24 +624,36 @@ class LatestProblemInfo(HiveAnswerTableFromQueryTask):
             ('question', 'STRING'),
             ('problem_display_name', 'STRING'),
             ('answer_uses_value_id', 'STRING'),   # change to TINYINT sometime...
-            ('time', 'STRING'),
+            ('earliest_time', 'STRING'),
+            ('latest_time', 'STRING'),
         ]
 
     @property
     def insert_query(self):
         return """
-        SELECT aat.course_id, aat.part_id, aat.problem_id, aat.question, aat.problem_display_name,
-                     aat.answer_uses_value_id, aat.time
-        FROM all_answers aat
-        INNER JOIN (
-            SELECT course_id, part_id, max(time) as latest_time
-                  FROM all_answers
-                  GROUP BY course_id, part_id
-            ) latest
-          ON (aat.course_id = latest.course_id
-              AND aat.part_id = latest.part_id
-              AND aat.time = latest.latest_time)
-        WHERE aat.should_include_answer = 'True' AND aat.uses_submission = 'True'
+        SELECT all_answers.course_id,
+               all_answers.part_id,
+               all_answers.problem_id,
+               all_answers.question,
+               all_answers.problem_display_name,
+               all_answers.answer_uses_value_id,
+               extremes.earliest_time,
+               extremes.latest_time
+        FROM   all_answers
+               INNER JOIN (
+                   SELECT   course_id,
+                            part_id,
+                            max(time) AS latest_time,
+                            min(time) AS earliest_time
+                   FROM     all_answers
+                   GROUP BY course_id, part_id
+               ) extremes
+                       ON all_answers.course_id = extremes.course_id
+                      AND all_answers.part_id = extremes.part_id
+                      AND (all_answers.time = extremes.latest_time OR
+                           all_answers.time = extremes.earliest_time)
+        WHERE  all_answers.should_include_answer = 'True'
+          AND  all_answers.uses_submission = 'True'
         """
 
     def requires(self):
@@ -669,7 +681,8 @@ class LatestAnswerInfo(HiveAnswerTableFromQueryTask):
             ('grouping_key', 'STRING'),
             ('variant', 'STRING'),
             ('is_correct', 'STRING'),
-            ('time', 'STRING'),
+            ('earliest_time', 'STRING'),
+            ('latest_time', 'STRING'),
             ('answer_value', 'STRING'),
             ('value_id', 'STRING'),
         ]
@@ -677,21 +690,37 @@ class LatestAnswerInfo(HiveAnswerTableFromQueryTask):
     @property
     def insert_query(self):
         return """
-        SELECT aat.course_id, aat.part_id, aat.grouping_key, aat.variant, aat.correct, aat.time,
-            IF(lpi.answer_uses_value_id='True' OR aat.uses_submission='True', aat.answer, aat.answer_value_id) as answer_value,
-            IF(lpi.answer_uses_value_id='True' OR aat.uses_submission='True', aat.answer_value_id, aat.answer) as value_id
-        FROM all_answers aat
-        INNER JOIN (
-            SELECT max(time) as max_time, course_id, part_id, grouping_key
-                    FROM all_answers
-                    GROUP BY course_id, part_id, grouping_key
-            ) latest
-          ON (aat.course_id = latest.course_id
-            AND aat.part_id = latest.part_id
-            AND aat.grouping_key = latest.grouping_key
-            AND aat.time = latest.max_time)
-        INNER JOIN latest_problem_info lpi
-          ON (lpi.course_id = aat.course_id AND lpi.part_id = aat.part_id)
+        SELECT all_answers.course_id,
+               all_answers.part_id,
+               all_answers.grouping_key,
+               all_answers.variant,
+               all_answers.correct,
+               extremes.earliest_time,
+               extremes.latest_time,
+               IF(latest_problem_info.answer_uses_value_id='True' OR all_answers.uses_submission='True',
+                  all_answers.answer,
+                  all_answers.answer_value_id) AS answer_value,
+               IF(latest_problem_info.answer_uses_value_id='True' OR all_answers.uses_submission='True',
+                  all_answers.answer_value_id,
+                  all_answers.answer) AS value_id
+        FROM   all_answers
+               INNER JOIN (
+                   SELECT   course_id,
+                            part_id,
+                            grouping_key,
+                            min(time) AS earliest_time,
+                            max(time) AS latest_time
+                   FROM     all_answers
+                   GROUP BY course_id, part_id, grouping_key
+               ) extremes
+                       ON all_answers.course_id = extremes.course_id
+                      AND all_answers.part_id = extremes.part_id
+                      AND all_answers.grouping_key = extremes.grouping_key
+                      AND (all_answers.time = extremes.max_time
+                           OR all_answers.time = extremes.min_time)
+               INNER JOIN latest_problem_info
+                       ON latest_problem_info.course_id = all_answers.course_id
+                      AND latest_problem_info.part_id = all_answers.part_id
         """
 
     def requires(self):
@@ -720,23 +749,35 @@ class LatestAnswers(HiveAnswerTableFromQueryTask):
             ('user_id', 'STRING'),
             ('grouping_key', 'STRING'),
             ('course_user_tags', 'STRING'),
-            ('time', 'STRING'),
+            ('earliest_time', 'STRING'),
+            ('latest_time', 'STRING'),
         ]
 
     @property
     def insert_query(self):
         return """
-        SELECT aat.course_id, aat.part_id, aat.user_id, aat.grouping_key, aat.course_user_tags, aat.time
-        FROM all_answers aat
-        INNER JOIN (
-            SELECT max(time) as max_time, course_id, part_id, user_id
-                    FROM all_answers
-                    GROUP BY course_id, part_id, user_id
-            ) latest
-        ON (aat.course_id = latest.course_id
-            AND aat.part_id = latest.part_id
-            AND aat.user_id = latest.user_id
-            AND aat.time = latest.max_time)
+        SELECT all_answers.course_id,
+               all_answers.part_id,
+               all_answers.user_id,
+               all_answers.grouping_key,
+               all_answers.course_user_tags,
+               extremes.earliest_time,
+               extremes.latest_time
+        FROM   all_answers
+               INNER JOIN (
+                   SELECT   course_id,
+                            part_id,
+                            user_id,
+                            min(time) AS earliest_time,
+                            max(time) AS latest_time
+                   FROM     all_answers
+                   GROUP BY course_id, part_id, user_id
+                   ) extremes
+               ON all_answers.course_id = extremes.course_id
+              AND all_answers.part_id = extremes.part_id
+              AND all_answers.user_id = extremes.user_id
+              AND (all_answers.time = extremes.earliest_time OR
+                   all_answers.time = extremes.latest_time)
         """
 
     def requires(self):
@@ -765,7 +806,8 @@ class LatestAnswerDist(HiveAnswerTableFromQueryTask):
             ('is_correct', 'STRING'),
             ('answer_value', 'STRING'),
             ('value_id', 'STRING'),
-            ('count', 'INT'),
+            ('first_response_count', 'INT'),
+            ('last_response_count', 'INT'),
             ('problem_id', 'STRING'),
             ('question', 'STRING'),
             ('display_name', 'STRING'),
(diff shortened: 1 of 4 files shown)


iloveagent57:
cc: @mulby @azafty468 @HassanJaveed84 @tobz

edx/analytics/tasks/tests/acceptance/test_answer_hive.py
```diff
@@ -0,0 +1,109 @@
...
+    """Acceptance tests for Hive Answer Distribution Tasks -> MySQL"""
+
+    def test_answer_distribution_mysql(self):
+        self.task.launch([
+            'AnswerDistributionFromHiveToMySQLTaskWorkflow',
+            '--source', self.test_src,
+            '--warehouse-path', url_path_join(self.test_root, 'dst'),
+            '--lib-jar', self.oddjob_jar,
+            '--n-reduce-tasks', str(self.NUM_REDUCERS),
+            '--credentials', self.export_db.credentials_file_url,
```
brianhw: Most of these don't need to be set, since default values are already provided by the config file constructed for acceptance tests in setup().  In particular, we shouldn't need to specify lib-jar, source, and warehouse-path on the command line.  But I think that credentials would also be found there correctly.

edx/analytics/tasks/tests/acceptance/test_answer_hive.py
```diff
@@ -0,0 +1,109 @@
...
+    INPUT_FORMAT = 'org.edx.hadoop.input.ManifestTextInputFormat'
+    NUM_REDUCERS = 1
+
+    def setUp(self):
+        super(BaseAnswerDistributionAcceptanceTest, self).setUp()
+
+        assert 'oddjob_jar' in self.config
+
+        self.oddjob_jar = self.config['oddjob_jar']
+        self.input_format = self.config.get('manifest_input_format', self.INPUT_FORMAT)
```
brianhw: These don't need to be here.

edx/analytics/tasks/tests/acceptance/test_answer_hive.py
```diff
@@ -0,0 +1,109 @@
...
+    """Base class for setting up answer dist acceptance tests"""
+
+    INPUT_FILE = 'answer_dist_acceptance_tracking.log'
+    INPUT_FORMAT = 'org.edx.hadoop.input.ManifestTextInputFormat'
+    NUM_REDUCERS = 1
+
+    def setUp(self):
+        super(BaseAnswerDistributionAcceptanceTest, self).setUp()
+
+        assert 'oddjob_jar' in self.config
```
brianhw: this is already in AcceptanceTestCase.setup()

edx/analytics/tasks/tests/acceptance/test_answer_hive.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class AnswerDistOneFilePerCourseHiveAcceptanceTest(BaseAnswerDistributionAcceptanceTest):
+    """Acceptance test for the CSV-generating Answer Distribution Task"""
+
+    def test_answer_distribution(self):
+        self.task.launch([
+            'AnswerDistOneFilePerCourseTask',
+            '--source', self.test_src,
+            '--warehouse-path', url_path_join(self.test_root, 'dst'),
```
brianhw: not needed:  this defaults already to test_root, 'warehouse'.   Also, source not needed -- already set in setup.

edx/analytics/tasks/tests/acceptance/test_answer_hive.py
```diff
@@ -0,0 +1,109 @@
...
+             'choice_2', '0', '1',
+             'block-v1:edX+DemoX+Test_2014+type@problem+block@932e6f2ce8274072a355a94560216d1a',
+             'Perchance to Dream'],
+            ['course-v1:edX+DemoX+Test_2014', '9cee77a606ea4c1aa5440e0ea5d0f618_2_1', 'False',
+             '[late penalties|instructor forgiveness]', '[choice_0|choice_1]', '1', '1',
+             'block-v1:edX+DemoX+Test_2014+type@problem+block@9cee77a606ea4c1aa5440e0ea5d0f618',
+             'Interactive Questions'],
+        ]
+
+        output_target = self.get_targets_from_remote_path(url_path_join(self.test_root, 'dst', 'latest_answer_dist'))[0]
```
brianhw: can just use self.warehouse_path

edx/analytics/tasks/tests/acceptance/test_answer_hive.py
```diff
@@ -0,0 +1,109 @@
...
+            ['course-v1:edX+DemoX+Test_2014', '932e6f2ce8274072a355a94560216d1a_2_1', 'True',
+             'There is an implication that the strangeness to follow can be considered like a dream.',
+             'choice_2', '0', '1',
+             'block-v1:edX+DemoX+Test_2014+type@problem+block@932e6f2ce8274072a355a94560216d1a',
+             'Perchance to Dream'],
+            ['course-v1:edX+DemoX+Test_2014', '9cee77a606ea4c1aa5440e0ea5d0f618_2_1', 'False',
+             '[late penalties|instructor forgiveness]', '[choice_0|choice_1]', '1', '1',
+             'block-v1:edX+DemoX+Test_2014+type@problem+block@9cee77a606ea4c1aa5440e0ea5d0f618',
+             'Interactive Questions'],
+        ]
```
brianhw: One thing to be careful of here is that floats are handled differently in output between Hive 1.0 and Hive 2.1.  This caused a problem with grades, where 0.7 != 0.70.  Looks like there are no floats this time, but it's an issue with string comparisons.

edx/analytics/tasks/insights/answer_hive.py
```diff
@@ -0,0 +1,1164 @@
...
+from edx.analytics.tasks.common.mysql_load import MysqlInsertTask, MysqlInsertTaskMixin
+from edx.analytics.tasks.insights.answer_dist import get_text_from_html, try_str_to_float
+import edx.analytics.tasks.util.eventlog as eventlog
+import edx.analytics.tasks.util.opaque_key_util as opaque_key_util
+from edx.analytics.tasks.util.url import get_target_from_url, url_path_join
+
+# TODO: move MultipartitionHiveTableTask to util.hive, and clean these out...
+import textwrap
+from luigi.hive import HiveTableTarget, HivePartitionTarget
+# from edx.analytics.tasks.util.datetime_util import all_dates_between
```
brianhw: If this is commented out, should it also be removed from datetime_util?   Or is it used elsewhere?

edx/analytics/tasks/insights/answer_hive.py
```diff
@@ -0,0 +1,1164 @@
...
+
+
+class AllProblemCheckEventsParamMixin(EventLogSelectionDownstreamMixin, MapReduceJobTaskMixin):
+    """Parameters for AllProblemCheckEventsTask."""
+
+
+class AllProblemCheckEventsTask(AllProblemCheckEventsParamMixin, EventLogSelectionMixin, MultiOutputMapReduceJobTask):
+    """Identifies last problem_check event for a user on a problem in a course, given raw event log input."""
+
+    # TODO: set this up to use warehouse_path instead...
```
brianhw, line 50: I'm assuming that most TODOs are still worth doing, but only once the code is working well enough to be worth the cleanup effort.

edx/analytics/tasks/insights/answer_hive.py
```diff
@@ -0,0 +1,1164 @@
...
+        event_source = event.get('event_source')
+        if event_source is None:
+            log.error("encountered event with no event_source: %s", event)
+            return None
+        if event_source != 'server':
+            return None
+
+        # Get the "problem data".  This is the event data, the context, and anything else that would
+        # be useful further downstream.
+        # TODO: make the timestamp logic more efficient than this (i.e. avoid the parse).
```
brianhw: doesn't seem like this TODO is in the right place.  Unless the timestamp stuff has already been addressed here.

edx/analytics/tasks/insights/answer_hive.py
```diff
@@ -0,0 +1,1164 @@
...
+# class HivePartitionInIntervalTarget(HivePartitionTarget):
+#     """
+#     This should see if all requested partitions in a specified
+#     interval are present.
+
+#     This is a short-cut for creating a separate list of HivePartitionTargets.
+#     """
+#     def __init__(self, table, partition, database='default', interval, fail_missing_table=False, client=default_client):
+#         super(HivePartitionInIntervalTarget).__init__(self, table, partition, database=database, fail_missing_table=fail_missing_table, client=client)
+#         self.interval = interval
```
brianhw: Pretty cryptic.  I'm assuming this is a failed experiment?

edx/analytics/tasks/insights/answer_hive.py
```diff
@@ -0,0 +1,1164 @@
...
+        # TODO: this breaks when the starting date of the interval doesn't actually have data.
+        #   This is requiring that the date of the first actual data be known,
+        #   rather than putting in something safe (e.g. 2012-01-01 or even 2013-10-01).
+        # starting_partition = HivePartition('dt', self.interval.date_a.isoformat()) # pylint: disable=no-member
+        ending_date = self.interval.date_b - datetime.timedelta(days=1)  # pylint: disable=no-member
+        ending_partition = HivePartition('dt', ending_date.isoformat())
+
+        return [
+            # HivePartitionTarget(self.table, starting_partition.as_dict(), database=hive_database_name()),
+            HivePartitionTarget(self.table, ending_partition.as_dict(), database=hive_database_name()),
```
brianhw, line 580: I'm hoping to get a first pass at a review done sooner, but I'm stumped by this.

edx/analytics/tasks/insights/answer_hive.py
```diff
@@ -0,0 +1,1164 @@
...
+        """.format(
+            table=self.table,
+            partition=self.partition,
+            insert_query=self.insert_query.strip(),  # pylint: disable=no-member
+        )
+
+        return create_table_statements + textwrap.dedent(full_insert_query)
+
+    def output(self):
+        return get_target_from_url(self.partition_location)
```
brianhw, line 609: I'm going to bet that all this Hive stuff is going to trip over having outputs that find files on disk when in fact we want data loaded in Hive.  So the dependencies are probably all wrong.

edx/analytics/tasks/insights/answer_hive.py
```diff
@@ -0,0 +1,1164 @@
...
+    @property
+    def insert_query(self):
+        return """
+        SELECT course_id,
+               part_id,
+               problem_id,
+               question,
+               problem_display_name,
+               answer_uses_value_id,
+               MAX(earliest_time) AS earliest_time,
```
brianhw: Is this MAX or MIN?

edx/analytics/tasks/insights/answer_hive.py
```diff
@@ -0,0 +1,1164 @@
...
+
+    @property
+    def insert_query(self):
+        return """
+        SELECT course_id,
+               part_id,
+               grouping_key,
+               variant,
+               correct,
+               max(earliest_time) AS earliest_time,
```
brianhw: max or min?

edx/analytics/tasks/insights/answer_hive.py
```diff
@@ -0,0 +1,1164 @@
...
+                       FROM     all_answers
+                       GROUP BY course_id, part_id
+                   ) latest_problem
+                          ON all_answers.course_id = latest_problem.course_id
+                         AND all_answers.part_id = latest_problem.part_id
+                         AND all_answers.time = latest_problem.time
+            WHERE  all_answers.should_include_answer = 'True'
+              AND  all_answers.uses_submission = 'True'
+        ) temp
+        GROUP BY course_id, part_id, problem_id, question, problem_display_name, answer_uses_value_id
```
brianhw: Ugh!   Who came up with this SQL formatting?   It's *so* unreadable.....

brianhw (commented):
I've only made a cursory pass through the actual code, and haven't really dug into the dependencies.   But that can also be cleaned up after we have more confidence that things work.

edx/analytics/tasks/insights/answer_hive.py
iloveagent57: yeah, this confused me when I first looked at. I'll kill the TODO.

edx/analytics/tasks/insights/answer_hive.py
iloveagent57: It's MAX to eliminate the NULLs that occur from left-joining in the `temp` relation.

edx/analytics/tasks/insights/answer_hive.py
iloveagent57: Can't tell if trolling or actual complaint.

edx/analytics/tasks/insights/answer_hive.py
iloveagent57: MAX for same reason as above.

edx/analytics/tasks/tests/acceptance/test_answer_hive.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+log = logging.getLogger(__name__)
+
+
+class BaseAnswerDistributionAcceptanceTest(AcceptanceTestCase):
+    """Base class for setting up answer dist acceptance tests"""
+
+    INPUT_FILE = 'answer_dist_acceptance_tracking.log'
+    INPUT_FORMAT = 'org.edx.hadoop.input.ManifestTextInputFormat'
```
brianhw, line 21: nit: guessing this is no longer needed?

edx/analytics/tasks/insights/answer_hive.py
```diff
@@ -0,0 +1,1150 @@
...
+    def extra_modules(self):
+        import six
+        return [html5lib, six]
+
+    def output_path_for_key(self, key):
+        """
+        Creates separate folders by date, so that processing can be performed
+        incrementally.
+        """
+        date_string, course_id = key
```
brianhw, line 146: Note that the splitting by course_id is not actually meaningful in any of the downstream processing.  I think I did it in part as a debugging aide.  If I found an issue in comparing something in a course on a date, I could just look at the problem_check events for that course on that date.   I have no idea what impact this has on Hive query performance.

edx/analytics/tasks/insights/answer_hive.py
```diff
@@ -0,0 +1,1150 @@
...
+        result = super(AllProblemCheckEventsTask, self).run()
+
+        # This makes sure that an output file exists for each date in the interval
+        # as downstream tasks require that they exist.
+        for date in self.interval:
+            date_directory = 'dt={}'.format(date.isoformat())
+            url = url_path_join(self.output_root, date_directory)
+            target = get_target_from_url(url)
+            if not target.exists():
+                target.open("w").close()  # touch the file
```
brianhw, line 441: If it proves to be too onerous to check all intermediate targets, we could return it to the way it was, and only check that the first and last dates of the interval exist.  And then this here would only need to make sure that the endpoints exist, rather than all dates in between.  Trade off resilience versus performance.

edx/analytics/tasks/insights/answer_hive.py
brianhw: Okay, I get this a bit more.  This is actually testing that the partition exists in Hive, and is presumably only used for implementing complete().  If we made sure that the upstream task created the starting partition, then it would be a better approximation than this.   But another issue with the load-everything approach this takes is that it loads whatever data is present at the data location, regardless of what the interval is.  That is, extra data outside the interval also gets loaded into Hive.  It may not matter, but the convenience has a cost.   Maybe if it then dropped partitions that weren't in the interval.  Or specfiically added every interval rather than using 'recover'.

edx/analytics/tasks/insights/answer_hive.py
brianhw: Yeah, I'm pretty sure this is wrong.  This was all written before we figured out some of the Hive dependency stuff, and I wasn't really striving to solve it until the results better matched anyway.  I still think that's the right order.

edx/analytics/tasks/insights/answer_hive.py
brianhw: It would be good to add that as a comment somewhere, since it's not at all obvious.

edx/analytics/tasks/insights/answer_hive.py
brianhw: Well, both, to be honest.   I really don't like the right-aligned keywords mixed with left-aligned keywords and column-aligned expressions, and have a hard time reading this.  I liked the way the code was formatted before, mostly (though I didn't actually originate the style).  The trolling tone is probably my reaction to your original explanation for why you decided to change it in other code you were modifying.   But this stuff is religious, and ultimately you're working in this code more than I will be.   My actual complaint, though, is that I don't like having multiple styles in the code base.  Perhaps you can propose a style guide for what you like, get folks to agree, and then switch the code over to use it.

brianhw (commented):
Nothing jumps out at me as being needed to be addressed before validation.  But the dependency stuff will need to be revisited after it works, to make it more resilient.

edx/analytics/tasks/insights/answer_hive.py
iloveagent57: Good point.

edx/analytics/tasks/insights/answer_hive.py
iloveagent57: Hah, "both" is good. +1 to a single style, we can probably comprise on the things we find most palatable.

mulby: My proposal would be to adopt the BI team's style, since they work in SQL pretty much 24/7.

Personally, I don't like parts of it, but I adapt for the sake of consistency: https://github.com/edx/edx-analytics-pipeline-scripts/blob/master/sql_style.md

iloveagent57: We adopted a lot of this at my last gig and liked it: http://www.sqlstyle.guide/
I'm all for a single style, and can probably stomach most of what's in our own sql style guide, with one exception: JOINs ain't their own section, they are a keyword whose result is a relation.  The top-level grammar of a SQL query is basically this:
```sql
SELECT
  columns
FROM
  relation
WHERE
  filter
GROUP BY aggregation
ORDER BY ordering
```

JOINs belong inside the relation part above.

PR #398, opened 2017-05-17
Add support for templating to Luigi configuration files.

tobz:
This approach is intended to utilize the fact that Luigi's configuration code is based on Python's `ConfigParser`, which has native support for templated variables.

The catch is that the variables must be present in the config already, and we intended to load multiple files, in order, to make these "base" variables available to the templatized configurations loaded after.

This PR adds support to the remote and local launchers, as well as changes the small bootstrap task run through Ansible, to allow passing in multiple configuration files from the secure configuration repository to be loaded.

### DIFF
--- a/edx/analytics/tasks/launchers/__init__.py
+++ b/edx/analytics/tasks/launchers/__init__.py
@@ -1,0 +1,9 @@
+import os
+
+REMOTE_DATA_DIR = '/var/lib/analytics-tasks'
+REMOTE_LOG_DIR = '/var/log/analytics-tasks'
+
+REMOTE_CONFIG_DIR_BASE = 'config'
+
+# The 'automation' bit is hard-coded. *shrug*
+REMOTE_CONFIG_DIR = os.path.join(REMOTE_DATA_DIR, 'automation', REMOTE_CONFIG_DIR_BASE)
--- a/edx/analytics/tasks/launchers/__init__.py
+++ b/edx/analytics/tasks/launchers/__init__.py
@@ -1,9 +1,0 @@
-import os
-
-REMOTE_DATA_DIR = '/var/lib/analytics-tasks'
-REMOTE_LOG_DIR = '/var/log/analytics-tasks'
-
-REMOTE_CONFIG_DIR_BASE = 'config'
-
-# The 'automation' bit is hard-coded. *shrug*
-REMOTE_CONFIG_DIR = os.path.join(REMOTE_DATA_DIR, 'automation', REMOTE_CONFIG_DIR_BASE)
--- a/edx/analytics/tasks/launchers/local.py
+++ b/edx/analytics/tasks/launchers/local.py
@@ -7,6 +7,7 @@ Main method for running tasks on a local machine.
 
 """
 
+import argparse
 from contextlib import contextmanager
 import logging
 import os
@@ -39,6 +40,10 @@ OVERRIDE_CONFIGURATION_FILE = 'override.cfg'
 
 
 def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--additional-config', help='additional configuration file to be loaded after default/override', default=None, action='append')
+    arguments, _extra_args = parser.parse_known_args()
+
     # In order to see errors during extension loading, you can uncomment the next line.
     logging.basicConfig(level=logging.DEBUG)
 
@@ -46,12 +51,23 @@ def main():
     # TODO: launch tasks by their entry_point name
     stevedore.ExtensionManager('edx.analytics.tasks')
 
+    # Load the override configuration if it's specified/exists.
     configuration = luigi.configuration.get_config()
     if os.path.exists(OVERRIDE_CONFIGURATION_FILE):
-        log.debug('Using %s', OVERRIDE_CONFIGURATION_FILE)
+        log.debug('Loading override configuration \'%s\'...', OVERRIDE_CONFIGURATION_FILE)
         configuration.add_config_path(OVERRIDE_CONFIGURATION_FILE)
     else:
-        log.debug('Configuration file %s does not exist', OVERRIDE_CONFIGURATION_FILE)
+        log.debug('Configuration file \'%s\' does not exist!', OVERRIDE_CONFIGURATION_FILE)
+
+    # Load any additional configuration files passed in.
+    if arguments.additional_config is not None:
+        for additional_config in arguments.additional_config:
+            if os.path.exists(additional_config):
+                log.debug('Loading additional configuration file \'%s\'...', additional_config)
+                configuration.add_config_path(additional_config)
+            else:
+                log.debug('Configuration file \'%s\' does not exist!', additional_config)
+
 
     # Tell luigi what dependencies to pass to the Hadoop nodes
     # - edx.analytics.tasks is used to load the pipeline code, since we cannot trust all will be loaded automatically.
--- a/edx/analytics/tasks/launchers/remote.py
+++ b/edx/analytics/tasks/launchers/remote.py
@@ -10,12 +10,10 @@ import sys
 from urlparse import urlparse, parse_qsl
 import uuid
 
+from . import *
 
 STATIC_FILES_PATH = os.path.join(sys.prefix, 'share', 'edx.analytics.tasks')
 EC2_INVENTORY_PATH = os.path.join(STATIC_FILES_PATH, 'ec2.py')
-
-REMOTE_DATA_DIR = '/var/lib/analytics-tasks'
-REMOTE_LOG_DIR = '/var/log/analytics-tasks'
 ANSIBLE_MAX_RETRY = 3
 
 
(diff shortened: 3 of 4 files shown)


edx/analytics/tasks/launchers/remote.py
```diff
@@ -35,7 +35,7 @@ def main():
     parser.add_argument('--user', help='remote user name to connect as', default=None)
     parser.add_argument('--private-key', help='a private key file to use to connect to the host', default=None)
     parser.add_argument('--override-config', help='config file to use to run the job', default=None)
-    parser.add_argument('--secure-config', help='config file in secure config repo to use to run the job', default=None)
+    parser.add_argument('--secure-config', help='config file in secure config repo to use to run the job', default=None, action='append')
```
tobz, line 39: This turns it into a multiple-use argument.

edx/analytics/tasks/launchers/remote.py
```diff
@@ -49,6 +49,12 @@ def main():
     arguments, extra_args = parser.parse_known_args()
     arguments.launch_task_arguments = extra_args
 
+    # Push in any secure config values that we got.
+    if arguments.secure_config:
+      for x in arguments.secure_config:
+        arguments.launch_task_arguments.insert(0, x)
+        arguments.launch_task_arguments.insert(0, '--extra-config')
+
```
tobz: We have to shove this in here because we treat the unknown arguments as the only ones we pass to the local scheduler call, so all the `--secure-config` stuff gets gobbled up, and we spit it back out as the more agnostic `--extra-config`.

edx/analytics/tasks/launchers/remote.py
```diff
@@ -188,7 +194,7 @@ def convert_args_to_extra_vars(arguments, uid):
     if arguments.override_config:
         extra_vars['override_config'] = arguments.override_config
     elif arguments.secure_config and 'secure' in repos:
-        extra_vars['secure_config'] = os.path.join(repos['secure']['dir_name'], arguments.secure_config)
+        extra_vars['secure_config'] = [os.path.join(repos['secure']['dir_name'], x) for x in arguments.secure_config]
```
tobz: This is a list now, so we treat it as such.

edx/analytics/tasks/launchers/local.py
```diff
@@ -39,20 +40,31 @@
...
         configuration.add_config_path(OVERRIDE_CONFIGURATION_FILE)
     else:
         log.debug('Configuration file %s does not exist', OVERRIDE_CONFIGURATION_FILE)
 
+    # Load any extra configuration files passed in.
+    if arguments.extra_config is not None:
+        for extra_config in arguments.extra_config:
+            log.debug('Loading extra configuration %s', extra_config)
+            configuration.add_config_path(extra_config)
+
```
tobz: Just load any extra configurations right on top.

edx/analytics/tasks/launchers/local.py
```diff
@@ -39,20 +40,31 @@
 
 
 def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--extra-config', help='extra configuration file to be loaded after default/override', default=None, action='append')
+    arguments, _extra_args = parser.parse_known_args()
```
mulby: I'm assuming Luigi ignores these arguments when it goes to parse the args?

mulby (approved):

edx/analytics/tasks/launchers/local.py
tobz: It seems to ignore them, yeah.

edx/analytics/tasks/launchers/local.py
brianhw: I assume that these preserve the ordering from left to right on the command-line.  It might be good to mention that in a comment, if so.

I'm wondering if it makes sense to skip reading from OVERRIDE_CONFIGURATION_FILE if extra_config is specified.  That way one gets what one configures, without having to figure out how to also unconfigure something.    I guess that would mean that it's no longer "extra_config" but just config.

edx/analytics/tasks/launchers/remote.py
```diff
@@ -49,6 +49,12 @@ def main():
     arguments, extra_args = parser.parse_known_args()
     arguments.launch_task_arguments = extra_args
 
+    # Push in any secure config values that we got.
+    if arguments.secure_config:
+      for x in arguments.secure_config:
```
brianhw: nit:  please use something more readable than x.

edx/analytics/tasks/launchers/remote.py
brianhw: Can these just be appended to the end, instead of being inserted at the beginning?

share/task.yml
```diff
@@ -185,8 +185,9 @@
       file: path={{ working_repo_dir }}/override.cfg state=absent
 
     - name: secure config installed
-      command: cp {{ working_dir }}/{{ secure_config }} {{ working_repo_dir }}/override.cfg
+      command: cp {{ working_dir }}/{{ item }} {{ working_repo_dir }}/
```
brianhw: nit:  just wondering why we use a command here instead of a copy, as used in the next directive.  And whether we should be setting mode on this too.

brianhw (commented):
Looks good!  Looks like a very clean and straight-forward approach, much better than jinja, assuming it's expressive enough.

azafty468 (approved):
👍

edx/analytics/tasks/launchers/__init__.py
```diff
@@ -0,0 +1,9 @@
+import os
+
+REMOTE_DATA_DIR = '/var/lib/analytics-tasks'
+REMOTE_LOG_DIR = '/var/log/analytics-tasks'
+
+REMOTE_CONFIG_DIR_BASE = 'config'
+
+# The 'automation' bit is hard-coded. *shrug*
+REMOTE_CONFIG_DIR = os.path.join(REMOTE_DATA_DIR, 'automation', REMOTE_CONFIG_DIR_BASE)
```
brianhw: What impact will this have on others?   It's one thing to introduce this hard-coded value into edx-analytics-configuration, which fewer people are using, and another to do so here.  Will this work on an analyticstack, for example?   Mine has this as /var/lib/analytics-tasks/<hash>/repo/config, if I'm understanding correctly.

brianhw (commented):
Does it work on analyticstack?   Running acceptance tests there?

edx/analytics/tasks/launchers/__init__.py
tobz: The only changes here were to make `'config'` in `remote.py` a constant (`REMOTE_CONFIG_DIR_BASE`), and then to build the full path to where that will actually be on-disk into `REMOTE_CONFIG_DIR`, which is brand new to this code.

So, I don't see how this would impact anything, as it stands.

tobz: And, FWIW, look at the removals in `remote.py`.  We already used these fully hard-coded paths, and have for a while.  This is just moving them into `__init__.py` and adding a new one.

brianhw: My suggestion was that you verify running this on analyticstack.  From what I read of the code, it does the following in run_task_playbook:
{code}
data_dir = os.path.join(REMOTE_DATA_DIR, uid)
command = 'cd {data_dir}/repo ....'
{code}
which does not match your hard-coded value of REMOTE_CONFIG_DIR.  But if you tell me that it works on analyticstack, and that acceptance tests also continue to work, that would be important to know.

brianhw: Also, is there a reason to pull these out into a separate file?  I'm not seeing these used anywhere else other than in remote.py.

brianhw (commented):
Please verify that this also works on analyticstack, and that it doesn't assume that edx-analytics-configuration is being used.

edx/analytics/tasks/launchers/remote.py
tobz: This change is old.  I ended up appending them because it didn't work the other way.

edx/analytics/tasks/launchers/__init__.py
tobz: Yeah, I could probably move them back.  I moved them instinctively, thinking I would inevitably have to reference them in `local.py`, but that wasn't the case.

In terms of the code you pasted, we always hard-code the `--remote-name` as `automation` when we run tasks normally, which is why I also have it hard-coded.  I could fix this up better in the code to consider `uid` properly.

I don't know if `uid` become an actual UUID (i.e. nothing specified for `--remote-name`) when running something on analyticsstack.  I'll take a look at that.

tobz:
So I ran the acceptance tests on a clean analyticstack with this code and it worked.  Are we willing to consider that good enough?

edx/analytics/tasks/launchers/remote.py
```diff
@@ -10,14 +10,17 @@
 from urlparse import urlparse, parse_qsl
 import uuid
 
+from . import *
```
brianhw: nit:  don't need this anymore, I assume.

brianhw (approved):
LGTM.    👍

edx/analytics/tasks/launchers/remote.py
tobz: Good catch.

PR #400, opened 2017-05-30
Add overwrite flag to LastCountryOfUserTableTask.

brianhw:
When testing ficus.master, it was found that running the location task was resulting in multiple entries in the resultstore.  The last_country_of_user table is supposed to only have one partition of data loaded into it, but when run with overwrite flag set, the table's contents weren't being deleted.  This resulted in a left join (against student_courseenrollment) would have multiple entries per student enrollment.  

Fix is to drop and recreate the table each time, if the overwrite flag is set.


brianhw:
This fix was tested in production by Pierre Mailhot, who also encountered the problem.

azafty468:
Question: are we going to have issues with open connections to last_country_of_user when the table is dropped & recreated?

brianhw:
It's the Hive table that is being dropped and recreated by this flag.

azafty468 (approved):
👍

PR #401, opened 2017-05-30
Handle problem-checks with bad answer_ids.

brianhw:
Rather than trying to output non-ascii answer id's, just screen out those that are not ascii.

### DIFF
--- a/edx/analytics/tasks/insights/answer_dist.py
+++ b/edx/analytics/tasks/insights/answer_dist.py
@@ -1031,3 +1031,8 @@ def _check_answer_ids(answer_dict):
     for answer_id in answer_dict:
         if '\n' in answer_id or '\t' in answer_id:
             raise ValueError('Malformed answer_id')
+
+        try:
+            answer_id.encode('ascii')
+        except UnicodeEncodeError:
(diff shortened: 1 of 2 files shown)


brianhw:
This seems to have fixed a problem introduced by an event on 5/24/2017 that caused answer-distribution to fail for several days.

azafty468 (approved):
Looks good!  Do we have any internationalized databases that would potentially be using unicode for their answer_ids?  Or are these all plainly ascii?

brianhw:
A fair question.  In theory, it used to be possible for problem_id values to have more arbitrary names, basically selected by the course author.  Later, these moved to being hash values created by Studio (or the underlying "content store").   Part of the reason I went the way I did with this fix was that the code has not been handling non-ascii characters in the time since the code was originally written, so I have reasonable confidence that there's no historical data with this issue that has been getting through.  We have apparently had examples of bogus answer_id values in the past, however, as there is a check performed to look for embedded tabs and newlines.   I just extended this check to confirm the data is ascii.   Were we to get a report from the open-source community complaining of such data in their systems, we could revisit this.   But it might be better to put that work into the new answer-dist.

mulby (approved):

PR #403, opened 2017-05-31
Added tasks to load warehouse data into BigQuery.

HassanJaveed84:


azafty468 (approved):
I'm 👍 for this code for our purposes.  If we do go with BigQuery I suspect we should have another pass for appropriate nits.  But this looks goooooood!

edx/analytics/tasks/common/bigquery_load.py
```diff
@@ -0,0 +1,607 @@
...
+            self.client = bigquery.Client(credentials=credentials, project=self.project_id)
+
+    def touch(self):
+        self.create_marker_table()
+        dataset = self.client.dataset(self.dataset_id)
+        table = dataset.table('table_updates')
+        table.reload() # Load the schema
+
+        # Use a tempfile for loading data into table_updates
+        # We deliberately don't use table.insert_data as it we cannot use delete on
```
brianhw, line 61: nit:  "as it we".   assume 'it' is extra.

edx/analytics/tasks/common/bigquery_load.py
```diff
@@ -0,0 +1,607 @@
...
+        # a bigquery table with streaming inserts.
+        tmp = tempfile.NamedTemporaryFile(delete=False)
+        table_update_row  = (self.update_id, "{dataset}.{table}".format(dataset=self.dataset_id, table=self.table))
+        tmp.write(','.join(table_update_row))
+        tmp.close()
+
+
```
brianhw, line 68: nit: extra line.

edx/analytics/tasks/common/bigquery_load.py
```diff
@@ -0,0 +1,607 @@
...
+        # table.upload_from_file requires the file to be opened in 'rb' mode
```
brianhw, line 69: nit: trailing period.

edx/analytics/tasks/common/bigquery_load.py
```diff
@@ -0,0 +1,607 @@
...
+
+
+def wait_for_job(job):
+    while True:
+        job.reload()
+        if job.state == 'DONE':
+            if job.error_result:
+                raise RuntimeError(job.errors)
+            return
+        time.sleep(1)
```
brianhw, line 39: Is there any reason we would need a maximum job length so that this doesn't run indefinitely due to some weirdness (i.e. it gets to a different state we don't know about -- aborted or something).  One could then just count the while loops and exit with a timeout error.

edx/analytics/tasks/common/bigquery_load.py
```diff
@@ -0,0 +1,607 @@
...
+
+    def run(self):
+        client = self.output().client
+        self.create_dataset(client)
+        self.init_copy(client)
+        self.create_table(client)
+
+        dataset = client.dataset(self.dataset_id)
+        table = dataset.table(self.table, self.schema)
+
```
brianhw, line 182: nit: extra line.

edx/analytics/tasks/common/bigquery_load.py
```diff
@@ -0,0 +1,607 @@
...
+        source_path = self.input()['source'].path
+        parsed_url = urlparse.urlparse(source_path)
+        destination_path = url_path_join('gs://{}'.format(parsed_url.netloc), parsed_url.path)
+
+        if source_path.endswith('.tsv'):
+            return_code = subprocess.call(['gsutil', 'cp', source_path, destination_path])
+        else:
+            log.debug(" ".join(['gsutil', '-m', 'rsync', source_path, destination_path]))
+            return_code = subprocess.call(['gsutil', '-m', 'rsync', source_path, destination_path])
```
brianhw, line 192: I'm assuming the logic here is to use the .tsv to distinguish between a file and a directory.  It seems a little brittle (and definitely not very general).  I'm assuming we have no better way of doing this.  Might be good to add a comment here, at least.

edx/analytics/tasks/common/bigquery_load.py
```diff
@@ -0,0 +1,607 @@
...
+        # with self.input()['source'].open('r') as source_file:
+        #     job = table.upload_from_file(
+        #             source_file,
+        #             source_format='text/csv',
+        #             field_delimiter=self.field_delimiter,
+        #             null_marker=self.null_marker,
+        #             quote_character=self.quote_character
+        #     )
+
+        #job_name = str(uuid.uuid4())
```
brianhw, line 204: I have no background on what this commented-out code is for.   Is it useful?  I'm assuming this was the original implementation (as is used for table_update), but that it didn't work at scale.  It might be good to replace this code with such a comment.   Unless there's a reason to support both from_file and from_storage methods.

edx/analytics/tasks/common/bigquery_load.py
```diff
@@ -0,0 +1,607 @@
...
+
+        if not destination_path.endswith('.tsv'):
```
brianhw, line 206: Might be good to pull out a method that is something like is_file().   Then the intention is clearer, and the assumption more consistently documented....

edx/analytics/tasks/common/bigquery_load.py
```diff
@@ -0,0 +1,607 @@
...
+                credentials_target=self.input()['credentials'],
+                dataset_id=self.dataset_id,
+                table=self.table,
+                update_id=self.update_id(),
+            )
+
+        return self.output_target
+
+    def update_id(self):
+        return '{task_name}(date={key})'.format(task_name=self.task_family, key=self.date.isoformat())
```
brianhw, line 239: I don't find task_family defined anywhere.   Where does it come from?   I'm guessing it's a Luigi thing.

edx/analytics/tasks/common/bigquery_load.py
```diff
@@ -0,0 +1,607 @@
...
+
+
+class LoadInternalReportingCertificatesToBigQuery(WarehouseMixin, BigQueryLoadTask):
```
brianhw, line 242: In the longer-term, I would like to see these moved to warehouse directory.

edx/analytics/tasks/common/bigquery_load.py
```diff
@@ -0,0 +1,607 @@
...
+                overwrite=self.overwrite,
+                warehouse_path=self.warehouse_path,
+                date=self.date
+            )
+        )
+
+
+class LoadInternalReportingCountryToBigQuery(WarehouseMixin, BigQueryLoadTask):
+
+
```
brianhw, line 276: nit: extra line.

requirements/default.txt
```diff
@@ -30,6 +30,7 @@ user-agents==0.3.2	# MIT
 vertica-python==0.6.11  # MIT
 yarn-api-client==0.2.3 # BSD
 
+git+https://github.com/HassanJaveed84/google-cloud-python.git@b225462292114c3e557cc84bd82277cfeefbc878#egg=google-cloud-python&subdirectory=bigquery
```
brianhw, line 33: Ouch!  So we needed to fork the repo...

edx/analytics/tasks/common/bigquery_load.py
```diff
@@ -0,0 +1,607 @@
...
+        return [
+            bigquery.SchemaField('user_id', 'INT64', mode='REQUIRED'),
+            bigquery.SchemaField('course_id', 'STRING', mode='REQUIRED'),
+            bigquery.SchemaField('is_certified', 'INT64'),
+            bigquery.SchemaField('certificate_mode', 'STRING'),
+            bigquery.SchemaField('final_grade', 'STRING'),
+            bigquery.SchemaField('has_passed', 'INT64'),
+            bigquery.SchemaField('created_date', 'DATETIME'),
+            bigquery.SchemaField('modified_date', 'DATETIME'),
+        ]
```
brianhw, line 261: Down the line, it would be good to get these from a shared Record object, used by various loaders.  Not sure if that would speed up the next connector, but I guess we only have three, so the cut-and-paste is not too bad in the shorter-term.

edx/analytics/tasks/common/bigquery_load.py
```diff
@@ -0,0 +1,607 @@
...
+    date = luigi.DateParameter()
+
+    def __init__(self, *args, **kwargs):
+        super(LoadInternalReportingUserActivityToBigQuery, self).__init__(*args, **kwargs)
+
+        path = url_path_join(self.warehouse_path, 'internal_reporting_user_activity')
+        path_targets = PathSetTask([path]).output()
+        paths = list(set([os.path.dirname(target.path) for target in path_targets]))
+        dates = [path.rsplit('/', 2)[-1] for path in paths]
+        latest_date = sorted(dates)[-1]
```
brianhw, line 464: maybe just add a comment saying that this finds the most-recent data for the source directory and loads that (unlike all the others, which will generate fresh data if it's not present).

edx/analytics/tasks/common/bigquery_load.py
```diff
@@ -0,0 +1,607 @@
+import logging
+import datetime
+import os
+import luigi
+import json
+import time
+import subprocess
+import urlparse
+import uuid
+import tempfile
```
brianhw, line 10: nit: these aren't in order.

brianhw (commented):
Hassan, this looks great!    

I only found nits and such -- I'm not sure if this is intended for master.   If it were, I would request the code split between common and warehouse first.

edx/analytics/tasks/common/bigquery_load.py
HassanJaveed84: I'll include something to cater that.

edx/analytics/tasks/common/bigquery_load.py
HassanJaveed84: I agree on the brittleness of things. 
BigQuery expects a fully qualified url, it can also contain a single wildcard.

Most of our workflows produce multiple files in the output directory, some produce a single .tsv file. The output path is either a path to a directory or path to a file.
I thought of just taking the directory path of such file paths and loading with a wildcard i-e path_to_subjects_tsv/* but I went this way to better control of what exactly gets loaded. 

If we were to do away with this logic, then we could also remove the 'gsutil cp' and just load whatever it is there under the directory of such file paths. (assuming that there is only one file under that directory).

edx/analytics/tasks/common/bigquery_load.py
HassanJaveed84: I thought of using it for generating a unique bigquery job name but then used current timestamp instead. I'll remove this.

edx/analytics/tasks/common/bigquery_load.py
HassanJaveed84: Yup, it defaults to the class name.

requirements/default.txt
HassanJaveed84: google-cloud-client is currently missing the option to pass a custom null marker('\N' in our case). 
I plan to open a PR once I'm able to get the testing environment setup for the cloud client.

brianhw:
Superceded by PR #434

PR #404, opened 2017-06-01
Make acceptance test work on hive 1x & 2x.

HassanJaveed84:
Not exactly a clean implementation, but I ran out of ideas for this one. 
The acceptance test passes on both EMR 4.7.2 and 5.5.0.

Suggestions would be appreciated!

### DIFF
--- a/edx/analytics/tasks/tests/acceptance/test_database_import.py
+++ b/edx/analytics/tasks/tests/acceptance/test_database_import.py
@@ -2,6 +2,21 @@ import os
 from unittest import TestCase
 
 from edx.analytics.tasks.tests.acceptance import AcceptanceTestCase
+from edx.analytics.tasks.util.record import Record, IntegerField, StringField, DateTimeField, FloatField
+
+
+class GradesPersistentCourseGradeRecord(Record):
+    id = IntegerField()
+    user_id = IntegerField()
+    course_id = StringField()
+    course_edited_timestamp = DateTimeField()
+    course_version = StringField()
+    grading_policy_hash = StringField()
+    percent_grade = FloatField()
+    letter_grade = StringField()
+    passed_timestamp = DateTimeField()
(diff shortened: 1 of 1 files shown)


edx/analytics/tasks/tests/acceptance/test_database_import.py
```diff
@@ -26,20 +41,35 @@ def test_import_from_mysql(self):
             '--database', self.import_db.database_name,
         ])
 
-        hive_output = self.hive.execute('SELECT * FROM grades_persistentcoursegrade;')
+        columns = [x[0] for x in GradesPersistentCourseGradeRecord.get_hive_schema()]
+
+        hive_output = self.hive.execute('SELECT {columns} FROM grades_persistentcoursegrade;'.format(
```
brianhw: nit: this could use self.TABLE, as on line 32.

brianhw: A larger question:  is there a need to enumerate the columns here, instead of using 'select *'?  It should be the same, I would expect, but would catch problems in future were new columns added.

brianhw (approved):
Yup, this is what I was proposing to try -- to use a Record to serve as the unit of comparison.  At some point, we can factor out the conversion of the 'select' output into a centralized method.  But this serves as a model.

edx/analytics/tasks/tests/acceptance/test_database_import.py
HassanJaveed84: Select * returns an extra column dt equal to the date string, I'm guessing it's related to partitioning.

brianhw: Yes, that makes sense.    Maybe add a comment -- that would be good to know.

tobz (approved):
LGTM.

PR #405, opened 2017-06-07
Task to calculate active users.

HassanJaveed84:
I am working on writing an acceptance test.

We could do away with the hive related tasks. I initially wrote them in order to take a join with auth_user.
The VerticaCopyTask does not depend on the map-reduce task rather an ExternalURL, as we plan on updating the table on weekly basis.  VerticaCopyTask can be run daily with the warehouse loading job and the actual data would be updated weekly(like user-activity). 

I plan on renaming stuff once we finalize the table name in the warehouse. I'm not sure if we have ?

TODO:

- Add acceptance test. DONE
- Docstrings and comments. DONE
- Rename table etc. DONE


edx/analytics/tasks/warehouse/load_internal_reporting_active_users.py
```diff
@@ -0,0 +1,153 @@
...
+        username = event.get('username', '').strip()
+        if not username:
+            return
+
+        event_data = eventlog.get_event_data(event)
+
+        if event_data is None:
+            return
+
+        user_id = event_data.get('user_id')
```
brianhw: Hmm.  Why are you relying only on getting user_id from event_data, and not from context (where it is more generally found)?   You should probably check here, and if not found here, then check context.   And then only if it's in neither place, return instead of yielding.

brianhw (commented):
I guess your other point about using Hive to join with auth_user depends on whether we want to rely only on events which have a user_id on them.   It might be worth adding some counters to see how many events that would skip.  I don't have a sense, but I'm guessing that Gabe's original code used username.

HassanJaveed84:
Ready for another review.

edx/analytics/tasks/warehouse/load_internal_reporting_active_users.py
```diff
@@ -0,0 +1,157 @@
+"""Computes active users over a time period"""
```
brianhw: nit:  Table name says "this_year", so I think that more should say that specifically.

edx/analytics/tasks/warehouse/load_internal_reporting_active_users.py
```diff
@@ -0,0 +1,157 @@
...
+
+log = logging.getLogger(__name__)
+
+
+class ActiveUsersDownstreamMixin(
+        WarehouseMixin,
+        OverwriteOutputMixin,
+        MapReduceJobTaskMixin,
+        EventLogSelectionDownstreamMixin):
+    """Common parameters needed for the workflow"""
```
brianhw: nit: trailing period.

edx/analytics/tasks/warehouse/load_internal_reporting_active_users.py
```diff
@@ -0,0 +1,157 @@
...
+    """Task to compute active users."""
+
+    def mapper(self, line):
+        value = self.get_event_and_date_string(line)
+
+        if value is None:
+            return
+        event, _date_string = value
+
+        event_type = event.get('event_type')
```
brianhw: this is not used.

edx/analytics/tasks/warehouse/load_internal_reporting_active_users.py
```diff
@@ -0,0 +1,157 @@
...
+
+        username = event.get('username', '').strip()
```
brianhw: nit: you could use eventlog.get_event_username()

edx/analytics/tasks/warehouse/load_internal_reporting_active_users.py
```diff
@@ -0,0 +1,157 @@
...
+        self.load_date = datetime.datetime.strptime(latest_date, "dt=%Y-%m-%d").date()
+
+    @property
+    def partition(self):
+        """The table is partitioned by date."""
+        return HivePartition('dt', self.load_date.isoformat())  # pylint: disable=no-member
+
+    @property
+    def insert_source_task(self):
+        hive_table = "active_users_this_year"
```
brianhw: nit:  pull this table name out into a class-level constant, for use here and in __init__.

brianhw (approved):
LGTM.   Just a few small nits, then 👍

mulby (approved):

tobz (approved):
Looks decent to me. :)

PR #406, opened 2017-06-12
Revert "Add support for templating to Luigi configuration files. (#398)"

tobz:
This reverts commit 82c99915b43a1eab6ac1de0e73d8c152efae4c43.

There's an issue with what Luigi expects to find for command-line arguments that doesn't quite mesh with what I was trying to do here, and the options to remedy it are all with their cons, so reverting this for now to not block up other things being mergable/deployable.


brianhw (approved):
LGTM.

PR #407, opened 2017-06-15
Add support for templating to Luigi configuration files, redux.

tobz:
This approach is intended to utilize the fact that Luigi's configuration code is based on Python's ConfigParser, which has native support for templated variables.

The catch is that the variables must be present in the config already, and we intended to load multiple files, in order, to make these "base" variables available to the templatized configurations loaded after.

This PR adds support to the remote and local launchers, as well as changes the small bootstrap task run through Ansible, to allow passing in multiple configuration files from the secure configuration repository to be loaded.

### DIFF
--- a/edx/analytics/tasks/launchers/local.py
+++ b/edx/analytics/tasks/launchers/local.py
@@ -46,10 +46,6 @@ def main():
         default=None, action='append')
     arguments, _extra_args = parser.parse_known_args()
 
-    # This is a relative level of terrible depending on who you are.  We're removing our --additional-config command
-    # arguments because Luigi will blow up if the underlying workflow isn't actually saying that it expects to see
-    # an argument called --additional-config.
-
     # We get a cleaned command-line arguments list, free of the arguments *we* care about, since Luigi will throw
     # errors when it sees arguments that it or the workflow didn't specify.  We pass these in when invoking Luigi.
     cmdline_args = get_cleaned_command_line_args()
--- a/edx/analytics/tasks/launchers/local.py
+++ b/edx/analytics/tasks/launchers/local.py
@@ -49,7 +49,10 @@ def main():
     # This is a relative level of terrible depending on who you are.  We're removing our --additional-config command
     # arguments because Luigi will blow up if the underlying workflow isn't actually saying that it expects to see
     # an argument called --additional-config.
-    reset_command_line_args()
+
+    # We get a cleaned command-line arguments list, free of the arguments *we* care about, since Luigi will throw
+    # errors when it sees arguments that it or the workflow didn't specify.  We pass these in when invoking Luigi.
+    cmdline_args = get_cleaned_command_line_args()
 
     # In order to see errors during extension loading, you can uncomment the next line.
     logging.basicConfig(level=logging.DEBUG)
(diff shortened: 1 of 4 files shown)


edx/analytics/tasks/launchers/local.py
```diff
@@ -76,6 +99,19 @@ def main():
         luigi.run()
```
mulby: It looks like this takes an argument called "cmdline_args" which could be used to pass in custom array. We might be able to use this to avoid changing the global state?

edx/analytics/tasks/launchers/local.py
```diff
@@ -76,6 +99,19 @@ def main():
...
 
+def reset_command_line_args():
+  arg_list = sys.argv
+  modified_arg_list = arg_list
+
+  for i, v in enumerate(arg_list):
+    if v == '--additional-config':
+      # Clear out the flag, and clear out the value attached to it.
+      modified_arg_list[i] = None
+      modified_arg_list[i+1] = None
```
mulby, line 112: Does this work if I specify:

"foo --additional-config=bar"

tobz: Nope, it definitely breaks on that.  I could modify it to try and detect that, but I'm not sure it's worth it.

edx/analytics/tasks/launchers/local.py
tobz: Good catch; yeah, this does look I could pass in the modified arguments list within having to modify `sys.argv` directly.

I'll try that out.

edx/analytics/tasks/launchers/local.py
```diff
@@ -73,7 +99,23 @@ def main():
     # Launch Luigi using the default builder
 
     with profile_if_necessary(os.getenv('WORKFLOW_PROFILER', ''), os.getenv('WORKFLOW_PROFILER_PATH', '')):
-        luigi.run()
+        luigi.run(cmdline_args)
```
azafty468, line 98: Just a blind question, are we opening ourselves to a security risk by putting these into a run command?  Or are the command args appropriately contained?

azafty468 (approved):
LGTM with the one question added

edx/analytics/tasks/launchers/local.py
tobz: This is overriding Luigi.  Instead of Luigi looking to `sys.argv` for command-line arguments to parse, we're telling it to use the list of arguments that we want it to use.

It is functionally equivalent, best that I can tell.

edx/analytics/tasks/launchers/local.py
```diff
@@ -39,19 +41,43 @@
 
 
 def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--additional-config', help='additional configuration file to be loaded after default/override',
+        default=None, action='append')
+    arguments, _extra_args = parser.parse_known_args()
+
+    # This is a relative level of terrible depending on who you are.  We're removing our --additional-config command
```
brianhw: I guess I'm one of those people (you know who you are) for whom this is not that great a level of terrible.  Particularly since we are extracting the args intended for local, and passing through the args intended for Luigi.  Seems straight-forward.  I'd be okay if you wanted to remove the editorial.   : )

brianhw (approved):
LGTM. 👍

edx/analytics/tasks/launchers/local.py
tobz: Oops! I meant to remove the above paragraph after writing the second, softer version below it. :)

PR #408, opened 2017-06-19
Redshift warehouse loading tasks

HassanJaveed84:
We would know more about the best distribution keys/styles after we do redshift evaluation. 
It would involve evaluating the query plan to optimize the distribution style.

### DIFF
--- a/edx/analytics/tasks/common/redshift_load.py
+++ b/edx/analytics/tasks/common/redshift_load.py
@@ -182,6 +182,10 @@ class LoadInternalReportingCertificatesToRedshift(WarehouseMixin, RedshiftS3Copy
     date = luigi.DateParameter()
 
     @property
+    def table_attributes(self):
+        return 'DISTKEY(user_id)'
+
+    @property
     def columns(self):
         return [
             ('user_id', 'INTEGER NOT NULL'),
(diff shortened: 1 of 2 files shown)


edx/analytics/tasks/common/redshift_load.py
```diff
@@ -0,0 +1,524 @@
...
+            date=self.date,
+            **kwargs
+        )
+
+        yield LoadInternalReportingCourseCatalogToRedshift(
+            date=self.date,
+            **kwargs
+        )
+
+        yield LoadUserCourseSummaryToRedshift(
```
brianhw, line 493: I assume this is still a work in progress -- I only seem to see this commented out down below.

edx/analytics/tasks/common/redshift_load.py
```diff
@@ -0,0 +1,524 @@
...
+    @property
+    def password(self):
+        return self.credentials()['password']
+
+    @property
+    def aws_access_key_id(self):
+        return ''
+
+    @property
+    def aws_secret_access_key(self):
```
tobz, line 78: Why is this (and `aws_access_key_id` above) here?  It looks like they're both unused.

tobz (approved):
Looks generally good to me. 👍

edx/analytics/tasks/common/redshift_load.py
brianhw: Nevermind, I see that it's also above.

edx/analytics/tasks/common/redshift_load.py
```diff
@@ -0,0 +1,524 @@
...
+            **kwargs
+        )
+
+        yield DailyLoadSubjectsToRedshift(
+            date=self.date,
+            **kwargs
+        )
+
+
+# class LoadUserCourseSummaryToRedshift(RedshiftS3CopyToTable):
```
brianhw, line 515: Just need to remove this -- it confused me. : )

brianhw (approved):
LGTM.

PR #409, opened 2017-06-19
Collect blast information from SailThru

brianhw:
DE-69.

This pulls information from SailThru about email blasts.  Two tables are created:  statistics about each email blast -- how many were opened and clicked and such; and information about whom the blast was sent to.

The current version of the code is not incremental.  Given how long blast information takes to be returned from SailThru, this would be a useful performance improvement.

Other shortcomings:
* Only twenty blasts will be returned for a request for info about blasts in a given time period.  This is done on a per-day basis, but is a serious hole.
* Email hashes are only useful if there is an additional table to link the hashes with the actual emails.  This will involve taking MD5 hashes of email addresses of all users, and putting that table up for joining.   It would be more convenient to have it already joined to the sailthru blast emails table.
* Once a job request has been submitted, I don't know how to cancel it.  Since only ten jobs are run at any time, long-running jobs can clog up Sailthru's queue for a long time.


tobz (approved):
LGTM with all the aforementioned caveats in mind.

👍

edx/analytics/tasks/warehouse/load_internal_reporting_sailthru.py
```diff
@@ -0,0 +1,594 @@
...
+import json
+import logging
+import pytz
+from Queue import PriorityQueue, Queue
+import requests
+import time
+from StringIO import StringIO
+
+import luigi
+# from luigi.configuration import get_config
```
tobz: nit: remove unused import

edx/analytics/tasks/warehouse/load_internal_reporting_sailthru.py
```diff
@@ -0,0 +1,594 @@
...
+    interval = luigi.DateIntervalParameter(
+        default=None,
+        description='Interval to pull data from Sailthru.',
+    )
+
+
+class DailyPullFromSailthruTask(PullFromSailthruTaskMixin, luigi.Task):
+    """
+    A task that fetches daily Sailthru blast information and writes to a file.
+
```
tobz: nit: unnecessary space

tobz: alternatively, just make this a single line docstring :P

edx/analytics/tasks/warehouse/load_internal_reporting_sailthru.py
```diff
@@ -0,0 +1,594 @@
...
+                        response.get_status_code(), requested_date
+                    )
+                    raise Exception(msg)
+
+                # TODO: decide whether to insert additional information about when the record was pulled.
+                output_file.write(response.get_body(as_dictionary=False))
+                output_file.write('\n')
+
+    def output(self):
+        """Output is NO LONGER in the form {output_root}/sailthru_raw/{CCYY-mm}/sailthru_blast_{CCYYmmdd}.json.
```
tobz: Could we move this to a code comment above the line generating the filename?

While I realize that we don't really generate API docs based on proper docstrings, if we're going to bother to use docstrings (instead of just normal code comments), then I think we should write them in a way that reflects what we'd actually want the API docs to say: this is an implementation detail that shouldn't necessarily matter to the caller, but only matters to us for understanding.. so it seems like it can live as a code comment.

brianhw: I agree that it's important to distinguish between documentation that belong in docstrings versus code comments, particularly in thinking about what makes sense to appear in API documentation generated from the docstrings.  However, I don't agree with the determination that describing where the output is generated by the task is an implementation detail.   I think it's very appropriate to include in API documents. 

I do agree that this particular comment is on the long side, because of the TODO-nature of the paragraph below.  But that's a symptom of the Spike-ness of this branch, and that would be addressed in productization (were we to get that far).

edx/analytics/tasks/warehouse/load_internal_reporting_sailthru.py
brianhw: Makes sense.

brianhw:
This is working now, in terms of being able to schedule jobs that would load Sailthru blast email and stats to Vertica.  At some point in the future, this should be revisited, to determine if these are the right fields to pull, and if Vertica is the right destination.

PR #410, opened 2017-06-23
Various fixes for ProblemResponseReportWorkflow and related tasks

pomegranited:
We encountered three production issues with the `ProblemResponseReportWorkflow` task after upgrading our client to master.

This PR fixes these issues, and verifies the fixes with updated acceptance tests.  All three fixes are included in a single PR, to make running the acceptance tests easier.

**1) Table not found: course_blocks**

cf 6fd14b7, e52f213

Error reported by [ProblemResponseLocationPartitionTask](https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/insights/problem_response.py#L426).

This task uses a hive query to join data from the `problem_response_latest` and `course_blocks` hive tables into a new table, `problem_response_location`, which contains the latest problem responses and their location in the course.

Our client runs the problem response task every 2 hours using an hourly-formatted data partition (i.e. `problem-response.partition_format = %Y-%m-%dT%H`), but the course blocks data uses the default daily partition (i.e. `course_blocks.partition_format = %Y-%m-%d`).  This is because the course blocks data rarely changes, but the problem responses come in all the time, and the client needs the latest data.

The problem response task was failing except for the first time it ran per day, because the course blocks are loaded during that run, and so the hive table was present.

  As diagnosed by @brianhw , the issue was caused by the unnecessary and erroneous [`CourseBlocksPartitionTask.complete()`](https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/insights/course_blocks.py#L383-L387) method, which was reporting the task to be complete because it found a `_SUCCESS` file, which is an artefact of that first daily partition task run.  Therefore on successive runs, the `CourseBlocksPartitionTask` wasn't being run, and so the hive table was not being loaded.

  This was not detected by the [acceptance tests](https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/insights/tests/test_problem_response.py#L51), because the acceptance tests emulate the first daily run, but not the subsequent runs.

**2) interval parameter propagation**

cf 6fd14b7, 856d4c2

Our client runs the `ProblemResponseReportWorkflow` every 2 hours so that the reports show the latest problem responses throughout the day.

The `interval_end` parameter, which decides what date range the tracking logs are read from, defaults to UTC today, 00:00.

We're overriding that parameter in our task run configuration by passing `--interval-end $TOMORROW`.  However, that `--interval-end` value isn't being passed all the way through the chain of tasks which require it. Specifically:

* [ProblemResponseReportWorkflow passes it to ProblemResponseReportTask](https://github.com/edx/edx-analytics-pipeline/blob/2ab4b708bec9d12d069394de3070fda63d3eced4/edx/analytics/tasks/insights/problem_response.py#L736-L738)
* But [ProblemResponseReportTask doesn't pass it to ProblemResponseLocationPartitionTask](https://github.com/edx/edx-analytics-pipeline/blob/2ab4b708bec9d12d069394de3070fda63d3eced4/edx/analytics/tasks/insights/problem_response.py#L611)
* And so the values passed by [ProblemResponseLocationPartitionTask to LatestProblemResponsePartitionTask](https://github.com/edx/edx-analytics-pipeline/blob/2ab4b708bec9d12d069394de3070fda63d3eced4/edx/analytics/tasks/insights/problem_response.py#L513-L515) aren't set from the Jenkins command line call, they fall back to the defaults.

**3) partition_format config typo**

cf a1b4c75

The `partition_format` configuration parameter for the Course Blocks and Course List tasks had a typo in the configuration section name.

**JIRA tickets**: OSPR

**Partner information**: 3rd party-hosted open edX instance

**Deployment targets**: These tasks are not run on insights.edx.org or insights.edge.edx.org, so these services should be unaffected.

**Merge deadline**: None

**Testing instructions**:

These changes can be verified using these acceptance tests:

```bash
make test-acceptance-local ONLY_TESTS=edx.analytics.tasks.tests.acceptance.test_course_list 
make test-acceptance-local ONLY_TESTS=edx.analytics.tasks.tests.acceptance.test_course_blocks 
make test-acceptance-local ONLY_TESTS=edx.analytics.tasks.tests.acceptance.test_problem_response 
```

**Author notes and concerns**:

1. 66bc7e9 adds 2 more acceptance tests to verify raw REST API JSON input (first run) and processed hive input (second run) use cases of these tasks.  @brianhw expressed performance concerns about this, since it increases the length of time required to run the acceptance tests.  If possible, I'd like to keep at least the added problem_response test, since this is our primary concern.
2. Was unsure how to test a1b4c75, but it is a minor change.

**Reviewers**
- [x] @bdero 
- [x] @edx/analytics reviewer[s]


openedx-webhooks:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=410&repo=edx%2Fedx-analytics-pipeline"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

bdero (approved):
:+1:

- [x] I tested this: I ran the `test_course_list`, `test_course_blocks`, and `test_problem_response` acceptance tests before and after checking out out this version of `edx-analytics-pipeline` to make sure they would still pass.
- [x] I read through the code
- [ ] ~I checked for accessibility issues~ N/A
- [ ] ~Includes documentation~ Includes comments, but in general doesn't seem to need additional external documentation because these fixes make existing functionality work properly with `problemResponseReportWorkflow`; no configuration change is present other than a typo fix.

pomegranited:
@edx/analytics This PR is now ready for review.  Thank you!

FYI re the codecov differences - they are due to additional code added to the acceptance tests, which aren't run to determine coverage levels, so it's not a real drop.

edx/analytics/tasks/insights/course_blocks.py
```diff
@@ -281,12 +278,6 @@ def order_by_sort_idx(key, blocks=blocks, default_sort_idx=no_sort_idx):
     def output(self):
         return get_target_from_url(self.output_root)
 
-    def complete(self):
-        """
-        The task is complete if the output_root file is present.
-        """
-        return get_target_from_url(url_path_join(self.output_root, '_SUCCESS')).exists()
-
```
brianhw: I'm confused here.  Is there a reason to take this out here?   This is a map-reduce task, so it will produce a _SUCCESS file, and because "enable_direct_output = True" it will write partial results to this output_root.  So I think we need this here.

edx/analytics/tasks/insights/course_blocks.py
```diff
@@ -379,9 +370,3 @@ def data_task(self):
     def output_root(self):
         """Expose the partition location path as the output root."""
         return self.partition_location
-
-    def complete(self):
-        """
-        The task is complete if the output_root/_SUCCESS file is present.
-        """
-        return get_target_from_url(url_path_join(self.output_root, '_SUCCESS')).exists()
```
brianhw, line 387: I think this is the one that needs to be removed.

edx/analytics/tasks/insights/course_list.py
```diff
@@ -202,15 +202,8 @@ def reducer(self, _key, values):
...
     def output(self):
-        """Expose the data location target as the output."""
         return get_target_from_url(self.output_root)
 
-    def complete(self):
-        """
-        The task is complete if the output_root/_SUCCESS file is present.
-        """
-        return get_target_from_url(url_path_join(self.output_root, '_SUCCESS')).exists()
-
```
brianhw: As mentioned for course_blocks, I think this is needed, as this is MR with enable_direct_output = True.

edx/analytics/tasks/insights/course_list.py
```diff
@@ -269,9 +262,3 @@ def data_task(self):
     def output_root(self):
         """Expose the partition location path as the output root."""
         return self.partition_location
-
-    def complete(self):
-        """
-        The task is complete if the output_root/_SUCCESS file is present.
-        """
-        return get_target_from_url(url_path_join(self.output_root, '_SUCCESS')).exists()
```
brianhw, line 277: ... but yes, this one needs to be deleted.

edx/analytics/tasks/insights/course_list.py
```diff
@@ -202,15 +202,8 @@ def reducer(self, _key, values):
             yield record.to_string_tuple()
```
brianhw: nit: apropos of nothing, there's a typo on line 163.

edx/analytics/tasks/insights/problem_response.py
```diff
@@ -364,12 +364,6 @@ def get_answer_data(self, event_string):
     def output(self):
         return get_target_from_url(self.output_root)
 
-    def complete(self):
-        """
-        The task is complete if the output_root/_SUCCESS file is present.
-        """
-        return get_target_from_url(url_path_join(self.output_root, '_SUCCESS')).exists()
-
```
brianhw, line 372: As mentioned for others above, I think this is needed, as this is MR with enable_direct_output = True.

edx/analytics/tasks/insights/problem_response.py
```diff
@@ -388,12 +382,6 @@ def output_root(self):
         """Expose the partition location path as the output root."""
         return self.partition_location
 
-    def complete(self):
-        """
-        The task is complete if the output_root/_SUCCESS file is present.
-        """
-        return get_target_from_url(url_path_join(self.output_root, '_SUCCESS')).exists()
-
```
brianhw, line 396: Correct:  this also needs to be deleted.

edx/analytics/tasks/insights/problem_response.py
```diff
@@ -525,12 +514,6 @@ def output_root(self):
         """Expose the partition location path as the output root."""
         return self.partition_location
 
-    def complete(self):
-        """
-        The task is complete if the output_root is present.
-        """
-        return get_target_from_url(self.output_root).exists()
-
```
brianhw, line 533: Correct:  this also should be deleted.

edx/analytics/tasks/insights/tests/test_course_blocks.py
```diff
@@ -66,21 +66,6 @@ def cleanup(self, dirname):
...
-    def test_complete(self):
-        self.create_task()
-        self.assertFalse(self.task.complete())
-
-        # Create the output_root/_SUCCESS file
-        with open(os.path.join(self.output_dir, '_SUCCESS'), 'w') as success:
-            success.write('')
-        self.assertTrue(self.task.output().exists())
-        self.assertTrue(self.task.complete())
-
```
brianhw: Restore this.

edx/analytics/tasks/insights/tests/test_problem_response.py
```diff
@@ -344,16 +344,6 @@ def create_task(self, *args, **kwargs):
...
-    def test_complete(self):
-        self.create_task()
-        self.assertFalse(self.task.complete())
-
-        # Create the output_root/_SUCCESS file
-        with open(os.path.join(self.output_dir, '_SUCCESS'), 'w') as success:
-            success.write('')
-        self.assertTrue(self.task.output().exists())
-        self.assertTrue(self.task.complete())
-
```
brianhw: restore this...

brianhw (commented):
I'm thinking that too many complete() methods were removed here.  

We also encountered an interesting race condition in the unit tests.  This just arose this week on our Jenkins test box.   In LatestProblemResponsePartitionTaskTest, there are two tests that check partition_value.   The expected value comes from utcnow() stored in self.timestamp, which is passed in as interval_end to test_partition_value_with_start_end() and test_partition_value_with_interval().  These compare that value with the value of self.task.partition_value, and this comes from the "date" parameter, which also defaults to utcnow().  However, it is apparently possible for these two values to differ enough that they belong to different minutes, and are therefore not equal.  I'm not sure how this should be fixed.  One possibility is to pass in "date" as well using the same self.timestamp, but I'm not really sure.

edx/analytics/tasks/insights/course_list.py
pomegranited: Gah..  I spent all of my time during the eucalyptus release misspelling it..  And when I finally think I've got it right, I'll be doing the same with <strike>gingko</strike> ginkgo!

Fixed with c1b190e.

edx/analytics/tasks/insights/course_blocks.py
pomegranited: Ok.. For some reason, I thought that setting `enable_direct_output = True` meant that the base class `complete()` method would check the `_SUCCESS` file for me, and so thought I was just being redundant.  I should have checked that.. thank you for noticing!

Restored with 0eb6fb4.

edx/analytics/tasks/insights/course_list.py
pomegranited: Restored with 0eb6fb4.

edx/analytics/tasks/insights/problem_response.py
pomegranited: Restored with 0eb6fb4.

edx/analytics/tasks/insights/tests/test_course_blocks.py
pomegranited: Restored with 0eb6fb4.

Also added a similar check for [`CourseListApiDataTaskTest`](https://github.com/edx/edx-analytics-pipeline/pull/410/files#diff-9093ee3452a6ee321da945bba52cf2baR49).

edx/analytics/tasks/insights/tests/test_problem_response.py
pomegranited: Restored with 0eb6fb4.

pomegranited (commented):
Thank you for your comments, @brianhw !  Have addressed them with the latest changes, but am also not sure how to test the flakiness with `LatestProblemResponsePartitionTaskTest`, since I'm not seeing it on my local devstack either.

Let me know if you find anything else?

brianhw (approved):
Thanks, Jill!   This looks good to me.    👍

pomegranited:
Thank you for helping me get this change done right, @brianhw !
I've squashed the changes down to something reasonable -- ok to merge now?

brianhw:
Okay on this end!   Merge away!

pomegranited:
@brianhw Wheee!

PR #412, opened 2017-06-30
Add backquotes for additional Hive columns.

brianhw:
This is needed for upgrading past Hive 1.2.

It turns out that AggregateInternalReportingUserActivityTableHive is the one task in user_activity stuff that is not included in acceptance tests, so we didn't catch this sooner.


azafty468 (approved):
👍

PR #413, opened 2017-06-30
Remove unneeded dependence on overwrite-n-days.

brianhw:
Fix problem with running ImportEnrollmentsIntoMysql out of box,
which encounters an error in a task because overwrite-n-days is not
provided.  It's not needed in two tasks, so this removes the
parameter from the tasks in question.

Tested in analyticstack.

### DIFF
--- a/edx/analytics/tasks/insights/enrollments.py
+++ b/edx/analytics/tasks/insights/enrollments.py
@@ -146,7 +146,17 @@ class CourseEnrollmentEventsTask(
         return tasks
 
     def run(self):
+        # This removes the marker file.
         self.remove_output_on_overwrite()
+        # We also want to remove the output files before running, in case output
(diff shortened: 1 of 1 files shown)


dcadams:
@brianhw If overwrite_n_days param is no longer required for this task could you please see to it that the docs get updated:

http://edx.readthedocs.io/projects/edx-installing-configuring-and-running/en/latest/installation/analytics/start_analytics.html

brianhw:
@dcadams Thanks for checking in on this.  Actually, this PR aims to make that page actually be accurate.  Right now, running the command on that documentation page doesn't work unless the overwrite_n_days parameter is also defined in the config file, just to get a task to work that doesn't actually use it.  So this should require no change in documentation.

brianhw:
Also tested on enrollment release candidate.

tobz (approved):
LGTM.

PR #414, opened 2017-07-05
Do not use --no-cache-dir for pre.txt.

brianhw:
The --no-cache-dir option is not defined on older versions of pip.
The pre.txt requirements specifically upgrade to a newer pip, so
the option may not work on that file, but should work fine after it
has been loaded.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ install: requirements uninstall
 	python setup.py install --force
 
 bootstrap: uninstall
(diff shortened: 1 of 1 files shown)


brianhw:
This is needed to fix master so that it works with the old version of pip we have on Jenkins.   Most jobs running off master are getting by because they already have a more up-to-date virtualenv present.  For jobs that are either older or have had their working directories deleted, this is not the case.

mulby (approved):

PR #415, opened 2017-07-11
Added course slot number to d_program_course.

HassanJaveed84:
Unsure if there is a better way of doing the join for getting slot number/order.
We could also make ProgramCourseOrderDataTask and PullProgramAPIData a single task that pulls from API and writes out what's needed in a single phase.

### DIFF
--- a/edx/analytics/tasks/warehouse/load_internal_reporting_course_catalog.py
+++ b/edx/analytics/tasks/warehouse/load_internal_reporting_course_catalog.py
@@ -7,10 +7,11 @@ import datetime
 import logging
 
 import luigi
+from luigi.hive import HiveQueryTask
 
 from edx.analytics.tasks.common.vertica_load import VerticaCopyTask, VerticaCopyTaskMixin
 from edx.analytics.tasks.util.edx_api_client import EdxApiClient
-from edx.analytics.tasks.util.hive import BareHiveTableTask, HivePartitionTask, WarehouseMixin
+from edx.analytics.tasks.util.hive import BareHiveTableTask, HivePartitionTask, WarehouseMixin, hive_database_name
 from edx.analytics.tasks.util.record import Record, StringField, FloatField, DateTimeField, IntegerField
 from edx.analytics.tasks.util.opaque_key_util import get_org_id_for_course
 from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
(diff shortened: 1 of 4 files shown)


tobz (approved):
LGTM. 👍

edx/analytics/tasks/warehouse/load_internal_reporting_course_catalog.py
```diff
@@ -88,6 +89,116 @@ def output(self):
...
+        with self.input().open('r') as input_file:
+            with self.output().open('w') as output_file:
+                for program_str in input_file:
+                    program = json.loads(program_str)
+                    current_slot_number = 1
+                    for course in program.get('courses', []):
+                        program_id = program['uuid']
+                        catalog_course = course['key']
+                        slot_number = current_slot_number
+                        output_file.write('\t'.join((program_id, catalog_course, str(slot_number))))
```
brianhw: I think we normally have to ensure that stuff is written out as utf-8.   The uuid and number should always be fine, and in the short term we can continue to assume that the course_id is also ascii.   Maybe add a comment to that effect?

edx/analytics/tasks/warehouse/load_internal_reporting_course_catalog.py
```diff
@@ -196,24 +308,93 @@ def process_course_run(self, course_run, output_file):
...
+        self.remove_output_on_overwrite()
+        super(JoinProgramCourseWithOrderTask, self).run()
+
+    @property
+    def table(self):
+        return 'program_course_with_order'
+
+    @property
+    def table_location(self):
+        return url_path_join(self.warehouse_path, self.table) + '/'
```
brianhw: So how does this work over time?   Is the idea that we always run it in overwrite mode, since it doesn't write out a separate version every day?  And if it requires overwrite, I assume it will always calculate everything over again from scratch, because overwrite gets propagated down to the API calls.  Or am I missing something?

brianhw (commented):
Confused about the versioning of the output.

edx/analytics/tasks/warehouse/load_internal_reporting_course_catalog.py
HassanJaveed84: I overlooked this. I'll change it to write out a separate version everyday.

brianhw (approved):
LGTM.   👍

PR #417, opened 2017-07-21
Active users per week.

HassanJaveed84:
Currently trying to make this incremental.
Looking to gather feedback on the implementation.


edx/analytics/tasks/warehouse/load_internal_reporting_active_users.py
```diff
@@ -47,19 +47,31 @@ def mapper(self, line):
...
+        week = isoweek.Week(iso_year, iso_weekofyear)
+        start_date = week.monday().isoformat()
+        end_date = (week.sunday() + datetime.timedelta(1)).isoformat()
+
+        yield (start_date, end_date, username), 1
 
     def reducer(self, key, values):
-        yield self.interval.date_a.isoformat(), self.interval.date_b.isoformat(), key
+        yield key
+        #yield self.interval.date_a.isoformat(), self.interval.date_b.isoformat(), key
```
brianhw: nit: I presume this can be removed.

edx/analytics/tasks/warehouse/load_internal_reporting_active_users.py
```diff
@@ -47,19 +47,31 @@ def mapper(self, line):
             self.incr_counter('Active Users last year', 'Discard Event Missing username', 1)
             return
 
-        yield username, 1
+        date = datetime.date(*[int(x) for x in date_string.split('-')])
+        iso_year, iso_weekofyear, iso_weekday = date.isocalendar()
+        week = isoweek.Week(iso_year, iso_weekofyear)
+        start_date = week.monday().isoformat()
+        end_date = (week.sunday() + datetime.timedelta(1)).isoformat()
```
brianhw, line 55: So this is an eight-day period, Monday to Monday, so I assume the end is exclusive.   Is it our general pattern to represent weeks as an inclusive beginning and an exclusive ending?   (This is not serving some API, so it matters less as long as the convention is understood.)

brianhw (approved):
This looks good, with just a nit and a question.    And I guess the general observation that we would need to always run this with an interval ending on a Monday, if we want to avoid having incomplete data for the last week.

edx/analytics/tasks/warehouse/load_internal_reporting_active_users.py
HassanJaveed84: You're correct, end is exclusive. That is indeed our general pattern, including how we handle tracking log data.

'course_activity' mysql table also follows this pattern, 'interval_end' column is an exclusive boundary.

tobz (approved):
This looks reasonable to me. :shipit:

brianhw:
I'm missing a detail about how this is being called.   There's a date parameter being passed to the workflow, and this is being used to represent the week, but it's not being normalized to the week anywhere until actually calculating the low-level data.  But if the date being passed in is today's date, then it will be flagged separately from yesterday's run, even though they involve calculating the same week.  Is this assuming that the date passed in is always a Monday?

HassanJaveed84:
I had developed this with the idea that we should schedule this workflow to run every Monday, and it would compute the active users over the last week, and recompute the week before that, overwriting in Vertica. 

Now that I look closely, it can possibly result in duplicated records. Maybe we can change it so that the 'interval' get's passed down to the VerticaCopyTask(instead of the date parameter), the 'update_id' can be changed to include 'start_date' and 'end_date' and we change the where clause in init_copy to include 'start_date' as well. We could still schedule to run it on only Mondays, but running on other days should also be safe, and would be a noop if already completed before.

Open to ideas and suggestions.

HassanJaveed84:
Also, does it make sense to create a IncrementalVerticaCopyTask similar to IncrementalMysqlInsertTask and then use it here ? We currently don't have any other workflow doing incremental copying into Vertica.

brianhw:
I like those suggestions.  Creating the IncrementalVerticaCopyTask, even if only used here, is much clearer as to what is different.  It took me a while to find the date clause.   And making the workflow safer by putting the interval into a common __init__ in the mixin (or something similar) would make this less likely to get corrupted.

HassanJaveed84:
@brianhw this is ready for a final review.

azafty468 (approved):
👍

brianhw:
I'm sorry I'm still not understanding this.  If this runs on a Monday and then again on Wednesday, it will be run with two different values for the date, no?  The start and end dates would be different, so it would still be run again.   It's relying on jobs being passed in the Monday date?  I was imagining that since the low-level task knows how to calculate the relevant full isoweek that can be calculated with this, that this is what the job would always run, relative to the date passed in.  If this is too hard to do, then skip it and document it at the workflow level as to the assumption of the date value being passed in, and we'll go with that.

HassanJaveed84:
If this runs on a Monday, then again on Wednesday, the start and end date for the low level tasks would be the same as the start and end date would always be an isoweek boundary. So suppose we pass in 2017-07-31(Monday), the week boundary is 2017-07-24-2017-07-31. If we run this workflow again on Wednesday(2017-08-02), the iso week it calculates is still 2017-07-24-2017-07-31. It would not run again till next Monday(or after) when the isoweek moves(apart from overwriting the week before last as overwrite=True). 

We can do the isoweek calculation in the ActiveUsersWorkflow task and pass down the interval to the vertica task and mapreduce task if it makes more sense.

brianhw:
On a plane now, but wouldn't it still write to Vertica on Monday and again
on Wednesday?  Base task would find the same interval, but above that would
run twice,  no?

HassanJaveed84:
Vertica task computes the isoweek interval(WeeklyIntervalMixin), and it sets the update_id as
LoadInternalReportingActiveUsersToWarehouse(start_date=start_date,end_date=end_date). As the start and end date would be the same on Monday and wednesday, it should not run again.

brianhw:
Sorry if I misread the code. I thought that the Vertica task did not use
the isoweek start and end, but used the input date and date minus one week,
and that the isoweek calculation happened relative to those dates only at
the lowest levels.

Either way, I don't really want to block this from Cleveland.

brianhw:
Just for closure on this, somehow I missed the use of WeeklyIntervalMixin.   👍

PR #418, opened 2017-07-24
A few tweaks to debug_emr_logs.py

mulby:
This worked great for me this morning!

I made a couple of tweaks:
* output the full command used to download the logs, at first I had put in a job flow ID in the "cluster name" argument, which didn't work... seeing the full URL made it very clear I had messed something up.
* support python 3.X - I've started using that as my default interpreter for running tools and had to make minimal changes to this script to make it work, so I figured I should commit them

### DIFF
--- a/scripts/debug_emr_logs.py
+++ b/scripts/debug_emr_logs.py
@@ -1,4 +1,5 @@
 import argparse
+from collections import deque
 import gzip
 import os
 import re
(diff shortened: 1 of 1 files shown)


HassanJaveed84 (approved):
LGTM 👍

mulby:
@HassanJaveed84 I just added another tweak here - a context buffer so you can see a few lines of context before the stack trace for each file

HassanJaveed84:
@mulby looks good!

PR #419, opened 2017-07-24
Enables contacting partner specific URLs for course information.

azafty468:
Enable the use of partner specific URLs to acquire course information.  Performed in conjunction with configuration variables that provide this information.


brianhw:
LGTM.  👍

PR #421, opened 2017-07-27
Update boto to support eu-central-1 usage.

tobz:
Long story short, the `eu-central-1` AWS region has some idiosyncrasies that were accounted for in code by boto, but not on the current version of it that we use.

This just uprevs the version of boto to something that supports talking to `eu-central-1`, which Open edX users have been trying to work with, to much dismay, on the code as-is.


UmanShahzad:
@tobz I've opened https://github.com/edx/edx-analytics-pipeline/pull/445 FYI

brianhw:
Superceded by #451 .

PR #423, opened 2017-07-31
Event Exporter Improved Encryption Logging & Error Handling 

azafty468:
This work is originally for bug fix DE-274, with the final work captured in DE-282.  The target is to improve logging during GPG encryption failures and to allow other organizations to continue to receive exported events when one encryption process fails.  Reviews are gladly welcomed on this one as I'm doing a bulk capture of IOErrors and it will do the job but feels like overkill.

### DIFF
--- a/edx/analytics/tasks/export/event_exports.py
+++ b/edx/analytics/tasks/export/event_exports.py
@@ -162,6 +162,12 @@ class EventExportTask(EventLogSelectionMixin, MultiOutputMapReduceJobTask):
             )
         )
 
+    def event_export_counter(self, counter_title, incr_value=1):
+        """
+        A shorthand hadoop counter incrementer that organizes counters under a common title.
+        """
+        self.incr_counter("Event Export", counter_title, incr_value)
+
     def multi_output_reducer(self, key, values, output_file):
         """
         Write values to the appropriate file as determined by the key.
@@ -173,20 +179,28 @@ class EventExportTask(EventLogSelectionMixin, MultiOutputMapReduceJobTask):
 
         def report_progress(num_bytes):
             """Update hadoop counters as the file is written"""
-            self.incr_counter('Event Export', 'Bytes Written to Output', num_bytes)
+            self.event_export_counter(counter_title='Bytes Written to Output', incr_value=num_bytes)
 
         key_file_targets = [get_target_from_url(url_path_join(self.gpg_key_dir, recipient)) for recipient in recipients]
-        with make_encrypted_file(output_file, key_file_targets, progress=report_progress) as encrypted_output_file:
-            outfile = gzip.GzipFile(mode='wb', fileobj=encrypted_output_file)
-            try:
-                for value in values:
-                    outfile.write(value.strip())
-                    outfile.write('\n')
-                    # WARNING: This line ensures that Hadoop knows that our process is not sitting in an infinite loop.
-                    # Do not remove it.
-                    self.incr_counter('Event Export', 'Raw Bytes Written', len(value) + 1)
-            finally:
-                outfile.close()
+        try:
+            with make_encrypted_file(output_file, key_file_targets, progress=report_progress,
+                                     hadoop_counter_incr_func=self.event_export_counter) as encrypted_output_file:
+                outfile = gzip.GzipFile(mode='wb', fileobj=encrypted_output_file)
+                try:
+                    for value in values:
+                        outfile.write(value.strip())
+                        outfile.write('\n')
+                        # WARNING: This line ensures that Hadoop knows that our process is not sitting in an infinite
+                        # loop.  Do not remove it.
+                        self.event_export_counter(counter_title='Raw Bytes Written', incr_value=(len(value) + 1))
+                finally:
+                    outfile.close()
+        except IOError as err:
+            log.error("Error encountered while encrypting and gzipping Organization: %s file: %s Exception: %s",
+                      org_id, key_file_targets, err)
+            # This counter is set when there is an error during the generation of the encryption file for an
+            # organization for any reason, including encryption errors related to an expired GPG key.
+            self.event_export_counter(counter_title="{} org with Errors".format(org_id), incr_value=1)
 
     def get_org_id(self, event):
         """
--- a/edx/analytics/tasks/util/encrypt.py
+++ b/edx/analytics/tasks/util/encrypt.py
@@ -4,6 +4,7 @@ Tasks for performing encryption on export files.
 from contextlib import contextmanager
 import logging
 import tempfile
+import datetime
 
 import gnupg
 
@@ -35,7 +36,8 @@ def get_key_from_target(key_file_target):
 
 
 @contextmanager
-def make_encrypted_file(output_file, key_file_targets, recipients=None, progress=None, dir=None):
+def make_encrypted_file(output_file, key_file_targets, recipients=None, progress=None, dir=None,
+                        hadoop_counter_incr_func=None):
     """
     Creates a file object to be written to, whose contents will afterwards be encrypted.
 
@@ -44,12 +46,15 @@ def make_encrypted_file(output_file, key_file_targets, recipients=None, progress
         key_file_targets: a list of luigi.Target objects defining the gpg public key files to be loaded.
         recipients:  an optional list of recipients to be loaded.  If not specified, uses all loaded keys.
         progress:  a function that is called periodically as progress is made.
+        hadoop_counter_incr_func:  A callback to a function that can generate MR counters so that non-critical GPG
+            messages can be promoted to a visible section of the MR run log.
     """
     with make_temp_directory(prefix="encrypt", dir=dir) as temp_dir:
         # Use temp directory to hold gpg keys.
         gpg = gnupg.GPG(gnupghome=temp_dir)
         gpg.encoding = 'utf-8'
-        _import_key_files(gpg, key_file_targets)
+        _import_key_files(gpg_instance=gpg, key_file_targets=key_file_targets,
+                          hadoop_counter_incr_func=hadoop_counter_incr_func)
 
         # Create a temp file to contain the unencrypted output, in the same temp directory.
         with tempfile.NamedTemporaryFile(dir=temp_dir, delete=False) as temp_input_file:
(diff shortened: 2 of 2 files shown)


edx/analytics/tasks/util/encrypt.py
```diff
@@ -75,17 +76,47 @@ def _import_key_files(gpg_instance, key_file_targets):
...
     )
-    log.info('Encryption complete.')
+
+    if not encryption_result.ok:
+        log.error("Encrypted file completed: " + str(encryption_result.ok))
+        status = ""
+        stderr = ""
+        if hasattr(encryption_result, "status"):
+            status = str(encryption_result.status)
+            log.error("Encryption status: " + str(encryption_result.status))
```
brianhw: nit: log messages should use %s instead of +.   And having set status, you could use that.

edx/analytics/tasks/util/encrypt.py
```diff
@@ -75,17 +76,47 @@ def _import_key_files(gpg_instance, key_file_targets):
...
+        if hasattr(encryption_result, "stderr"):
+            stderr = str(encryption_result.stderr)
+            log.error("Encryption error: " + str(encryption_result.stderr))
```
brianhw: nit: similar to status.

edx/analytics/tasks/util/encrypt.py
```diff
@@ -75,17 +76,47 @@ def _import_key_files(gpg_instance, key_file_targets):
...
         output=encrypted_filepath,
         armor=False,
     )
-    log.info('Encryption complete.')
+
+    if not encryption_result.ok:
+        log.error("Encrypted file completed: " + str(encryption_result.ok))
+        status = ""
+        stderr = ""
+        if hasattr(encryption_result, "status"):
```
brianhw: nit:  I think you could use status = getattr(encryption_result, "status", "") here, and then just test for "if status:" if you didn't want the empty log line.  (But maybe a blank log line is worthwhile too.)

edx/analytics/tasks/util/encrypt.py
```diff
@@ -75,17 +76,47 @@ def _import_key_files(gpg_instance, key_file_targets):
...
+
+        for key_fingerprint in import_result.fingerprints:
+            pub_keys = gpg_instance.list_keys()
+            for test_key in pub_keys:
+                if test_key["fingerprint"] == key_fingerprint:
+                    if len(test_key["expires"]) > 0:
+                        current_time = datetime.datetime.now()
+                        next_week = current_time + datetime.timedelta(days=7)
+                        key_expire = datetime.datetime.fromtimestamp(int(test_key["expires"]))
+                        # All of these messages will be buried in a container log and never see the light of day
```
brianhw: nit: This sounds pretty discouraging, and suggests that the logging is not worth doing here.   Maybe suggest using the debug-log tool here?  Or how would these messages be found?

edx/analytics/tasks/export/event_exports.py
```diff
@@ -176,17 +176,22 @@ def report_progress(num_bytes):
...
+                        outfile.write('\n')
+                        # WARNING: This line ensures that Hadoop knows that our process is not sitting in an infinite
+                        # loop.  Do not remove it.
+                        self.incr_counter('Event Export', 'Raw Bytes Written', len(value) + 1)
+                finally:
+                    outfile.close()
+        except IOError as err:
+            log.error("Error encountered while encrypting and gzipping Organization: {} file: {} Exception: {}"
+                      .format(org_id, key_file_targets, err))
+            self.incr_counter("Event Export", "{} org with Errors".format(org_id), 1)
```
brianhw: nit: It may be helpful to put a comment here saying that this is the line to be looked for when checking jobs using this task for possible failures.

brianhw (approved):
Just to check if I understand, this PR adds three improvements:  one is to add log messages, at the reducer level that an error happened, and at the utility level to indicate if a key expired.  Second is to explicitly raise an IOError to communicate between the utility level and the reducer level.  Third is to output a counter for each time there is a failure at the reducer level (e.g. number of days failing per organization).  

And the overall goal is to process as many organizations as possible, even though one or more organizations may have bad keys and not be capable of generating output successfully.  These errors are logged via stderr and counters, but do not cause the job to fail.  If rerun, the job will succeed and set a marker.  If the cause of a failure gets fixed, then rerunning with just the org specified will allow the job to rerun, since the marker will differ from that of the original job.

However, if the cause of the failure is not actually fixed, then rerunning with just the org specified will cause a marker to be generated for the single-org case, but actually fail to generate the output.  And it's not clear to me how to reset it, since I'm not sure we actually write out the location of the marker file in these cases.  (That would be very useful, if we really did want to retry rerunning it again.)  It would be great if the marker file path were output somewhere.

None of the nits I added are blocking.  But I'm wondering if there is a good place to put a note in the code describing the assumptions.  If it's only in a wiki, then it's not likely to be updated.   Some docstring somewhere?  Perhaps where the counter is incremented, since that's where one would expect the error to be raised.   It would be good to explain there why it's being caught and not propagated. 

Apart from adding that doc, I think it's good to go. 👍

HassanJaveed84:
LGTM. 👍 

We currently do not have the ability to delete marker output for a MultiOutputMapReduceJobTask. We do have a 'delete_output_root' parameter which deletes the marker as well as the output_root(which we don't want to do in case of event exports :) ). But it gets very tricky to rerun a failed MultiOutputMapReduceJobTask as you have to manually clear the marker output. Maybe we can capture that in a future story.

brianhw:
I'm assuming that when you squash the commits, you'll be rebasing on the tip of master, so hopefully the Travis builds will finally run.

edx/analytics/tasks/export/event_exports.py
```diff
@@ -176,17 +179,25 @@ def report_progress(num_bytes):
...
+                        # loop.  Do not remove it.
+                        self.incr_counter('Event Export', 'Raw Bytes Written', len(value) + 1)
+                finally:
+                    outfile.close()
+        except IOError as err:
+            log.error("Error encountered while encrypting and gzipping Organization: %s file: %s Exception: %s",
+                      org_id, key_file_targets, err)
+            # This counter is set when there is an error during the generation of the encryption file for an
+            # organization for any reason, including encryption errors related to an expired GPG key
+            self.incr_counter("Event Export", "{} org with Errors".format(org_id), 1)
```
brianhw: nit: this could use self.event_export_counter().

edx/analytics/tasks/export/event_exports.py
```diff
@@ -176,17 +179,25 @@ def report_progress(num_bytes):
...
+                        # WARNING: This line ensures that Hadoop knows that our process is not sitting in an infinite
+                        # loop.  Do not remove it.
+                        self.incr_counter('Event Export', 'Raw Bytes Written', len(value) + 1)
+                finally:
+                    outfile.close()
+        except IOError as err:
+            log.error("Error encountered while encrypting and gzipping Organization: %s file: %s Exception: %s",
+                      org_id, key_file_targets, err)
+            # This counter is set when there is an error during the generation of the encryption file for an
+            # organization for any reason, including encryption errors related to an expired GPG key
```
brianhw: nit: trailing period.

edx/analytics/tasks/export/event_exports.py
```diff
@@ -176,17 +179,25 @@ def report_progress(num_bytes):
...
+            with make_encrypted_file(output_file, key_file_targets, progress=report_progress,
+                                     hadoop_counter_callback=self.event_export_counter) as encrypted_output_file:
+                outfile = gzip.GzipFile(mode='wb', fileobj=encrypted_output_file)
+                try:
+                    for value in values:
+                        outfile.write(value.strip())
+                        outfile.write('\n')
+                        # WARNING: This line ensures that Hadoop knows that our process is not sitting in an infinite
+                        # loop.  Do not remove it.
+                        self.incr_counter('Event Export', 'Raw Bytes Written', len(value) + 1)
```
brianhw: nit: this cannot use self.event_export_counter as-is, but if it had a value added as an arg with 1 as a default, then it could be used here as well.   Effect is to centralize the key name.

edx/analytics/tasks/util/encrypt.py
```diff
@@ -44,12 +46,15 @@ def make_encrypted_file(output_file, key_file_targets, recipients=None, progress
         key_file_targets: a list of luigi.Target objects defining the gpg public key files to be loaded.
         recipients:  an optional list of recipients to be loaded.  If not specified, uses all loaded keys.
         progress:  a function that is called periodically as progress is made.
+        hadoop_counter_callback:  A callback to a function that can generate MR counters so that non-critical GPG
+            messages can be promoted to a visible section of the MR run log
```
brianhw: nit: trailing period.

edx/analytics/tasks/util/encrypt.py
```diff
@@ -67,25 +72,58 @@ def make_encrypted_file(output_file, key_file_targets, recipients=None, progress
...
+                        next_week = current_time + datetime.timedelta(days=7)
+                        key_expire = datetime.datetime.fromtimestamp(int(test_key["expires"]))
+
+                        if current_time > key_expire:
+                            # Ignore this error for now because a more serious and appropriate IOException will be
+                            # generated when the key is used for encryption
+                            log.error("Error key with fingerprint: '%s' and recipient '%s' has expired!!!",
+                                      key_fingerprint, test_key["uids"])
+                            for recipient in test_key["uids"]:
+                                hadoop_counter_callback("GPG Key for {} has expired".format(recipient))
```
brianhw: Is recipient here a list?  In the examples you gave in HipChat, this looks like one.  If so, I guess you could loop through them, and log them separately:   for recipient_list in test_key['uids'])  \n for recipient in recipient_list: \n ...

brianhw (approved):
LGTM.  👍     Just a few optional nits.

azafty468:
Whoa, hey this guy isn't ready for review yet.  I'm still testing the publication method, and look & feel in grafana before this guy should be examined.

edx/analytics/tasks/export/event_exports.py
azafty468: Great idea.  done!

edx/analytics/tasks/util/encrypt.py
azafty468: test_key["uids"] itself is a list of recipients.  In the examples I saw it was always a list of one but I'm not sure if this is always the case.  Each individual entry in test_key["uids"] will be a single recipient however.  Its been a few days and I'm not sure what I posted in HipChat, or if I did any editing on it beforehand.

azafty468:
All changes have been incorporated, pulling the trigger on this guy

PR #425, opened 2017-08-08
Speed up finance report

HassanJaveed84:
I've ran this and compared the numbers between our current workflow, I couldn't get an exact match as the current workflow last succeeded on 4th and I ran the optimized one on 7th, but the two result sets only differed by some extra data produced by the optimized version.

I plan to compare with data produced by the existing workflow for 7th with the one produced by this.

The problem I see is making sure we have a complete build every day or there could be missing data. We could also overwrite paypal/cybersource data for the last n days, similar to what we do for enrollments, module-engagement.

It completed in 3 hours.

### DIFF
--- a/edx/analytics/tasks/warehouse/financial/cybersource.py
+++ b/edx/analytics/tasks/warehouse/financial/cybersource.py
@@ -12,6 +12,7 @@ from luigi import date_interval
 from edx.analytics.tasks.util.hive import HivePartition, WarehouseMixin
 from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
 from edx.analytics.tasks.util.url import get_target_from_url, url_path_join
+from edx.analytics.tasks.common.pathutil import PathSelectionByDateIntervalTask
 
 log = logging.getLogger(__name__)
 
@@ -190,9 +191,6 @@ class DailyProcessFromCybersourceTask(PullFromCybersourceTaskMixin, luigi.Task):
 class IntervalPullFromCybersourceTask(PullFromCybersourceTaskMixin, WarehouseMixin, luigi.WrapperTask):
     """Determines a set of dates to pull, and requires them."""
 
-    interval = luigi.DateIntervalParameter(
-        default=None,
-    )
     interval_end = luigi.DateParameter(
         default=datetime.datetime.utcnow().date(),
         significant=False,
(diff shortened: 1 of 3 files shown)


edx/analytics/tasks/warehouse/financial/paypal.py
```diff
@@ -696,13 +697,23 @@ def __init__(self, *args, **kwargs):
...
     def requires(self):
-        for day in self.interval:
-            yield PaypalTransactionsByDayTask(
-                account_id=self.account_id,
-                output_root=self.output_root,
-                date=day,
-                overwrite=self.overwrite,
-            )
+        run_date = self.interval_end - datetime.timedelta(days=1)
+        selection_interval = date_interval.Custom(self.interval_start, run_date)
```
azafty468: It looks like you are replacing the functionality for self.interval here.  Can we remove it as a parameter then?

edx/analytics/tasks/warehouse/financial/cybersource.py
```diff
@@ -221,9 +222,19 @@ def requires(self):
             'overwrite': self.overwrite,
         }
 
-        for run_date in self.interval:
-            args['run_date'] = run_date
-            yield DailyProcessFromCybersourceTask(**args)
+        run_date = self.interval_end - datetime.timedelta(days=1)
+        selection_interval = date_interval.Custom(self.interval_start, run_date)
```
azafty468: Same comment as below

azafty468 (approved):
Other than the cleanup comments I think this looks good.

edx/analytics/tasks/warehouse/financial/cybersource.py
```diff
@@ -221,9 +222,19 @@ def requires(self):
...
+
+        yield PathSelectionByDateIntervalTask(
```
tobz: General question: what is this new task being run doing?  How does the existing `DailyProcessFromCybersourceTask` know what to do with the result of `PathSelectionByDateIntervalTask`?

HassanJaveed84: The way we have in master, IntervalPullFromCybersourceTask requires DailyProcessFromCybersourceTask for each date within the interval, and returns the flattened output. Task output existence check takes the most time.

The added PathSelectionByDateIntervalTask selects the historical data from source(assuming it already exists) and also requires DailyProcessFromCybersourceTask for the most recent day.
Then it returns the flattened output(of PathSelectionByDateIntervalTask & DailyProcessFromCybersourceTask).

HassanJaveed84:
Added validation task which we'll run daily/weekly for validation purporses. 
Added option to overwrite last n days of payment data.

edx/analytics/tasks/warehouse/financial/paypal.py
HassanJaveed84: Done.

edx/analytics/tasks/warehouse/financial/cybersource.py
HassanJaveed84: Done.

HassanJaveed84:
Added a validation task, which will be scheduled to run daily.

tobz (approved):
LGTM. 👍

tobz:
Just make sure you squash it.

PR #426, opened 2017-08-15
Default to None for hadoop counter callback.

HassanJaveed84:
We make use of make_encrypted_file method for obfuscation as well(where we don't provide a callback function). The workflow would fail if a key is expired or set to expire next week. 

We also make use of _import_key_files directly in test_encrypt(which currently fails as number of args don't match). 

This aims to fix both issues, open to ideas if this does not seem reasonable.


tobz (approved):
LGTM. 👍

tobz:
Just squash it before merging, obviously. :D

azafty468 (approved):
Looks like a great defensive addition 👍

PR #428, opened 2017-09-06
Use merchant close date to compute interval.

HassanJaveed84:
Default interval_end from config if present restricting look up to cybersource API. 

Only caveat I see here is that selection interval would go beyond the interval_end from config, but would still work correctly as it would only look up payment data files for the particular merchant.

### DIFF
--- a/edx/analytics/tasks/warehouse/financial/cybersource.py
+++ b/edx/analytics/tasks/warehouse/financial/cybersource.py
@@ -38,6 +38,11 @@ class PullFromCybersourceTaskMixin(OverwriteOutputMixin):
         self.password = config.get(section_name, 'password')
         self.interval_start = luigi.DateParameter().parse(config.get(section_name, 'interval_start'))
 
+        self.merchant_close_date = None
+        merchant_close_date = config.get(section_name, 'merchant_close_date', '')
+        if merchant_close_date:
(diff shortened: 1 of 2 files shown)


mulby (approved):

brianhw:
This looks like it will work.  But there are two things that I find funny about it.  One is that we rarely (if ever) have a parameter where the config file value takes precedence over the explicitly-passed value.  And second is the note you have about the selection interval.  

I'm wondering if it would work (as well) to have interval_close (or similar) added to PullFromCybersourceTaskMixin, to go with interval_start, indicating when the account "closed".   It would have to be optional, though, with a None value if it's not in the config.  (I can't remember if config.get() will fail or return None, but presumably the parse of None would fail.)  Then, in IntervalPullFromCybersourceTask, selection_interval can run to min(run_date, interval_close), and run_interval can run to min(interval_end, interval_close).   (I'm assuming we already know that a run_interval where the start is after the end is a no-op -- that's already built into your code change here.)  That way, if an interval_close is specified, it will only generate the data needed for both the selection and the run.

brianhw (approved):
LGTM (if it works :).

HassanJaveed84:
Found an issue with the code that checks for latest completion date. Fixed by adding include pattern that matches specific files only.

brianhw (approved):
Additional changes make sense. Good catch! 👍

PR #429, opened 2017-09-08
Incremental video

HassanJaveed84:
It took 1 hours 26 minutes to completion, overwriting last 3 days of viewing data.  
video table is exactly the same as non-incremental one. 
However, video_timeline table has 49 fewer rows than the non-incremental version. I expected incremental video_timeline to have more rows as I ran the workflow a few hours after video-timeline job had started. Currently investigating the cause. 

Some paths/table names have been changed for testing only.

### DIFF
--- a/edx/analytics/tasks/insights/video.py
+++ b/edx/analytics/tasks/insights/video.py
@@ -6,21 +6,24 @@ import math
 import re
 import textwrap
 import urllib
+import datetime
 
 import ciso8601
 import luigi
 from luigi import configuration
 from luigi.hive import HiveQueryTask
+from luigi.parameter import DateIntervalParameter
 
-from edx.analytics.tasks.common.mapreduce import MapReduceJobTask, MapReduceJobTaskMixin
+from edx.analytics.tasks.common.mapreduce import MapReduceJobTask, MultiOutputMapReduceJobTask, MapReduceJobTaskMixin
 from edx.analytics.tasks.common.mysql_load import MysqlInsertTask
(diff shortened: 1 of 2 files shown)


edx/analytics/tasks/insights/video.py
```diff
@@ -643,7 +745,7 @@ class VideoUsageTableTask(VideoTableDownstreamMixin, HiveTableTask):
 
     @property
     def table(self):
-        return 'video_usage'
+        return 'video_usage_test'
```
brianhw: I assume this is to facilitate testing.  Alternative is to run with different warehouse_path in config.

brianhw (commented):
Cursory review finds no offhand problems -- looks good.

HassanJaveed84:
Ready for a final review.

edx/analytics/tasks/insights/video.py
```diff
@@ -505,21 +606,20 @@ class VideoUsageTask(VideoTableDownstreamMixin, MapReduceJobTask):
...
+    def requires_local(self):
+        return UserVideoViewingByDateTask(
             mapreduce_engine=self.mapreduce_engine,
             n_reduce_tasks=self.n_reduce_tasks,
             source=self.source,
             interval=self.interval,
             pattern=self.pattern,
-            output_root=input_path,
+            overwrite_n_days=self.overwrite_n_days,
+            overwrite=True,
```
brianhw, line 617: Just curious if you think this will do the right thing when overwrite_n_days = 0 ...

brianhw (approved):
One small question, but non-blocking.  Looks good! 👍

PR #430, opened 2017-09-12
Upgrade setuptools, and other supporting packages.

brianhw:
The old pinned version of setuptools seemed to be breaking Travis builds.  This upgrades to the latest version of setuptools.

While in here changing dependencies, let the pinned values for supporting modules float to their most recent valid release.   This does not upgrade any of the explicitly required packages, only the packages that the required packages pull in.  

Test.txt should also be pinning the dependencies specified earlier, so that it doesn't install the latest of these packages (as it sometimes does, depending on the order of makefile calls).

This builds cleanly on Travis, and runs locally and on analyticstack.  Currently running acceptance tests.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -9,17 +9,24 @@ pycrypto==2.6.1 	# public domain
 # The following are dependencies pulled in by the above,
 # but are pinned here.
 
-cffi==1.9.1
-cryptography==1.7.1
+asn1crypto==0.22.0
+bcrypt==3.1.3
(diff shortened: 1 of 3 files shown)


brianhw:
This is ready for review.  Acceptance tests pass, as do several monitoring jobs on Jenkins.

azafty468 (approved):
All requirements and it runs?  I'm game!

HassanJaveed84 (approved):

PR #434, opened 2017-10-04
Added tasks to load warehouse data into BigQuery.

HassanJaveed84:
Question about the way I've defined the schema.
Is it okay to place technology specific code in record.py ? BigQuery schema definition is a bit different than our current mysql/hive/vertica stuff we have in record.py.

### DIFF
--- a/edx/analytics/tasks/common/bigquery_load.py
+++ b/edx/analytics/tasks/common/bigquery_load.py
@@ -1,0 +1,250 @@
+import logging
+import datetime
+import json
+import os
+import subprocess
+import tempfile
+import time
+import urlparse
+
+import luigi
+
+from google.cloud import bigquery
+from google.oauth2 import service_account
+from google.cloud.exceptions import NotFound
+
+from edx.analytics.tasks.util.url import ExternalURL, url_path_join
+from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
+from edx.analytics.tasks.util.hive import WarehouseMixin
+
+
+log = logging.getLogger(__name__)
+
+
+RETRY_LIMIT = 500
+WAIT_DURATION = 5
+
+def wait_for_job(job, check_error_result=True):
+    counter = 0
+    while True:
+        if counter == RETRY_LIMIT:
+            raise RuntimeError("Retry limit exceeded while waiting on job.")
+
+        job.reload()
+        if job.state == 'DONE':
+            if check_error_result and job.error_result:
+                raise RuntimeError(job.errors)
+            return
+
+        counter += 1
+        time.sleep(WAIT_DURATION)
+
+
+class BigQueryTarget(luigi.Target):
+
+    def __init__(self, credentials_target, dataset_id, table, update_id):
+        self.dataset_id = dataset_id
+        self.table = table
+        self.update_id = update_id
+        with credentials_target.open('r') as credentials_file:
+            json_creds = json.load(credentials_file)
+            self.project_id = json_creds['project_id']
+            credentials = service_account.Credentials.from_service_account_info(json_creds)
+            self.client = bigquery.Client(credentials=credentials, project=self.project_id)
+
+    def touch(self):
+        self.create_marker_table()
+        dataset = self.client.dataset(self.dataset_id)
+        table = dataset.table('table_updates')
+        table.reload() # Load the schema
+
+        # Use a tempfile for loading data into table_updates
+        # We deliberately don't use table.insert_data as we cannot use delete on
+        # a bigquery table with streaming inserts.
+        tmp = tempfile.NamedTemporaryFile(delete=False)
+        table_update_row  = (self.update_id, "{dataset}.{table}".format(dataset=self.dataset_id, table=self.table))
+        tmp.write(','.join(table_update_row))
+        tmp.close()
+
+        # table.upload_from_file requires the file to be opened in 'rb' mode.
+        with open(tmp.name, 'rb') as source_file:
+            job = table.upload_from_file(source_file, source_format='text/csv')
+
+        try:
+            wait_for_job(job)
+        finally:
+            tmp.close()
+            os.unlink(tmp.name)
+
+    def create_marker_table(self):
+        marker_table_schema = [
+            bigquery.SchemaField('update_id', 'STRING'),
+            bigquery.SchemaField('target_table', 'STRING'),
+        ]
+
+        dataset = self.client.dataset(self.dataset_id)
+        table = dataset.table('table_updates', marker_table_schema)
+        if not table.exists():
+            table.create()
+
+    def exists(self):
+        query_string = "SELECT 1 FROM {dataset}.table_updates where update_id = '{update_id}'".format(
+            dataset=self.dataset_id,
+            update_id=self.update_id,
+        )
+        log.debug(query_string)
+
+        query = self.client.run_sync_query(query_string)
+
+        try:
+            query.run()
+        except NotFound:
+            return False
+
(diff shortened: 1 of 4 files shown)


azafty468:
So normally I wouldn't put technology specific functions into our generalized classes, but that level of generalized purity sometimes doesn't work in practice.  I'm fine with what you've done in the code here.

azafty468 (approved):
Looks good to me!  Do we know if we need to put a cool down or wait period after deleting a table (on the overwrite)?  Or is the delete immediately reflected in the API?

edx/analytics/tasks/common/bigquery_load.py
```diff
@@ -0,0 +1,231 @@
+import logging
+import datetime
+import os
+import luigi
+import json
+import time
+import subprocess
+import urlparse
+import tempfile
```
brianhw: nit: imports not in order....

edx/analytics/tasks/common/bigquery_load.py
```diff
@@ -0,0 +1,231 @@
...
+            raise RuntimeError("Retry limit exceeded while waiting on job.")
+
+        job.reload()
+        if job.state == 'DONE':
+            if job.error_result:
+                raise RuntimeError(job.errors)
+            return
+
+        counter += 1
+        time.sleep(1)
```
brianhw: What's the normal runtime of these jobs?  I ran with a sleep of 5, which seemed like adequate granularity.

edx/analytics/tasks/common/bigquery_load.py
```diff
@@ -0,0 +1,231 @@
...
+def wait_for_job(job):
+    counter = 0
+    while True:
+        if counter == RETRY_LIMIT:
+            raise RuntimeError("Retry limit exceeded while waiting on job.")
+
+        job.reload()
+        if job.state == 'DONE':
+            if job.error_result:
+                raise RuntimeError(job.errors)
```
brianhw: I found that I wanted more control over the handling of job failures, so it was easier for me to move these two lines to the code after the wait_for_job() was called.  We've still got the job object, so it's no different.  And then this really is *only* waiting for the job.

edx/analytics/tasks/common/bigquery_load.py
```diff
@@ -0,0 +1,231 @@
...
+            return
+
+        counter += 1
+        time.sleep(1)
+
+
+class BigQueryTarget(luigi.Target):
+
+    def __init__(self, credentials_target, dataset_id, table, update_id):
```
brianhw, line 45: I have no idea, but is there a need for a super() call here?

edx/analytics/tasks/common/bigquery_load.py
brianhw: Maybe that should just be an option, then.  check_error_result=True by default, and I can then disable it.

edx/analytics/tasks/common/bigquery_load.py
```diff
@@ -0,0 +1,231 @@
...
+
+        source_path = self.input()['source'].path
+        parsed_url = urlparse.urlparse(source_path)
+        destination_path = url_path_join('gs://{}'.format(parsed_url.netloc), parsed_url.path)
+
+        if self.is_file(source_path):
+            return_code = subprocess.call(['gsutil', 'cp', source_path, destination_path])
+        else:
+            log.debug(" ".join(['gsutil', '-m', 'rsync', source_path, destination_path]))
+            return_code = subprocess.call(['gsutil', '-m', 'rsync', source_path, destination_path])
```
brianhw, line 191: Should there be something here that checks the return_code to see if it failed or succeeded?  Presumably we don't want to call load_table_from_storage if the data is not in google storage.

edx/analytics/tasks/common/bigquery_load.py
```diff
@@ -0,0 +1,231 @@
...
+                update_id=self.update_id(),
+            )
+
+        return self.output_target
+
+    def update_id(self):
+        return '{task_name}(date={key})'.format(task_name=self.task_family, key=self.date.isoformat())
+
+    def is_file(self, path):
+        if path.endswith('.tsv') or path.endswith('.csv'):
```
brianhw: This is adequate for this current use, but presumably it should also handle .gz as well?

edx/analytics/tasks/common/bigquery_load.py
```diff
@@ -0,0 +1,231 @@
...
+            load_uri
+        )
+        job.field_delimiter = self.field_delimiter
+        job.quote_character = self.quote_character
+        job.null_marker = self.null_marker
+
+        log.debug("Starting BigQuery Load job.")
+        job.begin()
+        wait_for_job(job)
+
```
brianhw: Might be nice to have some statistics logged here.  Something like:
    try:
        log.debug("   Load job started: %s   ended: %s   input_files: %s  output_rows: %s  output_bytes: %s",
            job.started, job.ended, job.input_files, job.output_rows, job.output_bytes)
    except KeyError as keyerr:
        log.debug("   Load job started: %s   ended: %s   No load stats.", job.started, job.ended)

    if job.error_result:
        for error in job.errors:
            log.debug("   Load error: %s", error)
    else:
        log.debug("   No errors encountered!")

edx/analytics/tasks/common/bigquery_load.py
brianhw: As I mentioned in Hipchat, I assume we can create a Google storage Client the same way we do for BigQuery. Then we just get a bucket, define a blob in it, and call upload_from_filename. Is what's missing the globbing?  Best I can tell, gsutil is required if we were streaming, and didn't have data in separate files.

brianhw (approved):
Looks good overall.  Mostly nits.  

Just a heads-up, however, that I'm going to have to figure out how to refactor some of this in order to work with events.  For them, we would be writing to partitions instead of tables.

edx/analytics/tasks/common/bigquery_load.py
HassanJaveed84: To clarify, by order do you mean alphabetically ? I've always been confused on how we are supposed to order imports

edx/analytics/tasks/common/bigquery_load.py
HassanJaveed84: A few seconds, let's default to 5.

edx/analytics/tasks/common/bigquery_load.py
HassanJaveed84: No, luigi.Target does not define a constructor and is derived from 'object'.

edx/analytics/tasks/common/bigquery_load.py
HassanJaveed84: Good catch, I'll add a check for return_code value.

edx/analytics/tasks/common/bigquery_load.py
```diff
@@ -0,0 +1,239 @@
...
+        destination_path = url_path_join('gs://{}'.format(parsed_url.netloc), parsed_url.path)
+
+        if self.is_file(source_path):
+            return_code = subprocess.call(['gsutil', 'cp', source_path, destination_path])
+        else:
+            log.debug(" ".join(['gsutil', '-m', 'rsync', source_path, destination_path]))
+            return_code = subprocess.call(['gsutil', '-m', 'rsync', source_path, destination_path])
+
+        if return_code != 0:
+            raise RuntimeError('Error while syncing {source} to {destination}'.foramt(
```
brianhw: foramt => format

PR #436, opened 2017-10-13
Brian/also load json events

brianhw:
This is the merging of what was in brian/load-json-events into master, so that it can coexist with outputting regular (non-JSON) events to Vertica.  

This includes an incremental improvement to the non-JSON events, fixing some early shortcomings by looking for user_id, username, and course_id values in additional places of Segment events.  The version number has been updated to 0.2.4, but no new columns are added so the data can coexist.  (In four months, it will be completely up-to-date.)

Unit tests pass, and regular events compare as expected.  Currently verifying JSON events and running the existing acceptance test.   BigQuery-related work will be done in a separate branch.

### DIFF
--- a/edx/analytics/tasks/common/bigquery_load.py
+++ b/edx/analytics/tasks/common/bigquery_load.py
@@ -9,16 +9,22 @@ import urlparse
 
 import luigi
 
-from google.cloud import bigquery
-from google.oauth2 import service_account
-from google.cloud.exceptions import NotFound
-
 from edx.analytics.tasks.util.url import ExternalURL, url_path_join
 from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
 from edx.analytics.tasks.util.hive import WarehouseMixin
 
-
 log = logging.getLogger(__name__)
+
+try:
+    from google.cloud import bigquery
+    from google.oauth2 import service_account
+    from google.cloud.exceptions import NotFound
+    bigquery_available = True  # pylint: disable=invalid-name
+except ImportError:
+    log.warn('Unable to import Bigquery libraries')
+    # On hadoop slave nodes we don't have bigquery libraries installed,
+    # so just fail noisily if we attempt to use these libraries there.
+    bigquery_available = False  # pylint: disable=invalid-name
 
 
 RETRY_LIMIT = 500
@@ -76,6 +82,11 @@ class BigQueryTarget(luigi.Target):
             tmp.close()
             os.unlink(tmp.name)
 
+    def marker_table_exists(self):
+        dataset = self.client.dataset(self.dataset_id)
+        table = dataset.table('table_updates')
+        return table.exists()
+
     def create_marker_table(self):
         marker_table_schema = [
             bigquery.SchemaField('update_id', 'STRING'),
@@ -87,10 +98,36 @@ class BigQueryTarget(luigi.Target):
         if not table.exists():
             table.create()
 
+    def clear_marker_table(self):
+        if not self.marker_table_exists():
+            return
+
+        query_string = "DELETE {dataset}.table_updates WHERE target_table='{dataset}.{table}'".format(
+            dataset=self.dataset_id, table=self.table
+        )
+        query = self.client.run_sync_query(query_string)
+        query.use_legacy_sql = False
+        query.run()
+
+    def clear_marker_table_entry(self):
+        if not self.marker_table_exists():
+            return
+
+        query_string = "DELETE {dataset}.table_updates WHERE update_id='{update_id}' AND target_table='{dataset}.{table}'".format(
+            dataset=self.dataset_id, update_id=self.update_id, table=self.table
+        )
+        query = self.client.run_sync_query(query_string)
+        query.use_legacy_sql = False
+        query.run()
+
     def exists(self):
-        query_string = "SELECT 1 FROM {dataset}.table_updates where update_id = '{update_id}'".format(
+        if not self.marker_table_exists():
+            return False
+
+        query_string = "SELECT 1 FROM {dataset}.table_updates WHERE update_id='{update_id}' AND target_table='{dataset}.{table}'".format(
             dataset=self.dataset_id,
             update_id=self.update_id,
+            table=self.table,
         )
         log.debug(query_string)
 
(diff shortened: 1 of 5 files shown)


brianhw:
Acceptance tests pass, and JSON files match.    Ready for review.

brianhw:
By the way, in order to do the next phase of this -- to put this data into BigQuery -- I need to get this PR merged or Hassan's PR 434, and then either get the other branch rebased so I can branch off of it.  Better yet would be to get both PRs merged, so then I can branch off of master instead of a branch.

edx/analytics/tasks/warehouse/load_internal_reporting_events.py
```diff
@@ -862,7 +1026,21 @@ def _get_time_from_segment_event(self, event, key):
...
+
+        if 'requestTime' in event:
+            self.incr_counter(self.counter_category_name, 'Supplementing requestTime for receivedAt', 1)
+            return self._get_time_from_segment_event(event, 'requestTime')
+
+        if 'timestamp' in event:
+            self.incr_counter(self.counter_category_name, 'Supplementing timestamp for receivedAt', 1)
+            return self._get_time_from_segment_event(event, 'timestamp')
+
+        self.incr_counter(self.counter_category_name, 'Neither timestamp, receivedAt nor requestTime present', 1)
```
tobz: While the metrics system/Hadoop can technically deal with arbitrary strings, can we rename this one to something a little more succinct? `No Timestamp Available` or something, maybe?

Mostly, I'm just not sure the name signals properly that this is effectively an error to have no available event timestamp.

tobz (approved):
LGTM, with the one nit about the counter name.

edx/analytics/tasks/warehouse/load_internal_reporting_events.py
```diff
@@ -364,7 +374,89 @@ class EventRecord(SparseRecord):
...
+        'For implicit events, this should be the full URL that the request was for.'
+    )
+
+    # use the length (and name) from segment version (referrer), and write tracking log 'referer' here as well.
+    referrer = StringField(length=8191, nullable=True, description='The HTTP referrer - also as a full URL.')
+
+    # was project:
+    source = StringField(length=255, nullable=False, description='The segment.com project the event was sent to.')
+
+    input_file = StringField(length=255, nullable=True, description='The full URL of the file that contains this event in S3.')
```
HassanJaveed84: Should this be nullable=False ? Are there cases where get_map_input_file() does not return the file name ?

edx/analytics/tasks/warehouse/load_internal_reporting_events.py
brianhw: Makes sense.  I'll fix it in the follow-up PR, if that's okay.

edx/analytics/tasks/warehouse/load_internal_reporting_events.py
brianhw: I don't know offhand, but I can certainly run an expensive query.  There are always values input_file in Vertica, at least.  But I'm willing to change it.

However, I wasn't planning on reloading all the JSON events in BigQuery, so any changes in schema would only be for the next time we define the schema.   We might do this were we to want to add additional columns to the JSON record for our convenience -- it's really a pain to have to write JSON queries to check multiple locations in a raw event for pretty common values, for example.   But then I think we would be defining a new record anyway.

edx/analytics/tasks/warehouse/load_internal_reporting_events.py
tobz: Yup, 100% cool with that.

HassanJaveed84 (approved):
LGTM

edx/analytics/tasks/common/bigquery_load.py
```diff
@@ -153,37 +212,43 @@ def create_table(self, client):
...
-
-        dataset = client.dataset(self.dataset_id)
-        table = dataset.table(self.table, self.schema)
-
-        source_path = self.input()['source'].path
+            if self.partitioning_type:
+                # Delete only the specific partition, and clear the marker only for the partition.
+                if table.exists():
+                    partition = self._get_table_partition(dataset, table)
+                    partition.delete()
```
brianhw: Hassan asked on the other PR whether this does the right thing if the partition doesn't actually exist.  That's a good question.  I think I tested that, but I'll check on that.

brianhw:
Okay, final review.   I addressed Hassan's question from the other PR by adding a comment into the code, and I squashed the commits down to one big one.

edx/analytics/tasks/common/bigquery_load.py
```diff
@@ -9,17 +9,23 @@
...
+try:
+    from google.cloud import bigquery
+    from google.oauth2 import service_account
+    from google.cloud.exceptions import NotFound
+    bigquery_available = True  # pylint: disable=invalid-name
+except ImportError:
+    log.warn('Unable to import Bigquery libraries')
+    # On hadoop slave nodes we don't have bigquery libraries installed,
+    # so just fail noisily if we attempt to use these libraries there.
+    bigquery_available = False  # pylint: disable=invalid-name
```
tobz, line 27: I like this, matching how we handle Vertica.  👍

brianhw: Cribbed from how we handle Vertica.  (No accident. : )

brianhw:
To elevate the comment on the commit:

It coexists with regular event output, and is controlled by an optional parameter.  By default it runs with event_record_type equal to 'EventRecord', but can be overridden by running with --event-record-type 'JsonEventRecord'.

Includes bug fix to timestamp handling:  Add validation screening dates < 1900.

Also includes support for event loading to BigQuery, by adding support for partitioning to bigquery_load.

* Use records for warehouse loading where defined.
* Check bigquery availability in load code.

Add support for loading to S3 by interval or by date.  PerDate loading checks whether the data already exists, which is good for incremental runs. Bulk loading just runs over an interval, and assumes the data isn't already present on S3.  This is better for processing many days more efficiently.

To address issue with loading into BigQuery, null characters in column values are encoded as the string '\0'.

HassanJaveed84 (approved):
LGTM 👍

PR #437, opened 2017-10-17
Add support for BigQuery event loading

brianhw:
Involves refactoring existing loading code into partition-based and non-partition-based tasks. 

Regular warehouse loading continues to use non-partition-based tasks.   The event loading uses partition-based tasks, for which overwrite is at the partition level rather than the table level.

### DIFF
--- a/edx/analytics/tasks/common/bigquery_load.py
+++ b/edx/analytics/tasks/common/bigquery_load.py
@@ -9,16 +9,22 @@ import urlparse
 
 import luigi
 
-from google.cloud import bigquery
-from google.oauth2 import service_account
-from google.cloud.exceptions import NotFound
-
 from edx.analytics.tasks.util.url import ExternalURL, url_path_join
 from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
 from edx.analytics.tasks.util.hive import WarehouseMixin
 
-
 log = logging.getLogger(__name__)
+
+try:
+    from google.cloud import bigquery
+    from google.oauth2 import service_account
+    from google.cloud.exceptions import NotFound
+    bigquery_available = True  # pylint: disable=invalid-name
+except ImportError:
+    log.warn('Unable to import Bigquery libraries')
+    # On hadoop slave nodes we don't have bigquery libraries installed,
+    # so just fail noisily if we attempt to use these libraries there.
+    bigquery_available = False  # pylint: disable=invalid-name
 
 
 RETRY_LIMIT = 500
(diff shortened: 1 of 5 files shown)


edx/analytics/tasks/common/bigquery_load.py
```diff
@@ -248,3 +287,47 @@ def is_file(self, path):
...
+        self.init_copy(client)
+        self.create_table(client, partitioning_type='DAY')
+
+        dataset = client.dataset(self.dataset_id)
+        table = dataset.table(self.table, self.schema)
+
+        source_path = self.input()['source'].path
+        destination_path = self._get_destination_from_source(source_path)
+        self._copy_data_to_gs(source_path, destination_path)
+        load_uri = self._get_load_url_from_destination(destination_path)
```
tobz: I feel like everything above, and including, this line for the method could be abstracted and used between the non-partitioning and partitioning job.

Maybe just take kwargs that correspond to table options and pass them through?

It's not huge, but, it does feel like a lot of duplication for what boils down to a single method to get the partition name, and passing a flag to enable it... everything else is effectively the same between the two versions.

tobz (changes requested):

edx/analytics/tasks/common/bigquery_load.py
brianhw: Yes, I agree.  When I started, there was a lot more different code, but I rearranged some of the base class as well, so the differences shrunk.  The only real remaining impediment is the adding of date as a parameter to the base class, but I think it should be there anyway (since it's actually used).

brianhw: Oh, but I think I would just parameterize it on a partition_type property.  If specified, then use that as a switch.

tobz: Yeah, whatever way you want to approach it, really.  The code duplication is really the only thing that makes me go "eeps".

brianhw:
Overwrite now works, and I've also gotten production to work (again).

edx/analytics/tasks/common/bigquery_load.py
brianhw: I think I've addressed that now.   There is a single class, differentiated by a property.

edx/analytics/tasks/common/bigquery_load.py
```diff
@@ -153,37 +212,43 @@ def create_table(self, client):
...
-        self.create_table(client)
-
-        dataset = client.dataset(self.dataset_id)
-        table = dataset.table(self.table, self.schema)
-
-        source_path = self.input()['source'].path
+            if self.partitioning_type:
+                # Delete only the specific partition, and clear the marker only for the partition.
+                if table.exists():
+                    partition = self._get_table_partition(dataset, table)
```
HassanJaveed84, line 231: Wondering what we would get here if the partition does not exist yet ?

HassanJaveed84 (approved):
One question, otherwise LGTM 👍

PR #438, opened 2017-10-17
Check Sqoop job completion before running.

brianhw:
Also check for the marker file when checking for completeness.

### DIFF
--- a/edx/analytics/tasks/common/sqoop.py
+++ b/edx/analytics/tasks/common/sqoop.py
@@ -181,6 +181,16 @@ class SqoopImportTask(OverwriteOutputMixin, SqoopImportMixin, luigi.hadoop.BaseH
             cred = json.load(credentials_file)
         return cred
 
+    def complete(self):
+        """
+        Wrap Task.complete() to check for metadata marker file as well as data.
+        """
+        data_complete = super(SqoopImportTask, self).complete()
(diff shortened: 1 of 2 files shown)


brianhw:
Unit tests pass, as does test_internal_reporting_database acceptance test.   Other acceptance tests are currently running.

tobz (approved):
LGTM if tests pass. :shipit:

HassanJaveed84 (approved):

brianhw:
Addresses both DE-99 and DE-100.

brianhw:
DE-99 had been tested this morning, with the success of the user-activity run.  I have now fixed the code so that DE-100 should also be properly addressed.   And I tested it by running a Sqoop task twice.   First I ran with the Sqoop initialization step removed from job's jenkins configuration so that Sqoop failed to run.  It produced the directory and a _metadata file, but no output.  I then reran with the initialization step properly included, and the sqoop task properly deleted the existing partial data and regenerated complete data.  

Once approved, I'll squash the two commits down into a single one before merging.

brianhw:
Oh, I also reran the job after it ran successfully, and it correctly detected that it had already succeeded. 

Ready for re-review.

azafty468 (approved):
Looks good!

PR #440, opened 2017-10-24
Inspect traceback for handling end of output error.

HassanJaveed84:


edx/analytics/tasks/common/mysql_load.py
```diff
@@ -336,6 +338,17 @@ def run(self):
 
             # commit only if both operations completed successfully.
             connection.commit()
+        except RuntimeError:
+            exc_type, exc_value, exc_traceback = sys.exc_info()
+            traceback_list =  traceback.format_exception(exc_type, exc_value, exc_traceback)
+            log.debug("========================")
+            log.debug(traceback_list)
+            log.debug("========================")
+            if "__iter__\n    self._finish()" in traceback_list[5]:
```
HassanJaveed84: Do we want to simplify this and only look for the _finish method call in the whole traceback, or do we want it to be really specific ?

Also I don't see any other place we can try catching the exception other than here.

tobz (approved):
LGTM, since we know it works in practice now 👍

brianhw (changes requested):
Code looks good to me.  I would like to see a comment inside the except clause that briefly explains why we are doing this, because it's going to be really cryptic for anyone later picking up this code.

brianhw:
Oh, and love those commit comments..... : )

HassanJaveed84:
Heh, squashed. Also added a comment explaining things. Does it look good ?

brianhw (approved):
LGTM.  👍

brianhw:
Turns out I'm running into similar issues on loading sailthru data to Vertica.  Looking forward to having this on master so I can rebase on it.

PR #441, opened 2017-10-25
Translation of MySQL's LONGBLOB to Vertica's LONG VARBINARY.  

azafty468:
Data loss is possible here, LONGBLOB's storage space is considerably larger than Vertica's LONG VARBINARY.  However this was the best option.


tobz (approved):
LGTM. 👍

PR #444, opened 2017-11-09
WIP: Incremental user activity

HassanJaveed84:


edx/analytics/tasks/insights/user_activity.py
```diff
@@ -143,18 +154,198 @@ def columns(self):
...
+    def hive_table_task(self):
+        return CourseActivityTableTask(
+            warehouse_path=self.warehouse_path,
+        )
+
+    def requires(self):
+        yield (
+            CourseActivityTableTask(
+                warehouse_path=self.warehouse_path,
+            ),
```
brianhw, line 275: Just curious, is this added automatically from hive_table_task?

edx/analytics/tasks/insights/user_activity.py
```diff
@@ -120,20 +129,22 @@ def reducer(self, key, values):
...
+        return super(UserActivityTask, self).run()
 
 
-class UserActivityTableTask(UserActivityDownstreamMixin, HiveTableTask):
-    """Hive table that stores the set of users active in each course over time."""
+class UserActivityTableTask(BareHiveTableTask):
 
     @property
     def table(self):
         return 'user_activity_daily'
```
brianhw, line 141: In terms of changing the definition of this table for people upgrading, it might be prudent to change the name of the table.  On the other hand, this is a really appropriate name.

edx/analytics/tasks/insights/user_activity.py
```diff
@@ -143,18 +154,198 @@ def columns(self):
...
+        return [
+            ('course_id', 'STRING'),
+            ('interval_start', 'TIMESTAMP'),
+            ('interval_end', 'TIMESTAMP'),
+            ('label', 'STRING'),
+            ('count', 'INT'),
+        ]
+
+
+class CourseActivityPartitionTask(UserActivityDownstreamMixin, HivePartitionTask):
```
brianhw: Would  "CourseWeeklyActivityPartitionTask" be more descriptive?

edx/analytics/tasks/insights/user_activity.py
```diff
@@ -143,18 +154,198 @@ def columns(self):
...
+                warehouse_path=self.warehouse_path,
+                overwrite_n_days=0,
+            ),
+            CalendarTableTask(
+                warehouse_path=self.warehouse_path,
+            )
+        )
+
+    def remove_output_on_overwrite(self):
+        super(HivePartitionTask, self).remove_output_on_overwrite()
```
brianhw: Wow!  What is this doing?  Looks like it needs a comment, since it's doing something tricky.  Is it skipping the implementation of remove_output_on_overwrite() in HivePartitionTask in order to do some other implementation?

edx/analytics/tasks/insights/user_activity.py
```diff
@@ -143,18 +154,198 @@ def columns(self):
...
+            ('interval_end',)
+        ]
+
+    @property
+    def insert_source_task(self):
+        return CourseActivityPartitionTask(
+            warehouse_path=self.warehouse_path,
+            interval=self.interval,
+            n_reduce_tasks=self.n_reduce_tasks,
+            overwrite=True,
```
brianhw: Overwrite is always a challenge to define and implement.  So if this is always hardcoded to overwrite, it will always re-query the activity for the specified week.  And will also recalculate some number of days of the week's data.  I think there's something not quite right here.

edx/analytics/tasks/insights/user_activity.py
```diff
@@ -143,18 +154,198 @@ def columns(self):
...
+                warehouse_path=self.warehouse_path,
+                overwrite=should_overwrite,
+            )
+
+
+class CourseActivityTableTask(BareHiveTableTask):
+
+    @property
+    def table(self):
+        return 'course_activity_test'
```
brianhw: I assume this is a temporary name?  Maybe put a TODO on this....

edx/analytics/tasks/insights/user_activity.py
```diff
@@ -143,18 +154,198 @@ def columns(self):
...
+    def requires(self):
+        yield (
+            CourseActivityTableTask(
+                warehouse_path=self.warehouse_path,
+            ),
+            UserActivityIntervalTask(
+                interval=self.interval,
+                n_reduce_tasks=self.n_reduce_tasks,
+                warehouse_path=self.warehouse_path,
+                overwrite_n_days=0,
```
brianhw: I find this confusing.  When is this non-zero, and how does it get that way?

edx/analytics/tasks/insights/user_activity.py
```diff
@@ -143,18 +154,198 @@ def columns(self):
...
+
+
+class CourseActivityWorkflow(WeeklyIntervalMixin, UserActivityDownstreamMixin, luigi.WrapperTask):
+
+    def requires(self):
+        yield InsertToMysqlCourseActivityTask(
+            interval=self.interval,
+            n_reduce_tasks=self.n_reduce_tasks,
+            warehouse_path=self.warehouse_path,
+            #credentials=self.credentials,
```
brianhw: Is this not settable here?  Are we missing a downstream mixin for Mysql?

edx/analytics/tasks/warehouse/load_internal_reporting_user_activity.py
```diff
@@ -48,22 +41,135 @@ def columns(self):
...
+        )
+
+        return query
+
+    def requires(self):
+        yield (
+            ImportAuthUserTask(overwrite=False, destination=self.warehouse_path),
+            InternalReportingUserActivityTableTask(
+                warehouse_path=self.warehouse_path,
+                overwrite=self.overwrite,
```
brianhw: Here it includes overwrite, but not in hive_table_task.   If the latter is automatically included in requires(), then it's unclear which version wins.  (I can't remember if overwrite is "significant".  I think it isn't.)

edx/analytics/tasks/warehouse/load_internal_reporting_user_activity.py
```diff
@@ -48,22 +41,135 @@ def columns(self):
...
+        return InternalReportingUserActivityPartitionTask(
+            date=self.date,
+            n_reduce_tasks=self.n_reduce_tasks,
+            warehouse_path=self.warehouse_path,
+            overwrite=self.overwrite,
+        )
+
+    @property
+    def table(self):
+        return 'f_user_activity_test'
```
brianhw, line 133: Add a TODO to update the table name.

edx/analytics/tasks/warehouse/load_internal_reporting_user_activity.py
```diff
@@ -48,22 +41,135 @@ def columns(self):
...
+
+    @property
+    def auto_primary_key(self):
+        return ('row_number', 'AUTO_INCREMENT')
```
brianhw, line 137: I'm just curious why we ever added this column to this table....

brianhw (commented):
I think it looks good.  I confess I've not thought through the overwrite behavior, though, and I expect it's pretty complicated.  But I'm glad that there's only a single user_activity_daily table serving both -- I think that makes more sense, assuming it doesn't screw up the overwrite logic too much.

edx/analytics/tasks/insights/user_activity.py
HassanJaveed84: HivePartitionTask defines requires as hive_table_task and data_task(if any). I've overridden the requires method here.
I'll change it to 'yield self.hive_table_task'

edx/analytics/tasks/insights/user_activity.py
HassanJaveed84: HivePartitionTask overrides remove_output_on_overwrite() and only sets self.attempted_removal = True.
However, in order to clear the actual data we need to call self.output().remove(). 

This line calls parent's method(remove_output_on_overwrite) of HivePartitionTask, which is defined in  OverwriteOutputMixin. Didn't see a better way of achieving the output removal without duplicating code.

brianhw: So that definitely would benefit from a comment, capturing what you just wrote for posterity.

azafty468 (approved):
Are there any tests that need to be updated for this, either acceptance or unit? And do we even have a good testing strategy for incremental vertica data?

HassanJaveed84:
So I've dumped this PR, currently working on another branch with a much simpler approach. I'll be putting up the PR today.

PR #448, opened 2017-11-14
Bump boto to latest version due to AWS EMR SSL/TLS update.

UmanShahzad:
According to an email from AWS:

> Amazon EMR is in the process of updating the SSL/TLS certificates it uses to secure communications with customers, with the changes scheduled to be made on December 3, 2017 4:00 PM PST. 

and, the important part for us:

> If you are using a version of the Python AWS SDK or CLI released before February 5, 2015, you must upgrade.

This PR bumps our use of `boto` to the latest version in preparation for this change.

### DIFF
--- a/edx/analytics/tasks/launchers/remote.py
+++ b/edx/analytics/tasks/launchers/remote.py
@@ -2,6 +2,7 @@
 """Execute tasks on a remote EMR cluster."""
 
 import argparse
+from boto.pyami.config import BotoConfigPath, UserConfigPath as BotoUserConfigPath
 import json
 import os
 import pipes
@@ -10,6 +11,8 @@ import sys
 from urlparse import urlparse, parse_qsl
 import uuid
 
+
+BOTO_FILE_PATH = os.path.join(sys.prefix, '.boto')
 STATIC_FILES_PATH = os.path.join(sys.prefix, 'share', 'edx.analytics.tasks')
 EC2_INVENTORY_PATH = os.path.join(STATIC_FILES_PATH, 'ec2.py')
 ANSIBLE_MAX_RETRY = 3
@@ -112,6 +115,7 @@ def run_task_playbook(inventory, arguments, uid):
     sudo_user = arguments.sudo_user
 
     env_vars = {}
+    env_vars['BOTO_PATH'] = os.pathsep.join([BotoConfigPath, BotoUserConfigPath, BOTO_FILE_PATH])
     if arguments.workflow_profiler:
         env_vars['WORKFLOW_PROFILER'] = arguments.workflow_profiler
         env_vars['WORKFLOW_PROFILER_PATH'] = log_dir
--- a/edx/analytics/tasks/launchers/remote.py
+++ b/edx/analytics/tasks/launchers/remote.py
@@ -2,7 +2,7 @@
 """Execute tasks on a remote EMR cluster."""
 
 import argparse
-from boto.pyami.config import BotoConfigPath, UserConfigPath as BotoUserConfigPath
+from boto.pyami.config import BotoConfigPath
 import json
 import os
 import pipes
@@ -12,7 +12,6 @@ from urlparse import urlparse, parse_qsl
 import uuid
 
 
-BOTO_FILE_PATH = os.path.join(sys.prefix, '.boto')
 STATIC_FILES_PATH = os.path.join(sys.prefix, 'share', 'edx.analytics.tasks')
 EC2_INVENTORY_PATH = os.path.join(STATIC_FILES_PATH, 'ec2.py')
 ANSIBLE_MAX_RETRY = 3
@@ -111,11 +110,12 @@ def run_task_playbook(inventory, arguments, uid):
 
     data_dir = os.path.join(REMOTE_DATA_DIR, uid)
     code_dir = os.path.join(data_dir, REMOTE_CODE_DIR_BASE)
+    boto_config_path = os.path.join(code_dir, '.boto')
+    boto_user_config_path = os.path.join(os.path.expanduser('~{}'.format(arguments.sudo_user)), '.boto')
     log_dir = os.path.join(REMOTE_LOG_DIR, uid)
-    sudo_user = arguments.sudo_user
 
     env_vars = {}
-    env_vars['BOTO_PATH'] = os.pathsep.join([BotoConfigPath, BotoUserConfigPath, BOTO_FILE_PATH])
+    env_vars['BOTO_PATH'] = os.pathsep.join([BotoConfigPath, boto_user_config_path, boto_config_path])
     if arguments.workflow_profiler:
         env_vars['WORKFLOW_PROFILER'] = arguments.workflow_profiler
         env_vars['WORKFLOW_PROFILER_PATH'] = log_dir
@@ -127,10 +127,8 @@ def run_task_playbook(inventory, arguments, uid):
         data_dir=data_dir,
         code_dir=code_dir,
         task_arguments=' '.join(arguments.launch_task_arguments),
-        log_dir=log_dir,
         bg='nohup ' if not arguments.wait else '',
         end_bg=' &' if not arguments.wait else '',
-        sudo_user=sudo_user,
     )
 
     result = run_remote_shell(inventory, arguments, command)
--- a/edx/analytics/tasks/util/tests/test_aws_elasticsearch_connection.py
+++ b/edx/analytics/tasks/util/tests/test_aws_elasticsearch_connection.py
@@ -36,7 +36,7 @@ class AwsElasticsearchConnectionTests(TestCase):
         self.assertTrue('my_access_key' in auth_header)
 
     def test_timeout(self):
-        def fake_connection(_address):
+        def fake_connection(_address, timeout=None):
             """Fail immediately with a socket timeout."""
             raise socket.timeout('fake error')
         socket.create_connection = fake_connection
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -1,10 +1,10 @@
 # These packages are required for bootstrapping.
 
 ansible==1.4.5 --no-binary ansible         # GPL v3 License
-boto==2.22.1            # MIT
-ecdsa==0.13 		# MIT
-Jinja2==2.8.1		# BSD
-pycrypto==2.6.1 	# public domain
+boto==2.48.0        # MIT
+ecdsa==0.13         # MIT
+Jinja2==2.8.1       # BSD
+pycrypto==2.6.1     # public domain
 
 # The following are dependencies pulled in by the above,
 # but are pinned here.
--- a/share/boto.cfg
+++ b/share/boto.cfg
@@ -1,3 +1,6 @@
 [Boto]
 metadata_service_timeout = 60.0
 metadata_service_num_attempts = 60
+
+[s3]
+host = s3.amazonaws.com


openedx-webhooks:
Thanks for the pull request, @UmanShahzad! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-analytics-pipeline&number=448"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

UmanShahzad:
@pomegranited @brianhw The code that checks for the host var is the following:

(ref https://github.com/boto/boto/blob/develop/boto/s3/connection.py#L176-L182, commit @ https://github.com/boto/boto/commit/221699bc561cc867f62f279d2ce9ed154b1b4369)
```python
        if host is NoHostProvided:
            host = boto.config.get('s3', 'host')
            if host is None:
                host = self.DefaultHost
                no_host_provided = True
```

So I've added a `s3` section with `host = s3.amazonaws.com` to `boto.cfg` so we don't have to change any code. Hopefully this works.

P.S. Apparently this [can't be done in `boto3`](https://stackoverflow.com/questions/32618216/overwrite-s3-endpoint-using-boto3-configuration-file), though.

openedx-webhooks:
Thanks for the pull request, @UmanShahzad! I've created [OSPR-2010](https://openedx.atlassian.net/browse/OSPR-2010) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

UmanShahzad:
@brianhw Btw, where are the acceptance tests being triggered? Any way I can see them?

brianhw:
No luck.  I ran just the enrollment acceptance tests, and got the following as two examples of S3 failures:
```
  File "/var/lib/analytics-tasks/f2c7086f821391d1ddb741986c718f7e/venv/local/lib/python2.7/site-packages/luigi/worker.py", line 218, in _add
    is_complete = self._check_complete(task)
  File "/var/lib/analytics-tasks/f2c7086f821391d1ddb741986c718f7e/venv/local/lib/python2.7/site-packages/luigi/worker.py", line 212, in _check_complete
    return task.complete()
  File "/var/lib/analytics-tasks/f2c7086f821391d1ddb741986c718f7e/venv/local/lib/python2.7/site-packages/luigi/task.py", line 426, in complete
    outputs = flatten(self.output())
  File "/var/lib/analytics-tasks/f2c7086f821391d1ddb741986c718f7e/venv/local/lib/python2.7/site-packages/edx/analytics/tasks/util/url.py", line 62, in output
    return get_target_from_url(self.url)
  File "/var/lib/analytics-tasks/f2c7086f821391d1ddb741986c718f7e/venv/local/lib/python2.7/site-packages/edx/analytics/tasks/util/url.py", line 130, in get_target_from_url
    cls, args, kwargs = get_target_class_from_url(url, marker)
  File "/var/lib/analytics-tasks/f2c7086f821391d1ddb741986c718f7e/venv/local/lib/python2.7/site-packages/edx/analytics/tasks/util/url.py", line 120, in get_target_class_from_url
    kwargs['client'] = ScalableS3Client()
  File "/var/lib/analytics-tasks/f2c7086f821391d1ddb741986c718f7e/venv/local/lib/python2.7/site-packages/luigi/s3.py", line 59, in __init__
    is_secure=True)
  File "/var/lib/analytics-tasks/f2c7086f821391d1ddb741986c718f7e/venv/local/lib/python2.7/site-packages/boto/__init__.py", line 141, in connect_s3
    return S3Connection(aws_access_key_id, aws_secret_access_key, **kwargs)
  File "/var/lib/analytics-tasks/f2c7086f821391d1ddb741986c718f7e/venv/local/lib/python2.7/site-packages/boto/s3/connection.py", line 200, in __init__
    "When using SigV4, you must specify a 'host' parameter."
HostRequiredError: BotoClientError: When using SigV4, you must specify a 'host' parameter.
```
And 

```
2017-11-14 17:23:19,620 WARNING 2164 [luigi-interface] worker.py:170 - Will not schedule PathSelectionByDateIntervalTask(source=('s3://edx-analytics-test-output/automation/f2c7086f821391d1ddb741986c718f7e/EnrollmentAcceptanceTest/src',), interval=2014-07-31-2014-08-07, expand_interval=2 days, 0:00:00, pattern=('.*tracking.log-(?P<date>\\d{8}).*\\.gz', '.*tracking.notalog-(?P<date>\\d{8}).*\\.gz'), date_pattern=%Y%m%d) or any dependencies due to error in complete() method:
  File "/var/lib/analytics-tasks/f2c7086f821391d1ddb741986c718f7e/venv/local/lib/python2.7/site-packages/luigi/worker.py", line 218, in _add
    is_complete = self._check_complete(task)
  File "/var/lib/analytics-tasks/f2c7086f821391d1ddb741986c718f7e/venv/local/lib/python2.7/site-packages/luigi/worker.py", line 212, in _check_complete
    return task.complete()
  File "/var/lib/analytics-tasks/f2c7086f821391d1ddb741986c718f7e/venv/local/lib/python2.7/site-packages/luigi/task.py", line 540, in complete
    return all(r.complete() for r in flatten(self.requires()))
  File "/var/lib/analytics-tasks/f2c7086f821391d1ddb741986c718f7e/venv/local/lib/python2.7/site-packages/edx/analytics/tasks/common/pathutil.py", line 166, in requires
    self.requirements = self._get_requirements()
  File "/var/lib/analytics-tasks/f2c7086f821391d1ddb741986c718f7e/venv/local/lib/python2.7/site-packages/edx/analytics/tasks/common/pathutil.py", line 192, in _get_requirements
    return [UncheckedExternalURL(url) for url_gen in url_gens for url in url_gen if self.should_include_url(url)]
  File "/var/lib/analytics-tasks/f2c7086f821391d1ddb741986c718f7e/venv/local/lib/python2.7/site-packages/edx/analytics/tasks/common/pathutil.py", line 196, in _get_s3_urls
    s3_conn = boto.connect_s3()
  File "/var/lib/analytics-tasks/f2c7086f821391d1ddb741986c718f7e/venv/local/lib/python2.7/site-packages/boto/__init__.py", line 141, in connect_s3
    return S3Connection(aws_access_key_id, aws_secret_access_key, **kwargs)
  File "/var/lib/analytics-tasks/f2c7086f821391d1ddb741986c718f7e/venv/local/lib/python2.7/site-packages/boto/s3/connection.py", line 200, in __init__
    "When using SigV4, you must specify a 'host' parameter."
```

The s3 setting didn't seem to have an effect.

brianhw:
Not clear why .boto is being created in the "current" directory (as seems to be happening) with the copy of boto.cfg you modified.  Best I can tell, that never gets read.  See http://boto.cloudhackers.com/en/latest/boto_config_tut.html#details.  On the EMR cluster master, the /etc/boto.cfg contains:
```
[s3]
use-sigv4 = True
[ec2]
use-sigv4 = True
```
And /home/hadoop/.aws/config contains:
```
[default]
s3 =
    signature_version = s3v4
```

And that's all I can find, apart from the .boto files copied from share/boto.cfg to the local repo directory (e.g. where the Makefile is).

UmanShahzad:
@brianhw That's the same confusion I'm having, yeah. I tried to make sure `boto.cfg` can be read locally by copying the `share/boto.cfg` to my `~/.boto` following it closely:

```
(Pdb) l
174                      suppress_consec_slashes=True, anon=False,
175                      validate_certs=None, profile_name=None):
176             no_host_provided = False
177             from pdb import set_trace;set_trace()
178             # Try falling back to the boto config file's value, if present.
179  ->         if host is NoHostProvided:
180                 host = boto.config.get('s3', 'host')
181                 if host is None:
182                     host = self.DefaultHost
183                     no_host_provided = True
184             if isinstance(calling_format, six.string_types):
(Pdb) n
> /home/mslm/dev/.envs/edx-analytics-pipeline/local/lib/python2.7/site-packages/boto/s3/connection.py(180)__init__()
-> host = boto.config.get('s3', 'host')
(Pdb) n
> /home/mslm/dev/.envs/edx-analytics-pipeline/local/lib/python2.7/site-packages/boto/s3/connection.py(181)__init__()
-> if host is None:
(Pdb) pp host
's3.amazonaws.com'
```

and so it does seem to work, given that it's in one of the paths as expected.

The only explanation to me is that the task play is copying it to the wrong directory. From `share/tasks.yml`:

```yml
    - name: boto configured
      copy: src=boto.cfg dest={{ working_repo_dir }}/.boto mode=644
```

It's copying things to the *repository*, which is why we see it in the same level as `Makefile` :\ Let's change that. But now I'm wondering if that'll have any effect anyway, since you say there exists a `/etc/boto.cfg`, and that seems to take precedence. If the only variables inside `/etc/boto.cfg` are those related to the signature, I think it'd be okay to just use `dest=/etc/boto.cfg` in the play then.

brianhw:
Actually, boto combines /etc/boto.cfg with ~/.boto, and overwriting /etc/boto.cfg is too global a change.  We should not be clobbering that. 

I just looked at the history of boto/boto/pyami/config.py, and I don't see that it *ever* read from ./.boto, so that copy in task.yml confuses me. Perhaps @mulby can remember something about this.

Options are setting config include overwriting whatever is in ~/.boto or /etc/boto.cfg, but another option would be to set either BOTO_CONFIG or BOTO_PATH in the environment.  (See https://github.com/boto/boto/blob/develop/boto/pyami/config.py#L32-L49.)

UmanShahzad:
@brianhw Ah, I didn't know they combined them, thanks!

I think the existence of `./.boto` is just a mistake. I've changed it to `{{ home }}/.boto` now.

brianhw:
Weird.  I ran the acceptance test again, and it didn't do the right thing.   It may be caching an older version of task.yml on the Jenkins box, because the copy wasn't made to /home/hadoop/.boto on the cluster, but the right version appeared in the repo on the cluster.  This might be easier to test in analyticstack, which I don't have set up here at home.  

I still wonder if the better solution is to define  BOTO_CONFIG or BOTO_PATH in the environment in this area:  https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/launchers/remote.py#L114-L121.

UmanShahzad:
@brianhw Thanks for taking a look at this while at home, btw!

I definitely think it's caching, unless you can find `.ssh` inside of the repository directory (doubt it), which would mean `{{ home }}` isn't being set properly.

The issue with `BOTO_CONFIG` is that then it wouldn't include `/etc/boto.cfg`, and `BOTO_PATH` would require us to hardcode in the path to both `/etc/boto.cfg` and `{repo}/.boto`, which is maybe more trouble than just letting the lookup happen safely on its own at all the relevant locations, including `~/.boto`. What do you think?

UmanShahzad:
@brianhw In the meantime, I'm downloading / setting up an analytic stack to test it locally, but that might take a *while*, and the caching probably would've already subsided.

mulby:
LGTM, I don't remember any details about the .boto file

UmanShahzad:
Can this be run one more time in the Jenkins acceptance tests? I can't get the devstack to work locally right now. And if things look good there, let's merge this and cherry-pick to ginkgo? This'll be required by the start of December.

UmanShahzad:
@brianhw @pomegranited Actually got 'em to run locally, and `~/.boto` exists:

```bash
vagrant@vagrant:~$ ls ~hadoop -al
total 64
...
-rw-r--r--  1 hadoop hadoop  104 Nov 15 14:49 .boto
...
```

Contents:

```cfg
[Boto]
metadata_service_timeout = 60.0
metadata_service_num_attempts = 60

[s3]
host = s3.amazonaws.com
```

Additionally, I see no tracebacks, so it's looking good! Will wait for it to finish and share the full output for review.

UmanShahzad:
@brianhw I actually didn't have the best luck locally -- some things seem to pass sometimes and fail other times due to random errors I believe are unrelated. Could you kick off the acceptance tests on Jenkins for me? I think this solution should do it.

brianhw:
Hmm.  I tried blowing away the workspace on Jenkins and rerunning the branch on the enrollment acceptance test.  Same failures.  And the .boto file is not appearing in /home/hadoop where I expect it would, but it's not elsewhere either.  Perhaps it's a permission issue?  But I'm really not liking overwriting ~/.boto anyway.   My preference again is to try setting BOTO_PATH = ['/etc/boto', '~/.boto', './.boto'] inside of remote.py when calling launch.

brianhw:
@macdiesel FYI.

UmanShahzad:
@brianhw We can try the `BOTO_PATH` strategy but at this point I went ahead and cherry-picked this onto a client branch and am testing it there in a real environment. It's working quite well hitherto.

Also, I'm wondering -- does `~/.boto` actually exist without this PR? If it does then you're right, we shouldn't be overwriting it. If it doesn't though then I think it should be safe, and I think it doesn't since you've mentioned it still doesn't show up in the acceptance tests, despite resetting the workspace.

UmanShahzad:
@brianhw Would you like me to update this PR to use the `BOTO_PATH` variable? I'd like to get this merged however we do it.

macdiesel:
@UmanShahzad Yes, we would like to use the BOTH_PATH environment variable. If you could make this change I can run the acceptance tests. Please ping me when you have the update and I'll take it for a spin.

macdiesel:
Also the desire to use the env variable stems from a desire to use caution and to not have side effects in the future. Should another job begin to write to the boto file in the home directory we could begin to have unintended job results. 

We feel it is better to be explicit rather than implicit when specifiying what boto file to use.

UmanShahzad:
@macdiesel Thanks, that makes sense to me.

I pushed the change, but am going to real quick do a sanity check on local acceptance tests first.

UmanShahzad:
@macdiesel If these changes look good to you, can you kick off the acceptance tests?

edx/analytics/tasks/launchers/remote.py
```diff
@@ -112,6 +115,7 @@ def run_task_playbook(inventory, arguments, uid):
     sudo_user = arguments.sudo_user
 
     env_vars = {}
+    env_vars['BOTO_PATH'] = os.pathsep.join([BotoConfigPath, BotoUserConfigPath, BOTO_FILE_PATH])
```
brianhw: These values are not quite right.  They should be the path values on the remote machine, not on the local machine.  BotoConfigPath should be the same for both, but BotoUserConfigPath uses the "jenkins" user in the current case, but I assume it should use the sudo_user being passed in.  Likewise, the BOTO_FILE_PATH should actually use code_dir.

brianhw: This also explains why the previous fix probably wasn't working on the acceptance tests, because it was trying to copy to a user that didn't exist on the remote server.  Why it didn't just fail I'm not sure.  Perhaps it worked for you if you were using the same user locally and remotely on analyticstack.

UmanShahzad: :+1: thanks for catching these. I'm quite new to this codebase so excuse my misunderstandings here -- learning as we go :)

UmanShahzad: @brianhw Yes that actually makes sense now.

UmanShahzad:
Any update on acceptance tests?

brianhw:
So we got further.   I kicked off an enrollment acceptance test after Uman's update.  It found the host this time, but failed in the reducer of the map-reduce job.  Twice.   Just for good measure, I kicked off a full suite of acceptance tests to run overnight.   And there were many more failures in various map-reduce jobs.  When looking at the EMR logs for the reducers, the problem was pretty clear: 

   boto.s3.connection.HostRequiredError: BotoClientError: When using SigV4, you must specify a 'host' parameter.

Basically, we have the same problem but now on the task instances, when they try to open a stream to S3 for their output.  They don't have the .boto file installed on them -- it's only being installed on the master node.

brianhw:
Oh, and of course, running acceptance tests locally won't uncover these issues in general, because they use HDFS instead of S3.

UmanShahzad:
@brianhw Any idea how we should go about that? At this point I'm thinking it'd just be easier to drop this s3 host var in the client configuration and put it manually as `host` into `connect_s3`? At least we'd be more sure about it working, and let it be more easily configurable.

UmanShahzad:
Also if there are any plans to move to boto3 completely some day, the `s3` section won't work there AFAIK, so maybe better to just keep it simple and in the code?

brianhw:
I don't know what the roadmap for boto->boto3 is, and particularly how that relates to AWS and Luigi stuff.  There are two uses of boto3 in the current Luigi (for SES and ECS), but not for S3 at this point.  

I'm rerunning my branch (brian/upgrade-boto) against enrollment acceptance just to see that it is sufficient.  One possibility is to just go with the hardcoded values in the short term, since I presume the resulting behavior is no different than the values that were hardcoded into boto.  And then in a separate PR, figure out how to parameterize the value for those that might want to override it.   (I don't actually have a sense of what different values people might want to choose.)

The immediate issue I have with reading from Luigi configuration is that it is still only available on the master instance, and would need to be passed to the task instances through the task definition(s).

UmanShahzad:
@brianhw Ah, if there's the same issue with the luigi configs then yeah, let's just go with hard-coded values short term (at this point I don't see any other option). Thanks for running the tests on your branch -- very interested to hear if it works out.

UmanShahzad:
@brianhw Closing this one out now.

openedx-webhooks:
@UmanShahzad Even though your pull request wasn’t merged, please take a moment to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=edx/edx-analytics-pipeline&entry.867055334=https://github.com/edx/edx-analytics-pipeline/pull/448&entry.1484655318=https://github.com/UmanShahzad&entry.752974735=2017-11-14+12:25&entry.1917517419=2017-11-18+23:57&entry.2133058324=No) so we can improve your experience in the future.

PR #452, opened 2017-11-20
Incrementalization of user-activity

HassanJaveed84:
Simplified approach using partition recovering. 

TODO:

- [x] Make the history generation faster.
- [x] Add docstrings/comments.
- [x] Document workflow.

### DIFF
--- a/docs/source/running_tasks.rst
+++ b/docs/source/running_tasks.rst
@@ -203,18 +203,59 @@ Engagement
 Notes
 ~~~~~
 
-* Intended to be run weekly
+* Intended to be run weekly or daily.
+* When using a persistent hive metastore, set overwrite_hive to True.
 
 Task
 ~~~~
 
 ::
 
-    CourseActivityWeeklyTask --local-scheduler \
+    InsertToMysqlCourseActivityTask --local-scheduler \
       --end-date $(date +%Y-%m-%d -d "$TO_DATE") \
       --weeks 24 \
       --credentials $CREDENTIALS \
+      --n-reduce-tasks $NUM_REDUCE_TASKS \
+      --overwrite_mysql
+
+Incremental implementation
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+On December 05, 2017 we merged a modification of the Engagement workflow to master.  The new code calculates Engagement *incrementally*, rather than entirely from scratch each time.  And it involves a new parameter: ``overwrite_n_days``.
+
+Also, the workflow has been renamed from ``CourseActivityWeeklyTask`` to ``InsertToMysqlCourseActivityTask``.
+
+The workflow now assumes that new Hive-ready data has been written persistently to the ``user_activity`` directory under warehouse_path by UserActivityTask.  The workflow uses the ``overwrite_n_days`` to determine how many days back to repopulate this data. The idea is that before this point, events are not expected to change, but perhaps there might be new events that have arrived in the last few days.  We are currently running the workflow daily with a value of 3, and we define that as an user-activity parameter in our override.cfg file.  You can define it there or on the command line.
+
+This means for us that only the last three days of raw events get scanned nightly.  It is assumed that the previous days' data has been loaded by previous runs, or by performing a historical load.
+
+If this workflow is run weekly, an ``overwrite_n_days`` value of 10 would be more appropriate.
+
+History task
+~~~~~~~~~~~~
+
+To load the historical user-activity counts, you would need to first run:
+
+::
+
+    UserActivityTask --local-scheduler \
+      --interval $(date +%Y-%m-%d -d "$FROM_DATE")-$(date +%Y-%m-%d -d "$TO_DATE") \
       --n-reduce-tasks $NUM_REDUCE_TASKS
+
+or you could run the incremental workflow with an ``overwrite_n_days`` value large enough that it would
+calculate the historical user-activity counts the first time it is ran:
+
+
+::
+
+    InsertToMysqlCourseActivityTask --local-scheduler \
+      --end-date $(date +%Y-%m-%d -d "$TO_DATE") \
+      --weeks 24 \
+      --credentials $CREDENTIALS \
+      --n-reduce-tasks $NUM_REDUCE_TASKS \
+      --overwrite_n_days 169
+
+After the first run, you can change ``overwrite_n_days`` to 3 or 10 depending on how you plan to run it(daily/weekly).
 
 Video
 ~~~~~
--- a/edx/analytics/tasks/insights/user_activity.py
+++ b/edx/analytics/tasks/insights/user_activity.py
@@ -2,18 +2,21 @@
 
 import datetime
 import logging
+from collections import Counter
 
 import luigi
 import luigi.date_interval
 
-from edx.analytics.tasks.common.mapreduce import MapReduceJobTask, MapReduceJobTaskMixin
+from edx.analytics.tasks.common.mapreduce import MapReduceJobTaskMixin, MultiOutputMapReduceJobTask
 from edx.analytics.tasks.common.pathutil import EventLogSelectionMixin, EventLogSelectionDownstreamMixin
 from edx.analytics.tasks.insights.calendar_task import CalendarTableTask
 from edx.analytics.tasks.util.decorators import workflow_entry_point
 import edx.analytics.tasks.util.eventlog as eventlog
-from edx.analytics.tasks.util.hive import WarehouseMixin, HiveTableTask, HivePartition, HiveQueryToMysqlTask
+from edx.analytics.tasks.util.hive import WarehouseMixin, BareHiveTableTask, HivePartitionTask, hive_database_name
 from edx.analytics.tasks.util.weekly_interval import WeeklyIntervalMixin
-from edx.analytics.tasks.util.url import get_target_from_url
+from edx.analytics.tasks.util.url import get_target_from_url, url_path_join
+from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
+from edx.analytics.tasks.common.mysql_load import MysqlInsertTask
 
 log = logging.getLogger(__name__)
 
@@ -23,7 +26,7 @@ PLAY_VIDEO_LABEL = "PLAYED_VIDEO"
 POST_FORUM_LABEL = "POSTED_FORUM"
 
 
-class UserActivityTask(EventLogSelectionMixin, MapReduceJobTask):
+class UserActivityTask(OverwriteOutputMixin, WarehouseMixin, EventLogSelectionMixin, MultiOutputMapReduceJobTask):
     """
     Categorize activity of users.
 
@@ -37,9 +40,8 @@ class UserActivityTask(EventLogSelectionMixin, MapReduceJobTask):
 
     """
 
-    output_root = luigi.Parameter(
-        description='String path to store the output in.',
-    )
+    output_root = None
+
 
     def mapper(self, line):
         value = self.get_event_and_date_string(line)
@@ -56,7 +58,7 @@ class UserActivityTask(EventLogSelectionMixin, MapReduceJobTask):
             return
 
         for label in self.get_predicate_labels(event):
-            yield self._encode_tuple((course_id, username, date_string, label)), 1
+            yield date_string, self._encode_tuple((course_id, username, date_string, label))
 
     def get_predicate_labels(self, event):
         """Creates labels by applying hardcoded predicates to a single event."""
@@ -111,27 +113,83 @@ class UserActivityTask(EventLogSelectionMixin, MapReduceJobTask):
         else:
             return values[0].encode('utf8')
 
-    def reducer(self, key, values):
-        """Cumulate number of events per key."""
-        num_events = sum(values)
-        if num_events > 0:
-            yield key, num_events
+    def multi_output_reducer(self, _date_string, values, output_file):
+        counter = Counter(values)
 
-    def output(self):
-        return get_target_from_url(self.output_root)
+        for key, num_events in counter.iteritems():
+            course_id, username, date_string, label = key
+            value = (course_id, username, date_string, label, num_events)
+            output_file.write('\t'.join([str(field) for field in value]))
+            output_file.write('\n')
+
+    def output_path_for_key(self, key):
+        date_string = key
+        return url_path_join(
+            self.hive_partition_path('user_activity', date_string),
+            'user_activity_{date}'.format(
+                date=date_string,
+            )
+        )
+
+    def run(self):
+        # Remove the marker file.
+        self.remove_output_on_overwrite()
+        # Also remove actual output files in case of overwrite.
+        if self.overwrite:
+            for date in self.interval:
+                url = self.output_path_for_key(date.isoformat())
+                target = get_target_from_url(url)
+                if target.exists():
+                    target.remove()
+
+        return super(UserActivityTask, self).run()
 
 
 class UserActivityDownstreamMixin(WarehouseMixin, EventLogSelectionDownstreamMixin, MapReduceJobTaskMixin):
     """All parameters needed to run the UserActivityTableTask task."""
-    pass
+
+    overwrite_n_days = luigi.IntParameter(
+        config_path={'section': 'user-activity', 'name': 'overwrite_n_days'},
+        description='This parameter is used by UserActivityTask which will overwrite user-activity counts '
+                    'for the most recent n days. Default is pulled from user-activity.overwrite_n_days.',
+        significant=False,
+    )
 
 
-class UserActivityTableTask(UserActivityDownstreamMixin, HiveTableTask):
-    """Hive table that stores the set of users active in each course over time."""
+class UserActivityTableTask(UserActivityDownstreamMixin, BareHiveTableTask):
+    """
+    The hive table for storing user activity data. This task also adds partition metadata info to the Hive metastore.
+    """
+
+    date = luigi.DateParameter()
+    interval = None
+
+    def requires(self):
+        # Overwrite n days of user activity data before recovering partitions.
+        if self.overwrite_n_days > 0:
+            overwrite_from_date = self.date - datetime.timedelta(days=self.overwrite_n_days)
+            overwrite_interval = luigi.date_interval.Custom(overwrite_from_date, self.date)
+
+            yield UserActivityTask(
+                interval=overwrite_interval,
+                warehouse_path=self.warehouse_path,
+                n_reduce_tasks=self.n_reduce_tasks,
+                overwrite=True,
+            )
+
+    def query(self):
+        query = super(UserActivityTableTask, self).query()
+        # Append a metastore check command with the repair option.
+        # This will add metadata about partitions that exist in HDFS.
+        return query + "MSCK REPAIR TABLE {table};".format(table=self.table)
 
     @property
     def table(self):
-        return 'user_activity_daily'
+        return 'user_activity'
+
+    @property
+    def partition_by(self):
+        return 'dt'
 
     @property
     def columns(self):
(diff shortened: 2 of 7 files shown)


edx/analytics/tasks/insights/user_activity.py
```diff
@@ -120,19 +129,52 @@ def reducer(self, key, values):
...
+            yield UserActivityTask(
+                date=date,
+                n_reduce_tasks=self.n_reduce_tasks,
+                output_root=partition_location,
+                overwrite=True
+            )
+
+    def query(self):
+        query = super(UserActivityTableTask, self).query()
+        return query + "MSCK REPAIR TABLE {table};".format(table=self.table)
```
azafty468: What's the benefit of using MSCK REPAIR TABLE?  Is this as a safeguard against the wild card add?

azafty468 (commented):
I think we also mentioned updating some documents that user-activity is now incremental.  Is that done somewhere?

edx/analytics/tasks/insights/user_activity.py
HassanJaveed84: This will add any partitions that exist on S3/HDFS to the Hive metastore. It basically loads all user_activity data partitions into this newly created Hive table.

azafty468 (approved):

edx/analytics/tasks/insights/user_activity.py
```diff
@@ -120,17 +113,35 @@ def _encode_tuple(self, values):
...
-    def reducer(self, key, values):
-        """Cumulate number of events per key."""
-        num_events = sum(values)
-        if num_events > 0:
-            yield key, num_events
-
-    def output(self):
-        return get_target_from_url(self.output_root)
+    def multi_output_reducer(self, _date_string, values, output_file):
+        c = Counter(values)
```
azafty468: Can we get a situation where we have a very large number of distinct values provided to this reducer?  Enough where this becomes a memory or processing time concern?

edx/analytics/tasks/insights/user_activity.py
```diff
@@ -120,17 +113,35 @@ def _encode_tuple(self, values):
...
+                date=date_string,
+            )
+        )
 
     def run(self):
+        # Remove the marker file.
         self.remove_output_on_overwrite()
+        # Also remove actual output files in case of overwrite.
+        if self.overwrite:
+            for date in self.interval:
```
azafty468: Do we still have the self.interval parameter?  It looks like it doesn't exist at the definition of UserActivityTask anymore

azafty468 (approved):
Just some trivial comments.  If it runs in a test environment I'm good with these changes.

edx/analytics/tasks/insights/user_activity.py
HassanJaveed84: It gets included from EventLogSelectionMixin.

edx/analytics/tasks/insights/user_activity.py
HassanJaveed84: Your concern is valid, but I couldn't see a way of achieving this with a MultiOutput MR task without using a dictionary. 

We could remove 'date_string' from the value tuple as we already group by the 'date_string', this would reduce the memory footprint somewhat. 
Running this task over the last year of eventlog data found no issues.

edx/analytics/tasks/insights/user_activity.py
```diff
@@ -269,26 +282,35 @@ def requires(self):
                 warehouse_path=self.warehouse_path,
                 overwrite=self.overwrite,
                 overwrite_n_days=self.overwrite_n_days,
```
azafty468: Will we ever get into a conflicting scenario with implementing both overwrite and overwrite_n_days?

On the enrollment jobs it looks like we've been selecting one or the other, and when we have both we've used overwrite_n_days and defaulted the overwrite flag as best fits.  I'm not saying this is the desired or optimal choice, but it does save us from weird scenarios where overwrite is False and overwrite_n_days is > 0

azafty468 (approved):
Other than my general comment I approve!

edx/analytics/tasks/insights/user_activity.py
```diff
@@ -111,27 +113,81 @@ def _encode_tuple(self, values):
...
+                warehouse_path=self.warehouse_path,
+                n_reduce_tasks=self.n_reduce_tasks,
+                overwrite=self.overwrite,
+            )
+
+    def query(self):
+        query = super(UserActivityTableTask, self).query()
+        # Append a metastore check command with the repair option.
+        # This will add metadata about partitions that exist in HDFS.
+        return query + "MSCK REPAIR TABLE {table};".format(table=self.table)
```
brianhw, line 184: I was concerned that there was a different version of this for EMR than for local Hadoop.   I think that finally got standardized, as EMR also uses this.   However, the doc for Hive still lists "ALTER TABLE table_name RECOVER PARTITIONS;" as the form supported on EMR.  (See https://cwiki.apache.org/confluence/display/Hive/LanguageManual+DDL.)  

I'm trying to remember why we didn't do this before.  Oh, yes.  I ran into this in the answer_hive.py work.  I had to pull this out into a platform-specific version of "recover_partitions".  https://github.com/edx/edx-analytics-pipeline/blob/brian/answer-hive/edx/analytics/tasks/insights/answer_hive.py#L502-L507.    Anyway, that no longer seems to be an issue, thankfully. 

The other issue with this approach is with the output() method.  (https://github.com/edx/edx-analytics-pipeline/blob/brian/answer-hive/edx/analytics/tasks/insights/answer_hive.py#L509-L513).  I guess the question is how it figures out whether the task is already complete or not.   If it just checks that the table exists, and not that all partitions are loaded, then it may not work well, particularly for those openedx installations that use a persistent metastore.

edx/analytics/tasks/tests/acceptance/test_user_activity.py
```diff
@@ -20,12 +20,14 @@ def test_user_activity(self):
...
         self.task.launch([
-            'CourseActivityWeeklyTask',
+            'InsertToMysqlCourseActivityTask',
             '--source', self.test_src,
             '--end-date', self.END_DATE.isoformat(),
             '--weeks', str(self.NUM_WEEKS),
             '--credentials', self.export_db.credentials_file_url,
+            '--overwrite-n-days', '43',
             '--n-reduce-tasks', str(self.NUM_REDUCERS),
+            '--overwrite'
```
brianhw: Yes, I'm confused by this as well.   What is the difference between overwrite_n_days and overwrite?  I'm guessing that overwrite-n-days only applies to the history, and that overwrite only applies to everything after the history.  If that were the case, is there any situation where one should *not* run with the --overwrite flag?

brianhw (commented):
Looks great!  And I'm glad that it seems to work.  I just had the question about overwrites and the partition recovering.

edx/analytics/tasks/insights/user_activity.py
HassanJaveed84: Yes, I realized the hive doc is outdated. 

Regarding the issue with Hive table output, we would always be running it with overwrite=True which would drop the table first before creation and partition restoration.

edx/analytics/tasks/tests/acceptance/test_user_activity.py
HassanJaveed84: Correct, overwrite_n_days is just for overwriting last n days of user activity data(generated from eventlog). 'overwrite' is for overwriting everything else(mysql, hive tables/s3 output). 
The way the workflow is structured, we would always run with overwrite=True.

edx/analytics/tasks/insights/user_activity.py
HassanJaveed84: Maybe it's less confusing if skip passing in overwrite here(it would default to False) and overwrite activity data if overwrite_n_days > 0.

brianhw:
So at this point, we have an agreement of having "overwrite" recompute Hive, "overwrite_n_days" for incremental data and default overwrite to True for Mysql.  Not sure if that's the case for user-activity yet.  

Also updating documentation for users to run this:  https://github.com/edx/edx-analytics-pipeline/blob/master/docs/source/running_tasks.rst

azafty468 (approved):
👍

edx/analytics/tasks/insights/user_activity.py
```diff
@@ -143,230 +199,141 @@ def columns(self):
...
-            ('interval_end',)
-        ]
-
-
-class CourseActivityDailyTask(CourseActivityTask):
-    """Number of users performing each category of activity each calendar day."""
+    def hive_table_task(self):
+        return CourseActivityTableTask(
+            warehouse_path=self.warehouse_path,
+        )
```
brianhw: I think this should also pass in overwrite, no?

edx/analytics/tasks/insights/user_activity.py
```diff
@@ -143,230 +199,141 @@ def columns(self):
...
 
-    @property
-    def table(self):
-        return 'course_activity_daily'
+    def requires(self):
+        yield (
+            CourseActivityTableTask(
+                warehouse_path=self.warehouse_path,
+                overwrite=self.overwrite,
+            ),
```
brianhw: This could return hive_table_task()?  A little DRYer.

edx/analytics/tasks/insights/user_activity.py
```diff
@@ -143,230 +199,141 @@ def columns(self):
...
-    )
-    months = luigi.IntParameter(
-        default=6,
-        description='The number of months to include in the analysis, counting back from the month that contains '
-        'the end_date.',
+    overwrite_hive = luigi.BooleanParameter(
+        default=False,
+        description='Overwrite the hive data used as source for this task. Users should override this to True '
+                    'when using a persistent Hive metastore.',
+        significant=False
```
brianhw: Interesting.  This is different than what we originally discussed, which was to hardcode Mysql updates to True, and map --overwrite to mean overwrite-hive.   And that is what I wrote about in reviewing Andrew's work on Enrollment, which is more complicated than this workflow.   This is certainly more explicit for hive, and more flexible for Mysql.  @azafty468 do you think we should define this for Enrollment, and look to move towards standardization among all insights workflows over time?  

Also, in your documentation, you suggest that overwrite_hive would be set to true when using a persistent Hive metastore.  While that is true, I also thought you were originally suggesting that we would also be always running it with it set to True?   Is that the case?

edx/analytics/tasks/warehouse/load_internal_reporting_user_activity.py
```diff
@@ -48,22 +43,72 @@ def columns(self):
...
-            , uad.`date`
-            , uad.category
-            , uad.count
-            FROM auth_user au
-            JOIN user_activity_daily uad ON au.username = uad.username
-            """
+    def hive_table_task(self):
+        return InternalReportingUserActivityTableTask(
+            warehouse_path=self.warehouse_path,
+        )
```
brianhw: overwrite here?  Like in requires()?  Or use this in requires()?

edx/analytics/tasks/warehouse/load_internal_reporting_user_activity.py
```diff
@@ -236,19 +270,28 @@ def requires(self):
         ]
 
 
-class UserActivityWorkflow(WeeklyIntervalMixin, luigi.WrapperTask):
+class UserActivityWorkflow(WeeklyIntervalMixin, WarehouseMixin, luigi.WrapperTask):
 
+    overwrite_hive = luigi.BooleanParameter(
+        default=False,
+        significant=False
+    )
```
brianhw: So for this workflow, there is no 'overwrite', just 'overwrite-hive' and 'overwrite-n-days'.   Again, that is clear, but it is not consistent with our usage elsewhere.

edx/analytics/tasks/warehouse/load_internal_reporting_user_activity.py
```diff
@@ -236,19 +270,28 @@ def requires(self):
...
         )
-        yield AggregateInternalReportingUserActivityTableHive(
-            interval=self.interval,
+        yield InsertToMysqlCourseActivityTask(
+            end_date=self.end_date,
+            weeks=self.weeks,
             n_reduce_tasks=self.n_reduce_tasks,
+            warehouse_path=self.warehouse_path,
+            overwrite_n_days=self.overwrite_n_days,
+            overwrite_hive=self.overwrite_hive,
```
brianhw, line 308: I'm confused why we are inserting to Mysql here.   Maybe the documentation will help me, once it's been added to the PR.  But I've lost track of which entry points are called just for Insights, and which are for our warehouse. It would be great if all the insights code could be in the insights directory, and the warehouse code be in the warehouse directory.  I'm hoping this is not the entry point for Insights. 

It may also just be that I need a little more documentation in the code to be reminded why things are arranged as they are.

Or that the entry point for openedx is actually the InsertToMysqlCourseActivityTask, and they would call it with ovewrite_n_days = 10, and we call it with 3 because we're expecting to run UserActivityTableTask daily instead of weekly.

edx/analytics/tasks/insights/user_activity.py
```diff
@@ -143,230 +199,141 @@ def columns(self):
...
-        ending_date = self.end_date.replace(day=1)  # pylint: disable=no-member
-        starting_date = ending_date - relativedelta(months=self.months)
-
-        return luigi.date_interval.Custom(starting_date, ending_date)
+    overwrite = luigi.BooleanParameter(
+        default=True,
+        description='Overwrite the table when writing to it by default. Allow users to override this behavior if they '
+                    'want.',
+        significant=False
+    )
```
brianhw: Just out of curiosity, how does one set a boolean parameter to False on the command line?  I know one can just call --overwrite and that sets it to True.  Can one also call '--overwrite False' and it will do the right thing?

brianhw (commented):
That took longer than I expected.  I think I figured out the entry points, but it would be good for us to all be clear about our expectations there.   Once we know that, it's easier to check on assumptions about overwrite and such.

edx/analytics/tasks/insights/user_activity.py
HassanJaveed84: Yes, originally it was supposed to always run with overwrite=True as the workflow passed in overwrite to InsertToMysqlCourseActivityTask(and all the upstream tasks). Now the mysql task defaults to True after separating overwrite and hive_overwrite. And hive_overwrite default of False would be okay for our temporary clusters(no persistent hive metastore).

edx/analytics/tasks/insights/user_activity.py
HassanJaveed84: Not entirely sure, IIRC '--overwrite False' doesn't do the right thing.

edx/analytics/tasks/warehouse/load_internal_reporting_user_activity.py
HassanJaveed84: Yes, I agree. But I am not sure how this can be made consistent. 
UserActivityWorkflow entry point calls InsertToMysqlCourseActivityTask, we want to default overwrite to True for this MysqlInsertTask but default to False for upstream hive related tasks. 

So if we only specify 'overwrite' here instead of 'overwrite_hive', we can't separate out the overwrite logic between mysql and hive. Open to suggestions.

edx/analytics/tasks/warehouse/load_internal_reporting_user_activity.py
HassanJaveed84: I modeled it after what we currently have in master.  We schedule UserActivityWorkflow to run on every Monday. This is what it looks like:
```
class UserActivityWorkflow(WeeklyIntervalMixin, luigi.WrapperTask):

    credentials = luigi.Parameter()

    def requires(self):
        yield CourseActivityWeeklyTask(
            end_date=self.end_date,
            weeks=self.weeks,
            credentials=self.credentials,
        )
        yield AggregateInternalReportingUserActivityTableHive(
            interval=self.interval,
        )
```
CourseActivityWeeklyTask does the mysql insertion and the aggregate task populates 'internal_reporting_user_activity'. 

The updated workflow(same entry point UserActivityWorkflow) is expected to run daily and would do the same thing that we currently do, populate the 'course_activity' mysql table and 'internal_reporting_user_activity' data.

Also, I chose not to remove warehouse stuff from this entry point as the load-warehouse job is configured to only launch a master node, not sure if the hive join query with auth_user would be performant on a single node cluster ? 
Otherwise we could have LoadInternalReportingUserActivityToWarehouse require InternalReportingUserActivityPartitionTask instead of ExternalUrl and have it do the aggregation there.

HassanJaveed84:
Responded to the comments, I'm gonna hold off on the documentation till we have a consensus on how we want to structure this workflow..

edx/analytics/tasks/insights/user_activity.py
brianhw: Hmm.  Having a parameter that cannot be changed to a different value on the command line strikes me as being of limited utility.   We would instead need to provide a flag that was --no-overwrite-mysql or something.

docs/source/running_tasks.rst
```diff
@@ -203,19 +203,58 @@ Engagement
...
 ~~~~
 
 ::
 
-    CourseActivityWeeklyTask --local-scheduler \
+    InsertToMysqlCourseActivityTask --local-scheduler \
       --end-date $(date +%Y-%m-%d -d "$TO_DATE") \
       --weeks 24 \
       --credentials $CREDENTIALS \
       --n-reduce-tasks $NUM_REDUCE_TASKS
```
HassanJaveed84: Should we add 'overwrite_mysql' here as one should run it with overwrite_mysql=True to overwrite mysql data.

HassanJaveed84:
@azafty468 @brianhw this is ready for a final review, hoping this can go as my last release 😔

docs/source/running_tasks.rst
brianhw: Yes, I think that would be good.

docs/source/running_tasks.rst
```diff
@@ -203,19 +203,58 @@ Engagement
...
 
+Incremental implementation
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+On December 04, 2017 we merged a modification of the Engagement workflow to master.  The new code calculates Engagement *incrementally*, rather than entirely from scratch each time.  And it involves a new parameter: ``overwrite_n_days``.
```
brianhw: This is not the Engagement workflow!   : )

docs/source/running_tasks.rst
azafty468: I'd say yes to this

docs/source/running_tasks.rst
```diff
@@ -203,19 +203,58 @@ Engagement
...
+
+Also, the workflow has been renamed from ``CourseActivityWeeklyTask`` to ``InsertToMysqlCourseActivityTask``.
+
+The workflow now assumes that new Hive-ready data has been written persistently to the ``user_activity`` directory under warehouse_path by UserActivityTask.  The workflow uses the ``overwrite_n_days`` to determine how many days back to repopulate this data. The idea is that before this point, events are not expected to change, but perhaps there might be new events that have arrived in the last few days.  We are currently running the workflow daily with a value of 3, and we define that as an user-activity parameter in our override.cfg file.  You can define it there or on the command line.
+
+This means for us that only the last three days of raw events get scanned nightly.  It is assumed that the previous days' data has been loaded by previous runs, or by performing a historical load.
+
+If this workflow is run weekly, an ``overwrite_n_days`` value of 3 would be more appropirate.
```
brianhw: Value of 10?   Or something else?    Also, s/appropirate/appropriate/.

azafty468:
Hmmm, looks like I lost the ability to review the difference between commits since this guy was squashed down to one, and I cannot open up a new review because my approval was already given.  But I'm a 👍 on this!  

And the HassanJaveed84 commits will live on in history for all time!

edx/analytics/tasks/warehouse/load_internal_reporting_user_activity.py
```diff
@@ -236,19 +268,42 @@ def requires(self):
...
+            date=self.end_date,
             n_reduce_tasks=self.n_reduce_tasks,
-            credentials=self.credentials,
+            warehouse_path=self.warehouse_path,
+            overwrite_n_days=self.overwrite_n_days,
+            overwrite=self.overwrite_hive,
         )
-        yield AggregateInternalReportingUserActivityTableHive(
-            interval=self.interval,
+        yield InsertToMysqlCourseActivityTask(
```
brianhw, line 302: Looks like overwrite_mysql is not being passed down here.

edx/analytics/tasks/warehouse/load_internal_reporting_user_activity.py
brianhw: Makes sense.

edx/analytics/tasks/insights/user_activity.py
brianhw: I assume this is okay, even if there are users who are not always running with overwrite_hive=True.

edx/analytics/tasks/insights/user_activity.py
```diff
@@ -143,230 +201,144 @@ def columns(self):
...
+        default=False,
+        description='Overwrite the table if set to True. Allow users to override this behavior if they '
+                    'want.',
+        significant=False
+    )
 
-        return luigi.date_interval.Custom(starting_date, ending_date)
+    def __init__(self, *args, **kwargs):
+        super(InsertToMysqlCourseActivityTask, self).__init__(*args, **kwargs)
+        self.overwrite = self.overwrite_mysql
```
brianhw, line 332: I'm wondering if we also want a "overwrite = None" at the class level so that --overwrite doesn't appear as a Parameter to this function in the automated documentation.

brianhw (approved):
Looks good!  A few small things (mostly doc), but fix those and we're ready to merge!

docs/source/running_tasks.rst
HassanJaveed84: Oh, isn't that what is in the docs currently ?
What should we change it to ?

edx/analytics/tasks/insights/user_activity.py
HassanJaveed84: Good catch.

edx/analytics/tasks/insights/user_activity.py
HassanJaveed84: If the table exists and overwrite_hive is False, it would be a noop.

edx/analytics/tasks/warehouse/load_internal_reporting_user_activity.py
HassanJaveed84: Fixed.

docs/source/running_tasks.rst
brianhw: Argh!  Sorry, I read this as "Enrollment" instead of "Engagement", and thought it was a cut-and-paste error.  Nevermind.

brianhw:
Looks good to me!   Thanks so much, Hassan!  👍

PR #454, opened 2017-11-22
Handle null-duration video events.

brianhw:
Some video events have {"duration": null} entries.  These are now mapped to the default value used when duration is not specified.


HassanJaveed84 (approved):
LGTM 👍

muhammad-ammar (approved):
@brianhw This is looking good. Thanks for taking care of this. But now I am worried why duration is not set for some videos. This should not happen as video component always have the duration. I will look into this.

muhammad-ammar:
@brianhw code coverage job is failing, I think you might need to add a unit test for the new code.

brianhw:
After finally getting the video history copied from prod to dev, I was able to successfully run the release candidate on this branch.

PR #455, opened 2017-11-25
Pin some google-related packages.

brianhw:
Some newer releases of google packages seem to be breaking existing imports, so pin them to their previous versions for now.


brianhw:
This has worked for a few days now, so I'm going to merge and release it.

PR #456, opened 2017-11-29
Added Dockerfile to aid local development

clintonb:
A Dockerfile is now included so that developers can use a Docker image/container for local testing. Travis now builds the Docker image and runs tests on the container. The image is pushed to Docker Hub for successful master builds.

LEARNER-1745

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,22 @@
+# docker build -t edxops/analytics-pipeline .
+
+FROM edxops/python:2.7
+ENV BOTO_CONFIG /dev/null
+
+RUN apt-get update && \
+    apt-get install --no-install-recommends -y \
+        libatlas-base-dev \
+        libblas-dev \
+        liblapack-dev \
+        libpq-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN mkdir -p /edx/app/analytics-pipeline/requirements
+COPY Makefile /edx/app/analytics-pipeline
+COPY requirements /edx/app/analytics-pipeline/requirements
+
+WORKDIR /edx/app/analytics-pipeline
+
+RUN make test-requirements
+
+VOLUME /edx/app/analytics-pipeline
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,12 @@ develop-local: uninstall
 	python setup.py develop
 	python setup.py install_data
 
+docker-build:
+	docker build -t edxops/analytics-pipeline:latest .
+
+docker-shell:
+	docker run -v `(pwd)`:/edx/app/analytics-pipeline -it edxops/analytics-pipeline:latest bash
+
 system-requirements:
 ifeq (,$(wildcard /usr/bin/yum))
 	# This is not great, we can't use these libraries on slave nodes using this method.
@@ -35,6 +41,9 @@ requirements:
 test-requirements: requirements
 	pip install -U -r requirements/test.txt --no-cache-dir
 
+test-docker:
+	docker run -v `(pwd)`:/edx/app/analytics-pipeline -it edxops/analytics-pipeline:latest make develop-local test-local
+
 test-local:
 	# TODO: when we have better coverage, modify this to actually fail when coverage is too low.
 	rm -rf .coverage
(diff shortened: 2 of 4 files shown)


brianhw:
Thanks for the docker contribution!   Perhaps it should go into a separate PR from the finance work?

clintonb:
Yep, I'll separate them shortly.

Makefile
```diff
@@ -51,6 +60,9 @@ test-acceptance-local:
 test-acceptance-local-all:
 	REMOTE_TASK=$(shell which remote-task) LUIGI_CONFIG_PATH='config/test.cfg' ACCEPTANCE_TEST_CONFIG="/var/tmp/acceptance.json" python -m coverage run --rcfile=./.coveragerc -m nose --nocapture --with-xunit -A acceptance -v
 
+coverage-docker:
+	docker run -v `(pwd)`:/edx/app/analytics-pipeline -it edxops/analytics-pipeline:latest coverage xml
+
```
brianhw, line 65: I tried this out, and didn't see that it did anything.   In comparison, I tried "docker-shell" and running "develop-local", "test-local", and then "coverage-local", and I got the current problem we have with version mismatch on pep8/diff-cover, for which we have a  bug to fix this sprint.  Should this really be equivalent to coverage-local, in the way that test-docker is equivalent to test-local?

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+FROM edxops/python:2.7
+ENV BOTO_CONFIG /dev/null
+
+RUN apt-get update && \
+    apt-get install --no-install-recommends -y \
+        libatlas-base-dev \
+        libblas-dev \
+        liblapack-dev \
+        libpq-dev \
```
brianhw, line 11: Just curious why this differs from the addons/apt/packages in .travis.yml -- why the libpq-dev was not needed there before?  I'm guessing it's for running acceptance tests?    And yet, the system-requirements defined in the Makefile are more extensive:   libmysqlclient-dev libpq-dev python-dev libffi-dev libssl-dev libxml2-dev libxslt1-dev.  But perhaps they should also be slimmed down if they are no longer needed?   (Not a bad thing, but we've probably not bothered to check.)

Makefile
```diff
@@ -51,6 +60,9 @@ test-acceptance-local:
 test-acceptance-local-all:
 	REMOTE_TASK=$(shell which remote-task) LUIGI_CONFIG_PATH='config/test.cfg' ACCEPTANCE_TEST_CONFIG="/var/tmp/acceptance.json" python -m coverage run --rcfile=./.coveragerc -m nose --nocapture --with-xunit -A acceptance -v
```
brianhw, line 61: Should this work in docker-shell as well?  I was finding that it didn't locate /var/tmp/acceptance.json.  But then again, I didn't expect it to work at all, as there's no hadoop in this docker image, best I can tell.  I'm assuming this PR is only to get unit tests working.

brianhw (approved):
Looks good to me -- just a few questions. 👍

Dockerfile
clintonb: I added `libpq-dev` to get the PostgreSQL dependencies needed to install psycopg2. This library was not needed when running on Travis because the Travis environment already includes the library.

As for the other libraries, your guess—they are needed for acceptance tests—seems right to me.

Makefile
clintonb: I never tested this command in the image. Your assumption is correct.

Makefile
clintonb: Yes, it should. How do you feel about moving the quality checks of `coverage-local` into their own target: `quality`?

Makefile
brianhw: Okay.

Makefile
brianhw: Sorry to be dense, but I don't understand.  Should this just call coverage-local?   Or what would you include in quality that is not in coverage?   And is it just so that travis doesn't call coverage?

Dockerfile
brianhw: Okay.

Makefile
clintonb: `coverage-local` should only be concerned with coverage. `quality` should only be concerned with code quality—pycodestyle, pylint.

brianhw:
Weird.   Diff-cover seems to fail on the Travis /push test.

Makefile
clintonb: I'm going to make these changes in a future iteration. What is present is good enough. The changes I would like to make—splitting coverage and quality—may break things for current users. I'd rather not do that now.

clintonb: Actually...I broke coverage reporting. Investigating.

clintonb: Nope. Codecov is working. Not sure why the check didn't run for Github.

PR #457, opened 2017-12-01
Move course-subjects to use the discovery API

brianhw:
Implemented.   

Working on getting tests to work.


brianhw:
Okay, unit tests and acceptance tests have been modified to deal with new API.   I also moved the various "raw" API pulls into a common "discovery_api_raw" directory, and updated acceptance tests that were affected by this (enrollment and grades).  Ready for final review.

brianhw:
Addresses DE-447.

azafty468 (approved):
👍 I like the standardization of the CourseSubjectRecord object and the clarity on the course_runs objects.  Just a confirmation, this is changing the name of the intermediate object on S3, and reloads previously build objects will not be possible using this version?

brianhw:
That's right.   But we rarely do such reruns, and if we need to, we can
copy the old data from the old locations to the new locations.

azafty468:
Ah, that's right, same format just different name.  I was going to then ask if we wanted to also wipe out the old data if it wasn't usable anymore once the new version is stable.  But there is still some value to keeping it around.

PR #458, opened 2017-12-04
Update pep8 to pycodestyle.

brianhw:
The pep8 package was renamed to pycodestyle.  See https://github.com/PyCQA/pycodestyle for more pointers.


brianhw:
This should address the failures to run pep8 that we've experienced since diff-cover upgraded.  I tested it out using @clintonb 's docker PR.

clintonb (approved):

PR #459, opened 2017-12-05
Improving code quality

clintonb:
LEARNER-1745

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,0 +1,3 @@
+[settings]
+line_length=120
+multi_line_output=5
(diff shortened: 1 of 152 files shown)


brianhw:
Interesting.   I sorted packages by package name, and not whether it was using an import or a from...import statement.   This seems to group the from's separately.   I'm fine either way, but just want something that can be consistently maintained. 

LGTM.  👍

clintonb:
[isort](https://pypi.python.org/pypi/isort) can help maintain the order. We use it in edx-platform and and the IDA repos.

Once I finish the entitlements work, I hope to return to the quality conversation from #456, and run quality checks with every commit. Right now they are not run, so none of the rules set by pycodestyle, pylint, or isort are enforced.

PR #460, opened 2017-12-05
Updated analyze script to support newer releases of EMR.

HassanJaveed84:


brianhw (approved):
LGTM, especially if this worked for you. 👍

HassanJaveed84:
Yup, works with logs from emr 2x as well as current.

PR #461, opened 2017-12-06
Fix source for course_subjects in BigQuery load.

brianhw:
Missed this in testing BQ load, and the ExternalURL failed silently.


azafty468 (approved):
good catch!

PR #464, opened 2017-12-14
Introduce pip-compile tooling

brianhw:
Some details:

* Add --upgrade-strategy only-if-needed, in order to avoid setuptools being automatically upgraded by google-cloud-bigquery's dependency on it.
* Repurpose extra.txt to be for dependencies that don't work well with pip-compile, rather than those that should not be used with doc-building.
* Upgrade requests and urllib3, because getting agreement across all packages on the current dependencies wasn't working.  
* Add some other package upgrades.    Some were buggy:  ciso8601==1.0.5 threw a seg fault in unit tests, for example.  Others were tested as part of PR #432 (which this supercedes).
* Moved pandas from production to test.  It appears to be only used at this point for validating test results.


brianhw:
By the way, unit tests and acceptance tests both pass.   Because of the changes to "requests", however, this should go through some amount of release candidate testing.   But in the meantime it's ready for review.

azafty468 (approved):
If unit and AC tests pass then I'm good!

rao-abdul-mannan (commented):
Looks good.

PR #466, opened 2017-12-21
Fix isort and pep8 errors.

brianhw:
There were two cases of "E731 do not assign a lambda expression, use a def", but I just added E731 to the exclusion list rather than changing them.

There were some indentation issues that were in disagreement between pycodestyle and pylint.  I went with the pycodestyle preference in those cases.


azafty468 (approved):
👍 you are a saint

PR #467, opened 2017-12-22
fixing a bad record ordering for EnrollmentDailyRecord that was causi…

azafty468:
…ng fields to be incorrectly read

### DIFF
--- a/edx/analytics/tasks/insights/enrollments.py
+++ b/edx/analytics/tasks/insights/enrollments.py
@@ -19,7 +19,7 @@ from edx.analytics.tasks.util.hive import (
     BareHiveTableTask, HivePartitionTask, OverwriteAwareHiveQueryDataTask, WarehouseMixin
 )
 from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
-from edx.analytics.tasks.util.record import BooleanField, DateTimeField, IntegerField, Record, StringField
+from edx.analytics.tasks.util.record import BooleanField, DateField, DateTimeField, IntegerField, Record, StringField
 from edx.analytics.tasks.util.url import ExternalURL, UncheckedExternalURL, get_target_from_url, url_path_join
 from edx.analytics.tasks.warehouse.load_internal_reporting_course_catalog import (
     CoursePartitionTask, LoadInternalReportingCourseCatalogMixin, ProgramCoursePartitionTask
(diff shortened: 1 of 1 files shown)


brianhw (approved):
LGTM. 👍

brianhw:
I also kicked off a manual acceptance test on Jenkins --  see edx-analytics-pipeline-acceptance-test-manual/1470.   The enrollment tests are still failing, but are closer.   Now they just differ in date format.  For example:

First has 1, Second has 0:  (u'2014-08-01', u'edX/Open_DemoX/edx_demo_course', 3, 4)
First has 0, Second has 1:  (datetime.date(2014, 8, 1), 'edX/Open_DemoX/edx_demo_course', 3, 4)

This is a side effect of switching to a Record, I suspect.  Abdul ran into a similar situation with student_engagement.

brianhw (changes requested):
It would be good to fix the acceptance test first.   The problem is that these date fields are defined as StringField, but they should be defined as DateField.  The call to get_hive_schema() should return 'STRING', yes, but the call to .get_sql_schema()  should return 'DATE NOT NULL' as before, not 'VARCHAR(255) NOT NULL' which this produces.  It just happened to affect this table first (in validate_base()), but I believe all the other tables similarly defined would also have failed.

edx/analytics/tasks/insights/enrollments.py
```diff
@@ -504,7 +504,7 @@ def change_state(self):
 
 class CourseEnrollmentRecord(Record):
     """A user's enrollment history."""
-    date = StringField(length=255, nullable=False, description='Enrollment date.')
+    date = DateField(nullable=False, description='Enrollment date.')
     course_id = StringField(length=255, nullable=False, description='The course the learner is enrolled in.')
     user_id = IntegerField(description='The user_id of the learner.')
     at_end = BooleanField(description='An indicator if the learner is still enrolled in the course by this date.')
```
brianhw: nit:   "by this date" => "at the end of this date"

azafty468: done and squashed back down to 1 commit

azafty468:
merging.  Actually, I it says @brianhw is requesting changes in the status box at the bottom of this history

brianhw (approved):
Make this official.

PR #468, opened 2017-12-28
Generate the Hive Table but not Table Data

azafty468:
Reverting this back to using the prior functionality of generating a hive table but not generating the associated data. Our recent release had squashed the hive table generation and caused the overnight failures of module-engagement.


azafty468:
So the underlying cause was that ExternalCourseEnrollmentTableTask used by module_engagement overrode the requires() call and returned an external URL.  As part of my work I transformed this into a Partition task, kept the override, but did not insert the table task into the requires() steps.  module-engagement-edge just successfully completed running on this branch so I think we are good to merge.

brianhw (approved):
Yes, that makes sense!

PR #472, opened 2018-01-19
Upgrade to fork of Luigi 2.7.

brianhw:
Upgrading from fork of Luigi 1.0.17 to fork of Luigi 2.7.1.

There are several features that are supported in Luigi 2.x that are not available in our current fork.

* Dynamic Dependencies.  In Luigi 1.x, dependencies must be declared in requires() methods.  In Luigi 2.x, dependencies can also be specified by yielding dependent tasks from run() methods.  This makes it easier to specify task ordering requirements.  Such ordering is a prerequisite for running with a central scheduler.
* Improved logging of workflow status:  "Luigi Execution Summary" includes reporting how many tasks were scheduled, and of them, how many succeeded, failed, were not run, and so forth.
* Better error handling:  return of error codes, and configurability of actions to take on various error conditions.  We are currently relying on the presence of tracebacks in console logs to determine the status of workflows, which is suboptimal.   We miss cases where tracebacks are not fatal, and also where failures do not result in tracebacks.
* Support for Spark tasks:  a base task that handles packaging up dependencies for remote execution.   This is relatively stand-alone, so we might be able to port it to our fork anyway.

The bulk of required changes are due to a redesign in the handling of list parameters.   In Luigi 1.x, list parameter values are specified in configuration files by strings on separate lines, and on command lines as multiple declarations of the parameter flag and value. In Luigi 2.x, list parameter values are specified as JSON strings, both in config files and command lines.  Because JSON lists require quotes and braces, there are issues of character escaping in order to get the right values through.  This has proved to be more an issue for acceptance tests. 

Other changes include:

* Change in S3 support.  Luigi 2.x provides support for multi-part writes, so our special S3ScalableClient code doesn't need to provide that anymore.  (However, we maintain our version of the client to support defaulting of values for host that boto requires.)
* Still require a fork of Luigi:  we found we needed to parse metadata about packages to determine what to copy to remote cluster nodes.   We needed to do this to get opaque-keys in particular to properly load. This is somewhat hacky code, so while we have put it on our fork (and still have to put it on our new branch of the fork), we don't anticipate contributing it back to Luigi. 

EdX changes in old fork:  https://github.com/spotify/luigi/compare/master...edx:edx/release
EdX changes in new fork:  https://github.com/spotify/luigi/compare/master...edx:edx/patch

Configuration files also are updated internally, on the brian/upgrade-luigi2 branch.  Again, this is mostly for specifying the value of list parameters. 

One change in functionality involves ProblemResponseReportWorkflow.   It uses the TimestampPartitionMixin, which was used to support partitioning by datetimes that were more finely defined than days.  It relied on a property of luigi.DateParameter that it also support datetimes in addition to dates.   This is no longer the case with Luigi 2.  I changed the definition of TimestampPartitionMixin.date from luigi.DateParameter to luigi.DateSecondParameter.   This requires the date of a job to always be specified to the minute.  We don't actually run this ourselves, so I'm reliant on the acceptance test.  I'm not sure if the interval or interval_end need to also support datetimes.  It turns out that luigi.date_interval.Custom never did support this, but ProblemResponseTableMixin.interval_end is still a luigi.DateParameter, so it now only holds the date.


brianhw:
@pomegranited Could you review the changes to ProblemResponseReportWorkflow in particular?  I summarized the issues above, but may not have provided enough detail.  But it would be good to know if this would break your use cases.

pomegranited:
Thanks for pinging me @brianhw  😄   I'm pretty sure the integration tests would reveal any issues, but would like to run the ProblemResponse reports on my analyticstack, just to be sure.

But I'm afraid I won't have time this week to review -- is next week ok?  Or if you need to merge this soon, would you be willing to do a follow-up PR if I find issues?  My client is running on the Ginkgo.1 release, so it won't break anything in production for us.

edx/analytics/tasks/util/s3_util.py
```diff
@@ -127,99 +128,20 @@ def func(name):
 
 class ScalableS3Client(S3Client):
```
azafty468, line 129: Yay!

edx/analytics/tasks/common/mysql_load.py
```diff
@@ -59,15 +59,22 @@ def requires(self):
...
+
         return self.required_tasks
 
     @property
     def insert_source_task(self):
         """Defines task that provides source of data for insertion."""
         raise NotImplementedError
 
+    @property
+    def insert_source_task_dynamically(self):
```
azafty468, line 74: Presumably we have an opportunity in many/most/all of our other tasks to use this as a template for dynamically sourced inputs?

azafty468 (approved):
Looks good!  I'm a little concerned at the sheer number of parameter changes for ListParameter and BooleanParameter but that's mostly because of my past reliance on statically typed languages.  I'm hoping our test coverage is good enough to detect any problems.

rao-abdul-mannan (approved):
Looks good, great work. Looks like you had to revise the whole project :)

brianhw:
@pomegranited So sorry for the late notice!  I think we would like to merge this sooner, if that's okay, but would absolutely do a follow-up PR for any issues that need to be addressed.  As it stood, the acceptance tests didn't uncover an issue -- it was a unit test I had initially overlooked and only addressed at the end. But to work around that, I needed to change the acceptance tests for problem_reports, course_blocks, and course_list to deal with the "date" parameter actually being a datetime instead of a date.  Going forward, it might be worthwhile to rename the parameter to be "datetime" to communicate that.  There are other alternatives, but we can discuss those as well when you have more time.

brianhw:
@azafty468 Our test coverage was not in fact complete for testing all of the ListParameter changes.  I ended up testing others by actually running RC jobs that paralleled our production jobs.  This was particularly the case with warehouse workflows.  I believe I covered all of the parameters between acceptance tests and RC runs.

pomegranited:
No worries @brianhw !  Thanks for getting this change in.
I'll run through my tests next week and submit a PR if needed.

And thank you for pinging me on this change!  It means configuration changes our Hawthorn Insights upgrades will need to be propagated down to our client config overrides, and so was great to have this heads up.

> I needed to change the acceptance tests for problem_reports, course_blocks, and course_list to deal with the "date" parameter actually being a datetime instead of a date. Going forward, it might be worthwhile to rename the parameter to be "datetime" to communicate that. 

Sure thing, will look at changing that parameter name in a PR too.

pomegranited:
@brianhw I ran my local tests, and this change doesn't break Problem Reports at all -- thank you!  Have created https://github.com/edx/edx-analytics-pipeline/pull/482 to follow up on the `datetime` parameter name change.

PR #476, opened 2018-01-24
WIP-Spark Tasks

rao-abdul-mannan:
This PR includes following tasks:

TotalEventsDailyTask
UserActivityTask
CourseActivityPartitionTask
InternalReportingUserActivityPartitionTask

Note: **Not to be merged with master branch yet**


edx/analytics/tasks/insights/location_per_course.py
```diff
@@ -163,6 +164,83 @@ def run(self):
...
+        validate_courseid = udf(validate_course_id, StringType())
+        df = df.withColumn('course_id', validate_courseid(df['course_id']))
+        df.createOrReplaceTempView('location')
+        query = """
+                SELECT
+                    timestamp, ip, user_id, course_id, dt
+                FROM (
+                    SELECT
+                        event_date as dt, user_id, course_id, timestamp, ip,
+                        ROW_NUMBER() over ( PARTITION BY event_date, user_id, course_id ORDER BY timestamp desc) as rank
```
brianhw: nit: for consistency, I would uppercase all keywords, including OVER and AS (here and the line above).

edx/analytics/tasks/insights/location_per_course.py
```diff
@@ -163,6 +164,83 @@ def run(self):
...
+                    FROM location
+                    WHERE ip <> '' AND timestamp <> '' AND user_id <> ''
+                ) user_location
+                WHERE rank = 1
+                """
+        result = self._spark.sql(query)
+        result.coalesce(4).write.partitionBy('dt').csv(self.output_dir().path, mode='append', sep='\t')
```
brianhw: It might be good to add a comment here, that this will produce four TSV files in the dt=<date> subdirectory under the output_dir().

But a question:  is there a reason why the mode is 'append'?   

And a second question:   I've noticed that write() generally creates a _SUCCESS file separate from the actual output files but in the same directory.   Will this use of partitionBy() do the same, and if so, where would it write this file?   In each partition, or in the output_dir()?

edx/analytics/tasks/insights/location_per_course.py
```diff
@@ -163,6 +164,83 @@ def run(self):
                 target.open("w").close()  # touch the file
 
 
+class LastDailyIpAddressOfUserTaskSpark(EventLogSelectionMixinSpark, WarehouseMixin, SparkJobTask):
+    """Spark alternate of LastDailyIpAddressOfUserTask"""
+
+    output_parent_dir = 'last_ip_of_user_id'
```
brianhw: nit:  how about "output_parent_dirname"?  When I first saw this below, I was unclear whether it was a name or a path.

edx/analytics/tasks/insights/location_per_course.py
```diff
@@ -163,6 +164,83 @@ def run(self):
...
+        config_path={'section': 'map-reduce', 'name': 'marker'},
+        significant=False,
+        description='A URL location to a directory where a marker file will be written on task completion.',
+    )
+
+    def output_dir(self):
+        """
+        Output directory for spark task
+        """
+        return get_target_from_url(
```
brianhw, line 181: Is there a reason this is a target instead of just a URL?  Is that the easiest way to get .exists() functionality?

edx/analytics/tasks/insights/location_per_course.py
```diff
@@ -163,6 +164,83 @@ def run(self):
...
+    def on_success(self):  # pragma: no cover
+        self.output().touch_marker()
+
+    def run(self):
+        self.remove_output_on_overwrite()
+        if self.output_dir().exists():      # only check partitions if parent dir exists
+            for target in self.output_paths():
+                if target.exists():
+                    target.remove()
+        super(LastDailyIpAddressOfUserTaskSpark, self).run()
```
brianhw, line 217: In the original, each date in the interval is checked to see if it produces actual data.   If the data is sparse enough, then not all date partitions will be created.  This was done because the downstream class needed to have all dates exist, even if empty, so as to know what dates had been processed.   This was also dealt with using the downstream_input_tasks() to do this check in the LastCountryOfUser task.  

But maybe this can be addressed when the follow-on (i.e. non-historic) workflow is written.

edx/analytics/tasks/insights/user_activity.py
```diff
@@ -146,6 +147,83 @@ def run(self):
...
+    def output(self):
+        """
+        Marker output path
+
+        There were 2 approaches to verify output from spark multi output task:
+        1) verify partitions for all dates in the interval
+        2) create marker in separate dir just like hadoop multi-mapreduce task
+        Former approach can fail in cases, when there is no data for some dates as spark will not generate empty
+        partitions.
+        Later approach is more consistent.
```
brianhw, line 177: I see.  So this is how you're doing user location.  You chose approach 2 over approach 1.    That may be fine, though I'm wary about using marker files where we can avoid it, because they are magical and hard to track down.  It is really hard to know why a job is not producing output when a marker file exists somewhere with an opaque hash for a name.  Also, because the hash comes from hashing the task, which effectively hashes the (significant) parameters, one will get a different hash for each interval.  So if I run different but overlapping intervals, it will do an overwrite, but it doesn't actually check that the intervals that have been run are contiguous, or overlapping, or whatever.  They only check that the same exact interval has or hasn't been run before. 

That's why the location workflow verified partitions for all dates in the interval, and that the run() method created empty partitions for dates that had no data after the multi-mapreduce was done.  (That's not just the issue of spark but an issue of multi-mapreduce in general, whether spark or map-reduce.)  

But perhaps there are other reasons to go with the marker over the partition, like performance.   But I'm wary that they're worth the complexity for the user.

edx/analytics/tasks/insights/user_activity.py
```diff
@@ -146,6 +147,83 @@ def run(self):
...
+        return map(
+            lambda date: get_target_from_url(
+                url_path_join(self.output_root, 'dt={}'.format(date.isoformat()))
+            ),
+            self.interval
+        )
+
+    def on_success(self):  # pragma: no cover
+        """Overload the success method to touch the _SUCCESS file.  Any class that uses a separate Marker file from the
+        data file will need to override the base on_success() call to create this marker."""
```
brianhw: nit:  make docstring multiline....

edx/analytics/tasks/insights/location_per_course.py
```diff
@@ -163,6 +164,83 @@ def run(self):
...
+
+    def on_success(self):  # pragma: no cover
+        self.output().touch_marker()
+
+    def run(self):
+        self.remove_output_on_overwrite()
+        if self.output_dir().exists():      # only check partitions if parent dir exists
+            for target in self.output_paths():
+                if target.exists():
+                    target.remove()
```
brianhw, line 216: It might be good to put in a comment here, because I'm not remembering if we did this because we were writing in append mode, or because of issues if we reduced the number of partitions we coalesced down to on output, or because Spark really doesn't want the files around when it's planning on writing them out.   Do you remember?

brianhw (commented):
Good progress!  Sorry to raise questions about marker files, but it's a challenge in general in any schema to figure out what has been done and what needs to be done.

edx/analytics/tasks/insights/location_per_course.py
rao-abdul-mannan: This is done to avoid duplicates in case of append mode.

edx/analytics/tasks/insights/location_per_course.py
rao-abdul-mannan: My approach is to handle it with non-historic task and if it isn't possible then I'll make adjustments here.

edx/analytics/tasks/insights/location_per_course.py
rao-abdul-mannan: Yes, it is for .exists() functionality.

edx/analytics/tasks/insights/location_per_course.py
rao-abdul-mannan: It will write to the output_dir() ( not the individual partition )

rao-abdul-mannan: With overwrite mode, it won't keep other partitions i.e. those partitions which aren't in this interval, will be deleted from the output_dir() path. So I'm using append mode and explicitly remove those partitions which will be written by the job to avoid duplicates.

edx/analytics/tasks/insights/user_activity.py
rao-abdul-mannan: Spark doesn't create empty partitions if there is no data when writing output using partitionBy(). In such cases, checking for partitions existence will fail so I opted for the marker approach.

edx/analytics/tasks/insights/user_activity.py
```diff
@@ -203,8 +289,138 @@ def columns(self):
...
+                UserActivityTaskSpark(
+                    interval=overwrite_interval,
+                    warehouse_path=self.warehouse_path,
+                    output_root=self.user_activity_hive_table_path(),
+                    overwrite=True,
+                )
+            )
+        yield required_tasks
 
+    def user_activity_hive_table_path(self, *args):
```
azafty468, line 330: I think as we convert jobs we'll not want to reference the word "hive".  It definitely makes sense now when you are interleaving Spark components.  But in the future we'll have no hive components and names like this will be confusing.  Again though, no change in this PR, just for future reference.

edx/analytics/tasks/insights/user_activity.py
azafty468: I'll echo Brian's thoughts on using an obscure named marker file to indicate if a job has succeeded in the past.  We've obviously done it in the past, and in this specific example we've done it.  I'd like to challenge engineers to see if we can come up with a better method even if all we find out is that our current implementation is the cleanest option.  In this case I'm fine with continuing our past behaviors.

edx/analytics/tasks/util/spark_util.py
```diff
@@ -0,0 +1,144 @@
...
+        default_value = event[key]
+    except KeyError:
+        pass
+    return default_value
+
+
+def get_event_time_string(event_time):
+    """Returns the time of the event as an ISO8601 formatted string."""
+    try:
+        # Get entry, and strip off time zone information.  Keep microseconds, if any.
```
azafty468, line 52: Dah, time string processing is a hobgoblin.  I suspect all of our event times are converted to UTC.  We may want to confirm that and not drop the timezone field.

azafty468 (approved):
I left a few comment as I read the code.  If I remember correctly the changes in the spark base PR was getting merged and these were being left unmerged.  I could be wrong.  The comments I wrote are all forward looking to future PRs.  If we aren't merging this guy no additional work is necessary.

edx/analytics/tasks/insights/user_activity.py
rao-abdul-mannan: Right, i'll keep it in mind.

edx/analytics/tasks/util/spark_util.py
rao-abdul-mannan: It is exactly the same as the one we use for hadoop.

PR #477, opened 2018-01-25
When writing to S3, don't validate the bucket.

brianhw:
If Boto is passed "validate=True", it will require an additional
permission to be present when asked to list all of the keys in the
bucket.  We want to minimize the set of required permissions so we get a
reference to the bucket without validating that it exists.  It should
only require PutObject and PutObjectAcl permissions in order to write to
the target bucket.

This is particularly the case with answer-distribution, which writes out
to a bucket that is used by LMS for course downloads.

It was awkward to maintain this change in the pipeline code, so the change
was instead made in the edx/luigi fork.    See the change here:

https://github.com/edx/luigi/commit/eb45bcc52243de11b2b16a81229ac584fe1e601b


rao-abdul-mannan (approved):

azafty468 (approved):
👍

PR #478, opened 2018-01-25
Doc generation requires a single requirements file

brianhw:
We currently have makefile targets for installing requirements for doc, but these are not actually used by our readthedocs configuration. Instead, it is configured to load only requirements/docs.txt. Before work done to use pip-compile to build dependencies, the docs.txt file included default.txt, which in turn included base.txt, so the build worked. The pip-compile work separated these, and so recent builds are failing because they are not finding Luigi. 

Addressing DE-524.


rao-abdul-mannan (approved):
Looks good

azafty468 (approved):
👍

PR #479, opened 2018-01-31
Doc for JSON list input and incremental video.

brianhw:
We didn't update the doc when incremental video went in, so I added it to this.

I also updated the examples for answer distribution, to reflect changes due to upgrade to Luigi2.


rao-abdul-mannan (approved):
Looks good

PR #480, opened 2018-02-06
ENT-777 Pipeline tasks to import enterprise enrollment data

brittneyexline:
This PR adds a new set of pipeline tasks for populating a table in the result store, which will be used to power the enterprise data api. It will read from multiple tables in the LMS and create a table with data similar to the one in vertica we use for automated csv reports.


edx/analytics/tasks/insights/enterprise_enrollments.py
```diff
@@ -0,0 +1,304 @@
...
+                END AS course_duration,
+                course.course_min_effort,
+                course.course_max_effort,
+                user_profile.date_joined,
+                user_profile.email,
+                user_profile.username,
+                YEAR(CURRENT_DATE() - user_profile.year_of_birth) AS user_age,
+                user_profile.level_of_education,
+                user_profile.gender,
+                user_profile.last_country,
```
brianhw: How important is this bit of information?   It comes at a cost of complexity.  If you look at load_internal_reporting_user.py, you'll see that this "internal_reporting_user" table is created using a join of auth_user, auth_userprofile, and a third table that is the result of the location run for Insights.   This third table is loaded using ExternalLastCountryOfUserToHiveTask, which assumes that the data has already been calculated.  So it introduces a dependency on another job -- the location info for insights job.   If instead you were able to make do without this column, you could use auth_userprofile instead of internal_reporting_user, and you would have no  dependencies on any other workflow.

edx/analytics/tasks/insights/enterprise_enrollments.py
```diff
@@ -0,0 +1,304 @@
...
+                        1, 2
+                ) social_auth
+            ON
+                enterprise_user.user_id = social_auth.user_id
+            JOIN
+                course_catalog course
+            ON
+                enterprise_course_enrollment.course_id = course.course_id
+            JOIN
+                internal_reporting_user user_profile
```
brianhw: This is the table I'd like to suggest replacing with auth_userprofile.  

<religious>So much vertical whitespace!   Do you really need a newline after every JOIN and ON ?  The join sequence itself doesn't even fit on a screen. : )    The other joins in this repo seem to have the JOIN and ON clauses all on the same line.   The table aliases are longer here, but could at least be put on the same line.</religious>

brianhw (commented):
This looks great, Brittney!  You figured this out really quickly -- nice work!

One organizational comment.  While I know this workflow reads from the LMS db and writes to the resultstore, just like Insights does, it seems to me to be different in its generality.  Would this be of use to someone who is setting up an openedx installation? 

I'm thinking that instead of putting in the insights subdirectory, the two files could be put in a new parallel enterprise subdirectory.   (The database_imports.py would be put in enterprise/enterprise_imports.py or something, along with enterprise_enrollments.py.)  That way, open source users installing insights wouldn't be confused about these additional workflows, just like they currently ignore everything in /warehouse.  

I know that ecommerce imports are also in the same database_imports.py file.   I think they should move as well, presumably to warehouse or warehouse/finance.   (But that's not part of this.)

My main question about implementation relates to the use of "internal_reporting_user".  Beyond that, I see the next steps being the creation of an acceptance test. 

Very cool!

brianhw:
LGTM.   👍    Great job!

PR #481, opened 2018-02-14
Use Luigi return codes.

brianhw:
Modify launch-task to fail if any error conditions are encountered.   Use the return code
values recommended by Luigi documentation.

Also add six, since it is also needed on remote clusters.  For AWS, six is already installed by default, but it is not always present for non-AWS open source users.


brianhw:
Acceptance tests are currently running.

rao-abdul-mannan (approved):
Looks good

azafty468 (approved):
Looks good!  I'm glad it appears so simple!

PR #483, opened 2018-02-21
Add quality check to travis builds.

brianhw:
When travis builds were dockerized, quality checking was left out.  This PR tries to add it back in so there is earlier feedback than failing CI quality builds on master. 

It also fixes basic quality issues so the build would pass.


rao-abdul-mannan (approved):
Looks good

azafty468 (approved):
Looks good, I'm surprised at all the internal changes needed for quality failures.  Good job!

PR #484, opened 2018-02-28
Delete manifest file on overwrite.

brianhw:
Implement for module engagement to begin with, since its ModuleEngagementDataTask class will often have the same parameters (i.e. the given day to calculate), and therefore uses the same manifest file when being rerun.


azafty468 (approved):
LGTM.  I like the info level printing of the manifest log file path.

PR #485, opened 2018-03-13
Use spark task in production jobs

rao-abdul-mannan:
Purpose of this PR is to convert an existing hadoop task to spark & use it in production for atleast a month and see how it performs as well as figure out any issues encountered during this trial run.  
Related ticket DE-573

### DIFF
--- a/config/devstack.cfg
+++ b/config/devstack.cfg
@@ -131,3 +131,8 @@ api_root_url = http://localhost:8000/api/courses/v1/courses/
 
 [course-blocks]
 api_root_url = http://localhost:8000/api/courses/v1/blocks/
+
+[spark]
+driver-memory=3g
+executor-memory=3g
+executor-cores=1
--- a/edx/analytics/tasks/common/spark.py
+++ b/edx/analytics/tasks/common/spark.py
@@ -1,0 +1,324 @@
+import ast
+import json
+import os
+import tempfile
+import zipfile
+
+import luigi.configuration
+from luigi.contrib.spark import PySparkTask
+
+from edx.analytics.tasks.common.pathutil import EventLogSelectionDownstreamMixin, PathSelectionByDateIntervalTask
+from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
+
+_file_path_to_package_meta_path = {}
+
+
+def get_package_metadata_paths():
+    """
+    List of package metadata to be loaded on EMR cluster
+    """
+    from distlib.database import DistributionPath
+
+    if len(_file_path_to_package_meta_path) > 0:
+        return _file_path_to_package_meta_path
+
+    dist_path = DistributionPath(include_egg=True)
+    for distribution in dist_path.get_distributions():
+        metadata_path = distribution.path
+        for installed_file_path, _hash, _size in distribution.list_installed_files():
+            absolute_installed_file_path = installed_file_path
+            if not os.path.isabs(installed_file_path):
+                absolute_installed_file_path = os.path.join(os.path.dirname(metadata_path), installed_file_path)
+            normalized_file_path = os.path.realpath(absolute_installed_file_path)
+            _file_path_to_package_meta_path[normalized_file_path] = metadata_path
+
+    return _file_path_to_package_meta_path
+
+
+def dereference(f):
+    if os.path.islink(f):
+        # by joining with the dirname we are certain to get the absolute path
+        return dereference(os.path.join(os.path.dirname(f), os.readlink(f)))
+    else:
+        return f
+
+
+def create_packages_archive(packages, archive_dir_path):
+    """
+    Create a zip archive for all the packages listed in packages and returns the list of zip file location.
+    """
+    import zipfile
+    archives_list = []
+    package_metadata_paths = get_package_metadata_paths()
+    metadata_to_add = dict()
+
+    package_zip_path = os.path.join(archive_dir_path, 'packages.zip')
+    package_zip = zipfile.ZipFile(package_zip_path, "w", compression=zipfile.ZIP_DEFLATED)
+    archives_list.append(package_zip_path)
+
+    def add(src, dst, package_name):
+        # Ensure any entry points and other egg-info metadata is also transmitted along with
+        # this file. If it is associated with any egg-info directories, ship them too.
+        metadata_path = package_metadata_paths.get(os.path.realpath(src))
+        if metadata_path:
+            metadata_to_add[package_name] = metadata_path
+
+        package_zip.write(src, dst)
+
+    def add_files_for_package(sub_package_path, root_package_path, root_package_name, package_name):
+        for root, dirs, files in os.walk(sub_package_path):
+            if '.svn' in dirs:
+                dirs.remove('.svn')
+            for f in files:
+                if not f.endswith(".pyc") and not f.startswith("."):
+                    add(dereference(root + "/" + f),
+                        root.replace(root_package_path, root_package_name) + "/" + f,
+                        package_name)
+
+    for package in packages:
+        # Archive each package
+        if not getattr(package, "__path__", None) and '.' in package.__name__:
+            package = __import__(package.__name__.rpartition('.')[0], None, None, 'non_empty')
+
+        n = package.__name__.replace(".", "/")
+
+        # Check length of path, because the attribute may exist and be an empty list.
+        if len(getattr(package, "__path__", [])) > 0:
+            # TODO: (BUG) picking only the first path does not
+            # properly deal with namespaced packages in different
+            # directories
+            p = package.__path__[0]
+
+            if p.endswith('.egg') and os.path.isfile(p):
+                raise 'Not going to archive egg files!!!'
+                # Add the entire egg file
+                # p = p[:p.find('.egg') + 4]
+                # add(dereference(p), os.path.basename(p))
+
+            else:
+                # include __init__ files from parent projects
+                root = []
+                for parent in package.__name__.split('.')[0:-1]:
+                    root.append(parent)
+                    module_name = '.'.join(root)
+                    directory = '/'.join(root)
+
+                    add(dereference(__import__(module_name, None, None, 'non_empty').__path__[0] + "/__init__.py"),
+                        directory + "/__init__.py",
+                        package.__name__)
+
+                add_files_for_package(p, p, n, package.__name__)
+
+        else:
+            f = package.__file__
+            if f.endswith("pyc"):
+                f = f[:-3] + "py"
+            if n.find(".") == -1:
+                add(dereference(f), os.path.basename(f), package.__name__)
+            else:
+                add(dereference(f), n + ".py", package.__name__)
+
+        # include metadata in the same zip file
+        metadata_path = metadata_to_add.get(package.__name__)
+        if metadata_path is not None:
+            add_files_for_package(metadata_path, metadata_path, os.path.basename(metadata_path), package.__name__)
+
+    return archives_list
+
+
+class EventLogSelectionMixinSpark(EventLogSelectionDownstreamMixin):
+    """
+    Extract events corresponding to a specified time interval.
+    """
+    path_targets = None
+
+    def __init__(self, *args, **kwargs):
+        """
+        Call path selection task to get list of log files matching the pattern
+        """
+        super(EventLogSelectionDownstreamMixin, self).__init__(*args, **kwargs)
+        self.lower_bound_date_string = self.interval.date_a.strftime('%Y-%m-%d')  # pylint: disable=no-member
+        self.upper_bound_date_string = self.interval.date_b.strftime('%Y-%m-%d')  # pylint: disable=no-member
+
+    def get_log_schema(self):
+        """
+        Get spark based schema for processing event logs
+        :return: Spark schema
+        """
+        from pyspark.sql.types import StructType, StringType
+        event_schema = StructType().add("POST", StringType(), True).add("GET", StringType(), True)
+        module_schema = StructType().add("display_name", StringType(), True) \
+            .add("original_usage_key", StringType(), True) \
+            .add("original_usage_version", StringType(), True) \
+            .add("usage_key", StringType(), True)
+        context_schema = StructType().add("command", StringType(), True) \
+            .add("course_id", StringType(), True) \
+            .add("module", module_schema) \
+            .add("org_id", StringType(), True) \
+            .add("path", StringType(), True) \
+            .add("user_id", StringType(), True)
+
+        event_log_schema = StructType() \
+            .add("username", StringType(), True) \
+            .add("event_type", StringType(), True) \
+            .add("ip", StringType(), True) \
+            .add("agent", StringType(), True) \
+            .add("host", StringType(), True) \
+            .add("referer", StringType(), True) \
+            .add("accept_language", StringType(), True) \
+            .add("event", event_schema) \
+            .add("event_source", StringType(), True) \
+            .add("context", context_schema) \
+            .add("time", StringType(), True) \
+            .add("name", StringType(), True) \
+            .add("page", StringType(), True) \
+            .add("session", StringType(), True)
+
+        return event_log_schema
+
+    def get_event_log_dataframe(self, spark, *args, **kwargs):
+        from pyspark.sql.functions import to_date, udf, struct, date_format
+        path_targets = PathSelectionByDateIntervalTask(
+            source=self.source,
+            interval=self.interval,
+            pattern=self.pattern,
+            date_pattern=self.date_pattern,
+        ).output()
+        self.path_targets = [task.path for task in path_targets]
+        dataframe = spark.read.format('json').load(self.path_targets, schema=self.get_log_schema())
+        dataframe = dataframe.filter(dataframe['time'].isNotNull()) \
+            .withColumn('event_date', date_format(to_date(dataframe['time']), 'yyyy-MM-dd'))
+        dataframe = dataframe.filter(
+            (dataframe['event_date'] >= self.lower_bound_date_string) &
+            (dataframe['event_date'] < self.upper_bound_date_string)
+        )
+        return dataframe
+
+
+class SparkJobTask(OverwriteOutputMixin, PySparkTask):
+    """
+    Wrapper for spark task
+    """
+
+    _spark = None
+    _spark_context = None
+    _sql_context = None
+    _hive_context = None
+    _tmp_dir = None
+
+    driver_memory = luigi.Parameter(
+        config_path={'section': 'spark', 'name': 'driver-memory'},
+        description='Memory for spark driver',
+        significant=False,
+    )
+    executor_memory = luigi.Parameter(
+        config_path={'section': 'spark', 'name': 'executor-memory'},
+        description='Memory for each executor',
+        significant=False,
+    )
+    executor_cores = luigi.Parameter(
+        config_path={'section': 'spark', 'name': 'executor-cores'},
+        description='No. of cores for each executor',
+        significant=False,
+    )
+    always_log_stderr = False  # log stderr if spark fails, True for verbose log
+
+    def init_spark(self, sc):
+        """
+        Initialize spark, sql and hive context
+        :param sc: Spark context
+        """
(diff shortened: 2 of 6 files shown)


brianhw:
Just noting that the travis builds failed:   "ERROR: /edx/app/analytics-pipeline/edx/analytics/tasks/insights/tests/test_user_activity.py Imports are incorrectly sorted."

edx/analytics/tasks/common/spark.py
```diff
@@ -0,0 +1,324 @@
...
+        return dereference(os.path.join(os.path.dirname(f), os.readlink(f)))
+    else:
+        return f
+
+
+def create_packages_archive(packages, archive_dir_path):
+    """
+    Create a zip archive for all the packages listed in packages and returns the list of zip file location.
+    """
+    import zipfile
```
brianhw, line 50: Do we need this here?   It's already imported at the top of the file.

edx/analytics/tasks/common/spark.py
```diff
@@ -0,0 +1,324 @@
...
+    archives_list = []
+    package_metadata_paths = get_package_metadata_paths()
+    metadata_to_add = dict()
+
+    package_zip_path = os.path.join(archive_dir_path, 'packages.zip')
+    package_zip = zipfile.ZipFile(package_zip_path, "w", compression=zipfile.ZIP_DEFLATED)
+    archives_list.append(package_zip_path)
+
+    def add(src, dst, package_name):
+        # Ensure any entry points and other egg-info metadata is also transmitted along with
```
brianhw, line 60: nit:  s/is/are/

rao-abdul-mannan:
@brianhw I sorted it with isort but it did not go away

edx/analytics/tasks/common/spark.py
```diff
@@ -0,0 +1,324 @@
...
+            .add("host", StringType(), True) \
+            .add("referer", StringType(), True) \
+            .add("accept_language", StringType(), True) \
+            .add("event", event_schema) \
+            .add("event_source", StringType(), True) \
+            .add("context", context_schema) \
+            .add("time", StringType(), True) \
+            .add("name", StringType(), True) \
+            .add("page", StringType(), True) \
+            .add("session", StringType(), True)
```
brianhw, line 175: So there are a bunch of schema entries here that we do not use for any of our workflows.   I would suggest that this base version be more minimal.   If there are other entries that are needed in special cases, this can be overridden, or added to using a super() call.  

I would suggest that all we need here are:  user_id, course_id, org_id from context, and username, event_type, ip, event_source, context, and time from the base level.

edx/analytics/tasks/common/spark.py
```diff
@@ -0,0 +1,324 @@
...
+        super(EventLogSelectionDownstreamMixin, self).__init__(*args, **kwargs)
+        self.lower_bound_date_string = self.interval.date_a.strftime('%Y-%m-%d')  # pylint: disable=no-member
+        self.upper_bound_date_string = self.interval.date_b.strftime('%Y-%m-%d')  # pylint: disable=no-member
+
+    def get_log_schema(self):
+        """
+        Get spark based schema for processing event logs
+        :return: Spark schema
+        """
+        from pyspark.sql.types import StructType, StringType
```
brianhw, line 148: Should this be at the top of the file?

edx/analytics/tasks/common/spark.py
```diff
@@ -0,0 +1,324 @@
...
+            .add("context", context_schema) \
+            .add("time", StringType(), True) \
+            .add("name", StringType(), True) \
+            .add("page", StringType(), True) \
+            .add("session", StringType(), True)
+
+        return event_log_schema
+
+    def get_event_log_dataframe(self, spark, *args, **kwargs):
+        from pyspark.sql.functions import to_date, udf, struct, date_format
```
brianhw, line 180: should this be at the top of the file?

edx/analytics/tasks/common/spark.py
```diff
@@ -0,0 +1,324 @@
...
+            .add("original_usage_key", StringType(), True) \
+            .add("original_usage_version", StringType(), True) \
+            .add("usage_key", StringType(), True)
+        context_schema = StructType().add("command", StringType(), True) \
+            .add("course_id", StringType(), True) \
+            .add("module", module_schema) \
+            .add("org_id", StringType(), True) \
+            .add("path", StringType(), True) \
+            .add("user_id", StringType(), True)
+
```
brianhw, line 160: nit:  either remove this newline, or add one after other schema definitions.

edx/analytics/tasks/common/spark.py
```diff
@@ -0,0 +1,324 @@
...
+            .add("event_source", StringType(), True) \
+            .add("context", context_schema) \
+            .add("time", StringType(), True) \
+            .add("name", StringType(), True) \
+            .add("page", StringType(), True) \
+            .add("session", StringType(), True)
+
+        return event_log_schema
+
+    def get_event_log_dataframe(self, spark, *args, **kwargs):
```
brianhw, line 179: This could use a docstring, to indicate at least what kind of object "spark" is that is being passed in.

edx/analytics/tasks/common/spark.py
```diff
@@ -0,0 +1,324 @@
...
+        dataframe = dataframe.filter(
+            (dataframe['event_date'] >= self.lower_bound_date_string) &
+            (dataframe['event_date'] < self.upper_bound_date_string)
+        )
+        return dataframe
+
+
+class SparkJobTask(OverwriteOutputMixin, PySparkTask):
+    """
+    Wrapper for spark task
```
brianhw, line 200: nit:  end doc sentences with a period.

edx/analytics/tasks/common/spark.py
```diff
@@ -0,0 +1,324 @@
...
+        config_path={'section': 'spark', 'name': 'executor-memory'},
+        description='Memory for each executor',
+        significant=False,
+    )
+    executor_cores = luigi.Parameter(
+        config_path={'section': 'spark', 'name': 'executor-cores'},
+        description='No. of cores for each executor',
+        significant=False,
+    )
+    always_log_stderr = False  # log stderr if spark fails, True for verbose log
```
brianhw, line 224: Is this something that should be configured, or is it to be changed in code by someone trying to debug something?  It might be worth a comment (on the line before).

edx/analytics/tasks/common/spark.py
```diff
@@ -0,0 +1,324 @@
...
+
+    def init_spark(self, sc):
+        """
+        Initialize spark, sql and hive context
+        :param sc: Spark context
+        """
+        from pyspark.sql import SparkSession, SQLContext, HiveContext
```
brianhw, line 231: Should these be moved to the top of the file now?

edx/analytics/tasks/common/spark.py
```diff
@@ -0,0 +1,324 @@
...
+        self._sql_context = SQLContext(sc)
+        self._spark_context = sc
+        self._spark = SparkSession.builder.getOrCreate()
+        self._hive_context = HiveContext(sc)
+
+    def spark_job(self):
+        """
+        Spark code for the job
```
brianhw, line 239: nit: end with a period.

edx/analytics/tasks/common/spark.py
```diff
@@ -0,0 +1,324 @@
...
+    def init_spark(self, sc):
+        """
+        Initialize spark, sql and hive context
+        :param sc: Spark context
+        """
+        from pyspark.sql import SparkSession, SQLContext, HiveContext
+        self._sql_context = SQLContext(sc)
+        self._spark_context = sc
+        self._spark = SparkSession.builder.getOrCreate()
+        self._hive_context = HiveContext(sc)
```
brianhw, line 235: Should this be commented out for now?  I don't believe we are using Hive in Spark, are we?

edx/analytics/tasks/util/constants.py
```diff
@@ -0,0 +1,7 @@
+class PredicateLabels(object):
+    """Constants for predicate labels."""
```
brianhw, line 2: This doesn't explain enough to me.  Having moved it out of user_activity means it has much less context.

edx/analytics/tasks/insights/user_activity.py
```diff
@@ -169,10 +236,9 @@ def requires(self):
             overwrite_from_date = self.date - datetime.timedelta(days=self.overwrite_n_days)
             overwrite_interval = luigi.date_interval.Custom(overwrite_from_date, self.date)
 
-            yield UserActivityTask(
+            yield UserActivityTaskSpark(
```
brianhw, line 239: I'm not sure if the purpose of this PR is to merge the code to master, but I do not want to change the default behavior of master by merging this in as-is.  We should be able to run the two workflows side-by-side, the spark and the hive.

Perhaps the spark user-activity workflow can be put in a separate file, and the relevant sequence of tasks can be recreated by overriding the requires methods of the dependent tasks.  If so, we could move the user-activity part of spark_utils into that file. 

But wait, I thought that you had implemented an alternative more of the workflow than just this one task -- that you had also replaced the Hive calculations with Spark SQL.  Am I misremembering?

edx/analytics/tasks/util/spark_util.py
```diff
@@ -0,0 +1,78 @@
+"""Support for spark tasks"""
+import edx.analytics.tasks.util.opaque_key_util as opaque_key_util
+from edx.analytics.tasks.util.constants import PredicateLabels
+
+
+def get_event_predicate_labels(event_type, event_source):
+    """
+    Creates labels by applying hardcoded predicates to a single event.
+    Don't pass whole event row to any spark UDF as it generates a different output than expected
```
brianhw, line 9: I don't understand this.  (Also nit: end sentence with period.)

edx/analytics/tasks/util/spark_util.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+    # We only want the explicit event, not the implicit form.
+    # return 'test'
```
brianhw, line 12: I assume this line should be removed?

edx/analytics/tasks/util/spark_util.py
```diff
@@ -0,0 +1,78 @@
...
+
+    # Get the course_id from the data, and validate.
+    course_id = opaque_key_util.normalize_course_id(get_key_value_from_event(event_context, 'course_id', ''))
+    if course_id:
+        if opaque_key_util.is_valid_course_id(course_id):
+            return course_id
+
+    return ''
+
+    # TODO : make it work with url as well
```
brianhw, line 61: When does this need to be done?   I'm guessing for event exports.

edx/analytics/tasks/util/spark_util.py
```diff
@@ -0,0 +1,78 @@
...
+    if event_source in ('browser', 'mobile'):
+        if event_type == 'play_video':
+            labels += ',' + PredicateLabels.PLAY_VIDEO_LABEL
+
+    return labels
+
+
+def get_key_value_from_event(event, key, default_value=None):
+    """
+    Get value from event dict by key
```
brianhw, line 34: nit: end with period.   
But is 'event' here a dict, or some other object?   If it were really a dict, this wouldn't be necessary, but it's something else.   It would be good to note here what that something is.....

But this code may not be called at present anyway, because the code in get_course_id() that called it is commented out?

edx/analytics/tasks/util/spark_util.py
```diff
@@ -0,0 +1,78 @@
...
+        default_value = event[key]
+    except KeyError:
+        pass
+    return default_value
+
+
+def get_course_id(event_context, from_url=False):
+    """
+    Gets course_id from event's data.
+    Don't pass whole event row to any spark UDF as it generates a different output than expected
```
brianhw, line 47: I don't understand this.  (Also nit: end sentence with period.)

edx/analytics/tasks/insights/user_activity.py
```diff
@@ -144,6 +141,76 @@ def run(self):
         return super(UserActivityTask, self).run()
 
 
+class UserActivityTaskSpark(EventLogSelectionMixinSpark, WarehouseMixin, SparkJobTask):
```
brianhw, line 144: As mentioned below, maybe this should move to a separate file, and define an entirely separate workflow?

edx/analytics/tasks/common/spark.py
```diff
@@ -0,0 +1,324 @@
...
+        if len(dependencies_list) > 0:
+            for file in dependencies_list:
+                self._spark_context.addPyFile(file)
+
+    def get_luigi_configuration(self):
+        """
+        Return luigi configuration as dict for spark task
+        luigi configuration cannot be retrieved directly from luigi's get_config method inside spark task
+        """
+
```
brianhw, line 298: Is this method doing anything?   It's not returning configuration as a dict.  nit: no blank line.  End comments with period.

edx/analytics/tasks/insights/tests/test_user_activity.py
```diff
@@ -6,15 +6,14 @@
...
 
+from ddt import data, ddt, unpack
 from edx.analytics.tasks.common.tests.map_reduce_mixins import MapperTestMixin, ReducerTestMixin
-from edx.analytics.tasks.insights.user_activity import (
-    ACTIVE_LABEL, PLAY_VIDEO_LABEL, POST_FORUM_LABEL, PROBLEM_LABEL, InsertToMysqlCourseActivityTask, UserActivityTask
-)
+from edx.analytics.tasks.insights.user_activity import InsertToMysqlCourseActivityTask, UserActivityTask
+from edx.analytics.tasks.util.constants import PredicateLabels
 from edx.analytics.tasks.util.tests.opaque_key_mixins import InitializeLegacyKeysMixin, InitializeOpaqueKeysMixin
+from mock import Mock, call
```
brianhw, line 16: Not clear on the reorganization here, but it breaks the isort test in the quality build.  We separate out python imports, third party imports, and then our local code imports into separate groups.  Presumably the ddt and mock imports should go back where they were.

edx/analytics/tasks/insights/user_activity.py
```diff
@@ -11,7 +11,9 @@
 from edx.analytics.tasks.common.mapreduce import MapReduceJobTaskMixin, MultiOutputMapReduceJobTask
 from edx.analytics.tasks.common.mysql_load import MysqlInsertTask
 from edx.analytics.tasks.common.pathutil import EventLogSelectionDownstreamMixin, EventLogSelectionMixin
+from edx.analytics.tasks.common.spark import EventLogSelectionMixinSpark, SparkJobTask
```
brianhw, line 14: I see, so this may answer my question about imports.  If we put them at the top of common/spark.py, and this is included in our regular code, then we're making *everything* dependent on spark code.   I think it should be pulled out into a separate workflow.

config/devstack.cfg
```diff
@@ -131,3 +131,8 @@ api_root_url = http://localhost:8000/api/courses/v1/courses/
 
 [course-blocks]
 api_root_url = http://localhost:8000/api/courses/v1/blocks/
+
+[spark]
+driver-memory=3g
+executor-memory=3g
+executor-cores=1
```
brianhw, line 138: nit: Add newline.

edx/analytics/tasks/common/spark.py
```diff
@@ -0,0 +1,324 @@
...
+        return None
+
+    def app_options(self):
+        """
+        List of options that needs to be passed to spark task
+        """
+        options = {}
+        task_config = self.get_luigi_configuration()  # load task dependencies first if any
+        if isinstance(task_config, dict):
+            options = task_config
```
brianhw, line 308: Should these three lines above be removed?   Was it an experiment that didn't work out?

edx/analytics/tasks/common/spark.py
```diff
@@ -0,0 +1,324 @@
...
+        """
+        Return luigi configuration as dict for spark task
+        luigi configuration cannot be retrieved directly from luigi's get_config method inside spark task
+        """
+
+        return None
+
+    def app_options(self):
+        """
+        List of options that needs to be passed to spark task
```
brianhw, line 303: Where does this get called?   Also, maybe provide more explanation as to why the call to Luigi's configuration is being done here.  It's not clear to me why we couldn't put this logic into _load_internal_dependency_on_cluster().

brianhw (changes requested):
I'm not sure if this is really going to master, or if the proposal was to run this in its branch.  If just the latter, then most of my nit-picking comments can be ignored.  (I have different standards for master than for experimental branches.)  

The main question I have is scope -- should this include as well all the Hive replacement that I thought you wrote?  I only see the initial history task included here.  

But if we were to run the hive replacement workflow, then I think it should be defined separately and pulled into a separate file, and possibly inheriting and overriding some original classes as necessary to be more DRY about it.

edx/analytics/tasks/util/spark_util.py
rao-abdul-mannan: This will be needed when running spark based tests.

edx/analytics/tasks/util/spark_util.py
rao-abdul-mannan: This function is spark equivalent of eventlog's course_id method and returns valid course id from the event context as opposed to the complete event row passed to eventlog's method.

edx/analytics/tasks/common/spark.py
rao-abdul-mannan: Luigi configuration cannot be retrieved directly inside a spark job. Due to this issue, any configuration that is required by spark job is provided as a command line argument to spark-submit. This method is called directly by luigi to map custom parameters to arguments. 

This method along with get_luigi_configuration aren't used in this particular workflow but will be used for other spark tasks as used in #476

edx/analytics/tasks/common/spark.py
rao-abdul-mannan: No, let me explain these functions in more detail. 
Since luigi configuration cannot be retrieved inside spark task, any spark task which will need to access those configuration should overwride get_luigi_configuration method and return required luigi configurations as a dictionary. 

app_options method will  then convert these dictionary configurations and pass it to spark task as command line arguments via spark-submit. Again, these methods aren't used in this task but they will be used later with other spark tasks.

edx/analytics/tasks/insights/tests/test_user_activity.py
rao-abdul-mannan: Isort fails the test in both cases, whether they are organized this way or not. Though when sorted on local, this is how isort rearranges it.

edx/analytics/tasks/common/spark.py
rao-abdul-mannan: This method should be overridden in spark jobs if they need to access luigi's configuration.

edx/analytics/tasks/insights/user_activity.py
rao-abdul-mannan: I think its more manageable and easy to reference their counterpart hadoop job if new spark jobs are placed near them and in future it will be easy to just remove these classes entirely once we are done with them.

edx/analytics/tasks/util/spark_util.py
rao-abdul-mannan: It is called to get the course id from the event context. I have updated the docs about the object type

edx/analytics/tasks/util/spark_util.py
rao-abdul-mannan: Yup, removed it

edx/analytics/tasks/util/spark_util.py
rao-abdul-mannan: Perhaps name is confusing here, so i'll rename it. It is basically spark equivalent method of get_predicate_labels.

edx/analytics/tasks/common/spark.py
rao-abdul-mannan: Right.

edx/analytics/tasks/common/spark.py
rao-abdul-mannan: No, we can't move it to the top since it is imported in user_activity file which still has hadoop/hive jobs and it will error it when running those jobs.

PR #489, opened 2018-03-22
Spike: Integrated test of sqoop from vertica

azafty468:
Encapsulates the sqoop command targeting a Vertica database.  Work for DE-388

### DIFF
--- a/edx/analytics/tasks/common/sqoop.py
+++ b/edx/analytics/tasks/common/sqoop.py
@@ -283,3 +283,79 @@ class SqoopImportRunner(luigi.contrib.hadoop.JobRunner):
             except Exception:
                 log.exception("Unable to dump metadata information.")
                 pass
+
+
+class SqoopImportFromVertica(SqoopImportTask):
+    """
+    An abstract task that uses Sqoop to read data out of a Vertica database and writes it to a file in CSV format.
+
+    * fields delimited by comma
+    * lines delimited by \n
+    * delimiters escaped by backslash
+    * fields optionally enclosed by single quotes (')
+    """
+    # Direct does not exist for the Vertica connector and so columns is now required
+    direct = luigi.BoolParameter(
+        default=None
+    )
+    columns = luigi.ListParameter(
+        description='A list of column names to be included.  Default is to include all columns.',
+    )
+    schema_name = luigi.Parameter(
+        description=''
+    )
+    null_string = luigi.Parameter(
+        default=None,
+        description='String to use to represent NULL values in output data.',
+    )
+    fields_terminated_by = luigi.Parameter(
+        default=',',
+        description='Defines the file separator to use on output.',
+    )
+    delimiter_replacement = luigi.Parameter(
+        default='\\',
+        description='Defines a character to use as replacement for delimiters '
+        'that appear within data values, for use with Hive.  Not specified by default.'
+    )
+
+    def connection_url(self, cred):
+        """Construct connection URL from provided credentials."""
+        return 'jdbc:vertica://{host}/{database}'.format(host=cred['host'], database=self.database)
+
+    def import_args(self):
+        if self.columns is None or len(self.columns) == 0:
+            raise RuntimeError('Error Vertica\'s connector requires specific columns listed or a wild card be present. '
+                               ' No columns were supplied.')
+
+        arglist = [
+            '--target-dir', self.destination,
+            '--driver', 'com.vertica.jdbc.Driver',
+        ]
+
+        if self.where is not None:
+            where_qry = "({}) and $CONDITIONS".format(self.where)
+        else:
+            where_qry = "$CONDITIONS"
+
+        select_qry = "SELECT {columns} FROM {schema}.{table} WHERE {where_qry}".format(columns=self.columns,
+                                                                                       schema=self.schema_name,
+                                                                                       table=self.table_name,
+                                                                                       where_qry=where_qry)
+        arglist.extend(['--query', str(select_qry)])
+
+        if self.num_mappers is not None:
+            arglist.extend(['--num-mappers', str(self.num_mappers)])
+        else:
+            arglist.extend(['--num-mappers', '1'])
+        if self.verbose:
+            arglist.append('--verbose')
+        if self.null_string is not None:
+            arglist.extend(['--null-string', self.null_string, '--null-non-string', self.null_string])
+        if self.fields_terminated_by is not None:
+            arglist.extend(['--fields-terminated-by', self.fields_terminated_by])
+        arglist.extend(['--lines-terminated-by', '\n'])
+        if self.delimiter_replacement is not None:
+            arglist.extend(['--escaped-by', self.delimiter_replacement])
+        arglist.extend(['--optionally-enclosed-by', '\''])
+
+        return arglist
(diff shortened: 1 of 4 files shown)


edx/analytics/tasks/common/sqoop.py
```diff
@@ -283,3 +283,79 @@ def run_job(self, job):
...
+    """
+    # Direct does not exist for the Vertica connector and so columns is now required
+    direct = luigi.BoolParameter(
+        default=None
+    )
+    columns = luigi.ListParameter(
+        description='A list of column names to be included.  Default is to include all columns.',
+    )
+    schema_name = luigi.Parameter(
+        description=''
```
HassanJaveed84, line 305: I'd say we either remove 'description' key, or add a description here.

edx/analytics/tasks/common/sqoop.py
```diff
@@ -283,3 +283,79 @@ def run_job(self, job):
...
+
+    * fields delimited by comma
+    * lines delimited by \n
+    * delimiters escaped by backslash
+    * fields optionally enclosed by single quotes (')
+    """
+    # Direct does not exist for the Vertica connector and so columns is now required
+    direct = luigi.BoolParameter(
+        default=None
+    )
```
brianhw, line 300: If direct doesn't exist, is there a reason to make this a parameter?

edx/analytics/tasks/common/sqoop.py
```diff
@@ -283,3 +283,79 @@ def run_job(self, job):
...
+        description='String to use to represent NULL values in output data.',
+    )
+    fields_terminated_by = luigi.Parameter(
+        default=',',
+        description='Defines the file separator to use on output.',
+    )
+    delimiter_replacement = luigi.Parameter(
+        default='\\',
+        description='Defines a character to use as replacement for delimiters '
+        'that appear within data values, for use with Hive.  Not specified by default.'
```
brianhw, line 318: Why is this "not specified" by default, when there is clearly a default of backslash?

edx/analytics/tasks/common/sqoop.py
```diff
@@ -283,3 +283,79 @@ def run_job(self, job):
...
+    * lines delimited by \n
+    * delimiters escaped by backslash
+    * fields optionally enclosed by single quotes (')
+    """
+    # Direct does not exist for the Vertica connector and so columns is now required
+    direct = luigi.BoolParameter(
+        default=None
+    )
+    columns = luigi.ListParameter(
+        description='A list of column names to be included.  Default is to include all columns.',
```
brianhw, line 302: Comment is wrong about default....

edx/analytics/tasks/common/tests/test_sqoop.py
```diff
@@ -203,3 +241,40 @@ def test_overwrite(self):
         self.assertTrue(task.output().exists.called)
         self.assertTrue(task.output().remove.called)
         self.assertTrue(task.attempted_removal)
+
+    def test_success_vertica_with_complete_credentials(self):
+        self.create_and_run_vertica_task(table_name='example_table',
+                                         schema_name='fake_schema',
+                                         columns='field1,field2,field3')
```
brianhw, line 248: I don't understand how this comma-delimited string gets turned into a list.

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,100 @@
...
+        description='The Vertica schema that is the source of the sqoop command.'
+    )
+    table = luigi.Parameter(
+        default=None,
+        description='The Vertica table that is the source of the sqoop command.'
+    )
+    warehouse_path = luigi.Parameter(
+        config_path={'section': 'vertica-export-sqoop', 'name': 'warehouse_path'},
+        description='A URL location of the data warehouse.',
+    )
```
brianhw, line 37: Wondering if we should just use the same warehouse_path as always, but write to a "vertica" subdirectory or something like that.   Similar to what we do for database exports.

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+class LoadVerticaToS3TableTask(VerticaSourcedSqoopMixin, luigi.Task):
+    """
+    Sample S3 loader to S3
+
+    A sample loader that reads a table from Vertica and persists the entry to S3.  In order to use
+    SqoopImportFromVertica the caller must already know the Vertica schema, table name, and column names. This
+    functionality should be added in future development cycles.
```
brianhw, line 46: By "this functionality", do you mean the automatic determination of the column names, given a schema and table?

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,100 @@
...
+    def insert_source_task(self):
+        """The sqoop command that manages the connection to the source datasource."""
+        target_url = url_path_join(self.warehouse_path, self.database, self.schema, self.table)
+
+        return SqoopImportFromVertica(
+            schema_name=self.schema,
+            table_name=self.table,
+            credentials=self.credentials,
+            database=self.database,
+            columns='course_id,course_org_id,course_number,course_run,course_start,course_end,course_name',
```
brianhw, line 74: again, I seem to have missed where this gets turned into a ListParameter value.

brianhw (commented):
FWIW, here are my comments on this spike.   I hope they're helpful for the final work.

edx/analytics/tasks/common/sqoop.py
azafty468: Yup, totally correct.  I've added a description on my other PR

edx/analytics/tasks/common/sqoop.py
azafty468: Totally right.  I've removed the luigi parameter here and just made this option None.

edx/analytics/tasks/common/sqoop.py
azafty468: Ah, an old comment that wasn't updated when I changed the default.  I've updated the description to reflect the default value.

edx/analytics/tasks/common/sqoop.py
azafty468: And has been updated.

edx/analytics/tasks/common/tests/test_sqoop.py
azafty468: hahah, and in fact it wasn't.  I got a little aggressive with my list comprehensions.  This was also a bug with my vertica_export code.  I've since updated both the tests and my usage in vertica_export to make this a proper list.

edx/analytics/tasks/common/vertica_export.py
azafty468: So I think that would be a good idea but let's discuss this.  I'm going to move this comment from here to the same line in my newer PR.

edx/analytics/tasks/common/vertica_export.py
azafty468: Correct, automatic determination of the column names should be done automatically.  This task is now done and this comment has been removed.

edx/analytics/tasks/common/vertica_export.py
azafty468: Yup, that has been corrected.  Thanks!

azafty468:
Eclipsed by PR-491

PR #490, opened 2018-03-29
Dockerize analytics pipeline

rao-abdul-mannan:
This is still work in progress. 

Before running following commands, make sure that you have assigned enough memory ( 6Gb memory recommended ) to docker via menu **Docker -> Preferences -> Advanced -> Memory** 

##### build analytics pipeline, hadoop & spark dockers first
`make dev.build.analyticstack`
`docker build -t edxops/hadoop-namenode:2.7.2 -f docker/hadoop_namenode/Dockerfile .`
`docker build -t edxops/hadoop-datanode:2.7.2 -f docker/hadoop_datanode/Dockerfile .`
`docker build -t edxops/hadoop-nodemanager:2.7.2 -f docker/hadoop_nodemanager/Dockerfile .`
`docker build -t edxops/hadoop-resourcemanager:2.7.2 -f docker/hadoop_resourcemanager/Dockerfile .`
`docker build -t edxops/spark-master:2.1.0-hadoop2.7 -f docker/spark_master/Dockerfile .`
##### provision pipeline containers
`make dev.provision`
##### access analytics pipeline container's shell
`make shell`

--------------------------------------------------------------------------------------------------
##### GeoLiteCountry.dat for Geography tasks
`make shell`
`curl -fSL http://geolite.maxmind.com/download/geoip/database/GeoLiteCountry/GeoIP.dat.gz -o /var/tmp/GeoIP.dat.gz`
`gunzip /var/tmp/GeoIP.dat.gz`
`mv GeoIP.dat geo.dat`
`hdfs dfs -put /var/tmp/geo.dat /edx-analytics-pipeline/;`

##### Build ( edx-opaque-keys & edx-ccx-keys ) egg files for spark tasks
```
cd /var/tmp/ && curl -fSL https://files.pythonhosted.org/packages/63/86/d4bf9c7e7a720125b0572c43dda002f72be2cc66313be601cd7b6cccb2ad/edx-opaque-keys-0.4.tar.gz -o /var/tmp/edx-opaque-keys-0.4.tar.gz && tar -xzf /var/tmp/edx-opaque-keys-0.4.tar.gz && cd /var/tmp/edx-opaque-keys-0.4 && python setup.py bdist_egg && mkdir /var/tmp/edx_egg_files/ && cp /var/tmp/edx-opaque-keys-0.4/dist/edx_opaque_keys-0.4-py2.7.egg /var/tmp/edx_egg_files/edx_opaque_keys.egg
```

```
cd /var/tmp/ && curl -fSL https://files.pythonhosted.org/packages/a4/03/444d30a3859e36ef2273ae6254c011e1fb3a02f4e3dba16008ecf0b4bdb3/edx-ccx-keys-0.2.1.tar.gz -o /var/tmp/edx-ccx-keys-0.2.1.tar.gz && tar -xzf /var/tmp/edx-ccx-keys-0.2.1.tar.gz && cd /var/tmp/edx-ccx-keys-0.2.1 && python setup.py bdist_egg cp /var/tmp/edx-ccx-keys-0.2.1/dist/edx_ccx_keys-0.2.1-py2.7.egg  /var/tmp/edx_egg_files/edx_ccx_keys.egg
```

##### For running acceptance tests
`make shell`
`make docker-test-acceptance-local`

## P.S. Do not merge


azafty468:
So this looks reasonable but I have to admit that I've never setup hadoop on docker, nor have I interfaced with the docker setup here at edX.

Dockerfile
```diff
@@ -1,22 +1,106 @@
-# docker build -t edxops/analytics-pipeline .
-
-FROM edxops/python:2.7
+FROM ubuntu:16.04
```
jmbowman: Could this derive from edxops/xenial-common instead?  I realize that adds ansible back to the image, but it also fixes a locale issue, adds Open edX release identification, and would automatically inherit any other customizations we make to our base image moving forward.

Dockerfile
```diff
@@ -1,22 +1,106 @@
...
+RUN curl -fSL "$HIVE_URL" -o /var/tmp/$HIVE_DIST_FILE \
+    && tar -xzf /var/tmp/$HIVE_DIST_FILE -C $HIVE_HOME --strip-components=1 \
+    && rm -f /var/tmp/$HIVE_DIST_FILE
+ADD scripts/hive-site.xml.template $HIVE_HOME/conf/hive-site.xml
+
+# SPARK
+RUN curl -fSL "$SPARK_URL" -o /var/tmp/$SPARK_DIST_FILE \
+    && tar -xzf /var/tmp/$SPARK_DIST_FILE -C $SPARK_HOME --strip-components=1 \
+    && echo 'spark.master  spark://sparkmaster:7077\nspark.eventLog.enabled  true\nspark.eventLog.dir  hdfs://namenode:8020/tmp/spark-events\nspark.history.fs.logDirectory  hdfs://namenode:8020/tmp/spark-events\nspark.sql.warehouse.dir hdfs://namenode:8020/spark-warehouse' > $SPARK_HOME/conf/spark-defaults.conf \
+    && rm -f /var/tmp/$SPARK_DIST_FILE
```
jmbowman, line 95: Given that Hadoop, Hive, and Spark are in separate containers, this seems like an awful lot of setup for them in this one.  Is this really all needed for the client libraries?

Makefile
```diff
@@ -106,3 +106,29 @@ docs: docs-requirements docs-local
 
 todo:
 	pylint --disable=all --enable=W0511 edx
+
+### Docker commands
+
+dev.build.permissions:
+	chmod +x scripts/hadoop-docker-command.sh scripts/provision.sh
```
jmbowman: Can't this just be done to the files in git, so the permissions don't have to be done for each checkout?

docker-compose.yml
```diff
@@ -0,0 +1,125 @@
+version: '2'
+services:
+
+  resultstore:
+    image: mysql:5.6
+    container_name: edx.devstack.analytics_pipeline.resultstore
```
jmbowman: Once moved to the devstack repo, the plan is to use the shared mysql container there, correct?

docker-compose.yml
```diff
@@ -0,0 +1,125 @@
...
+      CORE_CONF_fs_defaultFS: "hdfs://namenode:8020"
+      YARN_CONF_yarn_log___aggregation___enable: 'true'
+      YARN_CONF_yarn_nodemanager_aux___services: mapreduce_shuffle
+      YARN_CONF_yarn_nodemanager_aux___services_mapreduce_shuffle_class: 'org.apache.hadoop.mapred.ShuffleHandler'
+      MAPRED_CONF_mapreduce_framework_name: yarn
+    depends_on:
+      - namenode
+      - datanode
+    ports:
+      - 8088:8088      # resource manager web ui
```
jmbowman: Be aware that any exposed ports like this are accessible not just from the host OS, but from any computer on its network.  Any exposed services which offer significant flexibility in altering data or running arbitrary code should be protected by passwords which are not statically configured for all instances of the container; we randomly generate and log passwords on startup for the devstack browser containers' VNC services, for example.

scripts/provision.sh
```diff
@@ -0,0 +1,50 @@
...
+echo "Waiting for MySQL"
+until docker exec -i edx.devstack.analytics_pipeline.resultstore mysql -uroot -se "SELECT EXISTS(SELECT 1 FROM mysql.user WHERE user = 'root')" &> /dev/null
+do
+  printf "."
+  sleep 1
+done
+
+# In the event of a fresh MySQL container, wait a few seconds for the server to restart
+# This can be removed once https://github.com/docker-library/mysql/issues/245 is resolved.
+sleep 20
```
jmbowman, line 25: It looks like that ticket got closed as "working as intended".  It's probably not a huge deal since this happens only on provisioning, but we could potentially avoid the sleep by checking for an occurrence of "mysqld: ready for connections." in the container log after "MySQL init process done. Ready for start up."  I'd hold off on actually doing that until we determine if the sleep delay really poses a burden, though.

scripts/provision.sh
```diff
@@ -0,0 +1,50 @@
...
+
+echo -e "MySQL ready"
+
+echo -e "${GREEN}Creating databases and users...${NC}"
+docker exec -i edx.devstack.analytics_pipeline.resultstore mysql -uroot mysql < ./scripts/provision.sql
+sleep 20
```
jmbowman: Why the additional sleep here?

Dockerfile
rao-abdul-mannan: Unfortunately yes, without these libraries we can't connect to their containers & use them.

docker-compose.yml
rao-abdul-mannan: Yes, thats right.

Makefile
rao-abdul-mannan: Yeah that can be done, this is just a temporary solution for now.

scripts/provision.sh
rao-abdul-mannan: I'm just following devstack repo provision script since it will be moved there eventually, so I didn't add any other check besides using theirs.

scripts/provision.sh
rao-abdul-mannan: Initializing hive metastore on first try was complaining about access issues with hive user so I added this delay to try to solve that, which still fails sometimes though.

Dockerfile
rao-abdul-mannan: Good suggestion,I can give it a try and see how it works.

docker-compose.yml
rao-abdul-mannan: Just read more about it in detail, I will bind these exposed ports to localhost ( 127.0.0.1 ) only.

Dockerfile
rao-abdul-mannan: As for Hive, running a separate container seems like overkill as it only requires warehouse directory and metastore database to connect to  and for that hadoop & resultstore containers are enough. 
In short, hive package installed in pipeline container is configured to connect with other containers and everything works out.

scripts/provision.sh
rao-abdul-mannan: Issue has been resolved using docker volumes

config/docker.cfg
```diff
@@ -0,0 +1,136 @@
+[retcode]
```
macdiesel: Perhaps this file is better named luigi_docker.cfg?

docker-compose.yml
```diff
@@ -0,0 +1,131 @@
+version: '2'
+services:
+
+  resultstore:
```
macdiesel: Why not just name this mysql now and avoid having to make any changes when you port it to the devstack repo?

docker-compose.yml
```diff
@@ -0,0 +1,131 @@
...
+    depends_on:
+      - sparkmaster
+    environment:
+      - SPARK_MASTER=spark://sparkmaster:7077
+    ports:
+      - 127.0.0.1:8081:8081
+    env_file:
+      - ./scripts/hadoop.env
+
+  analyticspipelinedocker:
```
macdiesel: You can remove the docker part of this name.

config/docker.cfg
rao-abdul-mannan: Sounds good

docker-compose.yml
rao-abdul-mannan: Yeah I tried to port these containers today to devstack docker compose file and realized it needs to be changed. Will do it.

docker-compose.yml
rao-abdul-mannan: Sure!

scripts/provision.sh
macdiesel: I'm ok with leaving this as is for now.

scripts/provision.sh
macdiesel: @rao-abdul-mannan so the sleep can be removed?

Dockerfile
```diff
@@ -1,22 +1,137 @@
...
+FROM edxops/xenial-common:latest
+MAINTAINER edxops
 
-FROM edxops/python:2.7
-ENV BOTO_CONFIG /dev/null
+USER root
+ENV BOTO_CONFIG=/dev/null \
+    JDK_URL=http://download.oracle.com/otn-pub/java/jdk/8u131-b11/d54c1d3a095b4ff2b6607d096fa80163/jdk-8u131-linux-x64.tar.gz \
+    JDK_DIST_FILE=jdk-8u131-linux-x64.tar.gz \
+    JAVA_HOME=/usr/lib/jvm/java-8-oracle \
```
macdiesel, line 8: It may help the future maintainers of this if the versions could be specified as ARGs at the top of the file then the variables used here in the ENV part of the file.

macdiesel (approved):
Good work Abdul.

scripts/provision.sh
rao-abdul-mannan: Sure, will update it.

Dockerfile
rao-abdul-mannan: That's a good suggestion. I'll make the changes.

Dockerfile
```diff
@@ -123,14 +125,22 @@ RUN pip install $COMMON_PIP_PACKAGES_PIP $COMMON_PIP_PACKAGES_SETUPTOOLS $COMMON
...
+    && . $ANALYTICS_PIPELINE_VENV/analytics_pipeline/bin/activate \
     && make test-requirements requirements
 
 RUN sudo chown hadoop:hadoop $COMMON_CFG_DIR/edx-analytics-pipeline/ \
-    && echo "{\"username\": \"$COMMON_MYSQL_READ_ONLY_USER\", \"host\": \"resultstore\", \"password\": \"$COMMON_MYSQL_READ_ONLY_PASS\", \"port\": 3306}" > $COMMON_CFG_DIR/edx-analytics-pipeline/input.json \
-    && echo "{\"username\": \"$ANALYTICS_PIPELINE_OUTPUT_DATABASE_USER\", \"host\": \"resultstore\", \"password\": \"$ANALYTICS_PIPELINE_OUTPUT_DATABASE_PASSWORD\", \"port\": 3306}" > $COMMON_CFG_DIR/edx-analytics-pipeline/output.json
+    && echo "{\"username\": \"$COMMON_MYSQL_READ_ONLY_USER\", \"host\": \"mysql\", \"password\": \"$COMMON_MYSQL_READ_ONLY_PASS\", \"port\": 3306}" > $COMMON_CFG_DIR/edx-analytics-pipeline/input.json \
```
macdiesel: these items are quite hard to read, perhaps it would be better to have these JSON strings in files co-located with the Docker file. You could then add them to the image and cat them in to the necessary JSON files?

macdiesel (commented):
One comment, looks good!

Dockerfile
rao-abdul-mannan: I was going to provide them as templates but these dockerfiles are supposed to go to configuration repo and i'm not sure if that repo authors/contributors would want them there. 
I'll keep that in mind when i'll create a pull request there and if they don't mind extra files then i'll move these json files out to additional templates.

macdiesel: Moving them out to a different file will be fine there is a precedent for this.

PR #491, opened 2018-04-03
Vertica Data Extract Using Sqoop

azafty468:
This code *IS* intended to be merged to master.

I think I've pulled out all of my testing notes from the code commits.  I've manually tested exporting to S3, confirmed the data persisted, and manually loaded the S3 file into BigQuery using the google console and the field mappings in the vertica_export.py file.  I have not used our luigi jobs to manage the BQ importing.  There are 3 issues I have uncovered and an additional issue I suspect:

Issues: 
- BINARY fields are not exported properly.  There appears to be padding added to the extracted field that BigQuery does not interpret properly.  *There are currently no binary fields in Vertica except in my test table.*
- LONG VARBINARY fields cannot be exported using Vertica JDBC client.  The client returns an error: "Error: java.io.IOException: java.sql.SQLFeatureNotSupportedException: [Vertica][JDBC](10220) Driver not capable" when these fields are encountered.  I am currently removing the LONG VARBINARY fields from extracts and creating a log message.  VARBINARY and LONG VARCHAR can be successfully exported.  *LONG VARBINARY fields are used by the experimental, public and nicke schemas*
- Timestamp fields with Timezone information are not properly exported from Vertica.  A '2018-04-02 10:55:13-04' value in Vertica was exported '2018-04-02 10:55:13.0'.  Timetz fields however are adjusted to UTC.  I'm unsure if my local system's timezone or the EMR cluster's local timezone is aggravating this issue. *TIMESTAMPTZ fields are widely used*

Suspected Issue:
- VARBINARY may suffer from the same byte padding as above.  It's not immediately clear how to test for this because the VSQL client is a little too smart for it's own good.  I first tried what I expected to be binary (00101011) which it interpreted as ASCII and converted to binary.  I next used HEX_TO_BINARY(), which allowed me to test basic field exporting but did not let me test padding.  More research is needed here to figure out how to test this (there could be a super easy function and a simple google search might be all that is needed).

There are also questions open on my other pull request.  I will respond to them there but the conversation should continue on this PR as this is the code that will be merged.

### DIFF
--- a/edx/analytics/tasks/common/sqoop.py
+++ b/edx/analytics/tasks/common/sqoop.py
@@ -92,7 +92,7 @@ class SqoopImportTask(OverwriteOutputMixin, SqoopImportMixin, luigi.contrib.hado
     )
     fields_terminated_by = luigi.Parameter(
         default=None,
-        description='Defines the file separator to use on output.',
+        description='Defines the field separator to use on output.',
     )
     delimiter_replacement = luigi.Parameter(
         default=None,
@@ -197,7 +197,7 @@ class SqoopImportTask(OverwriteOutputMixin, SqoopImportMixin, luigi.contrib.hado
 
 class SqoopImportFromMysql(SqoopImportTask):
     """
-    An abstract task that uses Sqoop to read data out of a database and writes it to a file in CSV format.
+    An abstract task that uses Sqoop to read data out of a MySQL database and writes it to a file in CSV format.
 
     By default, the output format is defined by meaning of --mysql-delimiters option, which defines defaults used by
     mysqldump tool:
@@ -283,3 +283,75 @@ class SqoopImportRunner(luigi.contrib.hadoop.JobRunner):
             except Exception:
                 log.exception("Unable to dump metadata information.")
                 pass
+
+
+class SqoopImportFromVertica(SqoopImportTask):
+    """
+    An abstract task that uses Sqoop to read data out of a Vertica database and writes it to a file in CSV format.
+
+    * fields delimited by comma
+    * lines delimited by \n
+    * fields optionally enclosed by single quotes (')
+    """
+
+    # Direct is not supported by the Vertica JDBC connector.
+    direct = None
+    # A bug in the Sqoop process with this JDBC connector makes this parameter useless.
+    num_mappers = None
+    schema_name = luigi.Parameter(
+        description='The schema that contains the table being exported.'
+    )
+    timezone_adjusted_column_list = luigi.ListParameter(
+        default=[],
+        description='The list of columns that need to have their times manually adjusted to UTC.'
+    )
+
+    def connection_url(self, cred):
+        """Construct connection URL from provided credentials."""
+        return 'jdbc:vertica://{host}/{database}?searchpath={schema}'.format(
+            host=cred['host'],
+            database=self.database,
+            schema=self.schema_name
+        )
+
+    def import_args(self):
+        if self.columns is None or len(self.columns) == 0:
+            raise RuntimeError('Error Vertica\'s connector requires specific columns listed. No columns were supplied.')
+
+        arglist = [
+            '--target-dir', self.destination,
+            '--driver', 'com.vertica.jdbc.Driver',
+        ]
+
+        # The vertica JBDC client does not handle timestamptz fields or column names that are reserved words
+        column_list = []
+        for column in self.columns:
+            if column in self.timezone_adjusted_column_list:
+                column_list.append('"{}" AT TIME ZONE \'UTC\''.format(column))
+            else:
+                column_list.append('"{}"'.format(column))
+
+        query = 'SELECT {cols} FROM {tbl} WHERE $CONDITIONS'.format(cols=','.join(column_list), tbl=self.table_name)
+        arglist.extend(['--query', query])
+
+        # There appears to be a bug in the handling of the --num-mappers and --split-by options. if --num-mappers is
+        # omitted Sqoop terminates with errors.  If --num-mappers is included with any value other than 1 then Sqoop
+        # terminates requiring the --split-by field also be defined.  If --num-mappers is included and a valid
+        # --split-by field is included then in one example Sqoop exported only 1/3 of the proper number of records.
+        # For now disable num_mappers and default the parallelism to 1.
+        arglist.extend(['--num-mappers', '1'])
+
+        if self.verbose:
+            arglist.append('--verbose')
+        if self.null_string is not None:
+            arglist.extend(['--null-string', self.null_string, '--null-non-string', self.null_string])
+        if self.fields_terminated_by is not None:
+            arglist.extend(['--fields-terminated-by', self.fields_terminated_by])
+
+        # At some point we may want to promote this sqoop option to a luigi parameter option
+        arglist.extend(['--optionally-enclosed-by', '\''])
+        if self.delimiter_replacement is not None:
+            arglist.extend(['--hive-delims-replacement', self.delimiter_replacement])
+        arglist.extend(['--lines-terminated-by', '\n'])
+
+        return arglist
--- a/edx/analytics/tasks/common/tests/test_sqoop.py
+++ b/edx/analytics/tasks/common/tests/test_sqoop.py
@@ -6,11 +6,11 @@ import unittest
 
 from mock import MagicMock, Mock, patch, sentinel
 
-from edx.analytics.tasks.common.sqoop import SqoopImportFromMysql
+from edx.analytics.tasks.common.sqoop import SqoopImportFromMysql, SqoopImportFromVertica
 from edx.analytics.tasks.util.tests.target import FakeTarget
 
 
-class SqoopImportFromMysqlTestCase(unittest.TestCase):
+class SqoopImportTestCase(unittest.TestCase):
     """
     Ensure we can pass the right arguments to Sqoop.
     """
@@ -25,9 +25,9 @@ class SqoopImportFromMysqlTestCase(unittest.TestCase):
         self.mock_run = patcher2.start()
         self.addCleanup(patcher2.stop)
 
-    def create_task(self, num_mappers=None, where=None, verbose=False,
-                    columns=None, null_string=None, fields_terminated_by=None, delimiter_replacement=None,
-                    overwrite=False, direct=True, mysql_delimiters=True):
+    def create_mysql_task(self, num_mappers=None, where=None, verbose=False,
+                          columns=None, null_string=None, fields_terminated_by=None, delimiter_replacement=None,
+                          overwrite=False, direct=True, mysql_delimiters=True):
         """Create a SqoopImportFromMysql with specified options."""
         task = SqoopImportFromMysql(
             credentials=sentinel.ignored,
@@ -45,6 +45,35 @@ class SqoopImportFromMysqlTestCase(unittest.TestCase):
             direct=direct,
             mysql_delimiters=mysql_delimiters,
         )
+        return task
+
+    def create_vertica_task(self, num_mappers=None, where=None, table_name=None, schema_name=None, columns=None,
+                            null_string=None, fields_terminated_by=None, delimiter_replacement=None, overwrite=False):
+        """A generic task generator for Sqoop on Vertica"""
+        if columns is None:
+            column_list = []
+        elif isinstance(columns, list):
+            column_list = columns
+        else:
+            column_list = list(columns)
+
+        kw_args = {
+            "credentials": sentinel.ignored,
+            "schema_name": schema_name,
+            "database": 'exampledata',
+            "destination": "/fake/destination",
+            "table_name": table_name,
+            "num_mappers": num_mappers,
+            "where": where,
+            "columns": column_list,
+            "null_string": null_string,
+            "fields_terminated_by": fields_terminated_by,
+            "delimiter_replacement": delimiter_replacement,
+            "overwrite": overwrite,
+        }
+        # remove options marked as None
+        trimmed_kws = {k: v for k, v in kw_args.iteritems() if v is not None}
+        task = SqoopImportFromVertica(**trimmed_kws)
         return task
 
     def run_task(self, task, credentials=None):
@@ -67,11 +96,11 @@ class SqoopImportFromMysqlTestCase(unittest.TestCase):
 
         task.run()
 
-    def create_and_run_task(self, credentials=None, num_mappers=None, where=None, verbose=False,
-                            columns=None, null_string=None, fields_terminated_by=None, delimiter_replacement=None,
-                            overwrite=False, direct=True, mysql_delimiters=True):
+    def create_and_run_mysql_task(self, credentials=None, num_mappers=None, where=None, verbose=False,
+                                  columns=None, null_string=None, fields_terminated_by=None, delimiter_replacement=None,
+                                  overwrite=False, direct=True, mysql_delimiters=True):
         """Create a SqoopImportFromMysql task with specified options, and then run it."""
-        task = self.create_task(
+        task = self.create_mysql_task(
             num_mappers=num_mappers,
             where=where,
             verbose=verbose,
@@ -85,24 +114,41 @@ class SqoopImportFromMysqlTestCase(unittest.TestCase):
         )
         self.run_task(task, credentials)
 
+    def create_and_run_vertica_task(self, credentials=None, num_mappers=None, where=None, table_name=None,
+                                    schema_name=None, columns=None, null_string=None, fields_terminated_by=None,
+                                    delimiter_replacement=None, overwrite=False):
+        """Create a SqoopImportFromVertica task with specified options, and then run it."""
+        task = self.create_vertica_task(
+            num_mappers=num_mappers,
+            where=where,
+            columns=columns,
+            null_string=null_string,
+            table_name=table_name,
+            schema_name=schema_name,
+            fields_terminated_by=fields_terminated_by,
+            delimiter_replacement=delimiter_replacement,
+            overwrite=overwrite,
+        )
+        self.run_task(task, credentials)
+
     def get_call_args_after_run(self):
         """Returns args for first call to Hadoop."""
         return self.mock_run.call_args[0][0]
 
     def test_connect_with_missing_credentials(self):
         with self.assertRaises(KeyError):
-            self.create_and_run_task('{}')
+            self.create_and_run_mysql_task('{}')
         self.assertTrue(self.mock_sqoop_password_target().remove.called)
         self.assertFalse(self.mock_run.called)
 
     def test_connect_with_credential_syntax_error(self):
         with self.assertRaises(ValueError):
-            self.create_and_run_task('{')
+            self.create_and_run_mysql_task('{')
         self.assertTrue(self.mock_sqoop_password_target().remove.called)
         self.assertFalse(self.mock_run.called)
 
     def test_connect_with_complete_credentials(self):
-        self.create_and_run_task()
+        self.create_and_run_mysql_task()
         arglist = self.get_call_args_after_run()
         self.assertTrue(self.mock_run.called)
         expected_arglist = [
@@ -125,30 +171,30 @@ class SqoopImportFromMysqlTestCase(unittest.TestCase):
         self.assertTrue(self.mock_sqoop_password_target().remove.called)
 
     def test_verbose_arguments(self):
-        self.create_and_run_task(verbose=True)
+        self.create_and_run_mysql_task(verbose=True)
         arglist = self.get_call_args_after_run()
         self.assertIn('--verbose', arglist)
 
     def test_connect_with_where_args(self):
-        self.create_and_run_task(where='id < 50')
+        self.create_and_run_mysql_task(where='id < 50')
         arglist = self.get_call_args_after_run()
         self.assertEquals(arglist[-4], '--where')
         self.assertEquals(arglist[-3], 'id < 50')
 
     def test_connect_with_num_mappers(self):
-        self.create_and_run_task(num_mappers=50)
+        self.create_and_run_mysql_task(num_mappers=50)
         arglist = self.get_call_args_after_run()
         self.assertEquals(arglist[-4], '--num-mappers')
         self.assertEquals(arglist[-3], '50')
 
     def test_connect_with_columns(self):
-        self.create_and_run_task(columns=['column1', 'column2'])
+        self.create_and_run_mysql_task(columns=['column1', 'column2'])
         arglist = self.get_call_args_after_run()
         self.assertEquals(arglist[-4], '--columns')
         self.assertEquals(arglist[-3], 'column1,column2')
 
     def test_connect_with_null_string(self):
-        self.create_and_run_task(null_string='\\\\N')
+        self.create_and_run_mysql_task(null_string='\\\\N')
         arglist = self.get_call_args_after_run()
         self.assertEquals(arglist[-6], '--null-string')
         self.assertEquals(arglist[-5], '\\\\N')
@@ -156,29 +202,29 @@ class SqoopImportFromMysqlTestCase(unittest.TestCase):
         self.assertEquals(arglist[-3], '\\\\N')
 
     def test_connect_with_fields_terminations(self):
-        self.create_and_run_task(fields_terminated_by='\x01')
+        self.create_and_run_mysql_task(fields_terminated_by='\x01')
         arglist = self.get_call_args_after_run()
         self.assertEquals(arglist[-4], '--fields-terminated-by')
         self.assertEquals(arglist[-3], '\x01')
 
     def test_connect_with_delimiter_replacement(self):
-        self.create_and_run_task(delimiter_replacement=' ')
+        self.create_and_run_mysql_task(delimiter_replacement=' ')
         arglist = self.get_call_args_after_run()
         self.assertEquals(arglist[-4], '--hive-delims-replacement')
         self.assertEquals(arglist[-3], ' ')
 
     def test_connect_without_mysql_delimiters(self):
-        self.create_and_run_task(mysql_delimiters=False)
+        self.create_and_run_mysql_task(mysql_delimiters=False)
         arglist = self.get_call_args_after_run()
         self.assertNotIn('--mysql-delimiters', arglist)
 
     def test_connect_without_direct(self):
-        self.create_and_run_task(direct=False)
+        self.create_and_run_mysql_task(direct=False)
         arglist = self.get_call_args_after_run()
         self.assertNotIn('--direct', arglist)
 
     def test_metadata(self):
-        task = self.create_task()
+        task = self.create_mysql_task()
         self.run_task(task)
         self.assertFalse(task.complete())
         metadata_target = task.metadata_output()
@@ -189,7 +235,7 @@ class SqoopImportFromMysqlTestCase(unittest.TestCase):
 
     def test_overwrite(self):
         kwargs = {'overwrite': True}
-        task = self.create_task(**kwargs)
+        task = self.create_mysql_task(**kwargs)
         output_target = Mock()
         task.output = Mock(return_value=output_target)
         output_target.exists = Mock(return_value=True)
@@ -203,3 +249,44 @@ class SqoopImportFromMysqlTestCase(unittest.TestCase):
         self.assertTrue(task.output().exists.called)
         self.assertTrue(task.output().remove.called)
         self.assertTrue(task.attempted_removal)
+
+    def test_success_vertica_with_complete_credentials(self):
+        self.create_and_run_vertica_task(table_name='example_table',
+                                         schema_name='fake_schema',
+                                         delimiter_replacement=' ',
+                                         fields_terminated_by=',',
+                                         columns=['field1', 'field2', 'field3'])
+        arglist = self.get_call_args_after_run()
+        self.assertTrue(self.mock_run.called)
+        expected_arglist = [
+            'sqoop', 'import',
+            '--connect', 'jdbc:vertica://db.example.com/exampledata?searchpath=fake_schema',
+            '--username', u'exampleuser',
+            '--password-file', '/temp/password_file',
+            '--target-dir', '/fake/destination',
+            '--driver', 'com.vertica.jdbc.Driver',
+            '--query', 'SELECT "field1","field2","field3" FROM example_table WHERE $CONDITIONS',
+            '--num-mappers', '1',
+            '--fields-terminated-by', ',',
+            '--optionally-enclosed-by', '\'',
+            '--hive-delims-replacement', ' ',
+            '--lines-terminated-by', '\n',
+        ]
+        self.assertEquals(arglist, expected_arglist)
+        self.assertTrue(self.mock_sqoop_password_target().remove.called)
+
+    def test_success_vertica_with_custom_delimiters(self):
+        self.create_and_run_vertica_task(table_name='example_table', schema_name='fake_schema',
+                                         columns=['field1', 'field2', 'field3'], fields_terminated_by='\a',
+                                         delimiter_replacement='a')
+        arglist = self.get_call_args_after_run()
+        self.assertTrue(self.mock_run.called)
+        self.assertEquals(arglist[16], '--fields-terminated-by')
+        self.assertEquals(arglist[17], '\a')
+        self.assertEquals(arglist[20], '--hive-delims-replacement')
+        self.assertEquals(arglist[21], 'a')
+
+    def test_vertica_failure_from_missing_columnlist(self):
+        with self.assertRaises(RuntimeError) as context:
+            self.create_and_run_vertica_task(table_name='example_table', schema_name='fake_schema')
+        self.assertTrue('Error Vertica\'s connector requires specific columns listed' in str(context.exception))
--- a/edx/analytics/tasks/common/vertica_export.py
+++ b/edx/analytics/tasks/common/vertica_export.py
@@ -1,0 +1,396 @@
+"""
+Supports exporting data from Vertica.
+"""
+import datetime
+import json
+import logging
+import re
+
+import luigi
+from google.cloud import bigquery
+
+from edx.analytics.tasks.common.bigquery_load import BigQueryLoadTask
+from edx.analytics.tasks.common.sqoop import SqoopImportFromVertica
+from edx.analytics.tasks.util.decorators import workflow_entry_point
+from edx.analytics.tasks.util.hive import HivePartition
+from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
+from edx.analytics.tasks.util.url import ExternalURL, get_target_from_url, url_path_join
+
+log = logging.getLogger(__name__)
+
+try:
+    import vertica_python
+    from vertica_python.errors import QueryError
+    vertica_client_available = True  # pylint: disable=invalid-name
+except ImportError:
+    log.warn('Unable to import Vertica client libraries')
+    # On hadoop slave nodes we don't have Vertica client libraries installed so it is pointless to ship this package to
+    # them, instead just fail noisily if we attempt to use these libraries.
+    vertica_client_available = False  # pylint: disable=invalid-name
+
+
+def get_vertica_results(credentials, query):
+    """Run a single query in Vertica and return the results."""
+    credentials_target = ExternalURL(url=credentials).output()
+    cred = None
+    with credentials_target.open('r') as credentials_file:
+        cred = json.load(credentials_file)
+
+    # Externalize autocommit and read timeout
+    connection = vertica_python.connect(user=cred.get('username'), password=cred.get('password'), host=cred.get('host'),
+                                        port=cred.get('port'), database='warehouse', autocommit=False,
+                                        read_timeout=None)
+
+    if not vertica_client_available:
+        raise ImportError('Vertica client library not available')
+
+    try:
+        cursor = connection.cursor()
+        cursor.execute(query)
+        results = cursor.fetchall()
+    finally:
+        connection.close()
+
+    return results
+
+
+VERTICA_TO_BIGQUERY_FIELD_MAPPING = {
+    'boolean': 'bool',
+    'integer': 'int64',
+    'int': 'int64',
+    'bigint': 'int64',
+    'varbinary': 'bytes',
+    'long varbinary': 'bytes',
+    'binary': 'bytes',
+    'char': 'string',
+    'varchar': 'string',
+    'long varchar': 'string',
+    'money': 'float64',
+    'numeric': 'float64',
+    'float': 'float64',
+    'date': 'date',
+    'time': 'time',
+    # Due to an error with the Vertica JDBC client time zone information is stripped prior to extracting.  All
+    # timestamptz and timetz values are coerced to UTC (manually in the case of timestamptz).
+    'timetz': 'time',
+    'timestamptz': 'datetime',
+    'timestamp': 'datetime'
+}
+
+
+def get_vertica_table_schema(credentials, schema_name, table_name):
+    """
+    Returns the Vertica schema in the format (column name, column type, nullable?) for the indicated table.
+    """
+
+    query = "select column_name, data_type, is_nullable from columns where table_schema='{schema}' and " \
+            "table_name='{table}' order by ordinal_position;".format(schema=schema_name, table=table_name)
+    results = []
+    rows = get_vertica_results(credentials, query)
+
+    for row in rows:
+        column_name = row[0]
+        nullable = row[2]
+        field_type = row[1]
+        field_type = field_type.rsplit('(')[0]
+
+        if field_type.lower() in ['long varbinary']:
+            log.error('Error Vertica jdbc tool is unable to export field type \'%s\'.  This field will be '
+                      'excluded from the extract.', field_type.lower())
+        else:
+            results.append((column_name, field_type.lower(), nullable))
+
+    return results
+
+
+class VerticaTableToS3Task(OverwriteOutputMixin, luigi.Task):
+    """
+    Export a table from Vertica to S3 using sqoop.
+
+    An exporter that reads a table from Vertica and persists the data to S3.  In order to use
+    LoadVerticaToS3TableTask the caller must know the Vertica schema and table name. The columns are automatically
+    discovered at run time.
+    """
+    date = luigi.DateParameter(
+        description='A URL location of the data warehouse.'
+    )
+    intermediate_warehouse_path = luigi.Parameter(
+        description='A dictionary specifying the S3-centric configuration.'
+    )
+    sqoop_null_string = luigi.Parameter(
+        default='null',
+        description='A string replacement value for any (null) values encountered by Sqoop when exporting from Vertica.'
+    )
+    sqoop_fields_terminated_by = luigi.Parameter(
+        default=',',
+        description='The field delimiter used by Sqoop.'
+    )
+    sqoop_delimiter_replacement = luigi.Parameter(
+        default=' ',
+        description='The string replacement value for special characters encountered by Sqoop when exporting from '
+                    'Vertica.'
+    )
+    column_list = luigi.ListParameter(
+        description='The column names being extracted from this table.'
+    )
+    warehouse_name = luigi.Parameter(
+        description='The Vertica warehouse that houses the schema being copied.'
+    )
+    schema_name = luigi.Parameter(
+        description='The Vertica schema being copied. '
+    )
+    credentials = luigi.Parameter(
+        description='Path to the external Vertica access credentials file.',
+    )
+    table_name = luigi.Parameter(
+        description='The Vertica table being copied.'
+    )
+    timestamptz_column_list = luigi.ListParameter(
+        default=[],
+        description='The list of columns being copied that are of type timestamptz.  This is necessary because of a '
+                    'bug in the time zone processing of the Vertica JDBC client.  Note timetz fields are automatically'
+                    'converted to UTC.'
+    )
+
+    def __init__(self, *args, **kwargs):
+        super(VerticaTableToS3Task, self).__init__(*args, **kwargs)
+        self.required_tasks = None
+        self.table_schema = None
+        self.vertica_source_table_schema = None
+
+    def requires(self):
+        if self.required_tasks is None:
+            self.required_tasks = {
+                'credentials': ExternalURL(url=self.credentials),
+                'insert_source': self.insert_source_task,
+            }
+        return self.required_tasks
+
+    def complete(self):
+        return self.insert_source_task.complete()
+
+    def s3_output_path(self):
+        partition_path_spec = HivePartition('dt', self.date).path_spec
+        target_url = url_path_join(self.intermediate_warehouse_path,
+                                   self.warehouse_name,
+                                   self.schema_name,
+                                   self.table_name,
+                                   partition_path_spec) + '/'
+        return target_url
+
+    def output(self):
+        return get_target_from_url(self.s3_output_path())
+
+    @property
+    def insert_source_task(self):
+        """The sqoop command that manages the data transfer from the source datasource."""
+        if len(self.column_list) <= 0:
+            raise RuntimeError('Error Sqoop copy of {schema}.{table} found no viable columns!'.
+                               format(schema=self.schema_name,
+                                      table=self.table))
+
+        target_url = self.s3_output_path()
+
+        return SqoopImportFromVertica(
+            schema_name=self.schema_name,
+            table_name=self.table_name,
+            credentials=self.credentials,
+            database=self.warehouse_name,
+            columns=self.column_list,
+            destination=target_url,
+            overwrite=self.overwrite,
+            null_string=self.sqoop_null_string,
+            fields_terminated_by=self.sqoop_fields_terminated_by,
+            delimiter_replacement=self.sqoop_delimiter_replacement,
+            timezone_adjusted_column_list=self.timestamptz_column_list,
+        )
+
+
+class LoadVerticaTableToBigQuery(BigQueryLoadTask):
+    """Copies one table from Vertica through S3/GCP into BigQuery."""
+
+    intermediate_warehouse_path = luigi.Parameter(
+        description='The intermediate warehouse path used on S3 and GCP'
+    )
+    table_name = luigi.Parameter(
+        description='The Vertica table being copied.'
+    )
+    vertica_warehouse_name = luigi.Parameter(
+        default='warehouse', description='The Vertica warehouse that houses the schema being copied.'
+    )
+    vertica_schema_name = luigi.Parameter(
+        description='The Vertica schema being copied. '
+    )
+    vertica_credentials = luigi.Parameter(
+        description='Path to the external Vertica access credentials file.',
+    )
+    exclude = luigi.ListParameter(
+        description='The Vertica tables that are to be excluded from exporting.'
+    )
+
+    def __init__(self, *args, **kwargs):
+        super(LoadVerticaTableToBigQuery, self).__init__(*args, **kwargs)
+        self.vertica_source_table_schema = None
+        self.bigquery_compliant_schema = None
+        self.timestamptz_column_list = None
+
+    @property
+    def vertica_table_schema(self):
+        """The schema of the Vertica table."""
+        if self.vertica_source_table_schema is None:
+            self.vertica_source_table_schema = get_vertica_table_schema(self.vertica_credentials,
+                                                                        self.vertica_schema_name,
+                                                                        self.table_name)
+        return self.vertica_source_table_schema
+
+    @property
+    def table(self):
+        return self.table_name
+
+    @property
+    def table_description(self):
+        return "Copy of Vertica table {}.{} on {}".format(self.vertica_schema_name,
+                                                          self.table_name,
+                                                          self.date)
+
+    @property
+    def table_friendly_name(self):
+        return '{} from {}'.format(self.table_name, self.vertica_schema_name)
+
+    @property
+    def field_delimiter(self):
+        return '\x01'
+
+    @property
+    def null_marker(self):
+        return 'NNULLL'
+
+    @property
+    def delimiter_replacement(self):
+        return ' '
+
+    @property
+    def schema(self):
+        """The BigQuery compliant schema."""
+        if self.bigquery_compliant_schema is None:
+            res = []
+            timestamp_list = []
+            for field_name, vertica_field_type, nullable in self.vertica_table_schema:
+                # column_name, data_type, is_nullable
+                if vertica_field_type in VERTICA_TO_BIGQUERY_FIELD_MAPPING:
+                    res.append(bigquery.SchemaField(field_name,
+                                                    VERTICA_TO_BIGQUERY_FIELD_MAPPING[vertica_field_type],
+                                                    mode='NULLABLE' if nullable else 'REQUIRED'))
+                    if vertica_field_type in ['timestamptz']:
+                        timestamp_list.append(field_name)
+                else:
+                    raise RuntimeError('Error for field {field}: Vertica type {type} does not have a mapping to '
+                                       'BigQuery'.format(field=field_name, type=vertica_field_type))
+            self.bigquery_compliant_schema = res
+            self.timestamptz_column_list = timestamp_list
+        return self.bigquery_compliant_schema
+
+    @property
+    def insert_source_task(self):
+        return VerticaTableToS3Task(
+            intermediate_warehouse_path=self.intermediate_warehouse_path,
+            date=self.date,
+            overwrite=self.overwrite,
+            sqoop_null_string=self.null_marker,
+            sqoop_fields_terminated_by=self.field_delimiter,
+            sqoop_delimiter_replacement=self.delimiter_replacement,
+            column_list=[field.name for field in self.schema],
+            warehouse_name=self.vertica_warehouse_name,
+            schema_name=self.vertica_schema_name,
+            credentials=self.vertica_credentials,
+            table_name=self.table_name,
+            timestamptz_column_list=self.timestamptz_column_list
+        )
+
+
+@workflow_entry_point
+class VerticaSchemaToBigQueryTask(luigi.WrapperTask):
+    """
+    A task that copies all the tables in a Vertica schema to S3.
+
+    Reads all tables in a schema and, if they are not listed in the `exclude` parameter, schedules a
+    LoadVerticaToS3TableTask task for each table.
+    """
+    overwrite = luigi.BoolParameter(
+        default=False,
+        significant=False,
+        description='Indicates if the target data sources should be removed prior to generating.'
+    )
+    date = luigi.DateParameter(
+        default=datetime.datetime.utcnow().date(),
+        description='Current run date.  This parameter is used to isolate intermediate datasets.'
+    )
+    exclude = luigi.ListParameter(
+        default=[],
+        description='The Vertica tables that are to be excluded from exporting.'
+    )
+    vertica_warehouse_name = luigi.Parameter(
+        default='warehouse',
+        description='The Vertica warehouse that houses the schema being copied.'
+    )
+    vertica_schema_name = luigi.Parameter(
+        description='The Vertica schema being copied. '
+    )
+    vertica_credentials = luigi.Parameter(
+        config_path={'section': 'vertica-export', 'name': 'credentials'},
+        description='Path to the external Vertica access credentials file.',
+    )
+    bigquery_dataset = luigi.Parameter(
+        default=None,
+        description='The target dataset that will hold the Vertica schema.'
+    )
+    max_bad_records = luigi.IntParameter(
+        default=0,
+        description="Number of bad records ignored by BigQuery before failing a load job."
+    )
+    gcp_credentials = luigi.Parameter(
+        description='Path to the external GCP/BigQuery access credentials file.',
+    )
+    s3_warehouse_path = luigi.Parameter(
+        config_path={'section': 'hive', 'name': 'warehouse_path'},
+        description='The warehouse path to store intermediate data on S3.'
+    )
+
+    def __init__(self, *args, **kwargs):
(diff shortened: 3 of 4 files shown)


edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,229 @@
...
+    database = luigi.Parameter(
+        default='warehouse',
+        description='The Vertica database that is the source of the sqoop command.'
+    )
+    schema = luigi.Parameter(
+        default=None,
+        description='The Vertica schema that is going to be exported.'
+    )
+    warehouse_path = luigi.Parameter(
+        config_path={'section': 'vertica-export-sqoop', 'name': 'warehouse_path'},
```
azafty468: Brian W's comment from my spike PR: Wondering if we should just use the same warehouse_path as always, but write to a "vertica" subdirectory or something like that. Similar to what we do for database exports.

I don't mind doing that at all.  But I guess does that path belong in this mixin?  Is there going to be one export job per schema?  Is there going to be one central export job that manages all schemas and that class should set the shared S3 path?

edx/analytics/tasks/common/sqoop.py
```diff
@@ -283,3 +283,72 @@ def run_job(self, job):
...
+
+
+class SqoopImportFromVertica(SqoopImportTask):
+    """
+    An abstract task that uses Sqoop to read data out of a Vertica database and writes it to a file in CSV format.
+
+    * fields delimited by comma
+    * lines delimited by \n
+    * delimiters escaped by backslash
+    * fields optionally enclosed by single quotes (')
```
brianhw: Where are you hoping to go with these CSV files?   If you are hoping to load them into BigQuery, then you need to use different parameters.   BigQuery does not recognized delimiters as being escaped by a backslash, for example.    
 So using a comma for field delimiter is a non-starter.   I used  ^A, and I still ran into some problems, though far fewer.  You also don't mention here what to use for NULLs.   There is an issue with fields that contain the string "NULL" (or "Null" or "null") rather than really being NULL, and these strings may be in non-Nullable columns.  Perhaps you won't run into that.

edx/analytics/tasks/common/sqoop.py
```diff
@@ -283,3 +283,72 @@ def run_job(self, job):
...
+    def connection_url(self, cred):
+        """Construct connection URL from provided credentials."""
+        return 'jdbc:vertica://{host}/{database}?searchpath={schema}'.format(host=cred['host'],
+                                                                             database=self.database,
+                                                                             schema=self.schema_name)
+
+    def import_args(self):
+        if self.columns is None or len(self.columns) == 0:
+            raise RuntimeError('Error Vertica\'s connector requires specific columns listed or a wild card be present. '
+                               ' No columns were supplied.')
```
brianhw: What kind of wildcard?   Something like "*"?   If that works, then perhaps that could be made the default.

edx/analytics/tasks/common/sqoop.py
```diff
@@ -283,3 +283,72 @@ def run_job(self, job):
...
+            arglist.extend(['--num-mappers', '1'])
+        if self.verbose:
+            arglist.append('--verbose')
+        if self.null_string is not None:
+            arglist.extend(['--null-string', self.null_string, '--null-non-string', self.null_string])
+        if self.fields_terminated_by is not None:
+            arglist.extend(['--fields-terminated-by', self.fields_terminated_by])
+        arglist.extend(['--lines-terminated-by', '\n'])
+        if self.delimiter_replacement is not None:
+            arglist.extend(['--escaped-by', self.delimiter_replacement])
```
brianhw: --escaped-by is not the same as --hive-delims-replacement, which is what the description for self.delimiter_replacement still says above.  Escaped-by implies that the delimiter is escaped with a character added in front of it.   Hive doesn't support such escapes, so it requires the delimiter characters to be actually replaced with something else.  BigQuery only allows double-quotes to escape other double-quotes, but this escaped-by would be about delimiters, not quote characters specifically.

edx/analytics/tasks/common/tests/test_sqoop.py
```diff
@@ -203,3 +248,41 @@ def test_overwrite(self):
...
+        self.assertTrue(self.mock_sqoop_password_target().remove.called)
+
+    def test_success_vertica_with_custom_delimiters(self):
+        self.create_and_run_vertica_task(table_name='example_table', schema_name='fake_schema',
+                                         columns=['field1', 'field2', 'field3'], fields_terminated_by='\a',
+                                         delimiter_replacement='a')
+        arglist = self.get_call_args_after_run()
+        self.assertTrue(self.mock_run.called)
+        self.assertEquals(arglist[19], '\a')
+        self.assertEquals(arglist[23], 'a')
```
brianhw: It might be good here to also assert arglist[18] and arglist[22], for readability....

edx/analytics/tasks/common/vertica_export.py
brianhw: I know we already have a separate warehouse_path that is specified in [database-export] as well as in [hive].  It's not clear when these would have different defaults.  We run them with the same.  And yet, in database exports, we take this different value that is the same, and tack on a subdirectory.  

It's important to keep these various exports separate, to the extent that they have different delimiters and such.  It would be good if we can be more systematic with our defaults, I think, while providing useful overrides for testing and alternative deployments and such.  

I just worry about having the same parameter specified in multiple places, and the opportunity that raised for mistakes and confusion.

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,229 @@
...
+    'long varchar': 'string',
+    'money': 'float64',
+    'numeric': 'float64',
+    'float': 'float64',
+    'date': 'date',
+    'time': 'string',
+    'timetz': 'string',
+    'timestamptz': 'datetime',
+    'timestamp': 'datetime'
+}
```
brianhw: Interesting.   I have a similar MYSQL_TO_BIGQUERY mapping.

Is there a reason you use datetime instead of timestamp for timestamptz?   I'm wondering if there's a setting or something on vertica which would make it output in UTC, and then hopefully timestamp would be the right thing (assuming it would default to UTC).   

But then again, you're not actually writing to BigQuery yet.

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,229 @@
...
+        if self.table_schema is None:
+            self.table_schema = get_vertica_table_schema(self.credentials, self.schema, self.table)
+        return self.table_schema
+
+    @property
+    def insert_source_task(self):
+        """The sqoop command that manages the data transfer from the source datasource."""
+        partition_path_spec = HivePartition('dt', self.run_date.isoformat()).path_spec
+
+        target_url = url_path_join(self.warehouse_path, self.database, self.schema, self.table, partition_path_spec)
```
brianhw: Yes, I wonder if this should be called warehouse_path here, or something else.

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,229 @@
...
+    def requires(self):
+        if self.required_tasks is None:
+            self.required_tasks = {
+                'credentials': ExternalURL(url=self.credentials),
+                'insert_source': self.insert_source_task,
+            }
+        return self.required_tasks
+
+    def complete(self):
+        return self.insert_source_task.complete()
```
brianhw, line 170: I see -- so at the moment, this task doesn't actually do any real work, whether that's loading into BigQuery, or into Hive or into something else.  It's currently a wrapper.  Do you think you would then extend it when defining these other things?

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,229 @@
...
+
+def translate_vertica_schema_to_bigquery_schema(vertica_schema):
+    """Converts a Vertica schema to a BigQuery compliant schema. """
+    result = []
+    for column in vertica_schema:
+        if column[1] in VERTICA_TO_BIGQUERY_FIELD_MAPPING:
+            result.append(column[0], VERTICA_TO_BIGQUERY_FIELD_MAPPING[column[1]], column[2])
+        else:
+            raise RuntimeError('Error for field {field}: Vertica type {type} does not have a mapping to BigQuery'.
+                               format(field=column[0], type=column[1]))
```
brianhw: Eventually you will presumably need to use SchemaField objects, but this is a start.

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,229 @@
...
+        super(VerticaSchemaToS3Task, self).__init__(*args, **kwargs)
+
+    def requires(self):
+        yield ExternalURL(url=self.credentials)
+
+        query = "select table_name from all_tables where schema_name='{schema_name}'".format(schema_name=self.schema)
+        table_list = [row[0] for row in get_vertica_results(self.credentials, query)]
+
+        for table_name in table_list:
+            if table_name not in self.exclude:
```
brianhw: This is a different definition of exclude than is used in our other imports.  There we treat exclusions as regular expressions.  This requires enumeration and exact match.

brianhw (commented):
Good progress!   I have a few comments, mostly about consistency and about how it may pair with BigQuery loading in future.

edx/analytics/tasks/common/sqoop.py
azafty468: I've adopted the ^A and NNULL options you've used in your mysql extracts.

edx/analytics/tasks/common/sqoop.py
azafty468: I've updated the comment to remove any reference to a wildcard.

edx/analytics/tasks/common/vertica_export.py
azafty468: No reason except that I typically use timestamp when a table is defaulting a value (for instance current timestamp) and datetime otherwise.  This may be the source of my timestamptz offset issue.  I'm going to run a small test with the test_export table I have in Vertica and see if this resolves the issue.

edx/analytics/tasks/common/vertica_export.py
azafty468: Correct, I'd like to use this as a wrapper class because it represents a logical step and it makes it easier organize the configuration variables.  The Sqoop job handles all the actual data movement.  I don't really see much need to ever extend this class unless someone wants to change the Sqoop parameters or S3 folder path.  If we ever find a need to persist the Vertica schema as a physical object on S3 however, this class may serve some purpose there.

edx/analytics/tasks/common/sqoop.py
```diff
@@ -283,3 +283,81 @@ def run_job(self, job):
...
+        arglist = [
+            '--target-dir', self.destination,
+            '--driver', 'com.vertica.jdbc.Driver',
+        ]
+
+        if len(self.columns) == 0:
+            raise RuntimeError('Error cannot complete export of {schema}.{table} because no columns were specified'
+                               ''.format(schema=self.schema_name, table=self.table_name))
+
+        # The vertica JBDC client doesn't handle reserved words in characters correctly.  We must build the query
```
brianhw: I don't understand what "reserved words in characters" means. Do you mean "reserved words in column names"?   (So then "We must use the query argument instead of the columns argument." ? )

edx/analytics/tasks/common/sqoop.py
```diff
@@ -283,3 +283,81 @@ def run_job(self, job):
...
+        else:
+            arglist.extend(['--num-mappers', '1'])
+
+        if self.verbose:
+            arglist.append('--verbose')
+        if self.null_string is not None:
+            arglist.extend(['--null-string', self.null_string, '--null-non-string', self.null_string])
+        if self.fields_terminated_by is not None:
+            arglist.extend(['--fields-terminated-by', self.fields_terminated_by])
+        arglist.extend(['--optionally-enclosed-by', '\''])
```
brianhw: Are you sure that you want to hard-code this?   As far as I know, BigQuery does not recognize these as quotes when it reads in CSVs.  It only understands that delimiters should be treated as characters when they are escaped by a duplicate of themselves.   So if this were to actually be invoked, one should then see extra single-quote characters appearing at the beginning and end of those fields, if they got loaded into BigQuery at all. 

Actually, I suspect this is a side effect of defining delimiter_replacement as a space.   That is, once that replacement is defined, then there will never be an optional enclosure because there will never be anything requiring it anymore.   Worse than that, there is no way to disable it.   So I suggest leaving this here, but that we change the default of delimiter_replacement, and move the setting of it to be a space to the code doing the vertica table copies to bigquery.

Actually, at some point going forward, this should really be an option, so that one can disable the option if it's not desired.  And maybe put a comment here to that effect.  But I'm fine to hard-code it here for now.

edx/analytics/tasks/common/sqoop.py
```diff
@@ -283,3 +283,81 @@ def run_job(self, job):
...
+        default=' ',
+        description='Defines a character to use as replacement for delimiters '
+        'that appear within data values, for use with Hive.'
+    )
+
+    def connection_url(self, cred):
+        """Construct connection URL from provided credentials."""
+        return 'jdbc:vertica://{host}/{database}?searchpath={schema}'.format(host=cred['host'],
+                                                                             database=self.database,
+                                                                             schema=self.schema_name)
```
brianhw: nit:  suggest formatting this with "format(" here, then host, database, schema together on the next line, followed by the matching close paren on the second line.

edx/analytics/tasks/common/sqoop.py
```diff
@@ -283,3 +283,81 @@ def run_job(self, job):
...
+    columns = luigi.ListParameter(
+        description='A list of column names to be included.',
+    )
+    schema_name = luigi.Parameter(
+        description='The schema that contains the table being exported.'
+    )
+    null_string = luigi.Parameter(
+        default=None,
+        description='String to use to represent NULL values in output data.',
+    )
```
brianhw: This definition of null_string does not seem different from that used in the base class.   Should this be removed?

edx/analytics/tasks/common/sqoop.py
```diff
@@ -283,3 +283,81 @@ def run_job(self, job):
...
+    fields_terminated_by = luigi.Parameter(
+        default=',',
+        description='Defines the file separator to use in output data.',
+    )
```
brianhw: This is redefining an existing parameter, and it's helpful to note when that's being done and why.  (I.e. different defaults, because 'direct' is no longer supported, let alone the default...).  I am curious, however, what the behavior of sqoop with Vertica is if this were not specified.  Would it not create any fields at all?   Or complain because a required parameter was missing?  (An alternative approach is to check for None in import_args, and substitute a comma at that point, but this is clearer.)

edx/analytics/tasks/common/sqoop.py
```diff
@@ -283,3 +283,81 @@ def run_job(self, job):
...
+    delimiter_replacement = luigi.Parameter(
+        default=' ',
+        description='Defines a character to use as replacement for delimiters '
+        'that appear within data values, for use with Hive.'
+    )
```
brianhw: As I mention further down, I think this should be removed from here, so that the default returns to being None.   Then the behavior will really be "delimiters escaped by backslash, and fields optionally enclosed by single quotes (')", as documented above.   My reason is first that if a space is the default, then one never gets escaped delimiters nor fields being optionally enclosed, which leaves the documentation somewhat misleading.  But secondly, I don't think there is a mechanism to restore that behavior if it were desired -- there is no way to set a parameter to None on the command line.    So None should be the default, and this parameter should be overridden to be a space when it's called for Vertica-to-BigQuery purposes.

edx/analytics/tasks/common/sqoop.py
```diff
@@ -283,3 +283,81 @@ def run_job(self, job):
...
+    schema_name = luigi.Parameter(
+        description='The schema that contains the table being exported.'
+    )
+    null_string = luigi.Parameter(
+        default=None,
+        description='String to use to represent NULL values in output data.',
+    )
+    fields_terminated_by = luigi.Parameter(
+        default=',',
+        description='Defines the file separator to use in output data.',
```
brianhw: nit:  seems like "file separator" should be "field separator", here and in the code from which it was cut....

edx/analytics/tasks/common/sqoop.py
```diff
@@ -283,3 +283,81 @@ def run_job(self, job):
...
+                               ''.format(schema=self.schema_name, table=self.table_name))
+
+        # The vertica JBDC client doesn't handle reserved words in characters correctly.  We must build the query
+        column_list = ['"{}"'.format(b) for b in self.columns]
+        query = 'SELECT {cols} FROM {tbl} WHERE $CONDITIONS'.format(cols=','.join(column_list), tbl=self.table_name)
+
+        arglist.extend(['--query', query])
+
+        # Splitting is not successfully done by this version of sqoop or I was using it wrong.  For now do not add
+        # a --split-by field.
```
brianhw: Just so I understand, this comment is not actually about the num_mappers setting below, is it?   If it's separate, it would be good to put a blank line after it.  If it is related to the num_mappers setting, then I might want to understand more.  In the base code, we do not set an explicit default value for num_mappers to 1, but this does by default.   Does that mean that you only expect this to work for num_mappers = 1, because it will not try to do a split?  And unlike the mysql implementation, the Vertica stuff cannot figure out what to use for splitting automatically, so it would require a --split-by field?   (In which case, I would assume it's an issue with the driver in question, not with the version of sqoop : ).

edx/analytics/tasks/common/tests/test_sqoop.py
```diff
@@ -47,6 +47,34 @@ def create_task(self, num_mappers=None, where=None, verbose=False,
...
+                            null_string=None, fields_terminated_by=None, delimiter_replacement=None, overwrite=False):
+        """A generic task generator for Sqoop on Vertica"""
+        if columns is None:
+            column_list = []
+        elif isinstance(columns, list):
+            column_list = columns
+        else:
+            column_list = list(columns)
+
+        kw_args = {"credentials": sentinel.ignored,
```
brianhw: nit:  I would move "credentials" to the next line, and adjust the indentation.

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,402 @@
...
+    'long varbinary': 'bytes',
+    'binary': 'bytes',
+    'char': 'string',
+    'varchar': 'string',
+    'long varchar': 'string',
+    'money': 'float64',
+    'numeric': 'float64',
+    'float': 'float64',
+    'date': 'date',
+    'time': 'string',
```
brianhw: this doesn't convert to time?

edx/analytics/tasks/common/vertica_export.py
brianhw: Curious what the result of the test was.   I don't recall if you actually listed it, but this is still taking a 'datetime' as the BigQuery type, which is the timezone-less representation.   A 'timestamp' in Bigquery includes the timezone.  So if you write to a 'datetime' with timezone information, it's not likely to store it.

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,402 @@
...
+            "table_name='{table}' order by ordinal_position;".format(schema=schema_name, table=table_name)
+    results = []
+    rows = get_vertica_results(credentials, query)
+
+    for row in rows:
+        column_name = row[0]
+        nullable = row[2]
+        field_type = row[1]
+        if '(' in field_type:
+            field_type = field_type[:field_type.find('(')]
```
brianhw: Or you can just always do:   field_type = field_type.rsplit('(')[0]

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,402 @@
...
+
+        if field_type.lower() in ['long varbinary']:
+            log.error('Error Vertica jdbc tool is unable to export field type \'%s\'.  This field will be '
+                      'excluded from the extract.', field_type.lower())
```
brianhw: For the mysql-to-bigquery stuff, I added the ability to exclude by column name.  I would suggest that this filtering be done outside of this method, closer to where such external filtering would be done.   And in that case, it would be possible to avoid getting such error messages if one explicitly excludes the column.

Having now reviewed the rest, I see that this gets called in two locations at the moment.   This call could be moved to both those locations, but that's not ideal.   So I'll just suck this up for now.  : )

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,402 @@
...
+        }
+        vertica_configuration = {
+            'schema_name': self.vertica_schema_name,
+            'warehouse_name': self.vertica_warehouse_name,
+            'exclude_list': self.exclude,
+            'overwrite': self.overwrite,
+            'credentials': self.vertica_credentials,
+        }
+
+        query = "select table_name from all_tables where schema_name='{schema_name}' and table_type='TABLE' " \
```
brianhw: nit: suggest using uppercase on keywords...   select, from, where, and.

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,402 @@
...
+        default=False,
+        significant=False,
+        description='Indicates if the target data sources should be removed prior to generating.'
+    )
+    date = luigi.DateParameter(
+        default=datetime.datetime.utcnow().date(),
+        description='Current run date.  This parameter is used to isolate intermediate datasets.'
+    )
+    exclude = luigi.ListParameter(
+        default=[], description='The Vertica tables that are exclude from exporting.'
```
brianhw: nit:   that are to be excluded ....

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,402 @@
...
+
+        for table_name in table_list:
+            if not self.should_exclude_table(table_name):
+                vertica_configuration['table_name'] = table_name
+                bigquery_configuration['table_name'] = table_name
+
+                yield LoadVerticaTableToBigQuery(
+                    dataset_id=bigquery_configuration['dataset_id'],
+                    credentials=bigquery_configuration['credentials'],
+                    max_bad_records=bigquery_configuration['max_bad_records'],
```
brianhw: So these three needed to be passed separately, rather than as part of the bigquery_configuration that is also being passed?  Something about inheriting from BigQueryLoadTask, IIRC.  So the configuration is then still used for the table_name.  It might be good to put a comment in the code to that effect -- there's a fair amount of experience you've had that is not going to be obvious to someone reading the code.

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,402 @@
...
+            schema_name=self.vertica_configuration['schema_name'],
+            table_name=self.vertica_configuration['table_name'],
+            credentials=self.vertica_configuration['credentials'],
+            database=self.vertica_configuration['warehouse_name'],
+            columns=column_list,
+            destination=target_url,
+            overwrite=self.overwrite,
+            null_string='NNULLL',
+            fields_terminated_by='\x01',
+            delimiter_replacement=' ',
```
brianhw: These should ideally be passed in from the caller, so that the values align with those used to write to BigQuery.   Well, at least the first two, and probably the third.

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,402 @@
...
+            log.info(json.dumps(tmp_schema_output))
+        return self.bigquery_compliant_schema
+
+    @property
+    def insert_source_task(self):
+        return VerticaTableToS3Task(
+            vertica_configuration=self.vertica_configuration,
+            s3_configuration=self.s3_configuration,
+            date=self.date,
+            overwrite=self.overwrite,
```
brianhw: This should pass along self.field_delimiter and self.null_marker at least.   And probably the delimiter_replacement as well.  We're picking the particular values because we're writing the output to BigQuery.  As you note, if we were writing to S3 for a different reason, we might want to use different values.   (E.g. loading into Hive would use a different null value.)

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,402 @@
...
+    def insert_source_task(self):
+        """The sqoop command that manages the data transfer from the source datasource."""
+
+        partition_path_spec = HivePartition('dt', self.date).path_spec
+        target_url = url_path_join(self.s3_configuration['warehouse_path'],
+                                   self.vertica_configuration['warehouse_name'],
+                                   self.vertica_configuration['schema_name'],
+                                   self.vertica_configuration['table_name'],
+                                   partition_path_spec) + '/'
+        column_list = [row[0] for row in self.get_table_schema()]
```
brianhw: It would be nice if we didn't have to do this work twice, once for the writing to S3 and once again to read from S3 and write to BigQuery.    This can be avoided if the column_list can be passed in, and if not provided here, then get_table_schema can provide it like this.

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,402 @@
...
+    def null_marker(self):
+        return 'NNULLL'
+
+    @property
+    def schema(self):
+        """The BigQuery compliant schema."""
+        if self.bigquery_compliant_schema is None:
+            res = []
+            tmp_schema_output = []
+            for field_name, vertica_field_type, nullable in self.vertica_table_schema:
```
brianhw: As mentioned above, I think this should be the only call we need to make to get_vertica_table_schema().  We would then use this to calculate the column_list that would be passed in to VerticaTableToS3Task below.    In future, were we to implement an exclude_field functionality, it would be implemented here, and the corresponding columns would be written to S3 and then loaded to BigQuery.  Note that the column list is then just:  [field.name for field in self.schema].   This can even be inlined in the VerticaTableToS3Task call.

brianhw (commented):
Looks good!   I *think* my parameter-related suggestions are minor, but I'm happy to discuss scope if they seem like a bigger deal.

edx/analytics/tasks/common/sqoop.py
azafty468: Yup, column names not characters.

edx/analytics/tasks/common/sqoop.py
azafty468: Yup, so as you noted above this option is closely tied to the delimiter_replacement option.  This option was necessary when I was not replacing the newline characters.  The Vertica JDBC connector would use a single quotation mark to indicate that a string field bridges multiple lines.  Now that I'm replacing these characters with a single space this option is no longer necessary, nor is it even probably used.   

But I get what you are saying.  As it's written now the delimiter_replacement flag is tuned to BigQuery specific settings.  I've already got the delimiter_replacement parameter set in VerticaTableToS3Task so I'll default the luigi parameter to none in the SqoopImportFromVertica class.

edx/analytics/tasks/common/sqoop.py
azafty468: It's used in def import_args(self)

edx/analytics/tasks/common/sqoop.py
azafty468: A comma is in fact the default option used by sqoop if none is specified.  I can remove it but I was carrying through sqoop's default behavior for clarity.

edx/analytics/tasks/common/sqoop.py
azafty468: done!

edx/analytics/tasks/common/sqoop.py
azafty468: So I can tell you what I did to get to this comment.  It definitely covers the Vertica implementation of Sqoop but I'm not completely sure what the Sqoop code covers and what the JDBC code covers (for instance the --direct flag implementation is governed by the JDBC code).  Strictly speaking there is a relationship between --num-mappers and --split-by

During testing when I specified the --split-by option I also needed to specify the --num-mappers option otherwise sqoop terminated with errors.  That's why the num-mappers option and this comment are clustered near each other.  To my knowledge if you are splitting a dataset you can still set the number of mappers to 1.  The number determines the amount of concurrent parallelism.   

However, and this is a big caveat:  I used this to dig into the load errors on user_session_summary and I set the split_by field to username and the num-mappers to something like 20.  Sqoop create an expected number of files (20 or so).  All but 1 of them were 0 bytes, and one file was 9.6 GB.  The total directory size was less than half of what I was anticipating (24.6 GB).  At this point I stopped using split-by and wrote this note.  

Because I didn't get a successful run I'm a bit hesitant to make any broad conclusions about the usage of --split-by and --num-mappers other than that there is a relationship.

edx/analytics/tasks/common/tests/test_sqoop.py
azafty468: Sure, I'll make this change.  Do we have this as part of our linter setup?  I didn't think this line was flagged the last time I ran it.  But I could be mistaken as it looks like my vagrant image is borked right now.

edx/analytics/tasks/common/vertica_export.py
azafty468: Doh!  I'll add it .  I don't think we are using it anywhere other than in a test table.

edx/analytics/tasks/common/vertica_export.py
azafty468: Neat, I haven't used rsplit before.

edx/analytics/tasks/common/vertica_export.py
azafty468: I'm all for preemptively excluding.  In the case of a schema copy I'm not expecting to actually know that there are or are not long varbinary fields, and any long varbinary field causes the sqoop process to terminate with an error.  Although I guess the issue with the implementation I have here is that a field is being excluded with only a log message being raised.  

Ah, but I see what you are talking about.  I only allow the ability to exclude whole tables and not columns within tables.  I think I could go either way on this and we can talk in person if you'd like.

edx/analytics/tasks/common/vertica_export.py
azafty468: Ewww, there is no good reason for that.  I bet I started with those 3 parameters because LoadVerticaTableToBigQuery inherits from a BigQueryLoadTask task.  I've re-plumbed the construction & init calls so that those 3 parameters are pulled from the bigquery_configuration object by LoadVerticaTableToBigQuery instead of being duplicated on the constructor call.

edx/analytics/tasks/common/vertica_export.py
azafty468: Totally correct, I'm replumbing now

edx/analytics/tasks/common/vertica_export.py
azafty468: Yup, done!

edx/analytics/tasks/common/vertica_export.py
azafty468: I think I need to mull over this.  I do see what you are saying, and that we would be reducing the risk of something like schema drift between queries or any other error that can occur.  But I've got mixed feelings because I do like the isolation where tasks acquire the data they need to run at their run time.

edx/analytics/tasks/common/sqoop.py
brianhw: Yes, but why is it redefined in this class with a definition that is identical to the definition it already gets by inheriting from SqoopImportTask?  Neither null_string nor delimiter_replacement should be redefined in this class.

edx/analytics/tasks/common/sqoop.py
brianhw: Yup.  And so these comments should be approprlate for the output of this class.   Agreed.

edx/analytics/tasks/common/sqoop.py
brianhw: Sounds good.   I think it should be made an option, but we can do that work later.   I would normally put a TODO in the code, but since TODOs are treated like bugs these days, I find myself being reluctant.   Maybe just a comment saying that this could later be made an option.

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,433 @@
...
+    'char': 'string',
+    'varchar': 'string',
+    'long varchar': 'string',
+    'money': 'float64',
+    'numeric': 'float64',
+    'float': 'float64',
+    'date': 'date',
+    'time': 'time',
+    'timetz': 'time',
+    'timestamptz': 'datetime',
```
brianhw: Curious what the result of your testing was. I don't recall if you actually listed it, but this is still taking a 'datetime' as the BigQuery type, which is the timezone-less representation. A 'timestamp' in Bigquery includes the timezone. So if you write to a 'datetime' with timezone information, it's not going to be able to store it.

edx/analytics/tasks/common/vertica_export.py
brianhw: Okay.

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,433 @@
...
+        return target_url
+
+    def output(self):
+        return get_target_from_url(self.s3_output_path())
+
+    @property
+    def insert_source_task(self):
+        """The sqoop command that manages the data transfer from the source datasource."""
+        target_url = self.s3_output_path()
+        column_list = [row[0] for row in self.get_table_schema()]
```
brianhw: I've lost your comment to reply to here, but I disagree that this task should not allow the column_list to be passed in from outside.  I agree that it should have a default behavior here that it can turn to if it is not given such a list, but I don't agree that it should be hardcoded here.   We already know what we want, so why should this be forced to recalculate it, and possibly get a different result?  Instead, it could read:

  column_list = self.column_list if self.column_list else [row[0] for row in self.get_table_schema()]

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,433 @@
...
+
+    @property
+    def table(self):
+        return self.bigquery_configuration['table_name']
+
+    @property
+    def table_description(self):
+        return "Copy of '{}' table from Vertica on {}.{}".format(self.vertica_configuration['table_name'],
+                                                                 self.vertica_configuration['schema_name'],
+                                                                 self.vertica_configuration['table_name'])
```
brianhw: Saying "Copy .... on (schema).(table)" sounds odd to me.   The Mysql-to-BQ load uses:  "Copy of '{}' table from '{}' MySQL database on {}.{}".format(self.table_name, self.database, self.date.isoformat(), optional).    The "on" there was used to convey the date the copy was made.   The {}.{} looks like a table name, but it's actually a period, followed by an optional string that describes the deleted columns, if any.

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,433 @@
...
+    @property
+    def insert_source_task(self):
+        return VerticaTableToS3Task(
+            vertica_configuration=self.vertica_configuration,
+            s3_configuration=self.s3_configuration,
+            date=self.date,
+            overwrite=self.overwrite,
+            sqoop_null_string=self.null_marker,
+            sqoop_fields_terminated_by=self.field_delimiter,
+            sqoop_delimiter_replacement=self.delimiter_replacement,
```
brianhw, line 301: I suggest that "column_list=[field.name for field in self.schema]" be added here.

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,433 @@
...
+                                                    VERTICA_TO_BIGQUERY_FIELD_MAPPING[vertica_field_type],
+                                                    mode='NULLABLE' if nullable else 'REQUIRED'))
+                else:
+                    raise RuntimeError('Error for field {field}: Vertica type {type} does not have a mapping to '
+                                       'BigQuery'.format(field=field_name, type=vertica_field_type))
+                tmp_schema_output.append({'name': field_name,
+                                          'type': VERTICA_TO_BIGQUERY_FIELD_MAPPING[vertica_field_type],
+                                          'mode': 'NULLABLE' if nullable else 'REQUIRED'})
+            self.bigquery_compliant_schema = res
+            log.info(json.dumps(tmp_schema_output))
```
brianhw: I'm not clear that you need to maintain tmp_schema_output.  In part there is always a risk of it becoming out of step with self.bigquery_compliant_schema.  But also because you should just be able to pass self.bigquery_compliant_schema to the log.info() call directly, and it will output the same information using its repr() method.  That is, something like:   log.info("BigQuery schema: %s", self.bigquery_compliant_schema).   I'm assuming this is left over from before you introduced bigquery.SchemaField.

brianhw (commented):
I would really like to avoid the duplication of calls to generate the column list.

edx/analytics/tasks/common/sqoop.py
brianhw: Interesting.   But what confuses me is that our default Sqoop calls regularly set --num-mappers but never set --split-by.   That is, the default for Sqoop is generally to run with --num-mappers=4, and Sqoop figures out how it's going to try to split things itself.  In the regular sqoop code, there is no "else" below.  It uses the default behavior rather than hardcoding to 1 when no explicit override value is provided.  I'm wondering what behavior you got, or would get, if you also allowed Sqoop to choose its split-by, and also its num-mappers.

edx/analytics/tasks/common/sqoop.py
```diff
@@ -283,3 +283,79 @@ def run_job(self, job):
...
+        if self.num_mappers is not None:
+            arglist.extend(['--num-mappers', str(self.num_mappers)])
+        else:
+            arglist.extend(['--num-mappers', '1'])
```
brianhw: So I'm proposing the removal of the two lines above here.

edx/analytics/tasks/common/sqoop.py
azafty468: Ah, it must have been when I was defaulting NNULLL in for this parameter instead of None.  I'll see what other fields are essentially copies from the parent SqoopImportTask.

edx/analytics/tasks/common/sqoop.py
```diff
@@ -283,3 +283,79 @@ def run_job(self, job):
...
+            raise RuntimeError('Error Vertica\'s connector requires specific columns listed. No columns were supplied.')
+
+        arglist = [
+            '--target-dir', self.destination,
+            '--driver', 'com.vertica.jdbc.Driver',
+        ]
+
+        if len(self.columns) == 0:
+            raise RuntimeError('Error cannot complete export of {schema}.{table} because no columns were specified'
+                               ''.format(schema=self.schema_name, table=self.table_name))
```
brianhw: It looks like we do a test of this on line 325 above.  Maybe combine them.

edx/analytics/tasks/common/vertica_export.py
azafty468: Ah, this was a testing step while I was debugging table load errors.  I'll excise this.

edx/analytics/tasks/common/sqoop.py
```diff
@@ -283,3 +283,62 @@ def run_job(self, job):
...
                 pass
+
+
+class SqoopImportFromVertica(SqoopImportTask):
+    """
+    An abstract task that uses Sqoop to read data out of a Vertica database and writes it to a file in CSV format.
+
+    * fields delimited by comma
+    * lines delimited by \n
+    """
```
brianhw: The other comments were appropriate to leave, because it is the behavior of this:    optionally quotes with single quotes, and backslashes.    Is it not?

edx/analytics/tasks/common/sqoop.py
```diff
@@ -283,3 +283,62 @@ def run_job(self, job):
...
+
+    # Direct is not supported by the Vertica JDBC connector
+    direct = None
+    schema_name = luigi.Parameter(
```
brianhw: You had correctly redefined "columns" here in a way that was different from the base class.  In particular, it removed the default of None, and made it a required parameter.  And that is also presumably a result of --direct not being supported.    (Though good to also mention in the comment.)

edx/analytics/tasks/common/sqoop.py
```diff
@@ -283,3 +283,62 @@ def run_job(self, job):
...
+        else:
+            arglist.extend(['--num-mappers', '1'])
+
+        if self.verbose:
+            arglist.append('--verbose')
+        if self.null_string is not None:
+            arglist.extend(['--null-string', self.null_string, '--null-non-string', self.null_string])
+        if self.fields_terminated_by is not None:
+            arglist.extend(['--fields-terminated-by', self.fields_terminated_by])
+        #
```
brianhw: Missing the comment. : )

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,428 @@
...
+
+    @property
+    def table(self):
+        return self.bigquery_configuration['table_name']
+
+    @property
+    def table_description(self):
+        return "Copy of '{}' table from Vertica {}.{}".format(self.vertica_configuration['table_name'],
+                                                                 self.vertica_configuration['schema_name'],
+                                                                 self.vertica_configuration['table_name'])
```
brianhw: Is there a reason not to include the date, but to include the table name twice?

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,428 @@
...
+                if vertica_field_type in VERTICA_TO_BIGQUERY_FIELD_MAPPING:
+                    res.append(bigquery.SchemaField(field_name,
+                                                    VERTICA_TO_BIGQUERY_FIELD_MAPPING[vertica_field_type],
+                                                    mode='NULLABLE' if nullable else 'REQUIRED'))
+                else:
+                    raise RuntimeError('Error for field {field}: Vertica type {type} does not have a mapping to '
+                                       'BigQuery'.format(field=field_name, type=vertica_field_type))
+                tmp_schema_output.append({'name': field_name,
+                                          'type': VERTICA_TO_BIGQUERY_FIELD_MAPPING[vertica_field_type],
+                                          'mode': 'NULLABLE' if nullable else 'REQUIRED'})
```
brianhw: If you're not outputting this anymore, is there a reason to collect it?

edx/analytics/tasks/common/sqoop.py
azafty468: The delimiter replacement is not used by default, which means that yes the optional quotes will be used.

edx/analytics/tasks/common/sqoop.py
azafty468: So the default behavior has columns as an empty list which results in a runtime error, similar to making columns a required parameter.  I figured it was added lines that didn't actually result in a  different outcome.

edx/analytics/tasks/common/vertica_export.py
azafty468: I've removed my additional configuration variables and now I'm using the [hive]'s warehouse_path and appending /import/vertica/sqoop in the code base.  So I'm at least somewhat consistent with what we are already doing.

edx/analytics/tasks/common/vertica_export.py
azafty468: Indeed there is not.

edx/analytics/tasks/common/vertica_export.py
azafty468: Not particularly haha.

edx/analytics/tasks/common/vertica_export.py
azafty468: I nixed testing here.  The data had the timezone dropped without it being adjusted to UTC on S3.  There is nothing we can do to recover it at this point.  The last avenue open would be to tell sqoop which fields are timestamptz fields and modify the SQL query to adjust timezones.  But that looked pretty ugly.

edx/analytics/tasks/common/vertica_export.py
azafty468: Yup, this has been incorporated

edx/analytics/tasks/common/sqoop.py
brianhw: Yes, I guess that's true.

edx/analytics/tasks/common/sqoop.py
```diff
@@ -283,3 +283,61 @@ def run_job(self, job):
...
+
+        # The vertica JBDC client does not handle reserved words in columns correctly.  We must build the query.
+        column_list = ['"{}"'.format(b) for b in self.columns]
+        query = 'SELECT {cols} FROM {tbl} WHERE $CONDITIONS'.format(cols=','.join(column_list), tbl=self.table_name)
+        arglist.extend(['--query', query])
+
+        if self.num_mappers is not None:
+            arglist.extend(['--num-mappers', str(self.num_mappers)])
+        else:
+            arglist.extend(['--num-mappers', '1'])
```
brianhw: I don't understand why this else statement is necessary.   In the regular sqoop code, there is no "else" here. It uses the default behavior rather than hardcoding to 1 when no explicit override value is provided.  That is, the default for Sqoop is generally to run with --num-mappers=4.   Because there's no way to override this value, we're getting it turned off, removing the parallelization, and therefore slowing down the copy time significantly.

On the other hand, if there's a reason (e.g. we're using a query) that we must run with num-mappers=1, then it shouldn't be an option, but should be hardcoded to 1 and commented appropriately.

edx/analytics/tasks/common/sqoop.py
```diff
@@ -283,3 +283,61 @@ def run_job(self, job):
             except Exception:
                 log.exception("Unable to dump metadata information.")
                 pass
+
+
+class SqoopImportFromVertica(SqoopImportTask):
```
brianhw, line 288: Pardon my constant confusion, but why is this a SqoopImport as opposed to being a SqoopExport?   Import makes me think we're loading and export makes me think we're dumping.   This is a dump.   Maybe we should use Load and Dump to be a little clearer?

edx/analytics/tasks/common/vertica_export.py
brianhw: I see.   This is an issue with the Sqoop output, not with the loading into BigQuery.   That would be good information to document somewhere, probably in the docstring for either SqoopImportFromMysql or VerticaTableToS3Task.  It would particularly be good to note what the output is, because I already don't remember.   That is, TIMESTAMPTZ fields in Vertica are dumped by Sqoop to output files as containing the Local time?  Or whatever time it was stored as, with the timezone information stripped off?  Perhaps since we're running a query to get the data dumped, we could just set the timezone to use for output explicitly?   E.g. SET TIMEZONE TO 'GMT';

edx/analytics/tasks/common/sqoop.py
```diff
@@ -283,3 +283,61 @@ def run_job(self, job):
...
+            raise RuntimeError('Error Vertica\'s connector requires specific columns listed. No columns were supplied.')
+
+        arglist = [
+            '--target-dir', self.destination,
+            '--driver', 'com.vertica.jdbc.Driver',
+        ]
+
+        # The vertica JBDC client does not handle reserved words in columns correctly.  We must build the query.
+        column_list = ['"{}"'.format(b) for b in self.columns]
+        query = 'SELECT {cols} FROM {tbl} WHERE $CONDITIONS'.format(cols=','.join(column_list), tbl=self.table_name)
```
brianhw: As I mentioned below, perhaps we could set the timezone here, to improve the accurate dumping of TIMESTAMPTZ data:   SET TIMEZONE TO 'GMT';

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,398 @@
...
+            "table_name='{table}' order by ordinal_position;".format(schema=schema_name, table=table_name)
+    results = []
+    rows = get_vertica_results(credentials, query)
+
+    for row in rows:
+        column_name = row[0]
+        nullable = row[2]
+        field_type = row[1]
+        if '(' in field_type:
+            field_type = field_type.rsplit('(')[0]
```
brianhw: nit:  you don't even need the if statement.   field_type = row[1].rsplit('(')[0]       But not important, just fyi.

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,398 @@
...
+        description='A dictionary specifying the S3-centric configuration.'
+    )
+    vertica_configuration = luigi.DictParameter(
+        default={},
+        description='A dictionary specifying the Vertica-centric configuration.'
+    )
+    bigquery_configuration = luigi.DictParameter(
+        default={},
+        description='A dictionary specifying the BigQuery-centric configuration.'
+    )
```
brianhw: One downside about this list of configuration objects is that it's not clear what needs to be in any of them, let alone any enforcement of that.   Is there a place where this is documented, if not here?  At the very least, it would be good if there were at least one place, and if necessary, the other places could at least point to that place.   Looking at this, I have no idea how to call this Task.

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,398 @@
...
+
+    # I need to specify defaults for these two parameters because the luigi interpreter is validating the field before
+    # the __init__ call below
```
brianhw: I'm a little confused about how this should be called, even beyond the specification of configuration dicts.  Are these parameters needed to be passed in correctly, or should they not be passed in at all (and use the None defaults)?   Or they should be not passed in for now because the super().__init__ doesn't yet make use of them, but if it did, then they would have to be passed in then, either instead of or in addition to the bigquery_configuration.  It would be good if this comment made that a little clearer.

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,398 @@
...
+    credentials = luigi.Parameter(
+        default=None
+    )
+
+    def __init__(self, *args, **kwargs):
+        super(LoadVerticaTableToBigQuery, self).__init__(*args, **kwargs)
+        # Because of the way Luigi handles parameters the next three dictionary objects are not populated until after
+        # the super call.  This means we are changing 3 internal variables after a function call to the parent class.
+        # If these 3 parameters are ever used in the BigQueryLoadTask __init__ call we will have to replicate the logic
+        # here.
```
brianhw: I wasn't clear which logic needed to be replicated.   The contents of the super().__init__(), because it operated on the wrong values?   

I wonder if this would just be clearer if the values from bigquery_configuration were passed into this task explicitly, rather than trying to use a dict for these at this level.   As you had it earlier.  

```
               yield LoadVerticaTableToBigQuery(
                    dataset_id=bigquery_configuration['dataset_id'],
                    credentials=bigquery_configuration['credentials'],
                    max_bad_records=bigquery_configuration['max_bad_records'],
```
Oh, and I guess pass the bigquery table name explicitly as well.   It just seems that the dict is not being helpful here, in that nothing gets passed very far, and it requires weird overwriting behavior in __init__ and such.

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,398 @@
...
+    def insert_source_task(self):
+        return VerticaTableToS3Task(
+            vertica_configuration=self.vertica_configuration,
+            s3_configuration=self.s3_configuration,
+            date=self.date,
+            overwrite=self.overwrite,
+            sqoop_null_string=self.null_marker,
+            sqoop_fields_terminated_by=self.field_delimiter,
+            sqoop_delimiter_replacement=self.delimiter_replacement,
+            column_list=[field for field, type, nullable in self.schema]
```
brianhw: How does this work?   Isn't self.schema a list of SchemaField objects?

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,398 @@
...
+            return True
+        return False
+
+    def requires(self):
+        yield ExternalURL(url=self.vertica_credentials)
+        yield ExternalURL(url=self.gcp_credentials)
+
+        bigquery_configuration = {
+            'dataset_id': self.vertica_schema_name if self.bigquery_dataset is None else self.bigquery_dataset,
+            'overwrite': self.overwrite,
```
brianhw: Is this actually used?

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,398 @@
...
+    for row in rows:
+        column_name = row[0]
+        nullable = row[2]
+        field_type = row[1]
+        if '(' in field_type:
+            field_type = field_type.rsplit('(')[0]
+
+        if field_type.lower() in ['long varbinary']:
+            log.error('Error Vertica jdbc tool is unable to export field type \'%s\'.  This field will be '
+                      'excluded from the extract.', field_type.lower())
```
brianhw: It might be good to also include the schema_name, table_name, and column_name here, as well as the field type.  Otherwise it would be hard to debug this.

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,398 @@
...
+        description='The field delimiter used by Sqoop.'
+    )
+    sqoop_delimiter_replacement = luigi.Parameter(
+        default=' ',
+        description='The string replacement value for special characters encountered by Sqoop when exporting from '
+                    'Vertica.'
+    )
+    column_list = luigi.ListParameter(
+        description='The column names being extracted from this table.'
+    )
```
brianhw, line 135: Interesting.   I thought you wanted this to be able to stand more on its own.  I was suggesting that this be specified, but have a default=None, so that if no column_list is specified here, it would use the column values returned by get_vertica_table_schema().

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,398 @@
...
+
+    def output(self):
+        return get_target_from_url(self.s3_output_path())
+
+    @property
+    def insert_source_task(self):
+        """The sqoop command that manages the data transfer from the source datasource."""
+        target_url = self.s3_output_path()
+
+        if len(self.column_list) <= 0:
```
brianhw: I guess I would put this line above here:
column_list = self.column_list if self.column_list else [row[0] for row in self.get_table_schema()]

edx/analytics/tasks/common/sqoop.py
azafty468: Import is the nomenclature Sqoop uses because you are bringing data in from a remote cluster as opposed to exporting data from the local HDFS system.  I am also following the naming convention of SqoopImportFromMysql as in both cases we are moving from a specific datastore to an S3 intermediary.

edx/analytics/tasks/common/vertica_export.py
azafty468: Ah it is a SchemaField.  The last change I made was to move this from self.vertica_table_schema to self.schema.  It's funny, I thought I ran a test export of the test schema to confirm this new code worked.  I'll fix this and schedule the job again.

edx/analytics/tasks/common/sqoop.py
azafty468: Yup, I mentioned this in either this PR or in chat.  We are constructing the query and can adjust the syntax in the case a timezone needs manually adjustment.  Another alternative would be to seek out an updated jar that would properly handle timezone information as it appears that this a bug in the version of the jar.  

I'm not sure how much we care about this right now.

edx/analytics/tasks/common/vertica_export.py
azafty468: that is neat and I've updated it

edx/analytics/tasks/common/vertica_export.py
azafty468: Correct, that is the major drawback here is that it is not self documenting.  I'll add in a few notes of what the expected structure is supposed to contain.

azafty468: Although I would say that there is some level of documentation at the bottom of the file where the objects are constructed.

edx/analytics/tasks/common/vertica_export.py
azafty468: Yeah, I think this opens up risk for the sake of having niceties of configuration structure.  I'm unravelling the dict parameters now.

edx/analytics/tasks/common/vertica_export.py
azafty468: Overwrite is not which I'm assuming is what you are highlighting.  At one point I was pulling all configurations from these dictionaries.  That has been eroded over the past few commits, and now it doesn't particularly make any sense to keep them.  The grand experiment is over.

edx/analytics/tasks/common/vertica_export.py
azafty468: I did but in the strict case of writing code useful for now it didn't serve a purpose.  Down the road if we want to expand the use of this class we can pick up the idea of having it autogenerate it's schema if not provided.  But for now it was an unnecessary bit of code.

edx/analytics/tasks/common/vertica_export.py
azafty468: You are indeed correct

edx/analytics/tasks/common/sqoop.py
azafty468: I've taken out the else statement.

edx/analytics/tasks/common/sqoop.py
azafty468: I've carried forward the timezone conversion to sqoop.  We should keep an eye out for this the next time we upgrade this jar however.  Timetz fields are handled fine, it's just timestamptz.

edx/analytics/tasks/common/sqoop.py
azafty468: Aaaaand the lines are back in now.  When left out the sqoop command terminates with an error: When importing query results in parallel, you must specify --split-by.

I am in fact not importing in parallel so I suspect this is an error of the JDBC client.

edx/analytics/tasks/common/vertica_export.py
azafty468: I have added a note indicating that times are coerced to UTC

azafty468:
Ok, I think this is everything

edx/analytics/tasks/common/sqoop.py
```diff
@@ -283,3 +283,75 @@ def run_job(self, job):
...
+    # Direct is not supported by the Vertica JDBC connector
+    direct = None
+    # A bug in the Sqoop process with this JDBC connector makes this parameter useless.
+    num_mappers = None
+    schema_name = luigi.Parameter(
+        description='The schema that contains the table being exported.'
+    )
+    timezone_adjusted_column_list = luigi.ListParameter(
+        default=[],
+        description='The list of columns that need to have their time''s manually adjusted to UTC.'
```
brianhw: nit:  time's => times

edx/analytics/tasks/common/sqoop.py
```diff
@@ -283,3 +283,75 @@ def run_job(self, job):
...
+class SqoopImportFromVertica(SqoopImportTask):
+    """
+    An abstract task that uses Sqoop to read data out of a Vertica database and writes it to a file in CSV format.
+
+    * fields delimited by comma
+    * lines delimited by \n
+    * fields optionally enclosed by single quotes (')
+    """
+
+    # Direct is not supported by the Vertica JDBC connector
```
brianhw: nit:   trailing period.

edx/analytics/tasks/common/sqoop.py
```diff
@@ -283,3 +283,75 @@ def run_job(self, job):
...
+    def import_args(self):
+        if self.columns is None or len(self.columns) == 0:
+            raise RuntimeError('Error Vertica\'s connector requires specific columns listed. No columns were supplied.')
+
+        arglist = [
+            '--target-dir', self.destination,
+            '--driver', 'com.vertica.jdbc.Driver',
+        ]
+
+        # The vertica JBDC client does not handle timestamptz fields or columns names that are reserved words
```
brianhw: nit: "columns names" => "column names".   Also trailing period.

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,393 @@
...
+            table_name=self.table_name,
+            credentials=self.credentials,
+            database=self.warehouse_name,
+            columns=self.column_list,
+            destination=target_url,
+            overwrite=self.overwrite,
+            null_string=self.sqoop_null_string,
+            fields_terminated_by=self.sqoop_fields_terminated_by,
+            delimiter_replacement=self.sqoop_delimiter_replacement,
+            timezone_adjusted_column_list=self.timestamptz_column_list,
```
brianhw, line 205: not a big deal, but interesting that the parameter names are so different....

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -0,0 +1,393 @@
...
+    vertica_credentials = luigi.Parameter(
+        config_path={'section': 'vertica-export', 'name': 'credentials'},
+        description='Path to the external Vertica access credentials file.',
+    )
+    bigquery_dataset = luigi.Parameter(
+        default=None,
+        description='The target dataset that will hold the Vertica schema.'
+    )
+    max_bad_records = luigi.IntParameter(
+        default=0, description="Number of bad records ignored by BigQuery before failing a load job."
```
brianhw: nit: just curious why this is on one line, and all others are not...

brianhw (approved):
Yay!   Nice!  Only nits, and no blockers.....

edx/analytics/tasks/common/sqoop.py
brianhw: Thanks!

edx/analytics/tasks/common/vertica_export.py
azafty468: Yeah that one is a bit odd.  While all four parameters are intended to control Sqoop functionality, the three starting with sqoop_ are essentially command line functions.  The timestamptz list is information owned by Vertica being conveyed to the Sqoop wrapper.  So semantics but that's how I thought about it.

edx/analytics/tasks/common/vertica_export.py
azafty468: The same thing is in place for vertica_warehouse_name and exclude.  I think I did it just about everywhere the combined line wouldn't exceed the 80 character limit with the exclusion of overwrite.  I can break it out to keep it uniform with the other lines.

PR #492, opened 2018-04-04
Create an empty 'course_program_metadata' table for insights.

HassanJaveed84:
Wanted to get feedback whether this is a reasonable way to fix it.
Going to verify if Insights stops complaining after running this workflow.

### DIFF
--- a/edx/analytics/tasks/insights/enrollments.py
+++ b/edx/analytics/tasks/insights/enrollments.py
@@ -1978,3 +1978,19 @@ class ImportEnrollmentsIntoMysql(OverwriteHiveAndMysqlDownstreamMixin, CourseSum
         ]
         if self.enable_course_catalog:
             yield CourseProgramMetadataInsertToMysqlTask(**course_summary_kwargs)
+        else:
+            # Create an empty 'course_program_metadata' table for insights.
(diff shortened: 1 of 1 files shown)


azafty468 (approved):
This looks good.  Simple and good.  Is there a chance the output has anything other than the credentials enabled mysql target?  I couldn't find one but I only made a quick review.

brianhw (changes requested):
Very elegant!  

A couple of questions/comments:
* Maybe it's worth putting in a try/finally, so the close() gets called if the create_table fails.
* Maybe overkill, but is it worth adding the create_database call as well?   Is there any chance this might be the first table?
* Do you think it's safe to skip over the marker file stuff?    I'm guessing that it is, since stuff should still do the right thing were the course_catalog to be enabled later.  It might be worth adding a brief comment here that we're doing that, though.

HassanJaveed84:
- Added create_database() call.
- Added try/finally.
- Added comment regarding the marker table entry.
- Verified that insights works after triggering the enrollment workflow with enable_course_catalog=False.

@azafty468 correct, it's only the credentials enabled target. 

This is ready for a final review before merging.

brianhw (approved):
LGTM.  👍

PR #493, opened 2018-04-10
Copy mysql database to bigquery dataset

brianhw:
Some of the features:
* Special handling of the marker table on overwrite was required due to quotas by BigQuery on table modifications.  To get around the quotas, the marker table is deleted in its entirety, and appended after that.
* Direct mode was not used in Sqoop calls because the null strings needed to be overridden, and direct mode prevents that.  Null strings needed to be redefined because the default value of "NULL" appears in field values for fields that are defined to be not nullable.
* Per-table filtering was left in as is done for copying mysql databases to Vertica.
* Per-column filtering was added to avoid columns that contain ASCII 0 characters.  BigQuery returns errors when loading such values.   This should not have a sizeable performance effect, as we already had to disable direct mode. 
* This moved away from using dynamic dependencies, as it seemed this was putting more load on repeatedly checking whether tables had already been loaded.  It seemed that at times those checks that had earlier been successful started being unsuccessful, resulting in a particular table being loaded multiple times.  This is much less likely to happen with regular requirements.
* The easiest way to handle not loading metadata files (like _metadata and _SUCCESS) was to delete them after synchronizing directories to Google Storage but before loading into BigQuery.   This could have been done differently.

This finally works on copying lms, ecommerce, discovery, and credentials databases to BigQuery.  Jenkins jobs have currently been set up to run these daily.

### DIFF
--- a/edx/analytics/tasks/common/bigquery_load.py
+++ b/edx/analytics/tasks/common/bigquery_load.py
@@ -97,6 +97,12 @@ class BigQueryTarget(luigi.Target):
         if not table.exists():
             table.create()
 
+    def delete_marker_table(self):
+        dataset = self.client.dataset(self.dataset_id)
+        table = dataset.table('table_updates')
+        if table.exists():
+            table.delete()
+
     def clear_marker_table(self):
         if not self.marker_table_exists():
             return
@@ -154,6 +160,12 @@ class BigQueryLoadTask(BigQueryLoadDownstreamMixin, luigi.Task):
     # Regardless whether loading only a partition or an entire table,
     # we still need a date to use to mark the table.
     date = luigi.DateParameter()
+
+    skip_clear_marker = luigi.BoolParameter(
+        default=False,
+        description='Set this when pre-deleting the marker table, to avoid quota issues with clearing entries.',
+        significant=False,
+    )
 
     output_target = None
     required_tasks = None
@@ -233,12 +245,14 @@ class BigQueryLoadTask(BigQueryLoadDownstreamMixin, luigi.Task):
                 if table.exists():
                     partition = self._get_table_partition(dataset, table)
                     partition.delete()
-                self.output().clear_marker_table_entry()
+                if not self.skip_clear_marker:
+                    self.output().clear_marker_table_entry()
             else:
                 # Delete the entire table and all markers related to the table.
                 if table.exists():
                     table.delete()
-                self.output().clear_marker_table()
+                if not self.skip_clear_marker:
+                    self.output().clear_marker_table()
 
     def _get_destination_from_source(self, source_path):
         parsed_url = urlparse.urlparse(source_path)
@@ -257,6 +271,13 @@ class BigQueryLoadTask(BigQueryLoadDownstreamMixin, luigi.Task):
         else:
             log.debug(" ".join(['gsutil', '-m', 'rsync', source_path, destination_path]))
             return_code = subprocess.call(['gsutil', '-m', 'rsync', source_path, destination_path])
+            if return_code == 0:
+                # Remove any files that were copied whose names have leading underscores, since
+                # these files cannot be uploaded to BigQuery.  It is easier to remove them here
+                # than to exclude them either in the rsync or in the load steps.
+                underscore_path = url_path_join(destination_path, '_*')
+                log.debug(" ".join(['gsutil', 'rm', underscore_path]))
+                return_code = subprocess.call(['gsutil', 'rm', underscore_path])
 
         if return_code != 0:
             raise RuntimeError('Error while syncing {source} to {destination}'.format(
--- a/edx/analytics/tasks/warehouse/load_internal_reporting_database.py
+++ b/edx/analytics/tasks/warehouse/load_internal_reporting_database.py
@@ -7,16 +7,24 @@ import re
 
 import luigi
 
+from edx.analytics.tasks.common.bigquery_load import BigQueryLoadDownstreamMixin, BigQueryLoadTask, BigQueryTarget
 from edx.analytics.tasks.common.mysql_load import get_mysql_query_results
 from edx.analytics.tasks.common.sqoop import SqoopImportFromMysql
 from edx.analytics.tasks.common.vertica_load import SchemaManagementTask, VerticaCopyTask
 from edx.analytics.tasks.util.hive import HivePartition, WarehouseMixin
 from edx.analytics.tasks.util.url import ExternalURL, url_path_join
 
+try:
+    from google.cloud.bigquery import SchemaField
+    bigquery_available = True  # pylint: disable=invalid-name
+except ImportError:
+    bigquery_available = False  # pylint: disable=invalid-name
+
+
 log = logging.getLogger(__name__)
 
 
-class MysqlToVerticaTaskMixin(WarehouseMixin):
+class MysqlToWarehouseTaskMixin(WarehouseMixin):
     """
     Parameters for importing a mysql database into the warehouse.
     """
@@ -30,7 +38,7 @@ class MysqlToVerticaTaskMixin(WarehouseMixin):
     )
 
 
-class LoadMysqlToVerticaTableTask(MysqlToVerticaTaskMixin, VerticaCopyTask):
+class LoadMysqlToVerticaTableTask(MysqlToWarehouseTaskMixin, VerticaCopyTask):
     """
     Task to import a table from mysql into vertica.
     """
@@ -188,7 +196,7 @@ class PostImportDatabaseTask(SchemaManagementTask):
         )
 
 
-class ImportMysqlToVerticaTask(MysqlToVerticaTaskMixin, luigi.WrapperTask):
+class ImportMysqlToVerticaTask(MysqlToWarehouseTaskMixin, luigi.WrapperTask):
     """Provides entry point for importing a mysql database into Vertica."""
 
     schema = luigi.Parameter(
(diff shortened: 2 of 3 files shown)


edx/analytics/tasks/warehouse/load_internal_reporting_database.py
```diff
@@ -267,3 +275,268 @@ def run(self):
...
+    # Sqoop outputs these as two character hex-codes separated by a space.  BigQuery reads these in and displays them as
+    # the corresponding hex values without the space separation.  This doesn't seem like the desired behavior, but the
+    # problem probably lies with Sqoop.  At present, LMS has only two columns that are longblobs, and they are passwords
+    # in a table that is excluded.
+    'binary': 'BYTES',
+    'varbinary': 'BYTES',
+    'tinyblob': 'BYTES',
+    'blob': 'BYTES',
+    'mediumblob': 'BYTES',
+    'longblob': 'BYTES',
```
brianhw, line 317: So, for example, a longblob password that (hypothetically) begins with 'fB...' would be dumped by Sqoop as '66 41 ...', and this gets displayed in BigQuery after import as '6641...'.    Given the Sqoop dump, it's not clear to me what would make this better, so I'm leaving it as-is.   If someone eventually needs blob support in BigQuery on imported data like this, it should be possible to write a temporary function to parse the resulting data as needed.

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
```diff
@@ -267,3 +275,268 @@ def run(self):
...
+    # DATE
+    'date': 'DATE',
+    # TIME
+    'time': 'TIME',
+    # DATETIME
+    'datetime': 'DATETIME',
```
brianhw, line 323: It turns out that *all* tables in the LMS use datetime, and none use timestamp.  So no dates get output with timezone information, so it would be left up to the BigQuery user to know that the dates without timezones are UTC.  I think that's appropriate.

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
```diff
@@ -267,3 +275,268 @@ def run(self):
...
+    @property
+    def null_marker(self):
+        """The null sequence in the data to be copied."""
+        # Using "NULL" doesn't work, because there are values in several tables
+        # in non-nullable fields that store the string value "NULL" (or "Null" or "null").
+        # As with the field delimiter, we must use something here that we don't expect to
+        # appear in a field value.
+        # Note that this option only works if "direct" mode is disabled, otherwise it will be ignored
+        # and "NULL" will always be output.
+        return 'NNULLL'
```
brianhw: The value of this string was completely arbitrary.

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
```diff
@@ -267,3 +275,268 @@ def run(self):
...
+
+        return self.table_schema
+
+    @property
+    def field_delimiter(self):
+        """The delimiter in the data to be copied."""
+        # Select a delimiter string that will not occur in field values.  Unfortunately,
+        # there probably is no failsafe string, but maybe making it longer than a single
+        # character would be safer.  But slower?
+        return '\x01'
```
brianhw: I didn't actually find \x01 to be an issue, so the comment was unnecessary.

edx/analytics/tasks/common/bigquery_load.py
```diff
@@ -257,6 +271,11 @@ def _copy_data_to_gs(self, source_path, destination_path):
         else:
             log.debug(" ".join(['gsutil', '-m', 'rsync', source_path, destination_path]))
             return_code = subprocess.call(['gsutil', '-m', 'rsync', source_path, destination_path])
+            if return_code == 0:
+                # Remove any files that were copied whose names have leading underscores.
+                underscore_path = url_path_join(destination_path, '_*')
+                log.debug(" ".join(['gsutil', 'rm', underscore_path]))
+                return_code = subprocess.call(['gsutil', 'rm', underscore_path])
```
azafty468: I'm curious why does a return_code == 0 result in needed to remove files starting with an underscore?  And if this delete needs to happen and fails should that result in a RuntimeError a few lines below this?

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
```diff
@@ -7,16 +7,24 @@
...
+    bigquery_available = False  # pylint: disable=invalid-name
+
+
 log = logging.getLogger(__name__)
 
 
-class MysqlToVerticaTaskMixin(WarehouseMixin):
+class MysqlToWarehouseTaskMixin(WarehouseMixin):
     """
     Parameters for importing a mysql database into the warehouse.
```
azafty468, line 29: I'm not sure if we want to align the comment to the class name here.

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
```diff
@@ -267,3 +275,268 @@ def run(self):
...
+                    field_type = field_type.rsplit('(')[0]
+
+                bigquery_type = MYSQL_TO_BIGQUERY_TYPE_MAP.get(field_type)
+                mode = 'REQUIRED' if field_null == 'NO' else 'NULLABLE'
+                description = ''
+
+                if self.should_exclude_field(field_name):
+                    self.deleted_fields.append(field_name)
+                else:
+                    self.table_schema.append(SchemaField(field_name, bigquery_type, description=description, mode=mode))
```
azafty468, line 391: Neat, I wasn't automatically generating a description, I'll "borrow" this in my PR.

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
```diff
@@ -267,3 +275,268 @@ def run(self):
...
+}
+
+
+class MysqlToBigQueryTaskMixin(MysqlToWarehouseTaskMixin):
+    """
+    Parameters for importing a mysql database into BigQuery.
+    """
+
+    exclude_field = luigi.ListParameter(
+        default=(),
```
azafty468, line 335: Is there a reason this is an empty tuple rather than an empty list?

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
```diff
@@ -267,3 +275,268 @@ def run(self):
...
+        destination = url_path_join(
+            self.warehouse_path,
+            self.warehouse_subdirectory,
+            self.database,
+            self.table_name,
+            partition_path_spec
+        ) + '/'
+        return SqoopImportFromMysql(
+            table_name=self.table_name,
+            credentials=self.db_credentials,
```
azafty468, line 435: I'm going to cherry-pick your changes on bigquery_load.py and some of your changes here are going to influence my design.  I think I will try to rework db_credentials and credentials to be source_credentials and target_credentials.

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
```diff
@@ -267,3 +275,268 @@ def run(self):
...
+        # There are quotas on deletes and upserts on a table -- no more than 96 per day.
+        # And we're trying to work around that.
+        # Note that we have to do this early, before scheduling begins, so that nothing is present then.
+        if self.overwrite:
+            # First, create a BigQueryTarget object, so we can connect to BigQuery.
+            credentials_target = ExternalURL(url=self.credentials).output()
+            target = BigQueryTarget(
+                credentials_target=credentials_target,
+                dataset_id=self.dataset_id,
+                table="dummy_table",
```
azafty468, line 498: Is this typically how we do this?  I know we pull the client from the target, but I don't recall using a dummy configuration variable used outside of test.  I'm also not positive that I overlooked a previous use of it.

azafty468 (approved):
Looks good 👍.  A few comments on non-material sections of the codebase.

edx/analytics/tasks/common/bigquery_load.py
brianhw: An initial return_code != 0 is something we want to always fail on and return a RuntimeError, whether from the cp, the rsync or the rm.   The particulars here is that all files except those beginning with underscores need to be synced, because the underscored files cannot be loaded into BigQuery.  I thought it was easier to sync everything and then remove the underscored files, rather than figuring out how to sync only the non-underscored files, or only load the non-underscored files.

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
brianhw: I don't understand what you mean by "align".   Physically, or in terms of the comment's content?

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
brianhw: But note that I don't actually set the description here to anything.   I've got no source for it, so it's left blank.  I do make a description for the table, however.

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
brianhw: I believe it's safer, being immutable.   Lists as defaults are more problematic, in that you can modify them, and then those modified defaults become the default for later instantiations of the parameter.   Bad things can happen.

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
brianhw: That sounds like a good direction.  The regular read-replica jobs will need to be updated, but that should be straight-forward.  And I think I'm fine with letting you do that, rather than trying to do it in this PR, in part because of that side-effect.

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
brianhw: We don't typically do this anywhere, but I needed access to a target in order to delete the marker table.  The dummy references here were to make clear the limited use of this target.

PR #494, opened 2018-04-11
Changed video-timeline workflow to use user_id instead of username.

HassanJaveed84:
Validation pending on history generation(in progress).

### DIFF
--- a/edx/analytics/tasks/insights/video.py
+++ b/edx/analytics/tasks/insights/video.py
@@ -183,13 +183,13 @@ class UserVideoViewingTask(EventLogSelectionMixin, MapReduceJobTask):
         # This has already been checked when getting the event, so just fetch the value.
         timestamp = eventlog.get_event_time_string(event)
 
-        # Strip username to remove trailing newlines that mess up Luigi.
-        username = event.get('username', '').strip()
-        if not username:
-            log.error("Video event without username: %s", event)
-            # Slow: self.incr_counter(self.counter_category_name, 'Discard Video Missing Something', 1)
-            # Slow: self.incr_counter(self.counter_category_name, 'Discard Video Missing username', 1)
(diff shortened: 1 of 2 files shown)


edx/analytics/tasks/insights/video.py
```diff
@@ -183,13 +183,21 @@ def mapper(self, line):
...
-            log.error("Video event without username: %s", event)
-            # Slow: self.incr_counter(self.counter_category_name, 'Discard Video Missing Something', 1)
-            # Slow: self.incr_counter(self.counter_category_name, 'Discard Video Missing username', 1)
+        # # Strip username to remove trailing newlines that mess up Luigi.
+        # username = event.get('username', '').strip()
+        # if not username:
+        #     log.error("Video event without username: %s", event)
+        #     # Slow: self.incr_counter(self.counter_category_name, 'Discard Video Missing Something', 1)
+        #     # Slow: self.incr_counter(self.counter_category_name, 'Discard Video Missing username', 1)
+        #     return
```
brianhw: This should just be removed, I think.

brianhw (commented):
Did we want to change the names of the history directories?   Instead of "user_video_viewing" and "user_video_viewing_by_date", maybe something like "userid_video_viewing"?   That way one could run this in parallel and compare, rather than having to overwrite.   And also so that when people upgrade, they don't get lots of errors from mixing the old with new.   Or "video_viewing_by_user", or just "video_viewing"?

HassanJaveed84:
I did change those, but in a separate branch(which is what I'm using for history generation). I thought after we're done validating, we could keep the original names of the directories. But your point about other people upgrading is very valid. I'll change them here as well.

brianhw (approved):
LGTM.   👍

PR #495, opened 2018-04-17
Change user-activity workflow to use user_id.

HassanJaveed84:
Still need to figure out how to do Vertica loading and add a couple of unit tests.


azafty468:
So I'm good with what you have so far.  Do you know what the error rate is for missing user_id's as opposed to user_name?  I see that you are emitting log lines when user_id is missing.  

It also looks like UserActivityTask and it's related classes are only used in this file and the UserActivityWorkflow workflow.  I'm curious how we are going to be handling this in Insights or Vertica and what additional testing will be needed there.

HassanJaveed84:
I haven't gotten to the validation part yet. I thought of adding a counter and doing row count comparisons between what we have now in production and this. That should give us a good understanding of the magnitude of differences. 

Adding data to vertica is proving rather tricky because of computation of 'internal_user_activity' would go away(not reflected in the PR yet). We story daily activity's data in a separate dated folder and to load all that data in Vertica would require us to append wildcard to the input's source path, which wouldn't be the best approach IMO.

HassanJaveed84:

edx/analytics/tasks/insights/tests/test_user_activity.py
```diff
@@ -80,32 +81,32 @@ def test_illegal_course_id(self):
...
+    # def test_empty_username(self):
+    #     line = self.create_event_log_line(username='')
+    #     self.assert_no_map_output_for(line)
 
-    def test_whitespace_username(self):
-        line = self.create_event_log_line(username='   ')
-        self.assert_no_map_output_for(line)
+    # def test_whitespace_username(self):
+    #     line = self.create_event_log_line(username='   ')
+    #     self.assert_no_map_output_for(line)
```
brianhw: I would probably just delete these tests entirely, rather than commenting them out.

edx/analytics/tasks/warehouse/load_internal_reporting_user_activity.py
```diff
@@ -117,16 +31,14 @@ class LoadInternalReportingUserActivityToWarehouse(WarehouseMixin, VerticaCopyTa
...
     def insert_source_task(self):
-        hive_table = "internal_reporting_user_activity"
-        partition_location = url_path_join(self.warehouse_path, hive_table, self.partition.path_spec) + '/'
-        return ExternalURL(url=partition_location)
+        hive_table = "user_activity_by_user"
+        # User activity data for each day is stored in a dated directory.
+        # We want to be able to load all that data into Vertica in one go, hence we use
+        # a wildcard('*') here.
+        url = url_path_join(self.warehouse_path, hive_table) + '/*/'
+        return ExternalURL(url=url)
```
brianhw: Does this now require that --overwrite be set, in order to not append everything each day (or each time) this is called?    

Also, since we don't actually maintain the view below anymore, we should probably just finally remove it (so it doesn't continue to confuse me at the very least : ).     And in fact, we probably only need this task in this file, and can remove all the tasks below this.

edx/analytics/tasks/warehouse/load_internal_reporting_user_activity.py
```diff
@@ -290,13 +202,6 @@ class UserActivityWorkflow(WeeklyIntervalMixin, WarehouseMixin, luigi.WrapperTas
     n_reduce_tasks = luigi.Parameter()
 
     def requires(self):
-        yield InternalReportingUserActivityPartitionTask(
-            date=self.end_date,
-            n_reduce_tasks=self.n_reduce_tasks,
-            warehouse_path=self.warehouse_path,
-            overwrite_n_days=self.overwrite_n_days,
-            overwrite=self.overwrite_hive,
-        )
```
brianhw, line 299: I wonder if it's worth getting rid of this task altogether, and just having the user-activity jenkins job call InsertToMysqlCourseActivityTask?

edx/analytics/tasks/warehouse/load_internal_reporting_user_activity.py
```diff
@@ -117,16 +31,14 @@ class LoadInternalReportingUserActivityToWarehouse(WarehouseMixin, VerticaCopyTa
...
-
     @property
     def insert_source_task(self):
-        hive_table = "internal_reporting_user_activity"
-        partition_location = url_path_join(self.warehouse_path, hive_table, self.partition.path_spec) + '/'
-        return ExternalURL(url=partition_location)
+        hive_table = "user_activity_by_user"
+        # User activity data for each day is stored in a dated directory.
+        # We want to be able to load all that data into Vertica in one go, hence we use
+        # a wildcard('*') here.
```
brianhw, line 39: You noted that you didn't like the wildcard approach.   Is there something that would improve it?   One could make the pattern more specific, perhaps, to match only dated subpaths?

brianhw (commented):
Looks good!   Just a few questions about task removal and about overwrite.

edx/analytics/tasks/warehouse/load_internal_reporting_user_activity.py
HassanJaveed84: We don't need to pass in overwrite, user-activity warehouse loading is part of load-warehouse task which does the schema swapping. I thought of removing the view, but wasn't sure. I'll remove it.

edx/analytics/tasks/warehouse/load_internal_reporting_user_activity.py
HassanJaveed84: I agree, I'll make the changes and update user-activity job as part of releasing this.

edx/analytics/tasks/warehouse/load_internal_reporting_user_activity.py
HassanJaveed84: We could use glob pattern, but it would be something like 
dt=[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9] or dt=[0-9]\*-[0-9]\*-[0-9]*(looks ugly still?)
Both would be a lot less performant as compared to using a plain wildcard, however it does not really matter as the overhead would still be in seconds when luigi does a 'hadoop fs -cat URI' for reading the input source. Thoughts ?

edx/analytics/tasks/warehouse/load_internal_reporting_user_activity.py
brianhw: Excellent!   So much simpler!

edx/analytics/tasks/warehouse/load_internal_reporting_user_activity.py
brianhw: It might be enough just to use something like '/dt=*/', as the likelihood of there being a badly-formatted-date in such a subdirectory seems really tiny to me.

brianhw (approved):
LGTM!   👍   Whatever you decide for the glob...

HassanJaveed84:
Fixed acceptance test.

brianhw:
LGTM.  👍

PR #496, opened 2018-04-20
Exclude unloadable files at rsync time

brianhw:
Avoid copying _SUCCESS and _metadata files from source directories, since these cannot be loaded into BigQuery.


azafty468 (approved):
Shoot I wasn't checking my email so I didn't see this come through.  Also super neat that there is an exclusion pattern available to use.  These changes look good!

HassanJaveed84 (approved):
👍

PR #499, opened 2018-05-03
add support for docker devstack

rao-abdul-mannan:
Support for docker devstack & also allow running acceptance tests with it.
Follow instructions on [Devstack PR](https://github.com/edx/devstack/pull/298) for docker devstack installation.

Reference [PR-490](https://github.com/edx/edx-analytics-pipeline/pull/490)


edx/analytics/tasks/tests/acceptance/services/hive.py
```diff
@@ -13,15 +16,22 @@ def reset(self):
...
+                '--shell', ". $HOME/.bashrc && hive --service cli{db} -e \"{stmt}\"".format(
+                    db=db_parameter,
+                    stmt=statement
+                ),
+            ])
+        else:
+            cmd = ['hive', '--service', 'cli']
+            if explicit_db:
+                cmd.extend(['--database', self.database_name])
+            cmd.extend(['-e', statement])
```
brianhw, line 33: nit:   the cmd generation here can be factored out, and used here and on line 24?  Well, not quite.   Line 24 puts the statement in quotes, and here it's not necessary.

edx/analytics/tasks/tests/acceptance/services/task.py
```diff
@@ -41,33 +46,55 @@ def launch(self, task_args, config_override=None):
...
+                command.extend(task_args)
+                command.append('--local-scheduler')
+            else:
+                # run the command in a shell since that's what is done by remote-task.
+                # Otherwise values like '"*"' cause problems since they are properly escaped for the "shell" case but
+                # malformed when not interpreted by a shell.
+                command = [
+                    '/bin/bash',
+                    '-c',
+                    '. ~/.bashrc && . /edx/app/analytics_pipeline/venvs/analytics_pipeline/bin/activate && {0} {1} --local-scheduler'.format(
```
brianhw: Is the path to the virtualenv a reasonable thing to hardcode here?   Or should this be passed into the local acceptance tests?

edx/analytics/tasks/tests/acceptance/services/vertica.py
```diff
@@ -69,5 +69,5 @@ def reset(self):
             return
 
         with self.cursor() as cur:
-            reset_query = 'DROP SCHEMA IF EXISTS {0} CASCADE; CREATE SCHEMA {0}'.format(self.schema_name)
+            reset_query = 'DROP SCHEMA IF EXISTS {0} CASCADE; CREATE SCHEMA {0}; SELECT MAKE_AHM_NOW();'.format(self.schema_name)
```
brianhw: nit: Can you add a comment that notes what MAKE_AHM_NOW() is for?   I'm guessing I'm not the only one to wonder.

edx/analytics/tasks/tests/acceptance/services/vertica.py
```diff
@@ -60,7 +60,7 @@ def connect(self):
         """
         Connect to the Vertica server.
         """
-        return vertica_python.connect(user=self.credentials.get('user'), password=self.credentials.get('password'),
+        return vertica_python.connect(user=self.credentials.get('username'), password=self.credentials.get('password'),
```
brianhw, line 63: Is there a reason why this hasn't been broken up until now?   Or is this a change that would also have to be made in our credentials files as well?

config/luigi_docker.cfg
```diff
@@ -0,0 +1,146 @@
...
+
+#[course-list]
+#api_root_url = http://localhost:8000/api/courses/v1/courses/
+
+#[course-blocks]
+#api_root_url = http://localhost:8000/api/courses/v1/blocks/
+
+[spark]
+master=spark://sparkmaster:7077
+edx_egg_files = ["/var/tmp/edx-ccx-keys-0.2.1/dist/edx_ccx_keys-0.2.1-py2.7.egg", "/var/tmp/edx-opaque-keys-0.4/dist/edx_opaque_keys-0.4-py2.7.egg"]
```
brianhw: Just out of curiosity, is the deep directory structure required here, or would it also work if these were copied in as ["/var/tmp/edx_egg_files/edx_ccx_keys.egg", "/var/tmp/edx_egg_files/edx_opaque_keys.egg"] ?   Then the config file would not need to be updated when version changes occur.

brianhw (commented):
It would be really nice if we didn't have to create so many almost-identical versions of the same config file, particularly docker-test.cfg and test.cfg.  I guess the problem is that we need to use the environment variable to get the config into the tests, rather than whatever mechanism we use for supporting multiple config files in launch-task.   Is that your take as well?

config/luigi_docker.cfg
rao-abdul-mannan: Right, I can link all egg files from 1 directory & specify in document to copy them there.

edx/analytics/tasks/tests/acceptance/services/vertica.py
rao-abdul-mannan: @brianhw Are you talking about production credentials ? They have username & not user so I used the same format for warehouse.json file.

config/luigi_docker.cfg
brianhw: Actually, this may be worth a comment here -- that the files need to be installed.

rao-abdul-mannan: I will document them with docker documentation.

edx/analytics/tasks/tests/acceptance/services/task.py
rao-abdul-mannan: Actually, I forgot to remove this  since I have updated the PATH variable in pipeline container so i'm going to remove this now.

rao-abdul-mannan:
@brianhw Do you have any suggestions to minimize the duplication b/w docker_test.cfg & test.cfg ?

edx/analytics/tasks/tests/acceptance/services/hive.py
rao-abdul-mannan: @brianhw Do you think I should refactor this code ?

edx/analytics/tasks/tests/acceptance/services/vertica.py
rao-abdul-mannan: Found the issue, reverting the changes.

edx/analytics/tasks/util/url.py
```diff
@@ -134,8 +134,8 @@ def get_target_class_from_url(url, marker=False):
     kwargs = {}
     if issubclass(target_class, HdfsTarget) and url.endswith('/'):
         kwargs['format'] = hdfs_format.PlainDir
-    if issubclass(target_class, luigi.LocalTarget) or parsed_url.scheme == 'hdfs':
-        # LocalTarget and HdfsTarget both expect paths without any scheme, netloc etc, just bare paths. So strip
```
rao-abdul-mannan, line 138: @brianhw I ran acceptance tests in docker devstack ( full suite ) and only few ( related to hadoop & hive ) in vagrant devstack and both worked perfectly fine. Just curious about the reason to place that check for hdfs targets?

edx/analytics/tasks/tests/acceptance/services/vertica.py
brianhw: I think this change may actually be worth making, as we noted in hipchat.  If we need to load a credentials file into place (or otherwise point to one) to make the tests work, it would be great if we didn't need to alter the credentials file to copy the username to user.    So we can fix this at the other end without a lot of heartache, and update the calls to our Jenkins acceptance tests if necessary, that seems like progress.

edx/analytics/tasks/util/url.py
brianhw: I'll poke around to see if I can reconstruct this.   Out of curiosity, have you tried running the branch on Jenkins acceptance tests?

rao-abdul-mannan: @brianhw not yet, I did run full acceptance test suite on docker devstack without errors.

edx/analytics/tasks/tests/acceptance/services/hive.py
brianhw: I guess not.   The quoting stuff can be a real headache.

edx/analytics/tasks/util/url.py
brianhw: Sure.  I'm just concerned that EMR might behave differently.

brianhw: This code was added by Gabe back in May, 2014, back before the analytics-tasks repository was open-sourced as edx-analytics-pipeline.  I don't see anything indicating why it was added, beyond the comment in the code, and I don't see anything remarkably different about the Luigi implementation of HdfsTarget between 1.0.17 and the current 2.7 version we are now using to explain a difference in behavior.  So if testing doesn't identify any issues, this seems like a good change to make.

brianhw:
I don't really have any suggestions for refactoring test.cfg and test_docker.cfg.   We would need some way of specifying multiple configs for that to work.   Launch-task provides such a mechanism.   If config files supported include statements, that would be another way.   But this is presumably using an environment variable to interface with Luigi, so presumably that doesn't work.  Unless I'm missing something?

edx/analytics/tasks/tests/acceptance/services/shell.py
```diff
@@ -14,7 +14,7 @@ def run(command):
     buf = []
 
     # Execute the process gathering the output from stdout
-    proc = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+    proc = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, env=env)
```
HassanJaveed84: This would overwrite the environment variables for the new process(defaults to None) instead of inheriting from the current one. 
I don't think we want to do that here as when we run the acceptance tests on Jenkins ? I have a feeling that this might be the change breaking acceptance tests on your branch.

rao-abdul-mannan: @HassanJaveed84 I'll try without environment overwrites and see how it goes.

rao-abdul-mannan: It is overwriting the environment variables for remote tasks.

azafty468 (approved):
I don't see anything that wasn't already mentioned.  I'm a bit curious how we'll set this up with a Jenkins CI job but that question doesn't need to be answered until we get to the Jenkins orchestration step.  Good work!

edx/analytics/tasks/util/url.py
brianhw: @mulby This is a shot in the dark, but do you happen to have any memory of why this special treatment of HDFS was necessary?

edx/analytics/tasks/tests/acceptance/services/vertica.py
rao-abdul-mannan: @brianhw In vertica, an epoch ( a logical time stamp for the data in Vertica ) is assigned to every row, which gets updated whenever data is committed via DML operation. Vertica advances the AHM at an interval of 5mins, but when there is a node down in vertica cluster, AHM does not advance. 

This function advances the epoch to the greatest allowable value, and lets you drop any projections that existed before the issue occurred, all history is lost and no historical queries cannot be made before current epoch.

brianhw (approved):
LGTM, given that it seems to work.   I would have liked to have understood why we used to strip the hdfs scheme from urls, but if it works when we don't do that, then it's a good thing.

👍

edx/analytics/tasks/util/url.py
mulby: heh - yeah I don't remember why this was there... sorry!

PR #500, opened 2018-05-08
Use analytics pipeline official docker for travis tests

rao-abdul-mannan:
This PR uses analytics's official docker for travis tests and removes the Dockerfile which was used earlier to run those tests.


brianhw:
I'm sorry to have not asked this sooner, but I don't understand how our various branches get tested once this change is made.   This Travis job is currently used to run unit tests on branches as commits are made to the branch.  It seems with this change, the Travis job will just run whatever was the last thing built and pushed to the docker repo.  Why do we want to do that?   What

rao-abdul-mannan:
@brianhw With this change it won't effect a thing, as we mount the volume of the branch ( against which travis build is going to run ) to that docker container when this build is triggered. Using the official docker or the old dockerfile does not effect the build process. 
Docker container in this scenario just ensures that we have proper python packages installed & is able to run the unit tests, code style checking etc.

brianhw:
Thanks for the explanation, but I'm still not understanding the boundaries of what is part of the test.   The tests will run "make test-docker" and such against the specified branch, using the released docker image as the starting point.   And I interpreted what you wrote above to mean that the "proper" python packages are already installed (at least those that are proper for the branch used to build the docker container).  

Since test-docker will still call "make develop-local" and "make test-local", it should also update itself with whatever dependencies have been *changed* with regard to development- and testing-related requirements files.   But it will *not* uninstall packages that were included in the docker image that were later removed in the local branch.  I think that's a rare occurrence, and one that we can live with, but I just want to make sure that I understand how this is working.

rao-abdul-mannan:
Yes that is correct and though its a rare occurrence, if we rerun test-requirements it will pick up those new packages and this case can be handled ,right ? Thoughts ?

macdiesel:
I do think we should rerun test-requirements. 

Brian you are correct that if a package is removed it would still show up in the image until a new one is built. We also frequently run in to this on the edx platform as well and there is a paver command to clear out these old dependencies. 

That said new docker builds of the image will not have these removed dependencies.

brianhw:
Oops, yes.  I definitely overlooked that.   I saw develop-local and test-local, and forgot that they do *not* update dependencies, only the pipeline code.  So yes, we should add 'test-requirements' somewhere.   But I'm thinking it should it go into something separate in .travis.yml, rather than being added to the test-docker target.  Either that, or we rename the current test-docker to be test-docker-local.   I think the functionality is still useful to the developer who is fixing tests, and needs to rerun them.  (Which is why I added the -local targets in the first place.)

rao-abdul-mannan:
@brianhw @macdiesel I have made the changes and now it will reset the virtualenv first, re-installs requirements for test from current branch and then executes the tests. 
Resetting the virtualenv is faster then uninstalling existing packages. Hope so this change covers the case scenarios that you guyz have mentioned.

Makefile
```diff
@@ -49,7 +53,10 @@ upgrade: ## update the requirements/*.txt files with the latest packages satisfy
 	echo "-r extra.txt" >> requirements/docs.txt
 
 test-docker:
-	docker run -v `(pwd)`:/edx/app/analytics-pipeline -it edxops/analytics-pipeline:latest make develop-local test-local
+	docker run -u root -v `(pwd)`:/edx/app/analytics_pipeline/analytics_pipeline -it edxops/analytics_pipeline:latest make develop-local test-local
+
+test-docker-local:
+	docker run -u root -v `(pwd)`:/edx/app/analytics_pipeline/analytics_pipeline -it edxops/analytics_pipeline:latest make reset-virtualenv test-requirements develop-local test-local
```
brianhw: "Local" should mean use the local requirements, so these targets are backwards.   This should be "test-docker" and the above one (that is currently test-docker) should be "test-docker-local".

.travis.yml
```diff
@@ -19,30 +19,25 @@ env:
...
-  - make docker-build
+  # Pull the docker image
+  - make docker-pull
 
   # Ensure we have a place to store coverage output
   - mkdir -p coverage
 
 script:
-  - make test-docker
+  - make test-docker-local
```
brianhw: And this should switch back to using test-docker....

brianhw (approved):
LGTM.   👍

PR #501, opened 2018-05-08
WIP: ENT-880 Add course key to enterprise enrollment task

mattdrayer:
@georgebabey here's the analytics task to add the course key field to the necessary vertica table.

### DIFF
--- a/edx/analytics/tasks/enterprise/enterprise_enrollments.py
+++ b/edx/analytics/tasks/enterprise/enterprise_enrollments.py
@@ -49,6 +49,7 @@ class EnterpriseEnrollmentRecord(Record):
     user_account_creation_timestamp = DateTimeField(description='')
     user_email = StringField(length=255, description='')
     user_username = StringField(length=255, description='')
+    course_key = StringField(length=255, description='')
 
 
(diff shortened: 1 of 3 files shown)


brianhw:
LGTM.   Two small comments.   First, this is not currently for a Vertica table, but rather for a Mysql table.    Second, you may find that you need to drop the table before running the job.  We don't have migrations.  I *think* that it's enough to run the job with --overwrite-mysql, but I'm not sure if that deletes the table (and recreates it) or if it merely removes its contents before reloading.

brianhw:
Yes, --overwrite-mysql only does a DELETE FROM table_name, not a DROP TABLE/CREATE TABLE.

georgebabey (approved):
This looks good to me, but I'm not terribly familiar with the pipeline side of the analytics-api. @brittneyexline is our resident expert, she have more insight (see what i did there..) into the needed changes

mattdrayer:
@brittneyexline thoughts?

brittneyexline (approved):
looks good to me!

mattdrayer:
Thanks!  Merging now 😄

PR #504, opened 2018-05-14
ENT-978 Add user country code and last activity date in enterprise en…

zubair-arbi:
…rollment hive table

* Add fields `user_country_code` and `last_activity_date` in `EnterpriseEnrollmentRecord` table.
* Update `insert_query`to populate above fields during `EnterpriseEnrollmentDataTask`

Remaining Task:
* Cannot add/populate field `country_name` as we are using django-countries which provide method for getting country name against the provided country code. We cannot do this with such ease in SQL. One way we can try is using manual dump for countries along with code and name.

### DIFF
--- a/edx/analytics/tasks/enterprise/enterprise_enrollments.py
+++ b/edx/analytics/tasks/enterprise/enterprise_enrollments.py
@@ -11,12 +11,13 @@ from edx.analytics.tasks.enterprise.enterprise_database_imports import (
     ImportEnterpriseCustomerUserTask, ImportUserSocialAuthTask
 )
 from edx.analytics.tasks.insights.database_imports import (
-    ImportAuthUserTask, ImportPersistentCourseGradeTask, ImportStudentCourseEnrollmentTask
+    ImportAuthUserProfileTask, ImportAuthUserTask, ImportPersistentCourseGradeTask, ImportStudentCourseEnrollmentTask
 )
 from edx.analytics.tasks.insights.enrollments import OverwriteHiveAndMysqlDownstreamMixin
+from edx.analytics.tasks.insights.user_activity import UserActivityTableTask
 from edx.analytics.tasks.util.decorators import workflow_entry_point
 from edx.analytics.tasks.util.hive import BareHiveTableTask, HivePartitionTask, OverwriteAwareHiveQueryDataTask
-from edx.analytics.tasks.util.record import BooleanField, DateTimeField, IntegerField, Record, StringField
+from edx.analytics.tasks.util.record import BooleanField, DateField, DateTimeField, IntegerField, Record, StringField
 from edx.analytics.tasks.warehouse.load_internal_reporting_course_catalog import (
     CoursePartitionTask, LoadInternalReportingCourseCatalogMixin
 )
@@ -50,6 +51,8 @@ class EnterpriseEnrollmentRecord(Record):
     user_email = StringField(length=255, description='')
     user_username = StringField(length=255, description='')
     course_key = StringField(length=255, description='')
+    user_country_code = StringField(length=2, description='')
+    last_activity_date = DateField(description='')
 
 
 class EnterpriseEnrollmentHiveTableTask(BareHiveTableTask):
@@ -130,7 +133,9 @@ class EnterpriseEnrollmentDataTask(
                     auth_user.date_joined AS user_account_creation_timestamp,
                     auth_user.email AS user_email,
                     auth_user.username AS user_username,
-                    course.catalog_course AS course_key
+                    course.catalog_course AS course_key,
+                    user_profile.country AS user_country_code,
+                    user_activity.latest_date AS last_activity_date
             FROM enterprise_enterprisecourseenrollment enterprise_course_enrollment
             JOIN enterprise_enterprisecustomeruser enterprise_user
                     ON enterprise_course_enrollment.enterprise_customer_user_id = enterprise_user.id
@@ -141,6 +146,21 @@ class EnterpriseEnrollmentDataTask(
                     AND enterprise_user.user_id = enrollment.user_id
             JOIN auth_user auth_user
                     ON enterprise_user.user_id = auth_user.id
+            JOIN auth_userprofile user_profile
+                    ON enterprise_user.user_id = user_profile.user_id
+            LEFT JOIN (
+                    SELECT
+                        user_id,
+                        course_id,
+                        MAX(`date`) AS latest_date
+                    FROM
+                        user_activity_by_user
+                    GROUP BY
+                        user_id,
+                        course_id
+                ) user_activity
+                    ON enterprise_course_enrollment.course_id = user_activity.course_id
+                    AND enterprise_user.user_id = user_activity.user_id
             LEFT JOIN consent_datasharingconsent consent
                     ON auth_user.username =  consent.username
                     AND enterprise_course_enrollment.course_id = consent.course_id
@@ -180,6 +200,7 @@ class EnterpriseEnrollmentDataTask(
         # the process that generates the source table used by this query
         yield (
             ImportAuthUserTask(),
+            ImportAuthUserProfileTask(),
             ImportEnterpriseCustomerTask(),
             ImportEnterpriseCustomerUserTask(),
             ImportEnterpriseCourseEnrollmentUserTask(),
@@ -192,6 +213,11 @@ class EnterpriseEnrollmentDataTask(
                 warehouse_path=self.warehouse_path,
                 api_root_url=self.api_root_url,
                 api_page_size=self.api_page_size,
+            ),
+            UserActivityTableTask(
+                warehouse_path=self.warehouse_path,
+                overwrite_n_days=0,
+                date=self.date
             ),
         )
 
--- a/edx/analytics/tasks/tests/acceptance/fixtures/input/enterprise/auth_userprofile.sql
+++ b/edx/analytics/tasks/tests/acceptance/fixtures/input/enterprise/auth_userprofile.sql
@@ -1,0 +1,43 @@
+--
+-- Table structure for table `auth_user`
+--
+
+DROP TABLE IF EXISTS `auth_userprofile`;
+CREATE TABLE `auth_userprofile` (
+  `id` int(11) NOT NULL AUTO_INCREMENT,
+  `name` varchar(255) NOT NULL,
+  `meta` longtext NOT NULL,
+  `courseware` varchar(255) NOT NULL,
+  `language` varchar(255) NOT NULL,
+  `location` varchar(255) NOT NULL,
+  `year_of_birth` int(11) DEFAULT NULL,
+  `gender` varchar(6) DEFAULT NULL,
+  `level_of_education` varchar(6) DEFAULT NULL,
+  `mailing_address` longtext,
+  `city` longtext,
+  `country` varchar(2) DEFAULT NULL,
+  `goals` longtext,
+  `allow_certificate` tinyint(1) NOT NULL,
+  `bio` varchar(3000) DEFAULT NULL,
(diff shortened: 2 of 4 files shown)


zubair-arbi:
@brittneyexline @georgebabey Please let me know if this is what we wanted for ENT-978. I am unable to get user country name which I mentioned in the PR description.

edx/analytics/tasks/enterprise/enterprise_enrollments.py
```diff
@@ -127,10 +129,12 @@ def insert_query(self):
                     END AS course_duration_weeks,
                     course.min_effort AS course_min_effort,
                     course.max_effort AS course_max_effort,
+                    course.catalog_course AS course_key,
```
georgebabey: Order matters here so be careful moving course_key. This order here needs to match the order in the fixture data here or else the acceptance tests will fail: edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py

See this PR from @mattdrayer https://github.com/edx/edx-analytics-pipeline/commit/16780bdba521035326d492d29cd156a8e843f0e9

georgebabey (changes requested):
You'll also need to update the acceptance tests to include these new fields fields, or they will also fail:
edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py

georgebabey:
Just realized we should call this `last_activity_timestamp` to align with the other datetime fields: `enrollment_created_timestamp`,  `passed_timestamp`, `user_account_creation_timestamp`

brianhw:
A few observations:
* You have added references to two additional tables in your Hive query, but you don't do what is needed to load that data into Hive.   In particular, the EnterpriseEnrollmentDataTask.requires() method should include tasks that make this happen.  That includes ImportAuthUserProfileTask() (to load auth_userprofile into Hive), and something to load user_activity_by_user into Hive.  
* The loading of user_activity_by_user into Hive is a little subtle.   You would use UserActivityTableTask (as defined in insights/user_activity.py), but you must pass in a parameter value for overwrite_n_days of zero, so that there is no attempt to recalculate any data.  This would simply use whatever data is currently available on the machine.  
* Your join "ON enterprise_user.user_id = user_profile.id" should be "ON enterprise_user.user_id = user_profile.user_id" instead? 
* In testing, you need to define user_activity_by_user fixture differently.   The data is not read from an LMS database, so there is no need to define such a table.  You have to write out files to the right warehouse_path location, since it is an intermediate output of a different pipeline job.   An example of this is CourseBlocksPartitionTaskAcceptanceTest.copy_hive_input() in test_course_blocks.py, though that's just the first I found.    You may want activity for more than one day.
* You can run acceptance tests on the test_ci server, using the ad-hoc manual job and putting in your branch and the name of the test (i.e. test_enterprise_enrollments).  Let me know if you need permission and don't have it.   Brittney and George do as well.

georgebabey (approved):
Updates look good but let me coordinate with @brianhw  to determine the best process for rolling these pipeline changes out before merging

brianhw:
I'm glad you tried to run the acceptance test.  It looks like it failed because of an extra character in one of your fixture files.  

Also, the Travis builds are failing due to an issue with ordering of includes:  
```
isort --check-only --recursive edx/
ERROR: /edx/app/analytics_pipeline/analytics_pipeline/edx/analytics/tasks/enterprise/enterprise_enrollments.py Imports are incorrectly sorted. 
```

zubair-arbi:
@brianhw Please have another look on the changes now.
Can you please tell why acceptance tests on jenkins failed: http://jenkins-ci.analytics.edx.org/view/ad-hoc/job/edx-analytics-pipeline-acceptance-test-manual/1566/console

edx/analytics/tasks/tests/acceptance/fixtures/input/enterprise/auth_userprofile.sql
```diff
@@ -0,0 +1,43 @@
...
+
+
+--
+-- Dumping data for table `auth_userprofile`
+--
+INSERT INTO `auth_userprofile` VALUES
+    ('5', 'Test User', '', 'course.xml', '', '', '1989', 'm', '', '', '', 'US', '', '1', NULL, NULL, '11'),
+    ('6', 'Test User2', '', 'course.xml', '', '', '1989', 'f', '', '', '', 'US', '', '1', NULL, NULL, '12'),
+    ('7', 'Test User3', '', 'course.xml', '', '', '1989', 'm', '', '', '', 'US', '', '1', NULL, NULL, '13'),
+    ('8', 'Test User4', '', 'course.xml', '', '', '1989', 'f', '', '', '', 'US', '', '1', NULL, NULL, '14'),
```
brianhw: So the problem with acceptance tests seems to be with loading this file.   At the bottom of the log, there is:
```
11:10:35 edx.analytics.tasks.tests.acceptance: DEBUG: Executing SQL fixture /var/lib/jenkins/jobs/edx-analytics-pipeline-acceptance-test-manual/workspace/analytics-tasks/edx/analytics/tasks/tests/acceptance/fixtures/input/enterprise/auth_userprofile.sql on acceptance_import_test_7d1cbdd16cac88a0b607536ba10f891d
```
And under the traceback, "11:10:35 ProgrammingError: 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '' at line 8".  

This is the only thing that really looks wrong.  It should end with a semi-colon, not a comma.

zubair-arbi:
@brianhw Thanks for pointing to syntax issue in fixture file. It look like `EnterpriseEnrollmentDataTask` failed now.
**Console log:** http://jenkins-ci.analytics.edx.org/view/ad-hoc/job/edx-analytics-pipeline-acceptance-test-manual/1567/console

rao-abdul-mannan:
@zubair-arbi Error seems to be 
`Failed to recognize predicate 'date'. Failed rule: 'identifier' in table or column identifier`
in the EnterpriseEnrollmentDataTask, which could be due to no whitespace being used in table name and alias `social_auth_usersocialauth`
https://github.com/edx/edx-analytics-pipeline/blob/3f3de316aae718348fc991343805dbf7c7c795c3/edx/analytics/tasks/enterprise/enterprise_enrollments.py#L176

rao-abdul-mannan:
Sorry about my earlier comment. 
So I have found 2 issues that **could** be the failure reason.
-  Due to hive reserved keywords, table & column names in the query should be wrapped in back ticks ( ` )

    e.g. date should be `date` in subquery
- `last_activity_timestamp` is a timestamp field and query is writing Date to it.

edx/analytics/tasks/enterprise/enterprise_enrollments.py
```diff
@@ -50,6 +51,8 @@ class EnterpriseEnrollmentRecord(Record):
     user_email = StringField(length=255, description='')
     user_username = StringField(length=255, description='')
     course_key = StringField(length=255, description='')
+    user_country_code = StringField(length=2, description='')
+    last_activity_timestamp = DateTimeField(description='')
```
brianhw: I think that if you are storing values like "2018-05-18", you may need to use a DateField here instead of a DateTimeField.    I don't know if that would involve changing the name of the field as well.   Or if we need to find a different way of getting the actual timestamp of the last activity, instead of merely the date.

edx/analytics/tasks/enterprise/enterprise_enrollments.py
```diff
@@ -141,6 +146,21 @@ def insert_query(self):
                     AND enterprise_user.user_id = enrollment.user_id
             JOIN auth_user auth_user
                     ON enterprise_user.user_id = auth_user.id
+            JOIN auth_userprofile user_profile
+                    ON enterprise_user.user_id = user_profile.user_id
+            LEFT JOIN (
+                    SELECT
+                        user_id,
+                        course_id,
+                        MAX(date) AS latest_date
```
brianhw: This may be the line that requires date to be back-quoted.

brianhw (commented):
Following up on Abdul's comments, here are some more specific locations and fixes.

brianhw:
Looks like it needs a rebase....

edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py
```diff
@@ -107,7 +114,8 @@ def validate_enterprise_enrollment_table(self):
                    'enrollment_created_timestamp', 'user_current_enrollment_mode', 'consent_granted', 'letter_grade',
                    'has_passed', 'passed_timestamp', 'enterprise_sso_uid', 'enterprise_site_id', 'course_title',
                    'course_start', 'course_end', 'course_pacing_type', 'course_duration_weeks', 'course_min_effort',
-                   'course_max_effort', 'user_account_creation_timestamp', 'user_email', 'user_username', 'course_key']
+                   'course_max_effort', 'user_account_creation_timestamp', 'user_email', 'user_username', 'course_key',
+                   'user_country_code', 'last_activity_timestamp']
```
brianhw: You probably need to update the field name here as well.

edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py
```diff
@@ -40,6 +40,10 @@ def setUp(self):
             os.path.join(self.data_dir, 'input', 'programs.json'),
             url_path_join(self.warehouse_path, 'discovery_api_raw', 'dt=' + self.DATE, 'programs.json')
         )
+        self.upload_file(
+            os.path.join(self.data_dir, 'input', 'internal_reporting_user_activity'),
```
brianhw: You are using a different file for user_activity here.   This has a different set of users than you have defined in auth_userprofile.  You define a user_activity_by_user fixture file which uses matching user_ids.  You could point to that here.

zubair-arbi: @brianhw I have verified that users defined in `user_activity_by_user` fixture file matches in `user_ids` and also the `upload_file` path matches the hive path `edx.analytics.tasks.insights.user_activity.UserActivityTask#output_path_for_key`. Still `last_activity_date` values are empty.

brianhw: RIght.   But this line uploads 'internal_reporting_user_activity' to the warehouse_path/'user_activity_by_user' location, rather than the 'user_activity_by_user' that you defined earlier.   You need to change this line to os.path.join(self.data_dir, 'input', 'user_activity_by_user'),

zubair-arbi:
@brianhw @rao-abdul-mannan I have updated the file path according to your suggestions now acceptance tests are passing with success http://jenkins-ci.analytics.edx.org/view/ad-hoc/job/edx-analytics-pipeline-acceptance-test-manual/1592/console
I have rebased and squashed the commits. Please have a final look on the changes.

brianhw (approved):
If the acceptance tests pass now, then 👍 .

Keep in mind, however, that there are two issues with merging to master.   One is that there are two PRs, and that only one can go to master.  The second one to merge will have to be rebased (with merge conflicts resolved), and re-tested.   The other is that the main job runs off of master, so it would break.   I'll suggest to Enterprise folks that they run off of our release instead of master.

brianhw:
I've changed the "enterprise-enrollment" job to run off of origin/release instead of origin/master, so there is less risk to production to have this merge to master.   Same with #512 .

zubair-arbi:
@brianhw the acceptance tests again passed with success after rebase http://jenkins-ci.analytics.edx.org/view/ad-hoc/job/edx-analytics-pipeline-acceptance-test-manual/1594/console

zubair-arbi:
@saleem-latif I have merged this branch please rebase your branch (https://github.com/edx/edx-analytics-pipeline/pull/512) now.

PR #505, opened 2018-05-21
Add docker analyticstack guides

rao-abdul-mannan:
This PR adds supporting documents for docker analyticstack.

Relevant [devstack PR](https://github.com/edx/devstack/pull/308).

### DIFF
--- a/docs/source/running_acceptance_tests_in_docker.rst
+++ b/docs/source/running_acceptance_tests_in_docker.rst
@@ -1,0 +1,45 @@
+..  _running_acceptance_tests_in_docker:
+
+Running Acceptance Tests in Docker Analyticstack
+================================================
+
+For docker analyticstack setup, follow instructions under **Getting Started on Analytics** section in `Devstack`_ repository.
+
+Pre-requisites
+--------------
+
+* Access analytics pipeline shell:
+
+   ::
+
+       make analytics-pipeline-shell
+
+* Before running the user-location workflows, a geolocation Maxmind data file must be downloaded. This file can be in
+  HDFS or S3, for example, and should be pointed to by the `geolocation_data` setting in the `geolocation` section of your
+  configuration file. To use the default location used by acceptance tests, execute the following:
+
+   ::
+
+       curl -fSL http://geolite.maxmind.com/download/geoip/database/GeoLiteCountry/GeoIP.dat.gz -o /var/tmp/GeoIP.dat.gz
+       cd /var/tmp/ && gunzip /var/tmp/GeoIP.dat.gz
+       mv GeoIP.dat geo.dat
+       hdfs dfs -put geo.dat /edx-analytics-pipeline/
+
+Running Acceptance Tests
+------------------------
+
+* To run the full test suite, execute the following command in the shell:
+
+  ::
+
+      make docker-test-acceptance-local-all
+
+* To run individual tests, execute the following:
+
+  ::
+
+      make docker-test-acceptance-local ONLY_TESTS=edx.analytics.tasks.tests.acceptance.<test_script_file>    # e.g.
+      make docker-test-acceptance-local ONLY_TESTS=edx.analytics.tasks.tests.acceptance.test_user_activity
+
+
+.. _Devstack: https://github.com/edx/devstack/tree/master/README.rst#getting-started-on-analytics
--- a/docs/source/running_spark_tasks.rst
+++ b/docs/source/running_spark_tasks.rst
@@ -1,0 +1,31 @@
+..  _running_spark_tasks:
+
+Running Spark Tasks in Docker Analyticstack
+===========================================
+
+Pre-requisites
+--------------
+
+* Access analytics pipeline shell:
+
+   ::
+
+       make analytics-pipeline-shell
+
+* Generate egg files
+
+  If you plan to run Spark workflows that use imports that in turn require the use of a plugin mechanism,
+  it is necessary to store those imports locally as egg files. These imports are then identified in the
+  configuration file in the `spark` section. Opaque keys is one of these imports, and the two egg files
+  used by Spark can be made as follows.
+
+   ::
+
+       make generate-spark-egg-files
+
+Task
+~~~~
+
+::
+
+    launch-task UserActivityTaskSpark --local-scheduler --interval 2017-03-16
--- a/docs/source/troubleshooting_docker_analyticstack.rst
+++ b/docs/source/troubleshooting_docker_analyticstack.rst
@@ -1,0 +1,26 @@
+..  _troubleshooting_docker_analyticstack:
+
+Troubleshooting Docker Analyticstack
+====================================
+
+* Make sure there are no errors during provision command. If there are errors, **do not rerun the provision command without first cleaning up after the failures.**
+* For cleanup, there are 2 options.
+
+  - Reset containers ( this will remove all containers and volumes )
+
+     ::
+
+         make destroy
+
+  - Manual cleanup
+
+     ::
+
+         make mysql-shell
+         mysql
+         DROP DATABASE reports
+         DROP DATABASE edx_hive_metastore
+         DROP DATABASE edxapp           # Only drop if provisioning failed while loading the LMS schema.
+         DROP DATABASE edxapp_csmh      # Only drop if provisioning failed while loading the LMS schema.
+         # exit mysql shell
+         make down
--- a/Makefile
+++ b/Makefile
@@ -120,3 +120,17 @@ docker-test-acceptance-local:
 
 docker-test-acceptance-local-all:
 	LAUNCH_TASK=$(shell which launch-task) REMOTE_TASK=$(shell which remote-task) LUIGI_CONFIG_PATH='config/docker_test.cfg' ACCEPTANCE_TEST_CONFIG="/edx/etc/edx-analytics-pipeline/acceptance.json" python -m coverage run --rcfile=./.coveragerc -m nose --nocapture --with-xunit -A acceptance -v
+
+generate-spark-egg-files:
+	# edx-opaque-keys egg
+	mkdir -p /var/tmp/edx_egg_files/
+	curl -fSL https://files.pythonhosted.org/packages/63/86/d4bf9c7e7a720125b0572c43dda002f72be2cc66313be601cd7b6cccb2ad/edx-opaque-keys-0.4.tar.gz -o /var/tmp/edx-opaque-keys.tar.gz
+	cd /var/tmp/ && mkdir -p /var/tmp/edx-opaque-keys/ && tar --strip-components=1 -xzf /var/tmp/edx-opaque-keys.tar.gz -C /var/tmp/edx-opaque-keys/
+	cd /var/tmp/edx-opaque-keys && python setup.py bdist_egg
+	cp /var/tmp/edx-opaque-keys/dist/edx_opaque_keys-0.4-py2.7.egg /var/tmp/edx_egg_files/edx_opaque_keys.egg
+
+	# edx-ccx-keys agg
+	curl -fSL https://files.pythonhosted.org/packages/a4/03/444d30a3859e36ef2273ae6254c011e1fb3a02f4e3dba16008ecf0b4bdb3/edx-ccx-keys-0.2.1.tar.gz -o /var/tmp/edx-ccx-keys.tar.gz
+	cd /var/tmp/ && mkdir -p /var/tmp/edx-ccx-keys/  && tar --strip-components=1 -xzf /var/tmp/edx-ccx-keys.tar.gz -C /var/tmp/edx-ccx-keys/
+	cd /var/tmp/edx-ccx-keys && python setup.py bdist_egg
+	cp /var/tmp/edx-ccx-keys/dist/edx_ccx_keys-0.2.1-py2.7.egg  /var/tmp/edx_egg_files/edx_ccx_keys.egg
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -13,6 +13,9 @@ This guide provides information about the edX data pipeline.
 
    running_tasks.rst
    faq.rst	     
+   running_acceptance_tests_in_docker.rst
+   running_spark_tasks.rst
+   troubleshooting_docker_analyticstack.rst
 
 The following documentation is generated dynamically from comments in the code. There is detailed documentation for every Luigi task.
 


docs/source/running_acceptance_tests_in_docker.rst
```diff
@@ -0,0 +1,35 @@
+..  _running_acceptance_tests_in_docker:
+
+Running Acceptance Tests in Docker Analyticstack
+================================================
+
+Pre-requisite
+-------------
+
+* Access analytics pipeline shell and run following commands.
```
macdiesel: end line with a colon :

docs/source/running_acceptance_tests_in_docker.rst
```diff
@@ -0,0 +1,35 @@
...
+
+* Geolocation maxmind data file (country granularity) to be uploaded to HDFS or S3 (see the ``geolocation`` section of the config file).
```
macdiesel: This might fit better as a ** line so it's indented under the list of commands to run.

docs/source/running_spark_tasks.rst
```diff
@@ -0,0 +1,38 @@
...
+-------------
+
+* Access analytics pipeline shell and run following commands.
+* Generate egg files
+
+  - Edx Opaque Keys
+
+     ::
+
+         curl -fSL https://files.pythonhosted.org/packages/63/86/d4bf9c7e7a720125b0572c43dda002f72be2cc66313be601cd7b6cccb2ad/edx-opaque-keys-0.4.tar.gz -o /var/tmp/edx-opaque-keys-0.4.tar.gz
```
macdiesel: Can we do this during provisioning or image building? If not what about a make command?

docs/source/troubleshooting_docker_analyticstack.rst
```diff
@@ -0,0 +1,27 @@
...
+
+     ::
+
+         make destroy
+
+  - Manual cleanup
+
+     ::
+
+         make mysql-shell
```
macdiesel, line 19: This should definitely be a make target.

macdiesel (commented):
A few questions/suggestions.

docs/source/running_spark_tasks.rst
rao-abdul-mannan: @macdiesel I remember discussing this in a scrum meeting when @brianhw suggested that we can document them ( at least for now ). Making it a make rule, will require making changes to that rule anytime this download link expires/changes or the package version changes.

docs/source/troubleshooting_docker_analyticstack.rst
rao-abdul-mannan: @macdiesel Some of the commands listed in here are optional which needs to be run if issue is related to them ( like dropping lms database, which makes sense if the issue occurred when provisioning lms ). 
I don't feel confident making it a make target yet.

macdiesel (approved):

docs/source/running_acceptance_tests_in_docker.rst
```diff
@@ -0,0 +1,34 @@
+..  _running_acceptance_tests_in_docker:
+
+Running Acceptance Tests in Docker Analyticstack
+================================================
+
+Pre-requisite
+-------------
+
+* Access analytics pipeline shell and run following commands:
```
brianhw: As mentioned below, the prerequisite here is probably just to have the shell.   Maybe mention how to get it:  (i.e. "make analytics_pipeline_shell" or whatever it is...)

docs/source/running_acceptance_tests_in_docker.rst
```diff
@@ -0,0 +1,34 @@
...
+
+         curl -fSL http://geolite.maxmind.com/download/geoip/database/GeoLiteCountry/GeoIP.dat.gz -o /var/tmp/GeoIP.dat.gz
+         cd /var/tmp/ && gunzip /var/tmp/GeoIP.dat.gz
+         mv GeoIP.dat geo.dat
+         hdfs dfs -put geo.dat /edx-analytics-pipeline/
+
+Running Acceptance Tests
+------------------------
+
+* Running full test suite will take several hours. To run full test suite, run the following command:
```
brianhw: nit:   "Running the full"....  "To run the full test suite, execute the following command in the shell:"

docs/source/running_acceptance_tests_in_docker.rst
```diff
@@ -0,0 +1,34 @@
...
+
+Running Acceptance Tests in Docker Analyticstack
+================================================
+
+Pre-requisite
+-------------
+
+* Access analytics pipeline shell and run following commands:
+
+  - Geolocation maxmind data file (country granularity) to be uploaded to HDFS or S3 (see the ``geolocation`` section of the config file). Getting a data file could look like this:
```
brianhw: In terms of the phrasing itself, I would suggest giving a little more context.  Something like:  "Before running the user-location workflows, a geolocation Maxmind data file must be downloaded.  This file can be in HDFS or S3, for example, and should be pointed to by the xxx setting in the ``geolocation`` section of your configuration file.   To use the default location used by acceptance tests, execute the following:"   I find "country granularity" to be confusing, so I'm guessing others might as well.

That said, I had a few larger questions:
* Have we figured out how to use S3 with Docker?  I thought there were issues, or at the very least more setup work that was necessary.  
* I'm a little confused by the flow of this section.   The main heading is "Pre-requisite" (only one), which has one main bullet (access a shell), and then this as a sub-bullet.  Is the opening of a shell the prerequisite?   It's one of them, since it's needed for running make below.  So I would probably make it "Pre-requisites", and move the geolocation setup to be a second prerequisite after the open the shell.

docs/source/running_acceptance_tests_in_docker.rst
```diff
@@ -0,0 +1,34 @@
...
+Running Acceptance Tests
+------------------------
+
+* Running full test suite will take several hours. To run full test suite, run the following command:
+
+  ::
+
+      make docker-test-acceptance-local-all
+
+* To run individual tests, run the following command:
```
brianhw: nit: just for variety:   "execute the following"....

docs/source/running_spark_tasks.rst
```diff
@@ -0,0 +1,38 @@
+..  _running_spark_tasks:
+
+Running Spark Tasks in Docker Analyticstack
+===========================================
+
+Pre-requisite
```
brianhw: Plural

docs/source/running_spark_tasks.rst
```diff
@@ -0,0 +1,38 @@
...
+-------------
+
+* Access analytics pipeline shell and run following commands.
```
brianhw: Again, I would leave off the "run the following commands" part here.

docs/source/running_spark_tasks.rst
```diff
@@ -0,0 +1,38 @@
...
+* Generate egg files
```
brianhw: I would give a little more explanation here.  Something like:  "In you plan to run Spark workflows that use imports that in turn require the use of a plugin mechanism, it is necessary to store those imports locally as egg files.  These imports are then identified in the configuration file in the ``whatever`` section.   Opaque keys is one of these imports, and the two egg files used by Spark can be made as follows."   I forget the name of the section and the parameter.  But it gives the reader a sense of what they need to do and why.

docs/source/running_spark_tasks.rst
brianhw: Interesting.   Yes, I definitely argued to have the final destination be more general:  e.g. "/var/tmp/edx_egg_files/edx_opaque_keys.egg".  My thinking was to try to limit the spread of version information too far across the system, so that if we change the version of these, we don't have to also update the configuration file.  

However, by putting the version information here, we are also spreading it out.   I might suggest that we find a way to limit the version information here as well.   For example, is it necessary to download to "/var/tmp/edx-opaque-keys-0.4.tar.gz", or can one download to "/var/tmp/edx-opaque-keys.tar.gz" instead?  I figure the version will still appear in the source location and probably in the untarred directory name.   But if we indicate that the first step is to identify the version number needed (by looking in the requirements/default.txt (or default.in) file, and then use '<version>' or something in the instructions below, then we avoid having to track future version changes here as well. 

What I don't know is if there is an easy way to explain to people how they can find the right tar file to download, given the package name and the version number.

docs/source/troubleshooting_docker_analyticstack.rst
```diff
@@ -0,0 +1,27 @@
+..  _troubleshooting_docker_analyticstack:
+
+Troubleshooting Docker Analyticstack
+====================================
+
+* Make sure there are no errors during provision command. If there are errors,
+  **do not rerun provision command before cleanup**.
+* For cleanup, there are 2 options.
+
+  - Reset containers ( this will remove all containers + volumes )
```
brianhw: nit: change the '+' to 'and'.

docs/source/troubleshooting_docker_analyticstack.rst
```diff
@@ -0,0 +1,27 @@
+..  _troubleshooting_docker_analyticstack:
+
+Troubleshooting Docker Analyticstack
+====================================
+
+* Make sure there are no errors during provision command. If there are errors,
+  **do not rerun provision command before cleanup**.
```
brianhw: I might suggest changing the word order a little:   "If there are errors, **do not rerun the provision command without first cleaning up after the failures**."

Just out of curiosity, are there a known list of ways in which provisioning might fail?   I'm wondering if that is worth trying to capture here.   We know that provisioning the LMS schema can fail, since it's mentioned below.

docs/source/troubleshooting_docker_analyticstack.rst
```diff
@@ -0,0 +1,27 @@
...
+  - Manual cleanup
+
+     ::
+
+         make mysql-shell
+         mysql
+         DROP DATABASE reports
+         DROP DATABASE edx_hive_metastore
+         DROP DATABASE edxapp           # only drop if provisioning command errored out while loading lms schema
+         DROP DATABASE edxapp_csmh      # only drop if provisioning command errored out while loading lms schema
```
brianhw: nit: I would change the comments to something like:  "# Only drop if provisioning failed while loading the LMS schema.

brianhw (commented):
This is a great addition!   Two things:
* I would also add these pages to the table of contents, so they can be found.  Look in docs/source/index.rst, and list these new files there along with the running_tasks.rst and faq.rst that are already there.
* You should be able to test this by running "make docs".  You can then inspect the resulting html, if it builds successfully.

docs/source/troubleshooting_docker_analyticstack.rst
rao-abdul-mannan: @brianhw The only issues I have seen is when LMS provisioning fails or something else stops the provisioning either due to internet timeout, unresponsiveness from docker etc.

docs/source/running_acceptance_tests_in_docker.rst
rao-abdul-mannan: @brianhw I couldn't yet test docker with s3.

docs/source/running_acceptance_tests_in_docker.rst
```diff
@@ -0,0 +1,40 @@
+..  _running_acceptance_tests_in_docker:
+
+Running Acceptance Tests in Docker Analyticstack
+================================================
+
+Pre-requisites
+--------------
+
+* Access analytics pipeline shell and run following commands:
```
brianhw: Remove "and run following commands"

docs/source/running_spark_tasks.rst
```diff
@@ -0,0 +1,57 @@
...
+
+* Access analytics pipeline shell:
+
+   ::
+
+       make analytics-pipeline-shell
+
+* Generate egg files
+
+  In you plan to run Spark workflows that use imports that in turn require the use of a plugin mechanism,
```
brianhw: s/In you/If you/

docs/source/running_spark_tasks.rst
```diff
@@ -0,0 +1,57 @@
...
+
+Task
+~~~~
+
+::
+
+    launch-task UserActivityTaskSpark --local-scheduler --interval 2017-03-16
+
+
+.. _pypi opaque keys: https://pypi.org/project/edx-opaque-keys
```
brianhw: More specifically https://pypi.org/project/edx-opaque-keys/<version>/#files.   Hmm.   No, that would break the linkability.  This is fine here.

docs/source/running_spark_tasks.rst
```diff
@@ -0,0 +1,57 @@
...
+     ::
+
+         curl -fSL https://files.pythonhosted.org/packages/63/86/d4bf9c7e7a720125b0572c43dda002f72be2cc66313be601cd7b6cccb2ad/edx-opaque-keys-0.4.tar.gz -o /var/tmp/edx-opaque-keys.tar.gz
+         cd /var/tmp/ && tar -xzf /var/tmp/edx-opaque-keys.tar.gz
+         cd /var/tmp/edx-opaque-keys
+         python setup.py bdist_egg
+         mkdir /var/tmp/edx_egg_files/
+         cp /var/tmp/edx-opaque-keys/dist/edx_opaque_keys-0.4-py2.7.egg /var/tmp/edx_egg_files/edx_opaque_keys.egg
+
+     **Note:** Download link can be copied from `pypi opaque keys`_
```
brianhw: Maybe add more instruction there:  "Choose the correct version, then select 'Download Files'.  " 

Then again, maybe this is all moot if this is moving into the Makefile.

docs/source/troubleshooting_docker_analyticstack.rst
brianhw: I'm fine with it being here and not in the makefile for now.

docs/source/running_acceptance_tests_in_docker.rst
```diff
@@ -0,0 +1,40 @@
+..  _running_acceptance_tests_in_docker:
+
+Running Acceptance Tests in Docker Analyticstack
+================================================
+
```
brianhw, line 5: I think it would be good to have a sentence or two here that points to the devstack repo (and its readme), so that people who find this documentation first know where to go to get docker analyticstack.  We should also state in the prerequisites below that the "docker analyticstack" has already been installed.  

Also, I'm wondering if we want to call it "Docker Analyticstack" here, if we give it a different name in the docker repo.  Or do we use that term there as well?

brianhw (commented):
Few more nits.

docs/source/running_acceptance_tests_in_docker.rst
brianhw: I should have just checked.  Yes, you call it analyticstack in the docker doc.   Cool.

rao-abdul-mannan: @brianhw We can mention that in README ?

brianhw: I don't understand.   Mention which, and in which README?

rao-abdul-mannan: i'll point devstack repo in this file.

brianhw:
LGTM. 👍   

We talked about getting the opaque key versions from the requirements file, but I think that can wait for another PR.

brianhw (approved):
LGTM 👍

PR #507, opened 2018-05-23
OEP implementation

HassanJaveed84:
Not sure how to deal with the 'mysql connector python' installation from a generic url in extra.txt, as pip compile does not support installations from generic urls.


scripts/post-pip-compile.sh
```diff
@@ -0,0 +1,35 @@
...
+
+function show_help {
+    echo "Usage: post-pip-compile.sh file ..."
+    echo "Replace the instructions for regenerating the given requirements file(s)."
+}
+
+function clean_file {
+    FILE_PATH=$1
+    TEMP_FILE=${FILE_PATH}.tmp
+	# If an editable VCS URL has a version number suffix, it was only editable for pip-compile's benefit;
```
HassanJaveed84: Initially added to remove -e flag from extra.txt, however it is not currently part of pip compile because of the generic url we use to install 'mysql connector python'

HassanJaveed84: Also, for our git based packages(luigi,pyinstrument) we don't tag it with a version.

azafty468 (approved):
LGTM.  Jeremy may have additional suggestions for the extra.txt requirement.

HassanJaveed84:
@jmbowman can you please review this when you get a chance ?

Makefile
```diff
@@ -34,23 +34,31 @@ else
 endif
 
 requirements:
-	pip install -U -r requirements/pre.txt
-	pip install -U -r requirements/default.txt --no-cache-dir --upgrade-strategy only-if-needed
-	pip install -U -r requirements/extra.txt --no-cache-dir --upgrade-strategy only-if-needed
+	pip install -r requirements/pip.txt
+	pip install -r requirements/default.txt --no-cache-dir
+	pip install -r requirements/extra.txt --no-cache-dir
```
brianhw, line 39: Interesting.   The -U arguments to "pip install" have been there since the beginning of the repo.     The upgrade-strategy was needed when the requirements files were more disjoint, but if we're including the upstream .txt files in the downstream .in files, then that's not an issue.

requirements/base.txt
```diff
@@ -2,24 +2,24 @@
 # This file is autogenerated by pip-compile
 # To update, run:
 #
-#    pip-compile --output-file requirements/base.txt requirements/base.in
+#    make upgrade
 #
 --no-binary ansible
```
jmbowman, line 7: Where did this option come from?  Do we still want it, and if so is it configured to be automatically included in the output whenever someone runs `make upgrade`?

scripts/post-pip-compile.sh
```diff
@@ -0,0 +1,35 @@
...
+	# this is a workaround for https://github.com/jazzband/pip-tools/issues/355
+    sed 's/-e \(.*==.*\)/\1/' ${FILE_PATH} > ${TEMP_FILE}
+    mv ${TEMP_FILE} ${FILE_PATH}
+    # Replace the instructions for regenerating the output file.
```
jmbowman: This morning I discovered a better way to do this: you can customize the command inserted into the output by setting the `CUSTOM_COMPILE_COMMAND` environment variable before running pip-compile: https://github.com/jazzband/pip-tools#configuration

scripts/post-pip-compile.sh
jmbowman: For `mysql-connector-python`, you could comment it out in the input file and include a section in `post-pip-compile.sh` that's something like this:

```bash
# Add back the mysql-connector-python package; this is a workaround for
# https://github.com/jazzband/pip-tools/issues/93
if [[ "${FILE_PATH}" == "requirements/extra.txt" ]]; then
    echo "http://cdn.mysql.com/Downloads/Connector-Python/mysql-connector-python-1.2.2.zip" >> ${FILE_PATH}
fi
```

I'd add an explanation above the commented-out line along the lines of

```
# pip-compile can't deal with this, it's added back by util/post-pip-compile.sh
```

requirements/base.txt
HassanJaveed84: base.in defines this here:
https://github.com/edx/edx-analytics-pipeline/blob/master/requirements/base.in#L3
I am not familiar with why we have it, maybe @mulby can shine some light on it.
It would automatically get included when running ```make upgrade```.

mulby: I don't remember why we have this in here. Sorry!

brianhw: I added this back in January, 2017 to master, and patched to Ficus branch in April, 2017.   This was a workaround for a "bug" in the old version of Ansible we're using.  The workaround was included in master along with a number of other changes, unfortunately.  The Ficus patch was more stand-alone:  https://github.com/edx/edx-analytics-pipeline/commit/60deb5360e3d1d1f693f800eca662da18ce4089e.   The comment:  "Addresses issue with ansible/ansible#8875, by pip installing with --no-binary flag."  So original issue is:  https://github.com/ansible/ansible/issues/8875.  If and when we upgrade ansible here, we can probably also remove this flag then.

requirements/pip-tools.txt
```diff
@@ -0,0 +1,10 @@
+#
+# This file is autogenerated by pip-compile
+# To update, run:
+#
+#    make upgrade
+#
+click==6.7                # via pip-tools
+first==2.0.1              # via pip-tools
+pip-tools==2.0.2
+six==1.11.0               # via pip-tools
```
brianhw: Note that this installs 1.11, and I *think* that if 1.11 is left in the virtualenv when trying to run the pipeline code that something breaks.  Hence all the hoops to use "--upgrade-strategy only-if-needed" to avoid this happening (before you took the different tack of removing the upgrade and including the .txt files instead of the .in files).  Perhaps this needs a comment in the pip-tools.in file?   Or maybe a suggestion somewhere that the "make upgrade" be done in a separate virtualenv?

requirements/base.txt
jmbowman: Thanks for the background.  Can we either try upgrading ansible or put that context in comments in `base.in`?

requirements/pip-tools.txt
jmbowman: Ok, then `six` should probably be pinned in `pip-tools.in`, with an explanation for why.

HassanJaveed84: Should we be pinning this to 1.10.0?

requirements/base.txt
brianhw: I agree that a comment at least would be useful.  I actually never tried to upgrade the ansible.  It might also not be too hard, but would involve more testing.  On the other hand, with the upgrades already happening, this should involve some thorough testing anyway.

Makefile
brianhw: So, what do we lose by removing the -U flags?  Is there some scenario where the packages would not get updated to the specified versions?   Perhaps it's a vestige of a time before all of our packages were pinned, and now it makes no sense anyway?

requirements/pip-tools.txt
brianhw: If pip-compile works with 1.10.0, then I would pin it with a comment.  But it would be better if it were possible to put six>=1.10.0 in the pip-tools.in file, and it came through in that form in the pip-tools.txt file.  Then it should the right thing if installed into a fresh virtualenv or into a requirements-loaded virtualenv.

Makefile
jmbowman: The latter.  It just told pip to upgrade to the latest available version satisfying the specified constraints.  Now that everything is pinned to exact versions, there shouldn't be any need for it.

scripts/post-pip-compile.sh
HassanJaveed84: @jmbowman Should we be tagging our git based installs or remove them from pip-compile ?
We have two forks specified in extra.in(luigi and pyinstrument), we do not have fixed versions for those on our forks and currently specify hash:
``git+https://github.com/edx/luigi.git@eb45bcc52243de11b2b16a81229ac584fe1e601b#egg=luigi``
``git+https://github.com/edx/pyinstrument.git@a35ff76df4c3d5ff9a2876d859303e33d895e78f#egg=pyinstrument``

Should extra.in be removed from pip-compile or should we add a version tag to the said repos and then specify it in extra.in ?

jmbowman: Ideally they should be tagged, even if just to ensure that the commit in use doesn't vanish later due to a force push or such.  But they should just work with pip-compile as is, given that they're installed as editable (the "-e" prefix).  We still have several dependencies like this in edx-platform also.

HassanJaveed84:
@brianhw do you think this looks good to go ?

Makefile
```diff
@@ -34,23 +34,34 @@ else
...
+	CUSTOM_COMPILE_COMMAND="make upgrade" pip-compile --upgrade -o requirements/extra.txt requirements/extra.in
 	echo "-r extra.txt" >> requirements/docs.txt
+	# Post process all of the files generated above to replace the instructions for recreating them
+	scripts/post-pip-compile.sh \
+		requirements/pip-tools.txt \
+		requirements/base.txt \
+		requirements/default.txt \
+		requirements/docs.txt \
+		requirements/test.txt \
+		requirements/extra.txt
```
brianhw: Is this still necessary?   I thought the purpose of the CUSTOM_COMPILE_COMMAND was to replace the instructions.    Oh, having looked, it also is used for the mysql dependency.   So perhaps the comment here needs to be updated.

requirements/extra.txt
```diff
@@ -1,12 +1,19 @@
...
+-e git+https://github.com/edx/pyinstrument.git@a35ff76df4c3d5ff9a2876d859303e33d895e78f#egg=pyinstrument
+backports-abc==0.5        # via tornado
+certifi==2018.4.16        # via tornado
+distlib==0.2.2
+docutils==0.14            # via python-daemon
+lockfile==0.12.2          # via python-daemon
 psycopg2==2.6.2
+python-daemon==2.1.2
+singledispatch==3.4.0.3   # via tornado
+six==1.11.0               # via singledispatch
```
brianhw: The motivation for having an extra.txt in the first place was for keeping psycopg2 out of docs.txt, but the file then got repurposed for things that didn't go through pip-compile.   If the intention is to put the git urls through pip-compile anyway, then perhaps they should be in something else, like default.in.   That way, you won't end up with a different set of dependencies here that are not constrained.   E.g. you have six=1.11.0 here. So if you're going to put the git urls into an .in file, it should not be the extra.in file, but probably move back to default.in.  

Except that docs.in includes default.txt but doesn't include extra.txt.  So maybe there's a renaming of this to something else, and have this include default.txt.  Or maybe docs.txt would work fine with the additions after all. 

So then that would just leave the mysql-connector-python entry.  Not sure what file it would be added to, since adding it into anything that is included elsewhere would cause that to fail in pip-compile.   

But I do wonder why it's better to post-process it into an extra.txt file if it ends up being the only thing in the extra.txt file.  Seems to me that we could just get rid of the extra.in file.   Such a mess.

scripts/post-pip-compile.sh
brianhw: Just for the record, the -e prefix was added in this PR, just to get them back through pip-compile.   When they were in extra.txt, they did not use the -e prefix.

HassanJaveed84:
@brianhw @jmbowman 

- Moved vcs urls to default.in, also moved psycopg2 to default.in. 
- Removed post-pip-compile script.
- Mysql connector python remains in extra.txt which is not part of pip-compile. 
- Tested doc generation using: ```make docs```

brianhw (approved):
I'll defer to Jeremy, but what I see LGTM. 👍

jmbowman (approved):
Thanks for getting this done!

PR #510, opened 2018-05-24
Update docker analyticstack information in README

rao-abdul-mannan:


azafty468 (approved):
I think what we have here is good.  I think most of the details should be in the "Setting up Docker Analyticstack" and less in here makes the document more readable.  So what you have is good.

README.md
```diff
@@ -62,6 +62,17 @@ If you are running a generic Open edX devstack, navigate to the directory that c
 
 Please follow the [analyticstack installation guide](http://edx.readthedocs.io/projects/edx-installing-configuring-and-running/en/latest/installation/analytics/index.html).
 
+Setting up Docker based Development Environment
+-----------------------------------------------
+
+We are moving our development setup from Vagrant to Docker, which uses a multi-container approach driven by Docker Compose. 
+There is a guide in place for [Setting up Docker Analyticstack](https://github.com/edx/devstack#getting-started-on-analytics) in
+devstack repository which can help you setup new analyticstack.
```
brianhw: nit:  devstack => "the devstack";  "setup new" => "set up a new".

README.md
```diff
@@ -62,6 +62,17 @@ If you are running a generic Open edX devstack, navigate to the directory that c
 
 Please follow the [analyticstack installation guide](http://edx.readthedocs.io/projects/edx-installing-configuring-and-running/en/latest/installation/analytics/index.html).
 
+Setting up Docker based Development Environment
```
brianhw: nit:  "Docker based" => "a Docker-based".  

And since I can't put comments up there, I suggest the previous section be renamed to "Setting up a Vagrant-based Development Environment".   And change the wording in the section to better differentiate:   'We call this environment the Vagrant "analyticstack".'   And fix the OEP-5 link.   

And I would like to propose that the whole Vagrant section be moved to come *after* the docker section.   I think Docker is ready to come first.   (Okay, so one could move the Docker section *before* the Vagrant section.   That's what I meant.)

README.md
```diff
@@ -62,6 +62,17 @@ If you are running a generic Open edX devstack, navigate to the directory that c
...
+-----------------------------------------------
+
+We are moving our development setup from Vagrant to Docker, which uses a multi-container approach driven by Docker Compose.
```
brianhw: Hmm.   I thought we should mention OEP-5 here, since it was mentioned in the Vagrant section, but I don't know enough whether the devstack work in general counts as adopting OEP-5.   That OEP seems to call for local targets, rather than ones in devstack, but maybe it's close enough.  Note that the location in the Vagrant section does not work -- the actual link is https://github.com/edx/open-edx-proposals/blob/master/oeps/oep-0005-arch-containerize-devstack.rst.  

Maybe this should say, "As part of our movement towards the adoption of [OEP-5](...), we have ported our development setup from Vagrant to Docker, which ...."

brianhw (commented):
Great!  A reorder suggestion, a few additions, and a few nits.  The one remaining question I have is whether to mention the word "deprecated" for Vagrant.  I'm willing to hold off for now, but we should think about it.

rao-abdul-mannan:
@brianhw I think we can add a note in vagrant section about our plan before we finally add deprecation to it.

macdiesel (approved):

brianhw (approved):
LGTM.  👍

PR #511, opened 2018-05-25
ENT-1021 Fix data discrepancy for enterprise enrollment table

brittneyexline:
The vertica version of the table had many more rows than the one in the result store,
and after investigating, it seems we were filtering out rows because of some joins
that should have been left joins.


georgebabey (approved):
Looks good - thanks for tracking this down!

brianhw:
Makes sense -- good find!  We're starting to find similar issues with finance reports as well.

PR #512, opened 2018-05-28
ENT-992:  Added additional fields in the analytics reports db

saleem-latif:
**JIRA Ticket:** [ENT-992](https://openedx.atlassian.net/browse/ENT-992)

__Description:__

This PR adds the following fields to analytics pipeline

1. coupon name
2. coupon code
3. final grade
4. course price
5. discount price

### DIFF
--- a/edx/analytics/tasks/enterprise/enterprise_database_imports.py
+++ b/edx/analytics/tasks/enterprise/enterprise_database_imports.py
@@ -105,3 +105,60 @@ class ImportUserSocialAuthTask(ImportMysqlToHiveTableTask):
             ('uid', 'STRING'),
             ('extra_data', 'STRING'),
         ]
+
+
+class ImportVoucher(ImportMysqlToHiveTableTask):
+    """
+    Ecommerce: Imports the voucher_voucher table from the ecommerce
+    database to a destination directory and a HIVE metastore.
+
+    A voucher is a discount coupon that can be applied to ecommerce purchases.
+    """
+    @property
+    def table_name(self):
+        return 'voucher_voucher'
+
+    @property
+    def columns(self):
+        return [
+            ('id', 'INT'),
+            ('name', 'STRING'),
+            ('code', 'STRING'),
+            ('usage', 'STRING'),
+            ('start_datetime', 'TIMESTAMP'),
+            ('end_datetime', 'TIMESTAMP'),
+            ('num_basket_additions', 'INT'),
+            ('num_orders', 'INT'),
+            ('total_discount', 'DECIMAL(12, 2)'),
+            ('date_created', 'TIMESTAMP'),
+        ]
+
+
+class ImportStockRecord(ImportMysqlToHiveTableTask):
+    """
+    Ecommerce: Imports the partner_stockrecord table from the ecommerce
+    database to a destination directory and a HIVE metastore.
+
+    A voucher is a discount coupon that can be applied to ecommerce purchases.
+    """
+    @property
+    def table_name(self):
+        return 'partner_stockrecord'
+
+    @property
+    def columns(self):
+        return [
+            ('id', 'INT'),
+            ('partner_sku', 'STRING'),
+            ('price_currency', 'STRING'),
+            ('price_excl_tax', 'DECIMAL(12, 2)'),
+            ('price_retail', 'DECIMAL(12, 2)'),
+            ('cost_price', 'DECIMAL(12, 2)'),
+            ('num_in_stock', 'INT'),
+            ('num_allocated', 'INT'),
+            ('low_stock_threshold', 'INT'),
+            ('date_created', 'TIMESTAMP'),
+            ('date_updated', 'TIMESTAMP'),
+            ('partner_id', 'INT'),
+            ('product_id', 'INT'),
+        ]
--- a/edx/analytics/tasks/enterprise/enterprise_enrollments.py
+++ b/edx/analytics/tasks/enterprise/enterprise_enrollments.py
@@ -8,16 +8,19 @@ import luigi.task
 from edx.analytics.tasks.common.mysql_load import MysqlInsertTask
 from edx.analytics.tasks.enterprise.enterprise_database_imports import (
     ImportDataSharingConsentTask, ImportEnterpriseCourseEnrollmentUserTask, ImportEnterpriseCustomerTask,
-    ImportEnterpriseCustomerUserTask, ImportUserSocialAuthTask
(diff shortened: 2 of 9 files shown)


edx/analytics/tasks/enterprise/enterprise_enrollments.py
```diff
@@ -139,7 +151,7 @@ def insert_query(self):
             JOIN student_courseenrollment enrollment
                     ON enterprise_course_enrollment.course_id = enrollment.course_id
                     AND enterprise_user.user_id = enrollment.user_id
-            JOIN auth_user auth_user
+            JOIN auth_user auth_usere
```
brianhw: Looks like a typo.

edx/analytics/tasks/insights/database_imports.py
```diff
@@ -789,6 +789,63 @@ def columns(self):
         ]
 
 
+class ImportVoucher(ImportMysqlToHiveTableTask):
```
brianhw: I would suggest that new imports that are only for enterprise should go into enterprise/enterprise_database_imports.py.    I am hoping at some point to remove the other imports that are already in this file but are not actually used by Insights, and move them to other locations, probably in warehouse.   (I think most are only used for financial reporting.)

edx/analytics/tasks/enterprise/enterprise_enrollments.py
```diff
@@ -11,12 +11,14 @@
     ImportEnterpriseCustomerUserTask, ImportUserSocialAuthTask
 )
 from edx.analytics.tasks.insights.database_imports import (
-    ImportAuthUserTask, ImportPersistentCourseGradeTask, ImportStudentCourseEnrollmentTask
+    ImportAuthUserTask, ImportPersistentCourseGradeTask, ImportStudentCourseEnrollmentTask, ImportVoucher,
+    ImportProductCatalog, ImportCurrentOrderLineState, ImportCurrentOrderDiscountState, ImportStockRecord,
+    ImportCurrentOrderState
```
brianhw: This is currently causing an error in the travis build.   You will probably need to run "isort --recursive edx/" to fix this.

brianhw (commented):
Good progress.   You may want to try using docker devstack to test out your acceptance test as you get it to work.  (With the additional fields, I hope it will fail.)

saleem-latif:
@brianhw please take another look?

edx/analytics/tasks/enterprise/enterprise_database_imports.py
```diff
@@ -105,3 +106,60 @@ def columns(self):
...
+        return [
+            ('id', 'INT'),
+            ('name', 'STRING'),
+            ('code', 'STRING'),
+            ('usage', 'STRING'),
+            ('start_datetime', 'TIMESTAMP'),
+            ('end_datetime', 'TIMESTAMP'),
+            ('num_basket_additions', 'INT'),
+            ('num_orders', 'INT'),
+            ('total_discount', hive_decimal_type(12, 2)),
```
brianhw: nit:  this was a really old thing when we were upgrading to Hive 0.13, and is no longer necessary.   You can just use 'DECIMAL(12, 2)' here explicitly, as we're well past supporting such old versions.

edx/analytics/tasks/enterprise/enterprise_enrollments.py
```diff
@@ -162,6 +173,18 @@ def insert_query(self):
...
+            JOIN order_line ecommerce_order_line
+                    ON ecommerce_catalogue_product.id = ecommerce_order_line.product_id
+            JOIN order_orderdiscount ecommerce_order_discount
+                    ON ecommerce_order_line.order_id = ecommerce_order_discount.order_id
+            JOIN voucher_voucher ecommerce_voucher
+                    ON ecommerce_order_discount.voucher_id = ecommerce_voucher.id
+            JOIN partner_stockrecord ecommerce_stockrecord
+                    ON ecommerce_order_line.stockrecord_id = ecommerce_stockrecord.id
+            JOIN order_order ecommerce_order
+                    ON ecommerce_order_line.order_id = ecommerce_order.id
```
brianhw: I certainly can't vouch for the correctness of all of this, since I don't really know what data is available in these tables. But I want to point out that these are all INNER JOINs, so if any of these ON statements does not come up with a match, then the record being joined to will be excluded from the output.   For example, is there always going to be voucher involved?    Alternative is to use a LEFT JOIN.

brianhw (approved):
Looks good.   Just a question about JOINs.   If that's not an issue, then 👍 .

edx/analytics/tasks/enterprise/enterprise_enrollments.py
saleem-latif: I think `order`, `order_line`, and `stockrecord` will always be present but `orderdiscount` and `voucher` are optional, So I have used LEFT JOIN for `orderdiscount` and `voucher` to make sure we get desired results. 
@georgebabey @brittneyexline can you confirm my assumption on `order`, `order_line`, and `stockrecord`?

zubair-arbi (approved):
Nice work 👍

edx/analytics/tasks/enterprise/enterprise_enrollments.py
```diff
@@ -162,6 +173,18 @@ def insert_query(self):
                     ON enterprise_user.user_id = social_auth.user_id
             JOIN course_catalog course
                     ON enterprise_course_enrollment.course_id = course.course_id
+            JOIN catalogue_product ecommerce_catalogue_product
+                    ON enterprise_course_enrollment.course_id = ecommerce_catalogue_product.course_id
+            JOIN order_line ecommerce_order_line
```
brittneyexline: I think it would be better for these to be left joins given not all enrollments are from coupons. Some are audit enrollments (which I'm not sure have orders anymore), and others might be through enterprise offers which would require different joins.

If you look at the vertica queries, there is a table called enterprise_user_course_redemption that pulls this data both for verified enrollments that go through coupons or offers. See https://github.com/edx/edx-analytics-pipeline-scripts/blob/master/enterprise_summary.sql#L420 for the query that pulls offer data and https://github.com/edx/edx-analytics-pipeline-scripts/blob/master/enterprise_summary.sql#L509 for the query that pulls the coupon data.

saleem-latif:
@brittneyexline I have used LEFT JOIN for order and order_line, please take another look.

georgebabey (approved):

saleem-latif:
@brianhw Can you please look at why coverage is failing? It was passing before I just added a new commit with a very small change. Can you guide me on how to fix it?

@georgebabey  -- FYI

brianhw:
Sorry about the coverage -- don't worry about it.

PR #513, opened 2018-05-29
Disable overwrite_n_days in ExternalCourseEnrollmentPartitionTask.

HassanJaveed84:
Do we need to get this change in before Hawthorn ?


brianhw (approved):
Yes, I think that's all that is needed.  LGTM.   👍

PR #514, opened 2018-05-30
Correcting coerced timestamp columns so that they do not use the default name

azafty468:
The production runs of vertica-to-bigquery-schema-copy-business-intelligence have been failing because duplicate field names are present due to a timestamp coercion process we are using in Sqoop.  This fixes the issue identified here: https://openedx.atlassian.net/browse/DE-763

Testing is being done on the production dataset here: http://scheduler.analytics.edx.org/view/Warehouse/job/vertica-to-bigquery-schema-copy-business-intelligence/6/ . 

I will update with the status when the job is complete and the code is ready for review.

### DIFF
--- a/edx/analytics/tasks/common/sqoop.py
+++ b/edx/analytics/tasks/common/sqoop.py
@@ -327,7 +327,10 @@ class SqoopImportFromVertica(SqoopImportTask):
         column_list = []
         for column in self.columns:
             if column in self.timezone_adjusted_column_list:
-                column_list.append('"{}" AT TIME ZONE \'UTC\''.format(column))
+                column_list.append('"{source_column}" AT TIME ZONE \'UTC\' AS "{exported_column_name}"'.format(
+                    source_column=column,
+                    exported_column_name=column
+                ))
             else:
                 column_list.append('"{}"'.format(column))
 
--- a/edx/analytics/tasks/common/tests/test_sqoop.py
+++ b/edx/analytics/tasks/common/tests/test_sqoop.py
@@ -48,7 +48,8 @@ class SqoopImportTestCase(unittest.TestCase):
         return task
 
     def create_vertica_task(self, num_mappers=None, where=None, table_name=None, schema_name=None, columns=None,
-                            null_string=None, fields_terminated_by=None, delimiter_replacement=None, overwrite=False):
+                            null_string=None, fields_terminated_by=None, delimiter_replacement=None, overwrite=False,
+                            timezone_adjusted_column_list=[]):
         """A generic task generator for Sqoop on Vertica"""
         if columns is None:
             column_list = []
@@ -70,6 +71,7 @@ class SqoopImportTestCase(unittest.TestCase):
             "fields_terminated_by": fields_terminated_by,
             "delimiter_replacement": delimiter_replacement,
             "overwrite": overwrite,
+            "timezone_adjusted_column_list": timezone_adjusted_column_list,
         }
         # remove options marked as None
         trimmed_kws = {k: v for k, v in kw_args.iteritems() if v is not None}
@@ -116,7 +118,7 @@ class SqoopImportTestCase(unittest.TestCase):
 
     def create_and_run_vertica_task(self, credentials=None, num_mappers=None, where=None, table_name=None,
                                     schema_name=None, columns=None, null_string=None, fields_terminated_by=None,
-                                    delimiter_replacement=None, overwrite=False):
+                                    delimiter_replacement=None, overwrite=False, timezone_adjusted_column_list=[]):
         """Create a SqoopImportFromVertica task with specified options, and then run it."""
         task = self.create_vertica_task(
             num_mappers=num_mappers,
@@ -128,6 +130,7 @@ class SqoopImportTestCase(unittest.TestCase):
             fields_terminated_by=fields_terminated_by,
             delimiter_replacement=delimiter_replacement,
             overwrite=overwrite,
+            timezone_adjusted_column_list=timezone_adjusted_column_list,
         )
         self.run_task(task, credentials)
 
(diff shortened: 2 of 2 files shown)


azafty468:
Production load job completed, confirmed the proper field names are present in the BigQuery table, and pushed a unit test to cover this issue going forward.  Once Travis results in a success I'll take the WIP off this PR and make it official.

edx/analytics/tasks/common/sqoop.py
```diff
@@ -327,7 +327,10 @@ def import_args(self):
         column_list = []
         for column in self.columns:
             if column in self.timezone_adjusted_column_list:
-                column_list.append('"{}" AT TIME ZONE \'UTC\''.format(column))
+                column_list.append('"{source_column}" AT TIME ZONE \'UTC\' as {exported_column_name}'.format(
```
brianhw: nit:   do you want this to be "AS" instead of "as"?   (If so, update the test as well. : )

edx/analytics/tasks/common/tests/test_sqoop.py
```diff
@@ -70,6 +71,7 @@ def create_vertica_task(self, num_mappers=None, where=None, table_name=None, sch
             "fields_terminated_by": fields_terminated_by,
             "delimiter_replacement": delimiter_replacement,
             "overwrite": overwrite,
+            "timezone_adjusted_column_list": timezone_adjusted_column_list
```
brianhw: nit:  trailing column

edx/analytics/tasks/common/tests/test_sqoop.py
```diff
@@ -275,6 +278,26 @@ def test_success_vertica_with_complete_credentials(self):
...
+        self.create_and_run_vertica_task(table_name='example_table',
+                                         schema_name='fake_schema',
+                                         delimiter_replacement=' ',
+                                         fields_terminated_by=',',
+                                         columns=['field1', 'field2', 'field3'],
+                                         timezone_adjusted_column_list=timezone_adjusted_column_list)
+        arglist = self.get_call_args_after_run()
+        self.assertTrue(self.mock_run.called)
+
+        # Confirm that timezone fields have a resulting AS renaming
```
brianhw: note: this was the line that made me notice that it's 'as' instead of 'AS'.

edx/analytics/tasks/common/tests/test_sqoop.py
```diff
@@ -128,6 +130,7 @@ def create_and_run_vertica_task(self, credentials=None, num_mappers=None, where=
             fields_terminated_by=fields_terminated_by,
             delimiter_replacement=delimiter_replacement,
             overwrite=overwrite,
+            timezone_adjusted_column_list=timezone_adjusted_column_list
```
brianhw: nit: trailing comma

edx/analytics/tasks/common/tests/test_sqoop.py
```diff
@@ -275,6 +278,26 @@ def test_success_vertica_with_complete_credentials(self):
...
+        arglist = self.get_call_args_after_run()
+        self.assertTrue(self.mock_run.called)
+
+        # Confirm that timezone fields have a resulting AS renaming
+        sqoop_query_fields = arglist[13].rsplit("SELECT")[1].rsplit("FROM")[0].strip().split(",")
+        for sqoop_query_column in sqoop_query_fields:
+            for timezone_field in timezone_adjusted_column_list:
+                if '"{}"'.format(timezone_field) in sqoop_query_column.split(' '):
+                    # A timezone field has been found, let's confirm the format
+                    self.assertTrue(' AT TIME ZONE \'UTC\' as ' in sqoop_query_column)
```
brianhw: Cool that you can parse this down, but was it possible to just look for the string "field1 AT TIME ZONE 'UTC' AS field1" in arglist[13]?   And is there a reason you exclude the actual column name here?   I guess if you wanted to test for more than one, you could just add field2 to timezone_adjusted_column_list as well, and then loop over an assertion that both "field1 AT TIME ZONE 'UTC' AS field1" and "field2 AT TIME ZONE 'UTC' AS field2" exist.

brianhw (approved):
Just nits.  Also wondering if testing should involve more than one field.  Otherwise LGTM.   👍

edx/analytics/tasks/common/sqoop.py
azafty468: Nuts I thought I caught

edx/analytics/tasks/common/tests/test_sqoop.py
azafty468: Er, is this part of linter?

macdiesel: Could we put these things in the linter instead of asking for adjustments in the code?

edx/analytics/tasks/common/tests/test_sqoop.py
azafty468: So I wanted to leave this unit test open in case we wanted to add more timestamp fields, modified how the query was created by Sqoop or found a new time issue.  A full string match would have worked as well but I think this form might be a bit more robust to future changes.

As for not checking the field name, I kind of didn't care about it.  The issue I'm safeguarding against is when Sqoop automatically generates a name.  Any unique name specified is valid.  But I can add in test for the name so that it's testing the generation and not safeguarding against an issue not spelled out in the test.

azafty468: Yuck, while adding in a fix to make sure that all fields expecting to have an AS I realized I didn't have a check for fields that should be AS.  And now I'm at the point where it's much easier to just do a full string comparison.  Oh well.

edx/analytics/tasks/common/tests/test_sqoop.py
brianhw: Which linter?   We use pycodestyle for pep8 to which we pay a little more attention, and pylint for lint errors that we rarely ever fix.    I didn't use a linter -- it's just me being anal, hence the non-blocking nit.

brianhw: I did not find any option in pycodestyle for this.   It's just in our Python style guide. http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python-guidelines.html#syntax-and-organization.   If I really wanted to push things, I'd invoke http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python-guidelines.html#breaking-long-lines more often, like on line 50-52 above.

PR #515, opened 2018-05-30
Add --dependency-tree option to launch-task

brianhw:
This option outputs a crude tree representation of a workflow's dependencies, instead of running it.

Uses task.deps() to gather dependencies correctly from Hadoop-based tasks as well as other (more normal) tasks.  It does not attempt to resolve dynamic dependencies.

### DIFF
--- a/edx/analytics/tasks/launchers/local.py
+++ b/edx/analytics/tasks/launchers/local.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 """
 Main method for running tasks on a local machine.
 
@@ -11,6 +12,7 @@ import argparse
 import logging
 import os
 import sys
+import warnings
 from contextlib import contextmanager
 
 import boto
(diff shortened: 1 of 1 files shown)


edx/analytics/tasks/launchers/local.py
```diff
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
```
HassanJaveed84, line 1: What is the significance of this line ?

HassanJaveed84 (approved):
Just one question, LGTM otherwise.

edx/analytics/tasks/launchers/local.py
brianhw: The characters used in displaying the tree, particularly the line that goes down and to the right, is a unicode character.   It is stored in the file as utf-8.  It could probably be stored with an escape sequence instead (e.g. \u0099 or whatever), and we could avoid adding this.  I can go either way -- it's nicer to have all ascii characters in a file, and it's also nice to see the actual symbols being output.

HassanJaveed84: Got it. 
I was reading the docs and saw utf-8 as the default if no encoding declaration is found, but those were for python 3.x. 
2.7 does not mention anything about default encoding.

PR #516, opened 2018-06-11
Add basic Spark support

brianhw:
This is a first step towards adding basic Spark support, independent of work done on Insights workflows.   Two main things this PR aims to provide are options support and dependency loading.   And these still need more work.  I am thinking of moving the dependency loading to be more similar to how the hadoop tasks handle it.   I think Abdul had ideas about improving options, but I don't remember what they were.   I would like to land this (and release it) before the Hawthorn release is tagged.

### DIFF
--- a/edx/analytics/tasks/common/spark.py
+++ b/edx/analytics/tasks/common/spark.py
@@ -1,0 +1,361 @@
+import ast
+import importlib
+import json
+import logging
+import os
+import shutil
+import tempfile
+import zipfile
+from collections import defaultdict
+
+import luigi
+import luigi.configuration
+from luigi.contrib.spark import PySparkTask
+
+_file_path_to_package_meta_path = {}
+
+log = logging.getLogger(__name__)
+
+
+def get_package_metadata_paths():
+    """
+    List of package metadata to be loaded on EMR cluster
+    """
+    from distlib.database import DistributionPath
+
+    if len(_file_path_to_package_meta_path) > 0:
+        return _file_path_to_package_meta_path
+
+    dist_path = DistributionPath(include_egg=True)
+    for distribution in dist_path.get_distributions():
+        metadata_path = distribution.path
+        for installed_file_path, _hash, _size in distribution.list_installed_files():
+            absolute_installed_file_path = installed_file_path
+            if not os.path.isabs(installed_file_path):
+                absolute_installed_file_path = os.path.join(os.path.dirname(metadata_path), installed_file_path)
+            normalized_file_path = os.path.realpath(absolute_installed_file_path)
+            _file_path_to_package_meta_path[normalized_file_path] = metadata_path
+
+    return _file_path_to_package_meta_path
+
+
+def dereference(f):
+    if os.path.islink(f):
+        # by joining with the dirname we are certain to get the absolute path
+        return dereference(os.path.join(os.path.dirname(f), os.readlink(f)))
+    else:
+        return f
+
+
+def create_packages_archive(packages, archive_dir_path):
+    """
+    Create a zip archive for all the packages listed in packages and returns the list of zip file location.
+    """
+    archives_list = []
+    package_metadata_paths = get_package_metadata_paths()
+    metadata_to_add = dict()
+
+    package_zip_path = os.path.join(archive_dir_path, 'packages.zip')
+    package_zip = zipfile.ZipFile(package_zip_path, "w", compression=zipfile.ZIP_DEFLATED)
+    archives_list.append(package_zip_path)
+
+    def add(src, dst, package_name):
+        # Ensure any entry points and other egg-info metadata is also transmitted along with
+        # this file. If it is associated with any egg-info directories, ship them too.
+        metadata_path = package_metadata_paths.get(os.path.realpath(src))
+        if metadata_path:
+            metadata_to_add[package_name] = metadata_path
+
+        package_zip.write(src, dst)
+
+    def add_files_for_package(sub_package_path, root_package_path, root_package_name, package_name):
+        for root, dirs, files in os.walk(sub_package_path):
+            if '.svn' in dirs:
+                dirs.remove('.svn')
+            for f in files:
+                if not f.endswith(".pyc") and not f.startswith("."):
+                    add(dereference(root + "/" + f),
+                        root.replace(root_package_path, root_package_name) + "/" + f,
+                        package_name)
+
+    for package in packages:
+        # Archive each package
+        if not getattr(package, "__path__", None) and '.' in package.__name__:
+            package = __import__(package.__name__.rpartition('.')[0], None, None, 'non_empty')
+
+        n = package.__name__.replace(".", "/")
+
+        # Check length of path, because the attribute may exist and be an empty list.
+        if len(getattr(package, "__path__", [])) > 0:
+            # TODO: (BUG) picking only the first path does not
+            # properly deal with namespaced packages in different
+            # directories
+            p = package.__path__[0]
+
+            if p.endswith('.egg') and os.path.isfile(p):
+                raise 'Not going to archive egg files!!!'
+                # Add the entire egg file
+                # p = p[:p.find('.egg') + 4]
+                # add(dereference(p), os.path.basename(p))
+
+            else:
+                # include __init__ files from parent projects
+                root = []
+                for parent in package.__name__.split('.')[0:-1]:
+                    root.append(parent)
+                    module_name = '.'.join(root)
+                    directory = '/'.join(root)
+
+                    add(dereference(__import__(module_name, None, None, 'non_empty').__path__[0] + "/__init__.py"),
+                        directory + "/__init__.py",
+                        package.__name__)
+
+                add_files_for_package(p, p, n, package.__name__)
+
+        else:
+            f = package.__file__
+            if f.endswith("pyc"):
+                f = f[:-3] + "py"
+            if n.find(".") == -1:
+                add(dereference(f), os.path.basename(f), package.__name__)
+            else:
+                add(dereference(f), n + ".py", package.__name__)
+
+        # include metadata in the same zip file
+        metadata_path = metadata_to_add.get(package.__name__)
+        if metadata_path is not None:
+            add_files_for_package(metadata_path, metadata_path, os.path.basename(metadata_path), package.__name__)
+
+    return archives_list
+
+
+class SparkMixin():
+    """
+    Provides `spark-submit` parameters that can be set on the command-line, as well as a configuration file.
+
+    Most `spark-submit` parameters defined in SparkSubmitTask are only pulled from config files.
+
+    See documentation for `spark-submit` for more details on the particular parameters.
+
+    """
+    # Note that all of these are hard-coded to pull from the 'spark' section of the configuration file.
+    # In SparkSubmitTask, the 'spark-version' that defines the section is a property that can be overridden
+    # in derived classes.  Because this is a standalone mixin, it doesn't have access to that property.
+
+    driver_memory = luigi.Parameter(
+        config_path={'section': 'spark', 'name': 'driver-memory'},
+        description='Memory for spark driver.',
+        significant=False,
+    )
+    executor_memory = luigi.Parameter(
+        config_path={'section': 'spark', 'name': 'executor-memory'},
+        description='Memory for each executor.',
+        significant=False,
+    )
+    executor_cores = luigi.Parameter(
+        config_path={'section': 'spark', 'name': 'executor-cores'},
+        description='Number of cores for each executor.',
+        significant=False,
+    )
+    spark_conf = luigi.Parameter(
+        config_path={'section': 'spark', 'name': 'conf'},
+        description="Spark configuration, specified with '|' as delimiter of key=value pairs.",
+        significant=False,
+        default=None
+    )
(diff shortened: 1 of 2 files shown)


config/devstack.cfg
```diff
@@ -131,3 +131,8 @@ api_root_url = http://localhost:8000/api/courses/v1/courses/
 
 [course-blocks]
 api_root_url = http://localhost:8000/api/courses/v1/blocks/
+
+[spark]
+driver-memory=3g
+executor-memory=3g
+executor-cores=1
```
rao-abdul-mannan: Though edx_egg_files is not required in vagrant devstack, but it should include a small note here explaining about its usage/importance while running it on a multi-node cluster ?

edx/analytics/tasks/common/spark.py
```diff
@@ -0,0 +1,286 @@
...
+        """
+        Returns `value` of `key` after parsing string argument.
+        """
+        default_value = kwargs.get('default_value', None)
+        str_arg = args[0]
+        config_dict = ast.literal_eval(str_arg)
+        value = config_dict.get(key, default_value)
+        return value
+
+    def get_luigi_configuration(self):
```
rao-abdul-mannan: Though I am still hesitant about loading luigi configuration this way before we merge spark work into master branch, but since we haven't worked on the other solution so I guess we don't have a choice.

edx/analytics/tasks/common/spark.py
```diff
@@ -0,0 +1,286 @@
...
+        # Import packages to be loaded on cluster.
+        # This list was taken from what was needed for Hadoop.  These may not all be needed still on Spark jobs.
+        # Not all jobs need this, so perhaps the list of many packages can be moved to a derived class.
+        import edx
+        import luigi
+        import opaque_keys
+        import stevedore
+        import bson
+        import ccx_keys
+        import cjson
```
rao-abdul-mannan: `cjson` package doesn't work on spark standalone cluster and I guess it won't work on yarn cluster either. I tried creating its egg but failed. Have you tried this in docker or yarn ?

edx/analytics/tasks/common/spark.py
```diff
@@ -0,0 +1,286 @@
...
+import luigi
+import luigi.configuration
+from luigi.contrib.spark import PySparkTask
+
+from edx.analytics.tasks.common.pathutil import EventLogSelectionDownstreamMixin, PathSelectionByDateIntervalTask
+from edx.analytics.tasks.util.manifest import (
+    ManifestInputTargetMixin, convert_to_manifest_input_if_necessary, remove_manifest_target_if_exists
+)
+from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
+from edx.analytics.tasks.util.url import get_target_from_url, url_path_join
```
rao-abdul-mannan: L13-18 are unused imports

config/devstack.cfg
brianhw: Hmm.   I just blindly brought this over from your branch, and had assumed that this was for docker.  (But that's what luigi_docker.cfg is for.)  I think I'll just take this out, because we don't currently support Spark on vagrant devstack.  (And add more comments in luigi_docker.cfg.)

edx/analytics/tasks/common/spark.py
brianhw: Oops.  Good catch.

edx/analytics/tasks/common/spark.py
brianhw: This code worked for me on my hackathon, both on docker and EMR.  But I didn't make any use of most of these, since I was not running on events.   I want to pare this down to the minimum (i.e. edx, luigi, stevedore), and allow other tasks to add imports to this list as we do for Hadoop jobs.   I also want to only load the egg files that are requested, so I would like to change the list to a dict in the config file for that.

edx/analytics/tasks/common/spark.py
```diff
@@ -0,0 +1,286 @@
...
+        """
+        from pyspark.sql import SparkSession
+        self._spark_context = sc
+        # Note that this doesn't actually use sc.  It just gets
+        # the currently-existing Spark session.
+        self._spark = SparkSession.builder.getOrCreate()
+
+        self._tmp_dir = tempfile.mkdtemp()
+
+        # TODO: pull definition of __name__ out into a class property.
```
brianhw: I'm still hoping to make this change too.

brianhw:
This is ready for another (hopefully final) review.

edx/analytics/tasks/common/spark.py
```diff
@@ -136,68 +129,88 @@ def add_files_for_package(sub_package_path, root_package_path, root_package_name
...
+class SparkJobTask(SparkMixin, PySparkTask):
     """
     Base class for running a launchable Spark task.
+
+    This adds functionality to PySparkTask for loading Python packages onto Spark workers to support
+    Python user-defined functions (UDFs) running there.  It also provides a mechanism for passing
+    configuration information from Luigi into the Spark 'spark-submit' subprocess.
+
+    Note that methods that begin with 'spark_' or '_spark_' are executed within the 'spark-submit'
+    subprocess, while others will be executed in the main process (that contains Luigi).
```
rao-abdul-mannan, line 178: This note is helpful especially which functions are executed within spark and luigi.

edx/analytics/tasks/common/spark.py
```diff
@@ -136,68 +129,88 @@ def add_files_for_package(sub_package_path, root_package_path, root_package_name
...
+        """
+        return []
 
-    @property
-    def conf(self):
-        """Adds spark configuration to spark-submit task."""
-        return self._dict_config(self.spark_conf)
+    def get_task_configuration_from_luigi(self):
+        """
+        Return Luigi configuration needed for this Spark task, as a dict.
```
rao-abdul-mannan: I have another ( a bit clean ) solution for luigi ( for now ). 
Instead of user fetching each configuration value, that they need, in the base class method,  why not we allow user to just return a dict of {"sections_names"=>"option_name"} instead and then fetch all these values from that dict in app_options method ?
I think it will be a lot cleaner that way where they just specifiy which values they would need in their spark task and they can get those values from spark_get_config_from_args().

edx/analytics/tasks/common/spark.py
```diff
@@ -206,81 +219,116 @@ def get_config_from_args(self, key, *args, **kwargs):
...
+        self.log = log4j_logger.LogManager.getLogger(self.spark_logger_name)
+
+    @property
+    def conf(self):
+        """
+        Adds spark configuration to spark-submit task.
+
+        This overrides the parameter defined in SparkSubmitTask, which reads configuration from Luigi configuration
+        files only. The override permits configuration to be specified instead on the command line, and only read
+        from configuration if the command line is not used.
```
rao-abdul-mannan: It is not just to override the configuraton values, it allows user to specify extra arguments which luigi has not provided support yet and there are dozens of such properties. So instead of specifying all of them in configuration, they have added a generic option which user can overwrite to add new configuration.
e.g. conf=spark.default.parallelism=100|spark.sql.shuffle.partitions=200 ( both these properties & many more cannot be added directly in configuration section with their own name ).

edx/analytics/tasks/common/spark.py
```diff
@@ -206,81 +219,116 @@ def get_config_from_args(self, key, *args, **kwargs):
...
 
     def main(self, sc, *args):
+        """
+        Defines the main extension point for PySparkTask, providing its own 'spark_job' extension point.
+        """
         try:
-            self.init_spark(sc)  # initialize spark contexts
-            self._load_internal_dependency_on_cluster(*args)  # load packages on EMR cluster for spark worker nodes
+            self._spark_initialize(sc)  # initialize spark contexts
+            self._spark_load_internal_dependency_on_cluster(*args)  # load packages on EMR cluster for spark worker nodes
```
rao-abdul-mannan: Comment needs to be reworded. Its not just for **EMR** anymore since we now have standalone cluster in docker devstack.

edx/analytics/tasks/common/spark.py
```diff
@@ -206,81 +219,116 @@ def get_config_from_args(self, key, *args, **kwargs):
...
-        import boto
-        import filechunkio
-        import ciso8601
-        import chardet
-        import urllib3
-        import certifi
-        import idna
-        import requests
-        import six
+        Loading of packages via a single zipfile is adequate for most packages.  However, it is not adequate for
```
rao-abdul-mannan: Shouldn't we need to define some base packages that will at least enable pyspark to run python code on worker nodes ? 
For extra packages, we can ask user to define them but having them specify basic package list is redundant + hectic for anyone to figure it out.

rao-abdul-mannan (changes requested):

edx/analytics/tasks/common/spark.py
brianhw: Right.  But this is allowing conf to be defined on the command line, rather than requiring that it be defined in the Luigi configuration file.  Is there a specific way in which the wording of this could make that clearer?

edx/analytics/tasks/common/spark.py
brianhw: Good point.

edx/analytics/tasks/common/spark.py
brianhw: No.  Nothing needs to be defined here in order to get generic Spark python code to run on worker nodes.   I did my hackathon project using Spark ML and had all of this commented out.  It is only necessary to add things here when the UDFs one is using require more custom imports.    I tested this mechanism by using get_org_id_from_course_id() in a udf, and found that I needed edx, stevedore, bson, opaque_keys, and optionally ccx_keys.  (If I used a course_id that was ccx-based, the org would be found or would be Null depending on whether ccx_keys were loaded -- as expected.)    But I think we will want another class on top of SparkJobTask that is used as the basis for all the event-based processing that the Insights tasks do, and that should define the packages that it needs for its event parsing and processing.  But that should be a separate class from this base class.

edx/analytics/tasks/common/spark.py
brianhw: I found that knowing where something was run was one of the major challenges of understanding the code, so I wanted to make it a little easier for others.

edx/analytics/tasks/common/spark.py
brianhw: I'm not understanding.  Users are already providing a dict, so you're suggesting a change in the convention of how that dict is structured.   Are you just suggesting that another level be added to the dict, so that its keys are section names, and within each section is another dict with key-value pairs?  (I didn't understand what {"sections_names"=>"option_name"} meant.)  And then spark_get_config_from_args() would have an args and key argument as it currently does, but also an additional section argument?  

Also, the first goal of this is to get configuration across the spark-submit boundary.   I am assuming that this mechanism does not work within UDFs.  I am assuming that any values need to be pulled out in the spark_job() level, and explicitly passed to UDFs.

edx/analytics/tasks/common/spark.py
rao-abdul-mannan: Just wanted to clarify, thats good to know if it is working fine.

edx/analytics/tasks/common/spark.py
rao-abdul-mannan: I just to make sure that if by reading these comments + the description of spark_conf luigi parameter, users will be able to understand that they can specify extra spark configurations via this option well.

rao-abdul-mannan: if it does, then i'm okay with it.

edx/analytics/tasks/common/spark.py
rao-abdul-mannan: `app_options` method maps user supplied configuration ( get_task_configuration_from_luigi )  to spark-submit args. `app_options` is called from luigi & not from spark hence we can get luigi configuration values directly here. 
I'm saying we can have this function return something like a dict ( {'interval': 'calendar'}, option name is key, section is value ) and then get their values inside  `app_options` method.

brianhw:
As requested, I rewrote the configuration handoff between Luigi and Spark, so it is a lot simpler and more intuitive now.   On the Luigi side, a task requests configuration values to save by providing section and parameter-name, and on the Spark side, the task can request the configuration values by providing the same section and parameter-name to get the configuration values returned.    Ready for review again. : )

edx/analytics/tasks/common/spark.py
```diff
@@ -192,32 +196,42 @@ def spark_remote_package_names(self):
...
         """
-        return None
+        self._requested_config[section].append(parameter)
 
-    def spark_get_config_from_args(self, key, *args, **kwargs):
+    def spark_get_config_from_args(self, section, key, default_value=None):
         """
         Allows parameters to be fetched from the arguments parameter passed to spark_job().
 
         This is the mirror of get_task_configuration_from_luigi.
```
rao-abdul-mannan: Comment needs to be updated => **This is the mirror of request_configuration_from_luigi**

rao-abdul-mannan (changes requested):
Only comment needs an update, apart from that everything looks good.

rao-abdul-mannan:
LGTM 👍

rao-abdul-mannan (approved):

azafty468 (approved):
Code looks good, I'm a 👍

PR #517, opened 2018-06-12
Created pull request template.

HassanJaveed84:


.github/pull_request_template.md
```diff
@@ -0,0 +1,9 @@
+Analytics Pipeline Pull Request
+---
+
+Make sure that the following steps are done before merging:
+
+  - [ ] If you have a migration please contact data engineering team before merging.
```
brianhw, line 6: Do we want to list a way of contacting the data engineering team in this PR?   Is there a tag or an email address, or should we just leave it to people to figure this out?

Also nit:  "contact the data engineering team" sounds better to my ear, both here and in the third checkbox.

HassanJaveed84: Do you have any suggestions on what needs to be mentioned here ? Do we have a tag for DE team or should we mention a specific person tag ?

.github/pull_request_template.md
```diff
@@ -0,0 +1,9 @@
...
+  - [ ] Before merging run full acceptance tests suite and provide URL for the acceptance tests run.
+  - [ ] A member of data engineering team has approved the pull request.
```
HassanJaveed84, line 8: Should this be 'Contact the data engineering team before merging?'

.github/pull_request_template.md
brianhw: I don't have a particular suggestion, which is why I asked a question. : )   But I definitely think we should not mention a specific person.   The migration needs to be raised more broadly.    If we don't have a group tag or email, then we can leave the onus on the PR author to figure out how to reach us.   I'm fine with the existing wording as being a good step forward.

.github/pull_request_template.md
brianhw: I liked your original sentiment that DE needs to approve, not just be notified.   That way DE can at least acknowledge that they were notified.

brianhw (approved):
I'm fine with it as-is, if there's no obvious contact info we want to add.

azafty468 (approved):
👍 LGTM!

PR #523, opened 2018-07-17
Load course_block_records table in Vertica using Course Blocks API

brianhw:
Building off of a previous spike in brian/course-struct, this fixes the API calls.   The structure of the tasks permits calls to fetch information for an individual course, as well as to fetch structure for all courses known to the API.


brianhw:
Acceptance test passes, on Jenkins and devstack.   This is ready for final review.

edx/analytics/tasks/util/record.py
```diff
@@ -767,3 +775,144 @@ def validate(self, value):
...
+        """
+        For a given field name, add the location in the source dictionary from which to fetch a value.
+
+        Key values for the mapping are represented using dot notation, beginning with the hard-coded string value 'root'.
+        So for a dictionary with { 'a': {'b': 'b-value' } }, fetching the 'b-value' into a record field would be
+        defined as 'root.a.b'.
+        """
+        raise NotImplementedError
+
+    def _add_entry(self, record_dict, record_key, record_field, label, obj):
```
HassanJaveed84, line 813: Can we have a docstring here ?

edx/analytics/tasks/warehouse/load_internal_reporting_course_structure.py
```diff
@@ -0,0 +1,660 @@
...
+
+        with self.output().open('w') as output_file:
+            output_file.write("DONE.")
+
+    def output(self):
+        return get_target_from_url(url_path_join(self.output_root, "_SUCCESS"))
+
+    def complete(self):
+        # TODO: do we still need this?
+        return self.output().exists()
```
HassanJaveed84: I think we can remove this method as OverwriteOutputMixin's complete() method would be able to handle this.

edx/analytics/tasks/warehouse/load_internal_reporting_course_structure.py
```diff
@@ -0,0 +1,660 @@
...
+
+
+class AllCourseBlocksTask(AllCourseMixin, BlocksPerCourseTask):
+    """Fetch info provided by course blocks API for a list of course_id values read from raw course_list JSON, and output as TSV."""
+
+    course_id = None
+
+    def requires(self):
+        return_value = {}
+        parent = super(AllCourseBlocksTask, self).requires()
```
HassanJaveed84, line 221: This would call BlocksPerCourseTask's requires() method which is defined as:
```
def requires(self):
        pass
```
Which would always be None, I think I'm not correctly making sense of it.

edx/analytics/tasks/warehouse/load_internal_reporting_course_structure.py
```diff
@@ -0,0 +1,660 @@
...
+
+    def do_action_per_course(self, course_id, output_root):
+        self.output_blocks_for_course(course_id, output_root)
+
+
+class CourseBlockRecord(SparseRecord):
+    """Represents a block in a course."""
+
+    # Metadata:
+    version = StringField(length=20, nullable=False, description='blah.')
```
HassanJaveed84: Do we want to update or remove the description ? 
Same for the fields that follow('blah').

edx/analytics/tasks/warehouse/load_internal_reporting_course_structure.py
```diff
@@ -0,0 +1,660 @@
...
+            course_id = course_blocks_info.get('course_id')
+            output_target = self.get_output_target(course_id)
+            with output_target.open('w') as output_file:
+                for record in self.record_mapper.generate_records_from_course_blocks(course_blocks_info):
+                    record_string = record.to_separated_values()
+                    output_file.write(record_string)
+                    output_file.write('\n')
+
+    def run(self):
+        self.remove_output_on_overwrite()
```
HassanJaveed84, line 624: Do we also want to remove the ```market_target``` in case of overwrite=True ?

edx/analytics/tasks/warehouse/load_internal_reporting_course_structure.py
```diff
@@ -0,0 +1,660 @@
...
+        _input_path = self.hive_partition_path(raw_table_name, self.date)
+        return _input_path
+
+    @property
+    def output_path(self):
+        record_table_name = 'course_block_records'
+        _output_path = self.hive_partition_path(record_table_name, self.date)
+        return _output_path
+
+    def get_output_target(self, course_id):
```
HassanJaveed84: I got a little confused here, initially thinking it was the output target of this task. Now I realize it's for a particular course. Do you think changing the method name to something like 'get_output_target_for_course' be more appropriate ?

edx/analytics/tasks/warehouse/load_internal_reporting_course_structure.py
```diff
@@ -0,0 +1,660 @@
...
+        return get_target_from_url(url_path_join(self.output_path, '_SUCCESS'))
+
+    def output(self):
+        return get_target_from_url(self.output_path)
+
+    def complete(self):
+        return self.marker_target.exists()
+
+
+class LoadCourseBlockRecordToVertica(CourseBlocksDownstreamMixin, VerticaCopyTask):
```
HassanJaveed84: How and when would this be scheduled ? Isn't it loading the same data ```LoadInternalReportingCourseStructureToWarehouse``` task is loading into ```course_structure``` table ?

edx/analytics/tasks/tests/acceptance/test_internal_reporting_course_structure.py
```diff
@@ -0,0 +1,77 @@
...
+            self.upload_file(
+                os.path.join(self.data_dir, 'input', 'course_structure', file_name),
+                url_path_join(self.warehouse_path, 'course_block_raw', 'dt=' + self.DATE, file_name)
+            )
+
+    @when_vertica_available
+    def test_internal_reporting_course_structure(self):
+        """Tests the workflow for the internal reporting course_structure table, end to end."""
+
+        # First generate the records in S3.
```
HassanJaveed84, line 37: Doesn't this also write to Vertica ?

HassanJaveed84:
Few questions, otherwise LGTM 👍(given my limited knowledge on course blocks : ))

Also, do we want to pass in overwrite parameter to the upstream tasks ?

edx/analytics/tasks/tests/acceptance/test_internal_reporting_course_structure.py
brianhw: Yes, it does that too, but the important side-effect of it is to leave a trail behind in S3 that can then be loaded into the Warehouse as part of the daily production run.  
And that's a valid point.  I could just run AllCourseBlockRecordsTask here to write course structure to S3, and not do the extra load to Vertica.  And I could do the same in the load-course-struct job that runs weekly.   But I guess I felt it was useful to just do the extra write to Vertica on the weekly job as well (to experimental), so that results could be seen right away if there were issues with production or something.  And since I was doing that in production, I also did it here.

edx/analytics/tasks/util/record.py
brianhw: Yup, thanks!

edx/analytics/tasks/warehouse/load_internal_reporting_course_structure.py
brianhw: I think you're right.   I missed answering the TODO.

edx/analytics/tasks/warehouse/load_internal_reporting_course_structure.py
brianhw: Yes, that's right.   This is more general, and I cut it from elsewhere where the underlying class did have its own requirements.

edx/analytics/tasks/warehouse/load_internal_reporting_course_structure.py
brianhw: Yup.   Updating using documentation on wiki.

edx/analytics/tasks/warehouse/load_internal_reporting_course_structure.py
brianhw: Sounds good to me!

edx/analytics/tasks/warehouse/load_internal_reporting_course_structure.py
brianhw: LoadInternalReportingCourseStructureToWarehouse is a daily job that is part of load_warehouse.   (There's a parallel method to that that is in load_warehouse_bigquery.)   So it runs daily, but loads whatever is the most recent data -- it doesn't actually make any API calls.   Instead, the LoadCourseBlockRecordToVertica (or the underlying AllCourseBlocksTask) task does the actual work of doing the pull for the day on which it is run, making the API calls and generating the records and writing them out to S3.   And that currently runs weekly (on Sunday night).

As I mentioned earlier (in the acceptance test), we could technically call AllCourseBlocksTask weekly and LoadInternalReportingCourseStructureToWarehouse daily, but I like to have the extra Vertica load call in there.   We can easily change that in the load-course-struct Jenkins job.  But for development purposes and such, I definitely find it helpful to have this class at least exist.

edx/analytics/tasks/warehouse/load_internal_reporting_course_structure.py
brianhw: Oh, this is a mess.   I didn't really test overwrite, and this wouldn't do the right thing.   I've now rewritten it and tested it.   Thanks!

brianhw (commented):
Thanks, Hassan!   I think I've responded to all your comments.

edx/analytics/tasks/warehouse/load_internal_reporting_course_structure.py
HassanJaveed84: Thanks for the clarification. Looks good.

PR #524, opened 2018-07-19
Simplied enterprise_enrolments by modularizing tables

jonalexvasquez:
enterprise_enrollments.py has a large query due to having one table where all enrollment info is stored. The point of this change is to split that large query inside enterprise_enrollments.py into smaller, easier to understand queries by creating different tables. Following this : [https://openedx.atlassian.net/browse/ENT-1075](url) I created multiple files/tasks instead of having one big task(enterprise_enrollments.py).


jonalexvasquez:
What I do not understand is the following code : 
```python
def requires(self):  
        for requirement in super(EnterpriseUserEnrollmentDataTask, self).requires():
            yield requirement

        yield (
            ImportAuthUserTask(),
            ImportAuthUserProfileTask(),
            ImportEnterpriseCourseEnrollmentUserTask(),
            ImportUserSocialAuthTask(),
            CoursePartitionTask(
                date=self.date,
                api_root_url=self.api_root_url,
                api_page_size=self.api_page_size,
            ),
            UserActivityTableTask(
                overwrite_n_days=0,
                date=self.date
            )
        )

        kwargs = {
            'credentials': self.otto_credentials,
            'database': self.otto_database,
        }
        yield (
            ImportProductCatalog(**kwargs),
            ImportCurrentOrderLineState(**kwargs),
            ImportCurrentOrderDiscountState(**kwargs),
            ImportVoucherTask(**kwargs),
            ImportStockRecordTask(**kwargs),
            ImportCurrentOrderState(**kwargs),
            ImportEcommerceUser(**kwargs),
            ImportConditionalOfferTask(**kwargs),
            ImportBenefitTask(**kwargs),
        ) 
```
I sort of took a stab at it and chose what "require" import tasks inside yield() belong with what file/task but am not fully sure which ones are really needed in each file/task. It seems like the second yield function belongs to purely ecommerce but am not if it is needed across all files. Thanks!

edx/analytics/tasks/enterprise/enterprise_user_enrollments.py
```diff
@@ -0,0 +1,246 @@
+"""Compute metrics related to user enrollments in courses"""
```
brittneyexline, line 1: this should say something like: "Gather data related to enterprise users"

edx/analytics/tasks/enterprise/enterprise_user_enrollments.py
```diff
@@ -0,0 +1,246 @@
...
+    BooleanField, DateField, DateTimeField, FloatField, IntegerField, Record, StringField
+)
+from edx.analytics.tasks.warehouse.load_internal_reporting_course_catalog import (
+    CoursePartitionTask, LoadInternalReportingCourseCatalogMixin
+)
+
+log = logging.getLogger(__name__)
+
+
+class EnterpriseUserEnrollmentRecord(Record):
```
brittneyexline, line 33: EnterpriseUserRecord

edx/analytics/tasks/enterprise/enterprise_user_enrollments.py
```diff
@@ -0,0 +1,246 @@
...
+    """
+    Creates the metadata for the enterprise_user_enrollment hive table
+    """
+    @property  # pragma: no cover
+    def partition_by(self):
+        return 'dt'
+
+    @property
+    def table(self):  # pragma: no cover
+        return 'enterprise_user_enrollment'
```
brittneyexline, line 55: enterprise_user

edx/analytics/tasks/enterprise/enterprise_user_enrollments.py
```diff
@@ -0,0 +1,246 @@
...
+    lms_user_id = IntegerField(nullable=False, description='')
+    enterprise_user_id = IntegerField(nullable=False, description='')
+    enterprise_sso_uid = StringField(length=255, description='')
+    user_account_creation_timestamp = DateTimeField(description='')
+    user_email = StringField(length=255, description='')
+    user_username = StringField(length=255, description='')
+    user_country_code = StringField(length=2, description='')
+    last_activity_date = DateField(description='')
+
+class EnterpriseUserEnrollmentHiveTableTask(BareHiveTableTask):
```
brittneyexline, line 45: EnterpriseUserHiveTableTask

edx/analytics/tasks/enterprise/enterprise_user_enrollments.py
```diff
@@ -0,0 +1,246 @@
...
+            SELECT DISTINCT enterprise_customer.uuid AS enterprise_id,
+                enterprise_user.user_id AS lms_user_id,
+                enterprise_user.id AS enterprise_user_id,
+                SUBSTRING_INDEX(social_auth.uid_full, ':', -1) AS enterprise_sso_uid,
+                auth_user.date_joined AS user_account_creation_timestamp,
+                auth_user.email AS user_email,
+                auth_user.username AS user_username,
+                user_profile.country AS user_country_code,
+                user_activity.latest_date AS last_activity_date,
+            FROM enterprise_enterprisecourseenrollment enterprise_course_enrollment
```
brittneyexline, line 110: i think this should actually be the enterprise_enterprisecustomeruser table - we don't need enrollment data here, and enterprise_user is referenced below but hasn't been pulled in

edx/analytics/tasks/enterprise/enterprise_user_enrollments.py
```diff
@@ -0,0 +1,246 @@
...
+            warehouse_path=self.warehouse_path,
+            overwrite=self.overwrite,
+        )
+
+    def requires(self):  # pragma: no cover
+        for requirement in super(EnterpriseUserEnrollmentDataTask, self).requires():
+            yield requirement
+
+        # the process that generates the source table used by this query
+        yield (
```
brittneyexline, line 154: the purpose of these yields is to enumerate the tasks that generate the tables that the query above depends on. you should keep the following, and remove the rest (including all of the ecommerce tables):
```
ImportAuthUserTask(),
ImportAuthUserProfileTask(),
ImportEnterpriseCustomerUserTask(), // not listed here, but it is necessary
ImportUserSocialAuthTask(),
UserActivityTableTask(),
```

brittneyexline (commented):
this is a good start, i'm sorry about the confusion!
i commented in the enterprise_user_enrollments.py file - i think everything in the file, including the filename itself, needs to be renamed to enterprise_user. i've noted a few examples but be sure to look at everything in the file.
you will also need to pull in the enterprisecustomeruser table instead of the enrollments table, since we are only focused on enterprise user data here.
i think the best option would be to push up a new, identical branch with this code, and to adjust this branch/PR with your updates, so we don't lose the comment history.

let me know if you have any questions about any of this!

PR #526, opened 2018-07-22
Added unique constraint to course_id.

HassanJaveed84:


brianhw:
Just wondering if this would work for Mysql as well as for Vertica.   The sql stuff would be used for both, and I'm not sure that declaring UNIQUE inline with the column would work for Mysql.   Also, if we wanted to express uniqueness constraints as being across more than one column, then we would have to use the post-column form for declaring uniqueness constraints.    I was originally thinking we would have to do this similar to how we declare indices on a table.

HassanJaveed84:
This is ready for another review.

edx/analytics/tasks/common/vertica_load.py
```diff
@@ -447,6 +456,23 @@ def copy_data_table_from_target(self, cursor):
             else:
                 raise
 
+    def analyze_constraints(self, cursor):
+        # Vertica does not check for constraint violations during data loading.
+        # We explicitly check for violations by calling ANALYZE_CONSTRAINTS function, and fail
+        # the workflow if a voilation is found.
```
brianhw: nit:   violation

edx/analytics/tasks/common/vertica_load.py
```diff
@@ -447,6 +456,23 @@ def copy_data_table_from_target(self, cursor):
...
+        if self.foreign_key_mapping or self.unique_columns:
+            query = "SELECT ANALYZE_CONSTRAINTS('{schema}.{table}')".format(
+                schema=self.schema,
+                table=self.table
+            )
+            cursor.execute(query)
+            row = cursor.fetchone()
+            if row:
```
brianhw, line 473: nit:  is there useful information in row, out of curiosity, such as what the constraint failure was?   It might be good to include this in the exception, if it weren't too difficult...

edx/analytics/tasks/common/vertica_load.py
```diff
@@ -131,6 +131,11 @@ def default_columns(self):
         """List of tuples defining name and definition of automatically-filled columns."""
         return [('created', 'TIMESTAMP DEFAULT NOW()')]
 
+    @property
+    def unique_columns(self):
+        """List of lists containing column or group of columns on which the UNIQUE constraint would be added."""
```
brianhw: Could use tuple instead of lists.  Both work in the iterating logic below -- the for loop and the join call.  Do you have a preference?   Somehow, list of tuples looks better to me, but I've really got no good reason.   Tuples are immutable and lists are mutable, but that shouldn't really matter here.   But it would make the comment clearer:   "List of tuples, each containing .... "

edx/analytics/tasks/warehouse/load_warehouse.py
```diff
@@ -149,7 +149,8 @@ def requires(self):
                 credentials=self.credentials,
                 marker_schema=self.marker_schema,
                 overwrite=self.overwrite,
-                n_reduce_tasks=self.n_reduce_tasks
+                n_reduce_tasks=self.n_reduce_tasks,
+                warehouse_path=self.warehouse_path
```
brianhw: nit:  maybe add trailing comma?  (like below : )

brianhw (approved):
Looks great!   Just a few optional nits.  👍

edx/analytics/tasks/common/vertica_load.py
HassanJaveed84: Went with list of lists as calling ```', '.join(('c1'))``` returns ```'c, 1'``` instead of ```'c1'```
To do that with tuples we would need to set it as 
```[('c1',)] ``` Notice the trailing comma.
Do you know of a better way of achieving this ?

brianhw: Yes, that's a standard thing with tuples with one element, that they need the trailing comma.   I guess I don't have a strong opinion, but like consistency.   And looking at default_columns() above shows that it's defined as a "List of tuples....".     Again, either will work.

HassanJaveed84:
Ready for a final review before merging.

azafty468 (approved):
Yay query building.  I'm a 👍

brianhw:
Still LGTM. 👍

PR #529, opened 2018-08-03
Modify Mysql-to-Vertica copies to use same sqoop format as Mysql-to-BigQuery copies

brianhw:
This gets Vertica to use the same sqoop format.    Once this works well enough, we can then look to use the same intermediate data (i.e. warehouse_subdirectory).  Currently, Vertica runs write to "database_import" while BigQuery writes to "import_mysql_to_bq".   This switches Vertica to write to "import_mysql_to_vertica".    When we think this works, we can later switch both to the same value (e.g. "import_mysql_to_warehouse" or similar).  That way, we know the format of the data in the various subdirectories.


brianhw:
This has now been tested on Vertica, compared against the original.  I wrote a script that I ran on all of credentials and discovery, most of otto, and some of LMS. 

* It looped through all tables in a pair of schemas, finding the common tables -- few tables get added.   
* It first compared the column definitions -- few columns get changed. The code change in this PR does result in 'tinyint(1)' fields being mapped to 'boolean' instead of 'int', and this was specifically checked for.  
* It joined the table in the two schemas if the first column is 'id' -- most tables have an 'id' column as their first column.   
* The join excluded where the test table doesn't join -- the test table is older than the ref table that is regenerated daily, so this screens out the additional entries in the ref table (but not the changed entries).  
* It then ran the same join, but added a 'where' clause that checks if all other column values are equal in the table.   If adding the 'where' does not change the count of rows found, then there are no differences in that table.   This was true about half the time.  Note that no special code was needed for comparing the boolean values with the int values for fields that were originally tinyint(1) fields.   Vertica did the right thing automatically.
* If differences were found overall, the script then tested each column by adding only that column to the 'where' clause.   Many columns had the same counts -- this would find the few with changes.   Most changes were in varchar or long varchar or timestamp fields.
* If a column was found with differences, the script then queried again to output the different values to a file for inspection.  
* Manual inspection found several causes of differences:
** Some values just change over time -- things get updated.   Orders changed status, number of orders per address increased, etc.   And a lot of 'modified' timestamps were later in the reference production schema than in the test schema.
** Some "history" tables had 'id' values that were not good key values -- there were multiple rows that shared the same 'id' value.   This is to be expected of a history table, but no further work was done to match their values.
** Some string fields (i.e. 'varchar' or 'long varchar') that contain newline or return characters in the original Mysql had that translated to different values in Vertica.   In the reference production schema, these values had a '\r' translated to 'r' and '\n' translated to 'n'.   In the test schema generated using this PR, both the '\r' and '\n' characters were translated to a space.  In my opinion, this is an improvement.

azafty468 (approved):
I'm good with these changes.  Also I believe this code will use a new format for records loaded to S3?  If so then we should consider cleaning old datasets which are no longer loadable using the new script.  And in general we might benefit from having a retention policy defined for MySQL database copies.  And of course this work would be defined and captured in future ticket(s).

brianhw:
Trevor would like us to keep all MySQL data copies for the foreseeable future, even if we don't use them to load into Vertica or BigQuery.   The important part there is to keep track of the format changes, which is currently done by using different subdirectories.   But it would still require a retention policy, for moving to Glacier.    (Of course with global warming, stuff put in Glacier would likely melt....)

PR #530, opened 2018-08-27
Don't require org_id in d_course.

brianhw:
This is motivated by a bogus course on our stage server causing this to break, but in fact there is no real reason to fail if the org_id is not available.   Better that it just go through as None.


HassanJaveed84 (approved):
👍

rao-abdul-mannan (approved):

PR #532, opened 2018-09-11
Run analyze statistics for all tables in read replica schema

rao-abdul-mannan:
This PR adds the functionality to run ANALYZE_STATISTICS function in read replicas import workflow after new schema has been loaded.

  - [x] Before merging run full acceptance tests suite and provide URL for the acceptance tests run.

### DIFF
--- a/edx/analytics/tasks/warehouse/load_internal_reporting_database.py
+++ b/edx/analytics/tasks/warehouse/load_internal_reporting_database.py
@@ -255,10 +255,26 @@ class PostImportDatabaseTask(SchemaManagementTask):
     roles = luigi.ListParameter(
         config_path={'section': 'vertica-export', 'name': 'business_intelligence_team_roles'},
     )
+    tables = luigi.ListParameter(
+        default=[],
+        description="List of tables in schema."
+    )
+
+    def analyze_stats_queries(self):
+        """Provides queries to run analyze statistics for each table in schema."""
+        queries = []
+        for table in self.tables:
+            queries.append(
+                "SELECT ANALYZE_STATISTICS('{schema}.{table}');".format(
+                    schema=self.schema,
+                    table=table
+                )
+            )
+        return queries
 
     @property
     def queries(self):
-        return [
+        queries = [
             "DROP SCHEMA IF EXISTS {schema} CASCADE;".format(schema=self.schema),
             "ALTER SCHEMA {schema_loading} RENAME TO {schema};".format(
                 schema_loading=self.schema_loading,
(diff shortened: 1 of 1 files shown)


edx/analytics/tasks/warehouse/load_internal_reporting_database.py
```diff
@@ -347,12 +365,14 @@ def run(self):
                     exclude_field=self.exclude_field,
                 )
 
+        table_white_list = [table for table in self.table_list if not self.should_exclude_table(table)]
```
brianhw: nit:   we could move this up to line 352 and then loop through it there as well.

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
```diff
@@ -255,10 +255,26 @@ class PostImportDatabaseTask(SchemaManagementTask):
...
     )
+    tables = luigi.ListParameter(
+        default=None,
+        description="List of tables in schema."
+    )
+
+    def analyze_stats_queries(self):
+        """Provides queries to run analyze statistics for each table in schema."""
+        queries = []
+        for table in self.tables:
```
brianhw, line 266: nit: what happens if no tables is provided?   The default is None -- would that fail?    Note that this is only a nit because we will always be providing a table list.

brianhw (approved):
LGTM. 👍    A couple of nits, but nothing blocking.  

It would be good to confirm that the relevant acceptance test works, if there is one (as well as unit tests).

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
rao-abdul-mannan: I changed the default value to an empty list. If no table is provided, no `analyze_statistics` query will run.

rao-abdul-mannan: I don't know if there is any specific AC test so I'm going to run full AC suite ( unit tests are running ok ).

brianhw:
Still looks good.   The acceptance test I was thinking of was test_internal_reporting_database, but it's too low-level to be covered by this.

rao-abdul-mannan:
@brianhw `test_internal_reporting_database` AC test passed.  
http://jenkins-ci.analytics.edx.org/job/edx-analytics-pipeline-acceptance-test-manual/1645/

rao-abdul-mannan:
FYI, AC test suite passed.
http://jenkins-ci.analytics.edx.org/job/edx-analytics-pipeline-acceptance-test-manual/1646/

PR #534, opened 2018-10-17
ENT-1232: Unenrollments in Learner Reports 

asadiqbal08:
Unenrollments in Learner Reports [ENT-1232](https://openedx.atlassian.net/browse/ENT-1232)

As an Enterprise customer, I want to be able to have my employees unenroll in a course within the 14 day refund period so they have the opportunity to see if the course is or is not the right one for their needs.
As an edX operator, when a learner unenrolls, I want to know the enrollment date, and whether it occurred within the refund period

### DIFF
--- a/edx/analytics/tasks/enterprise/enterprise_enrollments.py
+++ b/edx/analytics/tasks/enterprise/enterprise_enrollments.py
@@ -1,11 +1,14 @@
 """Compute metrics related to user enrollments in courses"""
 
+import datetime
 import logging
+import os
 
 import luigi
 import luigi.task
 
 from edx.analytics.tasks.common.mysql_load import MysqlInsertTask
+from edx.analytics.tasks.common.pathutil import PathSetTask
 from edx.analytics.tasks.enterprise.enterprise_database_imports import (
     ImportBenefitTask, ImportConditionalOfferTask, ImportDataSharingConsentTask,
     ImportEnterpriseCourseEnrollmentUserTask, ImportEnterpriseCustomerTask, ImportEnterpriseCustomerUserTask,
@@ -16,13 +19,16 @@ from edx.analytics.tasks.insights.database_imports import (
     ImportCurrentOrderState, ImportEcommerceUser, ImportPersistentCourseGradeTask, ImportProductCatalog,
     ImportStudentCourseEnrollmentTask
 )
-from edx.analytics.tasks.insights.enrollments import OverwriteHiveAndMysqlDownstreamMixin
+from edx.analytics.tasks.insights.enrollments import (
+    CourseEnrollmentSummaryPartitionTask, OverwriteHiveAndMysqlDownstreamMixin
+)
 from edx.analytics.tasks.insights.user_activity import UserActivityTableTask
 from edx.analytics.tasks.util.decorators import workflow_entry_point
 from edx.analytics.tasks.util.hive import BareHiveTableTask, HivePartitionTask, OverwriteAwareHiveQueryDataTask
 from edx.analytics.tasks.util.record import (
     BooleanField, DateField, DateTimeField, FloatField, IntegerField, Record, StringField
 )
+from edx.analytics.tasks.util.url import ExternalURL, url_path_join
 from edx.analytics.tasks.warehouse.load_internal_reporting_course_catalog import (
     CoursePartitionTask, LoadInternalReportingCourseCatalogMixin
 )
@@ -64,6 +70,7 @@ class EnterpriseEnrollmentRecord(Record):
     current_grade = FloatField(description='')
     course_price = FloatField(description='')
     discount_price = FloatField(description='')
+    unenrollment_timestamp = DateField(description='')
 
 
 class EnterpriseEnrollmentHiveTableTask(BareHiveTableTask):
@@ -100,6 +107,34 @@ class EnterpriseEnrollmentHivePartitionTask(HivePartitionTask):
     def partition_value(self):  # pragma: no cover
         """ Use a dynamic partition value based on the date parameter. """
         return self.date.isoformat()  # pylint: disable=no-member
+
+
+class ExternalCourseEnrollmentSummaryPartitionTask(CourseEnrollmentSummaryPartitionTask):
+    date = luigi.DateParameter()
+    overwrite_n_days = None
+
+    def __init__(self, *args, **kwargs):
+        super(ExternalCourseEnrollmentSummaryPartitionTask, self).__init__(*args, **kwargs)
+
+        # Find the most recent data for the source.
+        path = url_path_join(self.warehouse_path, 'course_enrollment_summary')
+        path_targets = PathSetTask([path]).output()
+        paths = list(set([os.path.dirname(target.path) for target in path_targets]))
+        dates = [path.rsplit('/', 2)[-1] for path in paths]
+        latest_date = sorted(dates)[-1]
+        self.load_date = datetime.datetime.strptime(latest_date, "dt=%Y-%m-%d").date()
+
+    def requires(self):
+        yield self.hive_table_task
+
+        yield ExternalURL(
+            url=url_path_join(self.warehouse_path, 'course_enrollment_summary', self.partition.path_spec) + '/'
+        )
+
+    @property
+    def partition_value(self):
+        """ Use a dynamic partition value based on the date parameter. """
+        return self.load_date.isoformat()  # pylint: disable=no-member
 
 
 class EnterpriseEnrollmentDataTask(
@@ -159,7 +194,8 @@ class EnterpriseEnrollmentDataTask(
                     ecommerce_data.offer AS offer,
                     grades.percent_grade AS current_grade,
                     ecommerce_data.course_price AS course_price,
-                    ecommerce_data.discount_price AS discount_price
+                    ecommerce_data.discount_price AS discount_price,
+                    enrollment_summary.last_unenrollment_time AS unenrollment_timestamp
             FROM enterprise_enterprisecourseenrollment enterprise_course_enrollment
             JOIN enterprise_enterprisecustomeruser enterprise_user
                     ON enterprise_course_enrollment.enterprise_customer_user_id = enterprise_user.id
@@ -172,6 +208,9 @@ class EnterpriseEnrollmentDataTask(
                     ON enterprise_user.user_id = auth_user.id
             JOIN auth_userprofile user_profile
                     ON enterprise_user.user_id = user_profile.user_id
+            LEFT JOIN course_enrollment_summary enrollment_summary
+                    ON enterprise_course_enrollment.course_id = enrollment_summary.course_id
+                    AND enterprise_user.user_id = enrollment_summary.user_id
             LEFT JOIN (
                     SELECT
                         user_id,
@@ -305,7 +344,10 @@ class EnterpriseEnrollmentDataTask(
                 warehouse_path=self.warehouse_path,
                 overwrite_n_days=0,
                 date=self.date
-            )
+            ),
+            ExternalCourseEnrollmentSummaryPartitionTask(
+                date=self.date
+            ),
         )
 
         kwargs = {
--- a/edx/analytics/tasks/enterprise/tests/test_enterprise_enrollments.py
+++ b/edx/analytics/tasks/enterprise/tests/test_enterprise_enrollments.py
@@ -42,6 +42,7 @@ class TestEnterpriseEnrollmentMysqlTask(TestCase):
             'current_grade',
             'course_price',
             'discount_price',
+            'unenrollment_timestamp',
         )
         import_task = EnterpriseEnrollmentMysqlTask(
             date=datetime(2017, 1, 1), warehouse_path='/tmp/foo'
--- a/edx/analytics/tasks/tests/acceptance/fixtures/input/course_enrollment_summary
+++ b/edx/analytics/tasks/tests/acceptance/fixtures/input/course_enrollment_summary
@@ -1,0 +1,2 @@
+course-v1:edX+Testing102x+1T2017	12	no-id-professional	True	no-id-professional	2016-09-09 00:00:00	2016-09-08 00:00:00	2016-09-09 00:00:00	2016-09-09 00:00:00	2016-09-09 00:00:00
+course-v1:edX+Open_DemoX+edx_demo_course2	13	no-id-professional	True	no-id-professional	2016-03-09 00:00:00	2016-09-08 00:00:00	2016-09-09 00:00:00	2016-03-09 00:00:00	2016-10-10 00:00:00
--- a/edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py
+++ b/edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py
@@ -50,6 +50,15 @@ class EnterpriseEnrollmentAcceptanceTest(AcceptanceTestCase):
                 'user_activity_by_user_' + self.DATE
             )
         )
+        self.upload_file(
+            os.path.join(self.data_dir, 'input', 'course_enrollment_summary'),
+            url_path_join(
+                self.warehouse_path,
+                'course_enrollment_summary',
+                'dt={}'.format(self.DATE),
+                'course_enrollment_summary_{}'.format(self.DATE),
+            )
+        )
 
     def prepare_database(self, name, database):
         sql_fixture_base_url = url_path_join(self.data_dir, 'input', 'enterprise', name)
@@ -64,7 +73,7 @@ class EnterpriseEnrollmentAcceptanceTest(AcceptanceTestCase):
         """Kicks off the summary task."""
         task_params = [
             'ImportEnterpriseEnrollmentsIntoMysql',
-            '--date', self.CATALOG_DATE,
+            '--date', self.DATE,
         ]
 
         self.task.launch(task_params)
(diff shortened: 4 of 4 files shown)


saleem-latif (commented):
Looking good so far

asadiqbal08:
@brittneyexline Can you please look around the general changes in this PR?

brittneyexline (commented):
this is a solid start - next steps would be to incorporate testing your changes in the acceptance test (https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py) and run the acceptance test.

I haven't worked in this repo recently, but I believe this repo can now be set up in docker and you can run the acceptance tests there. In the past, I have run the acceptance tests on the data team's jenkins server (http://jenkins-ci.analytics.edx.org/view/All/job/edx-analytics-pipeline-acceptance-test-manual/), but you will need access from Brian Wilson for that.

Also, keep in mind that you will have to make changes to the data api (https://github.com/edx/edx-enterprise-data/tree/master/enterprise_data) to include the unenrollment timestamp.

asadiqbal08:
Thanks @brittneyexline . Yes I have requested the Brian for the access. In parallel, I did the corresponding changes in enterprise data repo [HERE](https://github.com/edx/edx-enterprise-data/pull/76)

asadiqbal08:
@brittneyexline I have run the acceptance tests on the data teams's jenkins server. Build seems me successful. You can take a look on the [build parameters](http://jenkins-ci.analytics.edx.org/view/All/job/edx-analytics-pipeline-acceptance-test-manual/1648/) and My corresponding changes in enterprise-data repo are [HERE](https://github.com/edx/edx-enterprise-data/pull/76) that we will merge later. 

And let me know is anything else is required in the PR to add or go ahead with merge ?

brittneyexline (commented):
could you update the fixtures and acceptance test to actually pull in an unenrollment timestamp value?

asadiqbal08:
@brittneyexline can you guide me why a particular acceptance test `test_enterprise_enrollment_table_generation` is failing [HERE](http://jenkins-ci.analytics.edx.org/view/All/job/edx-analytics-pipeline-acceptance-test-manual/1650/testReport/junit/edx.analytics.tasks.tests.acceptance.test_enterprise_enrollments/EnterpriseEnrollmentAcceptanceTest/test_enterprise_enrollment_table_generation/)
Either it seems to wrong with code changes ?

brittneyexline:
@asadiqbal08 it looks like the acceptance test is failing because `CourseEnrollmentEventsTask` failed.
```
Scheduled 67 tasks of which:
* 18 present dependencies were encountered:
    - 1 ExternalURL(url=s3+https://edx-analytics-test-data/credentials/acceptance-db.json)
    - 2 PathSelectionByDateIntervalTask(...)
    - 1 PullDiscoveryCourseRunsAPIData(...)
    - 14 UncheckedExternalURL(...)
* 43 ran successfully:
    - 1 CourseDataTask(...)
    - 1 CourseEnrollmentSummaryTableTask(warehouse_path=s3://edx-analytics-test-output/automation/f2c7086f821391d1ddb741986c718f7e/EnterpriseEnrollmentAcceptanceTest/warehouse)
    - 1 CoursePartitionTask(...)
    - 1 CourseTableTask(warehouse_path=s3://edx-analytics-test-output/automation/f2c7086f821391d1ddb741986c718f7e/EnterpriseEnrollmentAcceptanceTest/warehouse)
    - 1 EnterpriseEnrollmentHivePartitionTask(warehouse_path=s3://edx-analytics-test-output/automation/f2c7086f821391d1ddb741986c718f7e/EnterpriseEnrollmentAcceptanceTest/warehouse, date=2016-09-08)
    ...
* 1 failed:
    - 1 CourseEnrollmentEventsTask(...)
* 5 were left pending, among these:
    * 5 was not granted run permission by the scheduler:
        - 1 CourseEnrollmentSummaryPartitionTask(...)
        - 1 CourseEnrollmentSummaryTask(...)
        - 1 EnterpriseEnrollmentDataTask(...)
        - 1 EnterpriseEnrollmentMysqlTask(...)
        - 1 ImportEnterpriseEnrollmentsIntoMysql(...)

This progress looks :( because there were failed tasks
```
Since you introduced this new dependency, the enterprise enrollments acceptance test hasn't set up what is necessary for the task in a test environment. Take a look at https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/tests/acceptance/test_enrollments.py to see what that task needs set up. You'll also have to change the expected data returned in the acceptance to include the new field's data, not just add it to the top level fields.

asadiqbal08:
@brianhw I am working on that PR from last couple of days. Can you please look around the changes in it ? one of the existing test `test_enterprise_enrollment_table_generation` is failing. I tried to provide the data that seems to me required for the execution. You can look on the failed build [HERE](http://jenkins-ci.analytics.edx.org/view/All/job/edx-analytics-pipeline-acceptance-test-manual/1670/testReport/edx.analytics.tasks.tests.acceptance.test_enterprise_enrollments/EnterpriseEnrollmentAcceptanceTest/test_enterprise_enrollment_table_generation/?auto_refresh=false) Can you please share your thoughts on it ?

brianhw:
@asadiqbal08 Adding a dependency on Enrollment is pretty squirrelly stuff -- good work so far.  Looking at the acceptance test output failure you pointed to, the task that failed was CourseEnrollmentEventsTask, which failed because it did not find any input tracking log files in the specified interval:  interval=2018-10-22-2018-11-05.   (There is a warning "Directly processing files since 0 inputs are less than the threshold 500" that is pretty cryptic, but indicates the problem.)  You have loaded a course_enrollment_summary fixture, dated to 2016-09-08, and the hope was that it would be enough to satisfy the dependency on CourseEnrollmentSummaryPartitionTask.   However, the CourseEnrollmentSummaryTask inherits from CourseEnrollmentTask, which causes the output to be deleted and recalculated if it is not "complete" (in the run() method), and the complete() method in turn checks for the presence of a "marker" file named "_SUCCESS" next to the data file you output.   So, create that marker file as part of your setup.  Also, the interval that gets passed for the test must end in the same date as the location of the course_enrollment_summary, so the interval passed in this test should end with 2016-09-08, not with today's date.

brianhw:
Hmm, as I think about this, I'm wondering if we need to decouple this from the Enrollment workflow.  I am concerned that if this job is run before enrollment completes for the current day, then it will try to calculate values rather than using the values that were calculated in the enrollment workflow and thus prevent the enrollment workflow from doing some of its calculation.  

In particular, as we saw above, if the per-user enrollment summary data is not present, the enterprise workflow will try to run CourseEnrollmentEventsTask with no overwriting.  I think that means that it won't include yesterday's events.   So I think we really want this to use an external task that somehow loads the data but would fail if it were missing, rather than running  CourseEnrollmentEventsTask to try to calculate missing data.

edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py
```diff
@@ -62,9 +75,14 @@ def test_enterprise_enrollment_table_generation(self):
 
     def launch_task(self):
         """Kicks off the summary task."""
+        fake_param = luigi.DateIntervalParameter()
         task_params = [
             'ImportEnterpriseEnrollmentsIntoMysql',
-            '--date', self.CATALOG_DATE,
+            '--date', self.DATE,
+            '--interval', fake_param.parse(self.DATE),
+            '--marker', self.marker_path,
```
brianhw: I don't know what this particular argument does, but I don't think it has any effect on the workflow.

edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py
```diff
@@ -50,6 +54,15 @@ def setUp(self):
...
         )
+        self.upload_file(
+            os.path.join(self.data_dir, 'input', 'course_enrollment_summary'),
+            url_path_join(
+                self.warehouse_path,
+                'course_enrollment_summary',
+                'dt=' + self.DATE,
+                'course_enrollment_summary_' + self.DATE
+            )
+        )
```
brianhw, line 61: Create a file in  url_path_join(
                self.warehouse_path,
                'course_enrollment_summary',
                'dt=' + self.DATE, '_SUCCESS')

edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py
brianhw: So I would remove it.

asadiqbal08:
@brianhw I did the changes as u suggested and re run the test but now it is complaining about the `--interval` arguments in [logs](http://jenkins-ci.analytics.edx.org/view/All/job/edx-analytics-pipeline-acceptance-test-manual/1683/testReport/edx.analytics.tasks.tests.acceptance.test_enterprise_enrollments/EnterpriseEnrollmentAcceptanceTest/test_enterprise_enrollment_table_generation/). 
`launch-task: error: unrecognized arguments: --interval 2014-07-30-2016-09-08`
Does it require to introduce this parameter in some other places / repos ? 
Can you take another look around the logs and code changes ?

brianhw:
One quick point:  @brittneyexline noted that she thought "this repo can now be set up in docker and you can run the acceptance tests there."   I just wanted to confirm that this is true.  Look for the analytics section in the docker repo's readme.

edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py
```diff
@@ -26,6 +30,10 @@ def setUp(self):
         """Loads enrollment and course catalog fixtures."""
         super(EnterpriseEnrollmentAcceptanceTest, self).setUp()
 
+        self.upload_tracking_log(self.INPUT_FILE, datetime.date(2014, 7, 30))
```
brianhw: I don't understand why you need this.

edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py
```diff
@@ -17,6 +20,7 @@ class EnterpriseEnrollmentAcceptanceTest(AcceptanceTestCase):
 
     CATALOG_DATE = '2016-09-08'
     DATE = '2016-09-08'
+    INPUT_FILE = 'enrollment_trends_tracking.log'
```
brianhw: I don't understand why you need this.

edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py
```diff
@@ -26,6 +30,10 @@ def setUp(self):
         """Loads enrollment and course catalog fixtures."""
         super(EnterpriseEnrollmentAcceptanceTest, self).setUp()
 
+        self.upload_tracking_log(self.INPUT_FILE, datetime.date(2014, 7, 30))
+        self.execute_sql_fixture_file('load_auth_userprofile.sql')
+
+        self.marker_path = url_path_join(self.test_out, 'marker-{}'.format(str(time.time())))
```
brianhw: I don't understand why you need this.

edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py
brianhw: You needed to create the _SUCCESS file *in addition* to the copy of course_enrollment_summary that you had there before.   One is the data, the other is the marker that the data is complete.   Note that you could set self.marker_path = url_path_join(self.warehouse_path, 'course_enrollment_summary', 'dt=' + self.DATE, '_SUCCESS'), and then create it.  

But, once you use ExternalCourseEnrollmentSummaryPartitionTask, you no longer need this marker file.

edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py
```diff
@@ -58,17 +77,28 @@ def prepare_database(self, name, database):
 
     def test_enterprise_enrollment_table_generation(self):
         self.launch_task()
+        #self.validate_marker(self.marker_path)
```
brianhw: wrong value of marker_path

edx/analytics/tasks/enterprise/enterprise_enrollments.py
```diff
@@ -305,6 +313,10 @@ def requires(self):  # pragma: no cover
                 warehouse_path=self.warehouse_path,
                 overwrite_n_days=0,
                 date=self.date
+            ),
+            CourseEnrollmentSummaryPartitionTask(
```
brianhw: I think this needs to change.   You need to create a modification of this class that does not depend on CourseEnrollmentSummaryTask, but instead points to the output of that task.   I suggest calling it ExternalCourseEnrollmentSummaryPartitionTask.  That way, if the data is not present, the ExternalCourseEnrollmentSummaryPartitionTask will fail instead of running enrollment with the wrong parameters.  

To do this, I would create ExternalCourseEnrollmentSummaryPartitionTask by inheriting from CourseEnrollmentSummaryPartitionTask, and overriding the requires() method.   Look at the definition of ExternalCourseEnrollmentPartitionTask in insights/enrollments.py for an example of how to do this.  Note that it also disables the overwrite_n_days parameter, which is good.

Once you have done this, you will find that you no longer need to create an interval and pass it around.  It should simplify the acceptance test a little in this respect.

edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py
```diff
@@ -58,17 +77,28 @@ def prepare_database(self, name, database):
...
         self.validate_enterprise_enrollment_table()
 
     def launch_task(self):
         """Kicks off the summary task."""
+        #fake_param = luigi.DateIntervalParameter()
         task_params = [
             'ImportEnterpriseEnrollmentsIntoMysql',
-            '--date', self.CATALOG_DATE,
+            '--interval', '2014-07-30-2016-09-08',
```
brianhw: Once you use ExternalCourseEnrollmentSummaryPartitionTask, you should no longer need to pass --interval.

brianhw (changes requested):
Here is my feedback.  Hope it makes sense.

asadiqbal08:
> One quick point: @brittneyexline noted that she thought "this repo can now be set up in docker and you can run the acceptance tests there." I just wanted to confirm that this is true. Look for the analytics section in the docker repo's readme.

@brianhw Yes I know but on local machine it takes a lot of time to execute than Jenkins machine.

edx/analytics/tasks/enterprise/enterprise_enrollments.py
```diff
@@ -102,6 +107,18 @@ def partition_value(self):  # pragma: no cover
...
 
+class ExternalCourseEnrollmentSummaryPartitionTask(CourseEnrollmentSummaryPartitionTask):
+    date = luigi.DateParameter()
+    overwrite_n_days = None
+
+    def requires(self):
+        yield self.hive_table_task
+
+        yield ExternalURL(
+            url=url_path_join(self.warehouse_path, 'course_enrollment_summary', 'dt={}'.format(self.date)) + '/_SUCCESS'
```
HassanJaveed84: I think this needs to point to the directory and not to the ```_SUCCESS``` file. 
```url=url_path_join(self.warehouse_path, 'course_enrollment_summary', 'dt={}'.format(self.date)) + '/'```
or one could possibly use this:
```url=self.hive_partition_path('course_enrollment_summary', self.date)```
See: https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/util/hive.py#L66

File: edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py (hunk 2 above)
Comment (HassanJaveed84) [reply edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py:61]: Need to append a file name at the end of the remote file path. 

```
    self.upload_file(
        os.path.join(self.data_dir, 'input', 'course_enrollment_summary'),
        url_path_join(
            self.warehouse_path,
            'course_enrollment_summary',
            'dt={}'.format(self.DATE),
            'course_enrollment_summary_{}'.format(self.DATE),
        )
    )
```

File: edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py [hunk 10]
```diff
@@ -58,17 +80,24 @@ def prepare_database(self, name, database):
...
         """Kicks off the summary task."""
         task_params = [
             'ImportEnterpriseEnrollmentsIntoMysql',
-            '--date', self.CATALOG_DATE,
+            '--date', self.DATE,
         ]
 
         self.task.launch(task_params)
 
+    def validate_marker(self, marker_path):
```
HassanJaveed84: Don't think we need this.

edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py
```diff
@@ -26,9 +29,17 @@ def setUp(self):
         """Loads enrollment and course catalog fixtures."""
         super(EnterpriseEnrollmentAcceptanceTest, self).setUp()
 
+        self.execute_sql_fixture_file('load_auth_userprofile.sql')
```
HassanJaveed84: Why do we need this ?

edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py
```diff
@@ -50,6 +61,17 @@ def setUp(self):
...
         )
+        self.upload_file(
+            os.path.join(self.data_dir, 'input', 'course_enrollment_summary'),
+            url_path_join(
+                self.warehouse_path,
+                'course_enrollment_summary',
+                'dt={}'.format(self.DATE),
+            )
+        )
+        self.import_db.execute_sql_file(
```
HassanJaveed84: Same for this ?

HassanJaveed84:
I think you would also need to update ```expected_enterprise_enrollment_results``` method to include the newly added column, and then add it to the ```columns``` list in ```validate_enterprise_enrollment_table``` method.

asadiqbal08:
@HassanJaveed84 I have incorporated your suggestions in the PR. Can you take another look around it ? [HERE](http://jenkins-ci.analytics.edx.org/view/All/job/edx-analytics-pipeline-acceptance-test-manual/1690/testReport/edx.analytics.tasks.tests.acceptance.test_enterprise_enrollments/EnterpriseEnrollmentAcceptanceTest/test_enterprise_enrollment_table_generation/) is the logs now.

edx/analytics/tasks/enterprise/enterprise_enrollments.py
```diff
@@ -102,6 +107,18 @@ def partition_value(self):  # pragma: no cover
...
 
+class ExternalCourseEnrollmentSummaryPartitionTask(CourseEnrollmentSummaryPartitionTask):
+    date = luigi.DateParameter()
+    overwrite_n_days = None
+
+    def requires(self):
+        yield self.hive_table_task
+
+        yield ExternalURL(
+            url=self.hive_partition_path('course_enrollment_summary', self.date)
```
HassanJaveed84: The path needs to have the date partition spec, like dt=date.

asadiqbal08:
@brianhw @HassanJaveed84 I have updated the PR, Can you tell me is anything else missing to address in the PR? [HERE](http://jenkins-ci.analytics.edx.org/view/All/job/edx-analytics-pipeline-acceptance-test-manual/1691/testReport/edx.analytics.tasks.tests.acceptance.test_enterprise_enrollments/EnterpriseEnrollmentAcceptanceTest/test_enterprise_enrollment_table_generation/?auto_refresh=false) is the current logs.

edx/analytics/tasks/enterprise/enterprise_enrollments.py
```diff
@@ -102,6 +107,18 @@ def partition_value(self):  # pragma: no cover
...
 
+class ExternalCourseEnrollmentSummaryPartitionTask(CourseEnrollmentSummaryPartitionTask):
+    date = luigi.DateParameter()
+    overwrite_n_days = None
+
+    def requires(self):
+        yield self.hive_table_task
+
+        yield ExternalURL(
+            url=url_path_join(self.warehouse_path, 'course_enrollment_summary', 'dt={}'.format(self.date)) + '/'
```
brianhw: I would prefer that this used:             
```
  url=url_path_join(self.warehouse_path, 'course_enrollment_summary', self.partition.path_spec) + '/'
```
And defined:
```
    @property
    def partition_value(self):
        """ Use a dynamic partition value based on the date parameter. """
        return self.date.isoformat()  # pylint: disable=no-member
```

edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py
brianhw: As written, this copies the TSV to the marker_path.   You could just use self.marker_path for the destination.  And you could use the _SUCCESS fixture file you created for the source.    If you're not using it here, then you could just not add it at all to the PR.  (I cannot put a comment on it, because it has no content, so I have to say it here.)

edx/analytics/tasks/tests/acceptance/fixtures/input/course_enrollment_summary
```diff
@@ -0,0 +1,2 @@
+course-v1:edX+Testing102x+1T2017   11   verified    True    verified     2015-09-09 2016-09-08  2015-09-09  2015-09-09  2015-09-09
+course-v1:edX+Open_DemoX+edx_demo_course2   13   honor    True    honor     2015-09-09 2016-09-08  2015-09-09  2015-09-09  2015-09-09
```
brianhw: This is not required for the current functionality, but it might be good if the dates were changed so that they made more sense if someone else were to use this, especially since the fixture is not in the "enterprise" subdirectory.   The five dates at the end are first_enrollment_time, last_unenrollment_time, first_verified_enrollment_time, first_credit_enrollment_time and end_time.    The second user did not ever verify, so technically they should not have a first_verified_enrollment_time.   And end_time should be the end time of the interval when the summary was calculated, so it would be after 2016-09-08 at any rate.

brianhw (changes requested):
Almost there!

brianhw:
Oh, and there are quality issues with the includes, according to the travis builds:
ERROR: /edx/app/analytics_pipeline/analytics_pipeline/edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py Imports are incorrectly sorted.
ERROR: /edx/app/analytics_pipeline/analytics_pipeline/edx/analytics/tasks/enterprise/enterprise_enrollments.py Imports are incorrectly sorted.

edx/analytics/tasks/tests/acceptance/fixtures/input/course_enrollment_summary
asadiqbal08: so do u mean, `first_verified_enrollment_time` should be empty and `end_time` should be after 2016-09-08 in this file ? right  ?

brianhw:
Yup.

asadiqbal08:
> Oh, and there are quality issues with the includes, according to the travis builds:
> ERROR: /edx/app/analytics_pipeline/analytics_pipeline/edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py Imports are incorrectly sorted.
> ERROR: /edx/app/analytics_pipeline/analytics_pipeline/edx/analytics/tasks/enterprise/enterprise_enrollments.py Imports are incorrectly sorted.

Yeah I know, Thanks. I would fix this at the end. First problem is with the acceptance test failure over the jenkins machine :)

edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py
```diff
@@ -29,6 +32,12 @@ def setUp(self):
         self.prepare_database('lms', self.import_db)
         self.prepare_database('otto', self.otto_db)
 
+        self.marker_path = url_path_join(
```
HassanJaveed84: I think this can be remove altogether ?

edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py
HassanJaveed84: And we can also probably remove this ? We only need to create the tsv file which is copied over down below.

edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py
```diff
@@ -77,61 +89,65 @@ def expected_enterprise_enrollment_results(self):
...
+             'test_user5', 'edX+Open_DemoX', 'US', None, None, None, 'Absolute, 100 (#7)', 0.5, 200.00, 20.00,
+             datetime.datetime(2016, 9, 8, 0, 0)],
 
             ['0381d3cb033846d48a5cb1475b589d7f', 'Enterprise 1', 15, 4, 'course-v1:edX+Testing102x+1T2017',
              datetime.datetime(2016, 3, 22, 21, 8, 8), 'credit', 1, '', 0, None, 'ginny', 1,
              'All about acceptance testing Part 3!', datetime.datetime(2016, 12, 1, 0, 0),
              datetime.datetime(2017, 2, 1, 0, 0), 'instructor_paced', '9', 2, 5,
              datetime.datetime(2015, 2, 12, 23, 14, 35), 'test5@example.com', 'test_user5', 'edX+Testing102',
-             'US', None, None, None, 'Percentage, 100 (#8)', 0.74, 100.00, 60.00],
+             'US', None, None, None, 'Percentage, 100 (#8)', 0.74, 100.00, 60.00, datetime.datetime(2016, 9, 8, 0, 0)]
```
brianhw: Missing a comma here -- causes an error.

edx/analytics/tasks/tests/acceptance/fixtures/input/course_enrollment_summary
```diff
@@ -0,0 +1,2 @@
+course-v1:edX+Testing102x+1T2017	2	verified	True	verified	2015-09-09	2016-09-08	2015-09-09	2015-09-09	2015-09-09
```
HassanJaveed84: As you're joining ```course_enrollment_summary``` table with ```enterprise_enterprisecustomeruser``` based on ```user_id```. This file needs to have the same user_ids as in https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/tests/acceptance/fixtures/input/enterprise/lms/enterprise_enterprisecustomeruser.sql

edx/analytics/tasks/enterprise/enterprise_enrollments.py
HassanJaveed84: @asadiqbal08 I think this is the culprit of your failing test. You have to define ```partition_value``` 
the default implementation is 
```
    @property
    def partition_value(self):  # pragma: no cover
        """ Use a dynamic partition value based on the date parameter. """
        return self.interval.date_b.isoformat()  # pylint: disable=no-member
```
Where self.interval.date_b defaults to Today's date. You need to override it like Brian suggested so it uses the correct date(passed from the acceptance test) and not Today's date. 
The hive table ```course_enrollment_summary``` which gets loaded is actually empty if you don't override this.

edx/analytics/tasks/enterprise/enterprise_enrollments.py
```diff
@@ -172,6 +195,8 @@ def insert_query(self):
                     ON enterprise_user.user_id = auth_user.id
             JOIN auth_userprofile user_profile
                     ON enterprise_user.user_id = user_profile.user_id
+            JOIN course_enrollment_summary course_enrollment_summary
```
HassanJaveed84: I'm wondering if this should be a ```LEFT JOIN``` ?

asadiqbal08:
@brianhw @HassanJaveed84 Acceptance tests have been passed, You can see the build details [HERE](http://jenkins-ci.analytics.edx.org/view/All/job/edx-analytics-pipeline-acceptance-test-manual/1709/)
Kindly take a look around the changes, If we are good then let me know. Thanks.

edx/analytics/tasks/enterprise/enterprise_enrollments.py
```diff
@@ -185,6 +208,16 @@ def insert_query(self):
...
                     ON enterprise_course_enrollment.course_id = user_activity.course_id
                     AND enterprise_user.user_id = user_activity.user_id
+            LEFT JOIN (
+                    SELECT
+                        user_id,
+                        course_id,
+                        last_unenrollment_time
+                    FROM
+                        course_enrollment_summary
+                ) enrollment_summary
```
brianhw: Why is there a sub-select here?   Unlike user_activity_by_user, I presume that you don't have to find a MAX, but just find a single row.   So why not just join with course_enrollment_summary?

edx/analytics/tasks/enterprise/tests/test_enterprise_enrollments.py
```diff
@@ -42,6 +42,7 @@ def test_query(self):
             'current_grade',
             'course_price',
             'discount_price',
+            'unenrollment_timestamp'
```
brianhw: nit:  add a comma, so that the next addition doesn't need to change it.

edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py
```diff
@@ -155,4 +169,5 @@ def validate_enterprise_enrollment_table(self):
             results = cursor.fetchall()
 
         expected = self.expected_enterprise_enrollment_results()
+        print("TEST:" + str(results))
```
brianhw: This can be removed.

brianhw (commented):
Note that once this merges, the scheduling of the enterprise jobs will need to be made much later, to run after the regular enrollment calculations have been run for the previous day.   This can take hours.  So a question I have is whether there is a quicker way to get roughly the same information:   can someone check to see if the values that this would find for unenrollment timestamps would match the last-modified time for an entry in student_courseenrollment?    The premise is that if a student were currently unenrolled, the last modified date would be the unenrollment time if we assume that they wouldn't do anything else to modify that table once they unenrolled.    @brittneyexline  @mattdrayer ?

asadiqbal08:
@brianhw @brittneyexline Are we good with these change set ? 
Waiting for the response. Thanks. 
Other supporting --> [edx-enterprise-data-pr](https://github.com/edx/edx-enterprise-data/pull/76)

zubair-arbi:
@brianhw @asadiqbal08 what about the coverage failure? Are we missing something in tests?

zubair-arbi (approved):
Overall changes looks good to me 👍

brianhw:
I raised a question two weeks ago on which I did not get an answer, about the external dependency on enrollment calculations having been run.    Note that once this merges, the scheduling of the enterprise jobs will need to be altered.   As they currently are, they would be running before enrollment completes, and it will fail because the expected enrollment data is not present.    There are several choices:

1.  Schedule it to run much later, to run after the regular enrollment calculations have been run for the previous day. This can take hours. 
2. Modify the code to use the most recent enrollment calculations available, which most likely be yesterday's.  Or hardcode it to always use yesterday's.
3. Consider discovery into a more heuristic way to get unenrollments.   (I mentioned the possibility that the values that this would find for unenrollment timestamps would match the last-modified time for an entry in student_courseenrollment.  If they did, that would be data that would be quicker to include.)

brittneyexline:
@asadiqbal08 i think you need to choose one of @brianhw's recommendations before this can merge. the first option seems problematic from a performance standpoint, but i think options 2 and 3 seem viable, but with different tradeoffs. i would think about those tradeoffs and how difficult it would be to make either change, and make a choice from there.
fyi @georgebabey

asadiqbal08:
@brianhw As @brittneyexline highlighted the performance concern regarding the first option, I think, we should no go with that. 
From point 2 and 3, I will choose to go with option 2 as it seems me simple to handle. For option 3, We need to do more changes and may be we need to modify the tests as well at that stage. 

As you suggested the option 2 that we use the most recent enrollment calculations available, which most likely be yesterday's then I think we can avail via modifying the [partition_value](https://github.com/edx/edx-analytics-pipeline/pull/534/files#diff-7ef89a9bf47ca9cc8336e883532141c7R124) property to go one day back . 

    @property  
    def partition_value(self):  # pragma: no cover
        """ Use a dynamic partition value based on the date parameter. """
        return (self.date - timedelta(days=1)).isoformat()  # pylint: disable=no-member

Please let me know is that approach would be helpful to handle the issue ? 

@shamck @georgebabey  FYI

edx/analytics/tasks/enterprise/enterprise_enrollments.py
```diff
@@ -102,6 +107,23 @@ def partition_value(self):  # pragma: no cover
...
+        yield self.hive_table_task
+
+        yield ExternalURL(
+            url=url_path_join(self.warehouse_path, 'course_enrollment_summary', self.partition.path_spec) + '/'
+        )
+
+    @property
+    def partition_value(self):
+        """ Use a dynamic partition value based on the date parameter. """
+        return (self.date - datetime.timedelta(days=1)).isoformat()  # pylint: disable=no-member
```
brianhw: I know I said that "or you can run with yesterday's enrollment data" (or words to that effect), but I really suggest instead that we find the latest date for which we have enrollment data.   To that end I point you to https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/warehouse/load_warehouse_bigquery.py#L203-L210 as an example of how to find the latest date.   Put that into an __init__ method (with 'course_enrollment_summary' in place of 'internal_reporting_user_activity') to find the self.load_date, and then return that value here instead of the calculation you have.

brianhw (changes requested):

asadiqbal08:
@brianhw ready for final review.

edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py
```diff
@@ -6,7 +6,7 @@
 import logging
 import os
 
-from edx.analytics.tasks.tests.acceptance import AcceptanceTestCase
+from edx.analytics.tasks.tests.acceptance import AcceptanceTestCase, get_target_for_local_server
```
brianhw: nit: I don't happen to see where get_target_for_local_server is being used.  Unneeded?

brianhw (approved):
LGTM!   Just one nit.  👍

asadiqbal08:
Guys so are we go ahead with merge. @brianhw @HassanJaveed84 ?

saleem-latif (approved):
Looks Good 👍

PR #535, opened 2018-10-22
Remove conditional remap.

HassanJaveed84:


brianhw (approved):
LGTM.   👍  

But this should not be released until either the legacy jobs on Scheduler would work with this (by passing 'edx' instead of 'prod'), or they are replaced with the new jobs (that do the same).

PR #537, opened 2018-10-31
Fix acceptance test.

HassanJaveed84:


brianhw (approved):
LGTM. 👍

rao-abdul-mannan (approved):
LGTM 👍

edx/analytics/tasks/tests/acceptance/test_event_export.py
```diff
@@ -82,12 +82,13 @@ def test_event_log_exports_using_manifest(self):
...
         }
         for environment in ['prod', 'edge']:
+            site = "edx" if environment == 'prod' else environment
             self.task.launch([
                 'EventExportTask',
                 '--source', as_list_param(url_path_join(self.test_src, environment)),
                 '--output-root', self.test_out,
                 '--config', self.test_config,
-                '--environment', environment,
+                '--environment', site,
```
brianhw: Actually, a quick question.   Does it make more sense to change the environment to be ['edx', 'edge'], and to then either change the --source to use 'prod' when the environment is 'edx', or just move the source file so it's found in url_path_join(self.test_src, 'edx')?

HassanJaveed84: I've modified it as per your suggestion.

PR #702, opened 2018-12-10
Add LoadAllVideoToVertica task.

brianhw:
Provides a basic load into Vertica of video data that was calculated for Insights.  Uses external URLs, so it assumes the data is present.    

For DE-18.


rao-abdul-mannan (approved):
LGTM 👍

PR #705, opened 2019-01-01
Load insights data into warehouse

rao-abdul-mannan:
---


brianhw:
I think there's an issue with timing for all of this, that involves specifying a date for each of the tables.  It implies that when this job runs, all the insights tables for today need to have been generated.   We know that at least for some tables that this is not even knowable, because of the length and variance of the run times for jobs.   In particular, both enrollment and answer-distribution.  

So I'm not sure if what we want is a single copy-insights job that copies the latest available copy of each table, or if we distribute the copy-to-Vertica to a wrapper task that servers as the entry point we use for each individual insights job, so that they write to Vertica right after they write to Mysql.    For the former, we'd have to add find-the-latest-date to each table.  For the latter, we'd change all the prod Insights jobs to call a different entry point.  Or change all the Insights jobs to call a different entry point, but with a flag for prod that says to write out to Vertica, and no flag for release-candidate, stage, etc.

HassanJaveed84:
I think it would be cleaner just to load the latest available data unless we really do care about latency.

edx/analytics/tasks/insights/user_activity.py
```diff
@@ -203,6 +204,15 @@ def columns(self):
         ]
 
 
+class CourseActivityRecord(Record):
+    """Represents count of users performing each category of activity each ISO week."""
+    course_id = StringField(length=255, nullable=False, description='The course the learner is enrolled in.')
+    interval_start = DateTimeField(nullable=True, description='Start time of ISO week.')
+    interval_end = DateTimeField(nullable=True, description='End time of ISO week.')
```
brianhw: Just curious why these two fields happen to be nullable -- they weren't in CourseActivityPartitionTask.

edx/analytics/tasks/warehouse/load_warehouse_insights.py
```diff
@@ -0,0 +1,171 @@
...
+from edx.analytics.tasks.util.url import ExternalURL, url_path_join
+
+log = logging.getLogger(__name__)
+
+
+class LoadHiveTableToVertica(WarehouseMixin, VerticaCopyTask):
+    """
+    Generic task to load hive table into Vertica.
+    """
+    date = luigi.DateParameter()
```
brianhw: Does this need to be a parameter?   A read of the code here suggests that self.date is only used when load_data_from_partition is set, and then self.date is explicitly calculated.   Is there somewhere else that self.date is used?    If not, this could be renamed to something like self.latest_date, to make that clearer.

edx/analytics/tasks/warehouse/load_warehouse_insights.py
```diff
@@ -0,0 +1,171 @@
...
+    table_name = luigi.Parameter(
+        description='Name of hive table to load into vertica.'
+    )
+    column_list = luigi.ListParameter(
+        description='A list of column names to be included.'
+    )
+    load_data_from_partition = luigi.Parameter(
```
brianhw: nit: perhaps "load_from_latest_partition" would be more descriptive.   An alternative would be to define "load_from_all_partitions" with a default of false, but I prefer "load_from_latest_partition" as better documenting the code block in __init__ below.

edx/analytics/tasks/warehouse/load_warehouse_insights.py
```diff
@@ -0,0 +1,171 @@
...
+
+class LoadHiveTableToVertica(WarehouseMixin, VerticaCopyTask):
+    """
+    Generic task to load hive table into Vertica.
+    """
+    date = luigi.DateParameter()
+    table_name = luigi.Parameter(
+        description='Name of hive table to load into vertica.'
+    )
+    column_list = luigi.ListParameter(
```
brianhw: nit:  wonder if "sql_schema" would be clearer, since what gets passed in is what Record.get_sql_schema() returns.   It's not just a list of column names, but the full specification.

edx/analytics/tasks/warehouse/load_warehouse_insights.py
```diff
@@ -0,0 +1,171 @@
...
+            LoadHiveTableToVertica(
+                table_name='enterprise_enrollment',
+                column_list=EnterpriseEnrollmentRecord.get_sql_schema(),
+                **kwargs
+            ),
+            LoadHiveTableToVertica(
+                table_name='enterprise_user',
+                column_list=EnterpriseUserRecord.get_sql_schema(),
+                **kwargs
+            )
```
brianhw, line 180: I'm thinking it would make sense to move the Video calls here, for the two video tables that get written to Insights.  The third is sort of redundant.   They belong in 'insights' schema as well, and should probably be copied at the same time.

brianhw (commented):
Looks good!   Just a few naming questions, and the suggestion that we add video to this.

edx/analytics/tasks/insights/user_activity.py
rao-abdul-mannan: Good point.

edx/analytics/tasks/warehouse/load_warehouse_insights.py
rao-abdul-mannan: Yup, I was going to refactor that since it is only used internally.

edx/analytics/tasks/warehouse/load_warehouse_insights.py
rao-abdul-mannan: right.

edx/analytics/tasks/warehouse/load_warehouse_insights.py
rao-abdul-mannan: right, `sql_schema` is more appropriate in this context.

rao-abdul-mannan:
@brianhw I have addressed feedback & suggestion, this could use another review.

edx/analytics/tasks/warehouse/load_warehouse_insights.py
rao-abdul-mannan: Right, My intention was to write a generic task & if that works for tables in insights, then we could later on refactor video tasks as well.

brianhw (approved):
LGTM.    One last request:  I suggest that you delete load_internal_reporting_video.py altogether, since it's no longer runnable.  Once that's done and squashed, feel free to merge.   Then remove the video job when adding the insights job.   👍

PR #709, opened 2019-01-31
Use temp table for overwriting data to reduce runtime

HassanJaveed84:
Open to suggestions on how to make this better. 

```delete from table``` is quite expensive, specially with answer distribution. This mitigates it by writing output to a different table and doing a rename, which is atomic.

### CODE REVIEW

File: edx/analytics/tasks/common/mysql_load.py [hunk 1]
```diff
@@ -361,8 +374,21 @@ def run(self):
...
 
+            if self.overwrite and self.use_temp_table_for_overwrite:
+                query = "RENAME TABLE {table} to {table_old}, {table_temp} to {table}".format(
+                    table=self.table, table_old=self.table + '_old', table_temp=self.table + '_temp'
+                )
+                log.debug(query)
+                connection.cursor().execute(query)
+
             # commit only if both operations completed successfully.
             connection.commit()
```
mulby: Overall this change looks good to me.

A couple of thoughts:

- I wonder if this commit is necessary or if RENAME does an implicit commit?
- It seems like the job could be terminated near the end here, either prior to the rename or the drop and end up in a wedged state where it wouldn't work on re-run without manually intervention.

HassanJaveed84: * You're correct about the RENAME causing an implicit commit, ```connection.commit()``` should be move to an else.
* I think the drop statement can be moved to the ```finally``` block including the old and temp table so in case the job gets terminated before the rename, re-run would still work while maintaining the prior data in the original table.
```
finally:
    for table in [self.table + '_old', self.table + '_temp']:
        query = "DROP TABLE IF EXISTS {table}".format(
            table=table
        )
        log.debug(query)
    connection.close()
```

edx/analytics/tasks/common/mysql_load.py
```diff
@@ -247,11 +248,18 @@ def init_copy(self, connection):
...
-            # commit the currently open transaction before continuing with the copy.
-            query = "DELETE FROM {table}".format(table=self.table)
-            log.debug(query)
-            connection.cursor().execute(query)
+            if self.use_temp_table_for_overwrite:
+                query = "CREATE TABLE {table_temp} LIKE {table}".format(
+                    table_temp=self.table + '_temp', table=self.table
+                )
+                log.debug(query)
+                connection.cursor().execute(query)
```
brianhw: nit:  the log and execute calls are common, and could be pulled out of the if statement.

edx/analytics/tasks/common/mysql_load.py
```diff
@@ -291,8 +299,13 @@ def _execute_insert_query(self, cursor, value_list, column_names):
         # traditional python "%" operator.
         parameters = "(" + ",".join(["%s"] * num_cols) + ")"
         all_parameters = ",".join([parameters] * num_rows)
+
+        table_name = self.table
+        if self.overwrite and self.use_temp_table_for_overwrite:
+            table_name=self.table + '_temp'
```
brianhw: nit: spaces around =

edx/analytics/tasks/common/mysql_load.py
```diff
@@ -361,12 +374,28 @@ def run(self):
             # mark as complete in same transaction
             self.output().touch(connection)
 
-            # commit only if both operations completed successfully.
-            connection.commit()
+            if self.overwrite and self.use_temp_table_for_overwrite:
+                # Note that this would cause an implicit commit.
+                query = "RENAME TABLE {table} to {table_old}, {table_temp} to {table}".format(
+                    table=self.table, table_old=self.table + '_old', table_temp=self.table + '_temp'
+                )
```
brianhw, line 386: Out of curiosity, is this likely to fail or block if the table is in use (i.e. has a lock on some part of it)?  

Also, what happens if table_old already exists?  I know the "finally" statement below is supposed to delete it, but maybe it would be safer to add an additional delete-if-exists here as well?

edx/analytics/tasks/common/mysql_load.py
```diff
@@ -361,12 +374,28 @@ def run(self):
...
+
+            else:
+                # commit only if both operations completed successfully.
+                connection.commit()
         except Exception:
             connection.rollback()
             raise
         finally:
+            # Drop the tables here in case the job hits a failure.
+            for table in [self.table + '_old', self.table + '_temp']:
```
brianhw: Hmm.  This may also be overkill, but maybe we should put the drop of table_old into the main clause, so that it's dropped if it succeeds in being renamed, but persists if it somehow got created but the temp rename part of it failed?   That seems really unlikely though.  Am I overthinking this?

brianhw (commented):
Looks great!  I'm trying to think why we didn't do this before, and the open locks is the only thing that comes to mind.

edx/analytics/tasks/common/mysql_load.py
HassanJaveed84: Thinking about it, I don't think we should do anything that causes an implicit commit before the actual rename, doing a ```DROP IF EXISTS TABLE_OLD``` before ```RENAME``` would cause an explicit commit. Then, If RENAME fails for some reason, we would end up in a weird state where the marker table entry would already have been committed with the ```DROP``` statement.

edx/analytics/tasks/common/mysql_load.py
```diff
@@ -247,9 +248,15 @@ def init_copy(self, connection):
                 else:
                     raise
 
-            # Use "DELETE" instead of TRUNCATE since TRUNCATE forces an implicit commit before it executes which would
-            # commit the currently open transaction before continuing with the copy.
-            query = "DELETE FROM {table}".format(table=self.table)
+            if self.use_temp_table_for_overwrite:
+                query = "CREATE TABLE {table_temp} LIKE {table}".format(
+                    table_temp=self.table + '_temp', table=self.table
+                )
```
brianhw: Just wondering if it would be safer to delete this table first before creating it here, just to be sure we're getting an empty table and not reusing  old contents.

I'm also realizing that class IncrementalMysqlInsertTask(MysqlInsertTask) overrides the init_copy() method.  We might want to put something in there that forces it to fail if self.user_temp_table_for_overwrite is set (because incremental updates don't work with renames).

edx/analytics/tasks/common/mysql_load.py
brianhw: Good point about the implicit commit -- that makes sense.    So maybe we should do it before setting the marker table (but after we confirm that the task needs to run.

My other question was about locks and blocking.   It's probably not a big deal, because databases do this all the time, and I assume that the RENAME will be atomic, so anyone who is accessing this table through the API while this table is being renamed will get the old data before and during, and the new data afterwards.  I'm just concerned we thought about doing this before, but didn't because of some test we performed where we found that the rename failed?     My memory is not good on this.

Looking at documentation:  https://dev.mysql.com/doc/refman/5.6/en/rename-table.html "To execute RENAME TABLE, there must be no active transactions or tables locked with LOCK TABLES. With the transaction table locking conditions satisfied, the rename operation is done atomically; no other session can access any of the tables while the rename is in progress."  So does a table read by the API constitute a lock, or only write locks?

Also, "Any privileges granted specifically for a renamed table or view are not migrated to the new name. They must be changed manually." That's not an issue for us, because we're probably not granting privileges on the "temp" table anyway. But if we rename onto the existing table, does it lose its assigned privileges as if it were deleted? I guess that also doesn't matter because we don't really mess with assigning any privileges on any of the Mysql tables anyway, right?

HassanJaveed84: Yes, I think we should do it before setting the marker table. I'll make the change, just need to make sure everything is consistent, with the transactions and all. 

Regarding the lock, I tested a long running SELECT statement with a rename command at the same time. The RENAME got blocked until the SELECT query finished running, so I think it would be okay. 

And for the privileges, like you mentioned, we don't change anything.

brianhw: Sounds great!   Thanks for checking the lock!

Thoughts on the incremental task?   Just a comment?   Or a fail in an __init__ method?   That's the last detail.

brianhw (approved):
LGTM.   👍

PR #710, opened 2019-02-28
Passing source parameter from workflow entry point.

HassanJaveed84:
I'm thinking for event loading tasks we can leverage the command line passing. Segment and tracking event loading tasks both have a source parameter. So it would be easier to just pass --TrackingEventRecordDataTask-source to the task in jenkins.


brianhw (approved):
LGTM.   👍

PR #711, opened 2019-03-11
Add monitoring to financial transactions pull to catch failures

rao-abdul-mannan:
Purpose of this PR is to add some monitoring when we pull financial transactions from cybersource or paypal.

---

### DIFF
--- a/edx/analytics/tasks/warehouse/financial/cybersource.py
+++ b/edx/analytics/tasks/warehouse/financial/cybersource.py
@@ -27,6 +27,10 @@ class PullFromCybersourceTaskMixin(OverwriteOutputMixin):
     output_root = luigi.Parameter(
         description='URL of location to write output.',
     )
+    is_empty_transaction_allowed = luigi.BoolParameter(
+        default=False,
+        description='Allow empty transactions from payment processors to be processsed, default is False.'
+    )
 
     def __init__(self, *args, **kwargs):
         super(PullFromCybersourceTaskMixin, self).__init__(*args, **kwargs)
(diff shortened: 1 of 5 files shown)


edx/analytics/tasks/warehouse/financial/cybersource.py
```diff
@@ -80,6 +80,9 @@ def run(self):
             msg = "Encountered status {} on request to Cybersource for {}".format(response.status_code, self.run_date)
             raise Exception(msg)
 
+        if self.merchant_id not in response.content: # if there is no response, there will be no merchant id
+            log.error("No transactions to process.")
```
brianhw: Yes, this seems like a good way to determine whether there are any transactions appearing in the report.  

I might suggest changing the comment "is no response" to "are no transactions in the response".   If the comment is too long, it can go on the previous line.

brianhw (commented):
I assume the monitoring approach here is to log the string "No transactions to process", and then to fail the Jenkins job if this string appears in the console.  

I would suggest instead that we add a flag to the job to permit empty transactions, but by default we would fail the job if it hits this "No transactions to process" condition.  That would save us from having to clean up after the bad data that would be run through the system, but we could still set a flag if we have confirmed that no transactions were actually expected.

rao-abdul-mannan:
@brianhw yes my intention was to filter text & fail the build if it finds "No transactions to process" in console.  
Your suggestion to add a flag seems more useful ( although from JIRA ticket I assumed we were only interested in alerting ) , I will make changes to reflect them.

brianhw (approved):
LGTM.   👍

pwnage101 (approved):

doctoryes (approved):
General nits:
- I'd prefer a more direct name for the flag, such as `error_upon_empty_transaction`, defaulting to `True`.
- Even when the the theoretically named flag above is `False`, shouldn't the code log the condition? Or is there another easy way to see that an empty transaction occurred?

PR #712, opened 2019-03-27
Use 'timestamp' as Segment timestamp value.

brianhw:
Fixes DE-1408.   Updates some comments as well.

### DIFF
--- a/edx/analytics/tasks/warehouse/load_internal_reporting_events.py
+++ b/edx/analytics/tasks/warehouse/load_internal_reporting_events.py
@@ -971,6 +971,7 @@ class SegmentEventRecordDataTask(SegmentEventLogSelectionMixin, BaseEventRecordD
         return self.project_names[project_id]
 
     def _get_time_from_segment_event(self, event, key):
+        # We are not actually seeing Quality counters being triggered. (Any more?).
         try:
             event_time = event[key]
             event_time = self.normalize_time(event_time)
@@ -1022,6 +1023,8 @@ class SegmentEventRecordDataTask(SegmentEventLogSelectionMixin, BaseEventRecordD
             return None
 
     def get_event_arrival_time(self, event):
+        # This gets the arrival time from 'receivedAt', which seems to always succeed.
+        # No counters are appearing for event arrival being found from other values.
         if 'receivedAt' in event:
             return self._get_time_from_segment_event(event, 'receivedAt')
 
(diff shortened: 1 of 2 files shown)


edx/analytics/tasks/warehouse/load_internal_reporting_events.py
```diff
@@ -971,6 +971,7 @@ def _get_project_name(self, project_id):
         return self.project_names[project_id]
 
     def _get_time_from_segment_event(self, event, key):
+        # We are not actually seeing Quality counters being triggered. (Any more?).
```
bmedx: Maybe put a date on this for future reference?

edx/analytics/tasks/warehouse/load_internal_reporting_events.py
```diff
@@ -1157,12 +1172,12 @@ def mapper(self, line):
             event_type = event.get('event')
 
             if event_type is None or date_received is None:
-                # Ignore if any of the keys is None
+                # Ignore if any of the keys is None.  We shouldn't expect this to happen at all.
```
bmedx: If we no longer expect these to happen is it worth keeping around? If they start happening again that seems like it might be worth temporarily breaking on to see what changed?

edx/analytics/tasks/warehouse/load_internal_reporting_events.py
```diff
@@ -1157,12 +1172,12 @@ def mapper(self, line):
...
                 self.incr_counter(self.counter_category_name, 'Discard Tracking with missing type', 1)
                 return
 
             if event_type.startswith('/'):
```
bmedx: Noob question- Why do we ignore these?

bmedx (approved):
Just nits and questions

macdiesel (approved):

edx/analytics/tasks/warehouse/load_internal_reporting_events.py
doctoryes: It's unclear to me as a DE n00b what this comment means in relation to the purpose of the method. Perhaps change it to provide context *or* to describe this method's intent.

edx/analytics/tasks/warehouse/load_internal_reporting_events.py
```diff
@@ -1039,6 +1042,22 @@ def get_event_arrival_time(self, event):
...
+        # It is the value provided by the host system that is emitting the event,
+        # which could be a browser with a wildly different time.  The "receivedAt"
+        # timestamp, in contrast, is a Segment server, so its timestamps are
+        # presumably of better quality.  Segment then assumes that the "sentAt" and
+        # "receivedAt" times are the same, and uses that to correct the original timestamp:
+        #
+        # "timestamp" = "originalTimestamp" + ("receivedAt" - "sentAt")
+        #
+        # So try to use the timestamp, if possible, and only fall back to 'sentAt' if
+        # it is not found.
```
doctoryes, line 1060: That comment is 💯 % clear. 👍

edx/analytics/tasks/warehouse/load_internal_reporting_events.py
brianhw: I'm going to remove this.  Implicit events occur when the LMS receives requests and generates an event with the event_type equal to the URL requested.  We ignore these in Vertica tracking logs because there were so many, and it results in an explosion of possible event_type values.   On JSON events, we move the event_type to the URL column and use a constant in its stead.  But Segment events should not have this property at all.  I think this was added defensively, but it's not appropriate.

edx/analytics/tasks/warehouse/load_internal_reporting_events.py
brianhw: All this code was "experimental", and the intention was at some point to go back and look at the events, and the counters, to assess quality of the input and output data.  It seems more prudent to me to set counters and inspect them when we care, rather than breaking, unless we're really going to look at this.

doctoryes (approved):

pwnage101 (approved):
After failing tests are fixed.

Also, I'm curious where we go to see the results from the counters?  In the console log?

brianhw:
Yes, I finally noticed the failing tests too.   

And yes, to see results from counters, one has to look in the console log, or figure out how to query them on metrics.analytics.edx.org (since the counters are exported there after every job).

PR #715, opened 2019-04-03
GA360 Importing into Snowflake

pwnage101:
DE-1374

analytics-secure PR: https://github.com/edx-ops/analytics-secure/pull/234

---


edx/analytics/tasks/common/snowflake_load.py
```diff
@@ -233,20 +262,6 @@ def columns(self):
...
-        The delimiter in the data to be copied. Default is tab (\t).
-        """
-        return "\t"
-
-    @property
-    def null_marker(self):
-        """
-        The null sequence in the data to be copied. Default is Hive NULL (\\N).
-        """
-        return r'\\N'
```
pwnage101, line 248: reviewers: I just moved these two methods into an abstract subclass.

setup.cfg
```diff
@@ -75,6 +75,7 @@ edx.analytics.tasks =
     run-vertica-sql-scripts = edx.analytics.tasks.warehouse.run_vertica_sql_scripts:RunVerticaSqlScriptTask
     test-vertica-sqoop = edx.analytics.tasks.common.vertica_export:VerticaSchemaToBigQueryTask
     load-ga-permissions = edx.analytics.tasks.warehouse.load_ga_permissions:LoadGoogleAnalyticsPermissionsWorkflow
+    ga-imports = edx.analytics.tasks.warehouse.ga_imports:SnowflakeLoadGATask
```
pwnage101, line 78: I'm not really sure which class to reference here.  It seems to work as-is, despite the fact that i'm calling a different task from the command line (`SnowflakeLoadGAIntervalTask`).

edx/analytics/tasks/util/s3_util.py
```diff
@@ -125,19 +125,17 @@ def func(name):
...
 
-    def __init__(self, aws_access_key_id=None, aws_secret_access_key=None, **kwargs):
+    DEPRECATED: Just specify `host` in the `[s3]` configuration section, e.g.:
 
-        if not aws_access_key_id:
-            aws_access_key_id = self._get_s3_config('aws_access_key_id')
-        if not aws_secret_access_key:
-            aws_secret_access_key = self._get_s3_config('aws_secret_access_key')
-        if 'host' not in kwargs:
-            kwargs['host'] = self._get_s3_config('host') or 's3.amazonaws.com'
```
pwnage101, line 138: reviewers: All of this logic was removed because:

1. The aws access/secret key was duplicate code (this logic already exists inside of the luigi.contrib.s3 library).
2. The host argument  can be supplied in the luigi .cfg file instead of in code.  Besides, once we upgrade to boto3 there will be no need for setting this default.

edx/analytics/tasks/common/bigquery_extract.py
```diff
@@ -0,0 +1,312 @@
...
+        The type of compression to use for the output files.
+
+        Return None for no compression.
+        """
+        return 'GZIP'
+
+    @property
+    def output_format(self):
+        """
+        The type of compression to use for the output files.
```
HassanJaveed84, line 104: Did you mean the output format here ?

pwnage101: It's the output file format, or the compression format for the output file, same thing right?

HassanJaveed84: I just found it a little ambiguous, as both  ```output_compression_type``` and ```output_format ``` methods have the same docstrings. One is the exported file format and the other is the compression type to use for exported files. Just thought we could use  different docstrings.

pwnage101: Oh, whoops, fixing

pwnage101:
Per a conversation with @brianhw I'm going to split this up into two branches: one including the updates that don't require a newer Luigi, and one including updates that do require a newer Luigi.

pwnage101:
Closing in favor of https://github.com/edx/edx-analytics-pipeline/pull/721

PR #717, opened 2019-04-09
Snowflake vertica schema loading

HassanJaveed84:
I've specified ExternalURL dependency instead of SqoopImportFromVertica. We already export vertica to s3 when we copy BI/Finance schemas to BigQuery.

### DIFF
--- a/edx/analytics/tasks/warehouse/load_vertica_schema_to_snowflake.py
+++ b/edx/analytics/tasks/warehouse/load_vertica_schema_to_snowflake.py
@@ -1,0 +1,177 @@
+"""
+Tasks to load a Vertica schema into Snowflake.
+"""
+
+import datetime
+import logging
+import re
+
+import luigi
+from edx.analytics.tasks.common.snowflake_load import SnowflakeLoadDownstreamMixin, SnowflakeLoadFromHiveTSVTask
+from edx.analytics.tasks.common.vertica_export import get_vertica_results
+from edx.analytics.tasks.util.hive import HivePartition, WarehouseMixin
+from edx.analytics.tasks.util.url import ExternalURL, url_path_join
+
+log = logging.getLogger(__name__)
+
+
+class LoadVerticaTableFromS3ToSnowflakeMixin(WarehouseMixin):
+    """
+    Common parameters for loading a vertica table from S3 to Snowflake.
+    """
+    vertica_warehouse_name = luigi.Parameter(
+        default='warehouse',
+        description='The Vertica warehouse that houses the schema being copied.'
+    )
+    vertica_schema_name = luigi.Parameter(
+        description='The Vertica schema being copied. '
+    )
+    vertica_credentials = luigi.Parameter(
+        config_path={'section': 'vertica-export', 'name': 'credentials'},
+        description='Path to the external Vertica access credentials file.',
+    )
+
+
+class LoadVerticaTableFromS3ToSnowflake(LoadVerticaTableFromS3ToSnowflakeMixin, SnowflakeLoadFromHiveTSVTask):
+    """
+    Task to load a vertica table from S3 into Snowflake.
+    """
+
+    date = luigi.DateParameter(
+        default=datetime.datetime.utcnow().date(),
+        description='Current run date.  This parameter is used to isolate intermediate datasets.'
+    )
+    table_name = luigi.Parameter(
+        description='The Vertica table being copied.'
+    )
+
+    def snowflake_compliant_schema(self):
+        """
+        Returns the Vertica schema in the format (column name, column type, nullable?) for the indicated table.
+        """
+
+        query = """
+        SELECT column_name, data_type FROM columns
+        WHERE table_schema='{schema}' AND table_name='{table}'
+        ORDER BY ordinal_position
+        """.format(
+            schema=self.vertica_schema_name,
+            table=self.table_name
+        )
+        results = []
+        rows = get_vertica_results(self.vertica_credentials, query)
+
+        for row in rows:
+            column_name = row[0]
+            field_type = row[1]
+
+            if column_name == 'start':
+                column_name = '"{}"'.format(column_name)
+
+            if field_type.lower() in ['long varbinary']:
+                log.error('Error Vertica jdbc tool is unable to export field type \'%s\'.  This field will be '
+                          'excluded from the extract.', field_type.lower())
+                continue
+            elif field_type.lower().startswith('long'):
+                field_type = field_type.lower().lstrip('long ')
+            elif '(40' in field_type.lower():
+                field_type = field_type.rsplit('(')[0]
+
+            results.append((column_name, field_type.lower()))
+
+        return results
+
+    @property
+    def insert_source_task(self):
+        """
+        We are already exporting vertica tables to S3 using SqoopImportFromVertica through VerticaSchemaToBigQueryTask
+        workflow, so we specify ExternalURL here instead. In the future we can change this to a
+        SqoopImportFromVertica task.
+        """
+        partition_path_spec = HivePartition('dt', self.date).path_spec
+        intermediate_warehouse_path = url_path_join(self.warehouse_path, 'import/vertica/sqoop/')
+        url = url_path_join(intermediate_warehouse_path,
+                            self.vertica_warehouse_name,
+                            self.vertica_schema_name,
+                            self.table_name,
+                            partition_path_spec) + '/'
+
+        return ExternalURL(url=url)
+
+    @property
+    def table(self):
+        return self.table_name
+
+    @property
+    def file_format_name(self):
+        return 'vertica_sqoop_export_format'
+
+    @property
+    def columns(self):
+        return self.snowflake_compliant_schema()
+
+    @property
+    def null_marker(self):
+        return "NNULLL"
+
+    @property
+    def pattern(self):
+        return '.*part-m.*'
+
+    @property
+    def field_delimiter(self):
+        return r'\x01'
+
+
+class VerticaSchemaToSnowflakeTask(LoadVerticaTableFromS3ToSnowflakeMixin,
+                                   SnowflakeLoadDownstreamMixin,
+                                   luigi.WrapperTask):
+    """
+    A task that copies all the tables in a Vertica schema to S3.
+
+    Reads all tables in a schema and, if they are not listed in the `exclude` parameter, schedules a
+    LoadVerticaTableFromS3ToSnowflake task for each table.
+    """
+    date = luigi.DateParameter(
+        default=datetime.datetime.utcnow().date(),
+        description='Current run date.  This parameter is used to isolate intermediate datasets.'
+    )
+    exclude = luigi.ListParameter(
+        default=[],
+        description='The Vertica tables that are to be excluded from exporting.'
+    )
+
+    def should_exclude_table(self, table_name):
+        """
+        Determines whether to exclude a table during the import.
+        """
+        if any(re.match(pattern, table_name) for pattern in self.exclude): # pylint: disable=not-an-iterable
+            return True
+        return False
+
+    def requires(self):
+        query = "SELECT table_name FROM all_tables WHERE schema_name='{schema_name}' AND table_type='TABLE' " \
+                "".format(schema_name=self.vertica_schema_name)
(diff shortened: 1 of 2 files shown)


edx/analytics/tasks/warehouse/load_vertica_schema.py
```diff
@@ -0,0 +1,167 @@
...
+        query = "select column_name, data_type from columns where table_schema='{schema}' and " \
+                "table_name='{table}' order by ordinal_position;".format(schema=self.vertica_schema_name, table=self.table_name)
+        results = []
+        rows = get_vertica_results(self.vertica_credentials, query)
+
+        for row in rows:
+            column_name = row[0]
+            field_type = row[1]
+
+            if column_name == 'start':
```
HassanJaveed84: @doctoryes this is what I did.

edx/analytics/tasks/warehouse/load_vertica_schema_to_snowflake.py
```diff
@@ -0,0 +1,171 @@
...
+    table_name = luigi.Parameter(
+        description='The Vertica table being copied.'
+    )
+
+    def snowflake_compliant_schema(self):
+        """
+        Returns the Vertica schema in the format (column name, column type, nullable?) for the indicated table.
+        """
+
+        query = "select column_name, data_type from columns where table_schema='{schema}' and " \
```
brianhw: nit: I'd suggest capitalizing the keywords here, for readability.  And maybe even making it a multiline string.

edx/analytics/tasks/warehouse/load_vertica_schema_to_snowflake.py
```diff
@@ -0,0 +1,171 @@
...
+try:
+    import vertica_python
+    from vertica_python.errors import QueryError
+    vertica_client_available = True  # pylint: disable=invalid-name
+except ImportError:
+    log.warn('Unable to import Vertica client libraries')
+    # On hadoop slave nodes we don't have Vertica client libraries installed so it is pointless to ship this package to
+    # them, instead just fail noisily if we attempt to use these libraries.
+    vertica_client_available = False  # pylint: disable=invalid-name
+
```
brianhw: nit: extra line.

edx/analytics/tasks/common/snowflake_load.py
```diff
@@ -325,7 +325,11 @@ def init_copy(self, connection):
             ))
 
     def copy(self, connection):
-        query = "COPY INTO {database}.{schema}.{table} FROM @{database}.{schema}.{table}_stage PATTERN='{pattern}'".format(
+        query = """
+        COPY INTO {database}.{schema}.{table} FROM
+        @{database}.{schema}.{table}_stage
```
brianhw: nit: if wrapping, I'd put FROM on the same line as the stage database.

brianhw (approved):
LGTM.   👍 

(Just a few optional nits.)

HassanJaveed84:
This is ready for review.

edx/analytics/tasks/warehouse/load_vertica_schema_to_snowflake.py
pwnage101: don't we need two spaces before the first class to pass pycodestyle? (as in, I think this is fine as-is).

edx/analytics/tasks/warehouse/load_vertica_schema_to_snowflake.py
```diff
@@ -0,0 +1,175 @@
...
+from edx.analytics.tasks.common.vertica_export import get_vertica_results
+from edx.analytics.tasks.util.hive import HivePartition, WarehouseMixin
+from edx.analytics.tasks.util.url import ExternalURL, get_target_from_url, url_path_join
+from edx.analytics.tasks.common.snowflake_load import SnowflakeLoadDownstreamMixin, SnowflakeLoadTask
+
+log = logging.getLogger(__name__)
+
+try:
+    import vertica_python
+    from vertica_python.errors import QueryError
```
pwnage101: Looks like QueryError is unused.

edx/analytics/tasks/warehouse/load_vertica_schema_to_snowflake.py
```diff
@@ -0,0 +1,175 @@
...
+    def file_format_name(self):
+        return 'vertica_sqoop_export_format'
+
+    @property
+    def columns(self):
+        return self.snowflake_compliant_schema()
+
+    @property
+    def null_marker(self):
+        return "NNULLL"
```
pwnage101, line 115: is this really how the sqoop csv output defines null?

edx/analytics/tasks/warehouse/load_vertica_schema_to_snowflake.py
```diff
@@ -0,0 +1,175 @@
...
+                    schema=self.schema,
+                    table_name=table_name,
+                    vertica_schema_name=self.vertica_schema_name,
+                    vertica_warehouse_name=self.vertica_warehouse_name,
+                    vertica_credentials=self.vertica_credentials,
+                )
+
+    def complete(self):
+        # OverwriteOutputMixin changes the complete() method behavior, so we override it.
+        return all(r.complete() for r in luigi.task.flatten(self.requires()))
```
pwnage101, line 177: This is the exact same as the one implemented in luigi:

https://luigi.readthedocs.io/en/stable/_modules/luigi/task.html#WrapperTask.complete

which we call already here:

https://github.com/edx/edx-analytics-pipeline/blob/3bee07288a79f10ca7a35a34e4ed90685d850136/edx/analytics/tasks/util/overwrite.py#L47

Can't we do without this function?

edx/analytics/tasks/warehouse/load_vertica_schema_to_snowflake.py
```diff
@@ -0,0 +1,175 @@
...
+        for row in rows:
+            column_name = row[0]
+            field_type = row[1]
+
+            if column_name == 'start':
+                column_name = '"{}"'.format(column_name)
+
+            if field_type.lower() in ['long varbinary']:
+                log.error('Error Vertica jdbc tool is unable to export field type \'%s\'.  This field will be '
+                          'excluded from the extract.', field_type.lower())
```
doctoryes, line 73: The pattern I've followed here (which I copied from BQ's pattern) is to save all these fields in a `deleted_fields` object variable. Then there's a `table_description` property method which dumps out all these excluded fields to a string with other info - then uses that string as a `COMMENT=` when creating a Snowflake table.

edx/analytics/tasks/warehouse/load_vertica_schema_to_snowflake.py
```diff
@@ -0,0 +1,175 @@
...
+        default=datetime.datetime.utcnow().date(),
+        description='Current run date.  This parameter is used to isolate intermediate datasets.'
+    )
+    exclude = luigi.ListParameter(
+        default=[],
+        description='The Vertica tables that are to be excluded from exporting.'
+    )
+
+    def should_exclude_table(self, table_name):
+        """Determines whether to exclude a table during the import."""
```
doctoryes: Nit: All docstrings should be multi-line?

doctoryes (approved):

edx/analytics/tasks/warehouse/load_vertica_schema_to_snowflake.py
pwnage101: Oh, I see now that we actually are passing `--overwrite` to this job.  Can't we just mixin `OverwriteOutputMixin` into `SnowflakeLoadTask` instead of `SnowflakeLoadDownstreamMixin`?

pwnage101: I see now that we're actually passing `--overwrite` to the wrapper task.  I still don't understand why we would want to bypass the method definition in OverwriteOutputMixin.  What's the difference, whether all the required tasks `complete()` return False causing the Wrapper task to return False, or the Wrapper task itself just returns false without first consulting any required tasks, the result is the same: this method returns False.  Even if all the required tasks would return True, but `--overwrite` was supplied on the wrapper task, then we should still return False here.

It seems like defining this method override makes the logic incorrect.

HassanJaveed84: We don't set self.attempted_removal on the wrapper task. So if we don't override this method in the wrapper task it would default to the implementation in OverwriteOutputMixin which executes ```if self.overwrite and not self.attempted_removal:```, which would always return False as self.attempted_removal would be False. Luigi would most likely throw an unfullfilled dependency at runtime error when it checks whether all the tasks are complete(after executing all the tasks).

pwnage101: `if self.overwrite and not self.attempted_removal: return False` would never run for a wrapper task, maybe you were thinking it was `if self.overwrite **or** not self.attempted_removal` ?

pwnage101: if `self.attempted_removal` is always False for a wrapper task, then `self.overwrite and not self.attempted_removal` is equivalent to `self.overwrite and True` which is equivalent to `self.overwrite`.  so, for wrapper tasks, the inherited complete() implementation is effectively:

```python
if self.overwrite:
    return False
else:
    return super(OverwriteOutputMixin, self).complete()
```

When luigi schedules the wrapper task and `--overwrite` is passed to it, we force complete() to return False, and so luigi will schedule the task and any requirements.  Isn't that correct?

HassanJaveed84: I can't seem to see your most recent comment. However, I think what I wrote is not exactly correct. I'm going to give it another look. @brianhw can you recall our reasoning to override this method when we use OverwriteOutputMixin ?

pwnage101: I wouldn't block on this question, since if my concern is valid, we typically mitigate it by passing the value of overwrite to all sub-tasks from the wrapper task.

HassanJaveed84: Assuming we don't override this method here:
* Luigi first checks for wrapper task's completeness by calling ```complete()``` of the wrapper task which would be the ```complete()``` method defined in OverwriteOutputMixin
```
if self.overwrite and not self.attempted_removal:
    return False
else:
    return super(OverwriteOutputMixin, self).complete()
```
* In case ```--overwrite``` is passed it would return ```False``` as ```attempted_removal``` is False(on the wrapper task).
* Luigi schedules and executes all the required tasks.
* Luigi checks for completeness again by calling ```complete()``` of the wrapper task.
* As ```attempted_removal``` is still False on the wrapper task level, ```complete()``` returns False again. We set ```attempted_removal``` to True on all the loading tasks but never on the wrapper task.

pwnage101: Oh, I see.  I didn't think Luigi would check complete() again after executing the required tasks.

edx/analytics/tasks/warehouse/load_vertica_schema_to_snowflake.py
pwnage101: nevermind, i read somewhere that this was our own invention.

HassanJaveed84:
any objections on merging this to master?

edx/analytics/tasks/warehouse/load_vertica_schema_to_snowflake.py
brianhw: This was chosen because there were actual text fields that contained "NULL", so we arbitrarily chose a value that was unlikely to exist in actual fields.

edx/analytics/tasks/warehouse/load_vertica_schema_to_snowflake.py
brianhw: Yes, Luigi will in fact check complete(), and if it's false, it will actually retry the task once by default.   One can override this retry count, I believe, but we haven't.

brianhw:
For some reason, the build failed the quality test on Travis.   It looks fine to me, unless it is complaining about a lack of newline after importing luigi?   Otherwise, everything else looks in order.

pwnage101 (approved):

PR #718, opened 2019-04-09
Handle malformed events when filtering from export

brianhw:
Fix issue with events with a string value for its payload.   If we can't get the flag to check it, then just don't filter.

### DIFF
--- a/edx/analytics/tasks/export/event_exports.py
+++ b/edx/analytics/tasks/export/event_exports.py
@@ -101,10 +101,11 @@ class EventExportTask(EventLogSelectionMixin, MultiOutputMapReduceJobTask):
             log.debug('Unrecognized organization: org_id=%s', org_id or '')
             return
 
-        # Do not export events that have been flaged as non export
-        # Preferred value to not export is 'false'.
-        # Any event without an '_export' key will be sent as part of the export by default.
-        if str(event.get('event', {}).get('_export', 'true')).lower() in ('n', 'f', '0', 'false', 'no'):
+        # Do not export events that have been explicitly flagged as not being for export.
+        # Any event without an '_export' key will be sent as part of the export by default,
+        # and likewise any event without a falsey value. The preferred value to not export is 'false'.
+        event_data = eventlog.get_event_data(event)
+        if event_data and str(event_data.get('_export', 'true')).lower() in ('n', 'f', '0', 'false', 'no'):
             return
 
         # Check to see if the org_id is one that should be grouped with other org_ids.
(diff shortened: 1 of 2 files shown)


edx/analytics/tasks/export/event_exports.py
```diff
@@ -101,11 +101,14 @@ def mapper(self, line):
             log.debug('Unrecognized organization: org_id=%s', org_id or '')
             return
 
-        # Do not export events that have been flaged as non export
+        # Do not export events that have been flagged as non export.
```
doctoryes: Nit: `non-export`

edx/analytics/tasks/export/event_exports.py
```diff
@@ -101,11 +101,14 @@ def mapper(self, line):
...
         # Preferred value to not export is 'false'.
         # Any event without an '_export' key will be sent as part of the export by default.
-        if str(event.get('event', {}).get('_export', 'true')).lower() in ('n', 'f', '0', 'false', 'no'):
-            return
+        try:
+            if str(event.get('event', {}).get('_export', 'true')).lower() in ('n', 'f', '0', 'false', 'no'):
+                return
+        except Exception:  # pylint: disable=broad-except
+            log.debug('Unable to determine export flag for event: %s', unicode(event).encode('utf8'))
```
doctoryes: I'd prefer we started using `six.text_type()` instead of `unicode()`.

HassanJaveed84 (approved):

brianhw:
event-export-incremental-edge/182 ran successfully.  I'll update the comment and merge.   I'd prefer to do the porting to six a bit more globally than piece-meal.

edx/analytics/tasks/export/event_exports.py
```diff
@@ -101,11 +101,14 @@ def mapper(self, line):
...
-        # Any event without an '_export' key will be sent as part of the export by default.
-        if str(event.get('event', {}).get('_export', 'true')).lower() in ('n', 'f', '0', 'false', 'no'):
-            return
+        # Do not export events that have been explicitly flagged as not being for export.
+        # Any event without an '_export' key will be sent as part of the export by default,
+        # and likewise any event without a falsey value. The preferred value to not export is 'false'.
+        try:
+            if str(event.get('event', {}).get('_export', 'true')).lower() in ('n', 'f', '0', 'false', 'no'):
+                return
+        except Exception:  # pylint: disable=broad-except
```
mulby: In other places, we parse the event string as json before processing if it's a string. Maybe we need that here too?

edx/analytics/tasks/export/event_exports.py
mulby: This will send my new events to the tracking logs, I think. Since I'm sending browser events.

doctoryes:
I don't want to hold up this PR - but I feel that we as a team should be attempting to write only Python2/3-compatible code from this point forward (since we can't yet write Python3 code in this repo). It will get us all in the practice of doing so. (I, for one, need the practice 😄.)

mulby (approved):

PR #719, opened 2019-04-18
Add MySQL -> S3 -> Snowflake import via Luigi/Sqoop.

doctoryes:
Also:
- Differentiate between Snowflake DB name and MySQL DB name.
- Improve logging upon failed Snowflake query.
- Insert a date into each stage name, so different stages can point to different dates of the same DB export.

---

### DIFF
--- a/edx/analytics/tasks/warehouse/load_internal_reporting_database.py
+++ b/edx/analytics/tasks/warehouse/load_internal_reporting_database.py
@@ -9,6 +9,7 @@ import luigi
 
 from edx.analytics.tasks.common.bigquery_load import BigQueryLoadDownstreamMixin, BigQueryLoadTask, BigQueryTarget
 from edx.analytics.tasks.common.mysql_load import get_mysql_query_results
+from edx.analytics.tasks.common.snowflake_load import SnowflakeLoadDownstreamMixin, SnowflakeLoadTask, SnowflakeTarget
 from edx.analytics.tasks.common.sqoop import SqoopImportFromMysql
 from edx.analytics.tasks.common.vertica_load import SchemaManagementTask, VerticaCopyTask
 from edx.analytics.tasks.util.hive import HivePartition, WarehouseMixin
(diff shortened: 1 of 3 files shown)


brianhw (approved):
LGTM.   👍

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
```diff
@@ -605,3 +606,235 @@ def output(self):
...
+                credentials_target=credentials_target,
+                database='dummy_db',
+                schema='dummy_schema',
+                table='dummy_table',
+                role='dummy_role',
+                warehouse="dummy_wh",
+                update_id="dummy_id",
+            )
+            # Now ask it to delete the marker table completely.
+            target.delete_marker_table()
```
pwnage101, line 788: Does the `SnowflakeTarget.delete_marker_table` method even exist?

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
```diff
@@ -605,3 +606,235 @@ def output(self):
...
+        Output all the required tasks.
+        """
+        return [task.output() for task in self.requires()]
+
+    def complete(self):
+        """
+        Reduces the result of the required tasks into a single complete/not-complete.
+        """
+        # OverwriteOutputMixin changes the complete() method behavior, so we override it.
+        return all(r.complete() for r in luigi.task.flatten(self.requires()))
```
pwnage101, line 840: I have similar doubts about defining this method for the same reason as explained in: https://github.com/edx/edx-analytics-pipeline/pull/717#discussion_r276420506

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
```diff
@@ -605,3 +606,235 @@ def output(self):
...
+    def __init__(self, *args, **kwargs):
+        """
+        Inits this Luigi task.
+        """
+        super(ImportMysqlDatabaseToSnowflakeDatasetTask, self).__init__(*args, **kwargs)
+        self.table_list = []
+        self.is_complete = False
+        self.required_tasks = None
+
+        # If we are overwriting the database output, then delete the entire marker table.
```
HassanJaveed84, line 769: I thought we did this for BigQuery as a workaround for delete quotas on tables.

HassanJaveed84 (approved):
Just one comment, otherwise LGTM 👍

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
brianhw: Good catch.   That was true.   We would end up failing on BigQuery because updating the table_updates markers on each job was resulting in our going over the threshold.   I assume that Snowflake doesn't have the same limitation, so we could work this more like Vertica.

PR #720, opened 2019-04-18
Update snowflake and ga includes for isort.

brianhw:
Let's see if putting these in the right place will pass Travis (or help us figure out what Travis needs to accept these).


pwnage101 (approved):
after replacing the `pip instal -r requirements/test.txt` part of the line with `make test-requirements`

HassanJaveed84 (approved):

brianhw:
I changed the makefile to make running "quality" on docker look more like running "test" on docker.  So it cleans out the virtualenv and rebuilds it each time.   For more local docker dev, one should just be able to run "make quality-docker-local" and it will use whatever is currently in the virtualenv.

PR #721, opened 2019-04-19
Tasks for loading GA data into Snowflake (PART 1)

pwnage101:
This is the first part of the work to add a pipeline to load GA360 data into Snowflake.

This part of the code change DOES NOT depend on a Luigi upgrade, and is ready for review/merging now.

Other PRs:
* Code for Part 1: https://github.com/edx/edx-analytics-pipeline/pull/721 (this PR)
* Code for Part 2: https://github.com/edx/edx-analytics-pipeline/pull/722
* Config for Part 1: https://github.com/edx-ops/analytics-secure/pull/237
* Config for Part 2: https://github.com/edx-ops/analytics-secure/pull/238

---

  - [x] Before merging run full acceptance tests suite and provide URL for the acceptance tests run.

### DIFF
--- a/edx/analytics/tasks/common/snowflake_load.py
+++ b/edx/analytics/tasks/common/snowflake_load.py
@@ -11,6 +11,7 @@ from cryptography.hazmat.primitives import serialization
 from snowflake.connector import ProgrammingError
 
 from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
+from edx.analytics.tasks.util.s3_util import canonicalize_s3_url
 from edx.analytics.tasks.util.url import ExternalURL
 
 log = logging.getLogger(__name__)
@@ -90,9 +91,22 @@ class SnowflakeTarget(luigi.Target):
         self.create_marker_table()
 
         connection.cursor().execute(
-            """INSERT INTO {database}.{schema}.{marker_table} (update_id, target_table)
-               VALUES (%s, %s)""".format(database=self.database, schema=self.schema, marker_table=self.marker_table),
-            (self.update_id, "{database}.{schema}.{table}".format(database=self.database, schema=self.schema, table=self.table))
+            """
+            INSERT INTO {database}.{schema}.{marker_table} (update_id, target_table)
+            VALUES (%s, %s)
+            """.format(
+                database=self.database,
+                schema=self.schema,
+                marker_table=self.marker_table,
+            ),
+            (
+                self.update_id,
+                "{database}.{schema}.{table}".format(
+                    database=self.database,
+                    schema=self.schema,
+                    table=self.table,
+                ),
+            )
         )
 
         # make sure update is properly marked
@@ -112,8 +126,17 @@ class SnowflakeTarget(luigi.Target):
                 return False
 
             cursor = connection.cursor()
-            query = "SELECT 1 FROM {database}.{schema}.{marker_table} WHERE update_id='{update_id}' AND target_table='{database}.{schema}.{table}'".format(
-                database=self.database, schema=self.schema, marker_table=self.marker_table, update_id=self.update_id, table=self.table)
+            query = """
+            SELECT 1
+            FROM {database}.{schema}.{marker_table}
+            WHERE update_id='{update_id}' AND target_table='{database}.{schema}.{table}'
+            """.format(
+                database=self.database,
+                schema=self.schema,
+                marker_table=self.marker_table,
+                update_id=self.update_id,
+                table=self.table,
+            )
             log.debug(query)
             cursor.execute(query)
             row = cursor.fetchone()
@@ -135,8 +158,8 @@ class SnowflakeTarget(luigi.Target):
                 schema=self.schema,
             ))
             row = cursor.fetchone()
-        except ProgrammingError as e:
-            if "does not exist" in e.msg:
+        except ProgrammingError as err:
+            if "does not exist" in err.msg:
                 # If so then the query failed because the database or schema doesn't exist.
                 row = None
             else:
@@ -169,8 +192,14 @@ class SnowflakeTarget(luigi.Target):
         Delete all markers related to this table update.
         """
         if self.marker_table_exists(connection):
-            query = "DELETE FROM {database}.{schema}.{marker_table} where target_table='{database}.{schema}.{table}'".format(
-                database=self.database, schema=self.schema, marker_table=self.marker_table, table=self.table,
+            query = """
+            DELETE FROM {database}.{schema}.{marker_table}
+            WHERE target_table='{database}.{schema}.{table}'
+            """.format(
+                database=self.database,
+                schema=self.schema,
+                marker_table=self.marker_table,
+                table=self.table,
             )
             connection.cursor().execute(query)
 
@@ -234,20 +263,6 @@ class SnowflakeLoadTask(SnowflakeLoadDownstreamMixin, luigi.Task):
         raise NotImplementedError
 
     @property
-    def field_delimiter(self):
-        """
-        The delimiter in the data to be copied. Default is tab (\t).
-        """
-        return "\t"
-
-    @property
-    def null_marker(self):
-        """
-        The null sequence in the data to be copied. Default is Hive NULL (\\N).
-        """
-        return r'\\N'
-
-    @property
     def pattern(self):
         """
         Files matching this pattern will be used in the COPY operation.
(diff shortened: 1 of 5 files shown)


edx/analytics/tasks/common/snowflake_load.py
```diff
@@ -377,3 +382,79 @@ def output(self):
 
     def update_id(self):
         return '{task_name}(date={key})'.format(task_name=self.task_family, key=self.date.isoformat())
+
+
+class SnowflakeLoadCSVTask(SnowflakeLoadTask):  # pylint: disable=abstract-method
+    """
+    Abstract Task for loading CSV data from s3 into a table in Snowflake.
```
brianhw, line 389: CSV or TSV?

edx/analytics/tasks/util/s3_util.py
```diff
@@ -125,19 +125,17 @@ def func(name):
 class ScalableS3Client(S3Client):
     """
     S3 client that adds support for defaulting host name.
-    """
-    # TODO: Make this behavior configurable and submit this change upstream.
 
-    def __init__(self, aws_access_key_id=None, aws_secret_access_key=None, **kwargs):
+    DEPRECATED: Just specify `host` in the `[s3]` configuration section, e.g.:
```
brianhw: I don't think that this will work on task instances within an EMR cluster.   They don't have the boto or luigi config files available locally, so config must come from the host.   And so this needs to be standalone, and not dependent on config files.

brianhw (commented):
I've not done a full review, but I expect the S3 changes to not work generally.    Run acceptance tests, for example.

edx/analytics/tasks/util/s3_util.py
pwnage101: okay, didn't know that.  I was planning on figuring out how to run acceptance tests next anyway.

edx/analytics/tasks/common/snowflake_load.py
brianhw: Or HiveTSV, to be distinct from other TSVs coming from other placed (e.g. Mysql Sqoop) that use a different null marker.

edx/analytics/tasks/util/s3_util.py
```diff
@@ -162,3 +160,27 @@ def open(self, mode='r'):
             if not hasattr(self, 's3_client'):
                 self.s3_client = ScalableS3Client()
             return AtomicS3File(safe_path, self.s3_client, policy=DEFAULT_KEY_ACCESS_POLICY)
+
+
+def canonicalize_s3_url(url):
+    """
+    Convert the given s3 URL into a form which is safe to use with external tools.
+
+    Specifically, URL Schemes such as "s3+https" are urecognized by gsutil and Snowflake, and must
```
brianhw: nit: "unrecognized"

edx/analytics/tasks/common/snowflake_load.py
pwnage101: I envisioned that this would superclass any task that makes use of the "CSV" snowflake format type.  Maybe I should instead return `NotImplementedError` for the `field_delimiter` and `null_marker` properties, then define another abstract subclass called SnowflakeLoadHiveTSVTask, instead of making it arbitrarily work by default with HiveTSV.

pwnage101:
Currently running acceptance tests against this branch: http://jenkins-ci.analytics.edx.org/view/ad-hoc/job/edx-analytics-pipeline-acceptance-test-manual/1727/console

pwnage101:
Acceptance tests passed.

One test failed initially, but it was apparently flaky because I re-ran that module and it passed: http://jenkins-ci.analytics.edx.org/view/ad-hoc/job/edx-analytics-pipeline-acceptance-test-manual/1729/console

doctoryes (approved):

edx/analytics/tasks/util/s3_util.py
```diff
@@ -122,22 +122,19 @@ def func(name):
...
-        if not aws_access_key_id:
-            aws_access_key_id = self._get_s3_config('aws_access_key_id')
-        if not aws_secret_access_key:
-            aws_secret_access_key = self._get_s3_config('aws_secret_access_key')
+    def __init__(self, **kwargs):
         if 'host' not in kwargs:
             kwargs['host'] = self._get_s3_config('host') or 's3.amazonaws.com'
 
-        super(ScalableS3Client, self).__init__(aws_access_key_id=aws_access_key_id, aws_secret_access_key=aws_secret_access_key, **kwargs)
+        super(ScalableS3Client, self).__init__(**kwargs)
```
brianhw: I'm not sure this is equivalent, is it?   Will the super pull the values for access key out of kwargs, or do they need to be specified in their positional location?

edx/analytics/tasks/util/s3_util.py
```diff
@@ -162,3 +159,29 @@ def open(self, mode='r'):
             if not hasattr(self, 's3_client'):
                 self.s3_client = ScalableS3Client()
             return AtomicS3File(safe_path, self.s3_client, policy=DEFAULT_KEY_ACCESS_POLICY)
+
+
+def canonicalize_s3_url(url):
```
brianhw, line 164: Is it possible to provide some unit tests for this, since you're modifying test_s3_util.py?

edx/analytics/tasks/common/snowflake_load.py
brianhw: I like having a SnowflakeLoadFromHiveTSVTask class, because it's clearer what it's reading from.  (I would suggest "LoadFromHiveTSV" instead of "LoadHiveTSV", because the latter suggests to me that data is being loaded into Hive, not Snowflake.)  In that case, I'm not sure I see the value of defining SnowflakeLoadCSVTask -- it's a marker that you could use isinstance on, but I can't see when you would want to do that.  And when you did, you would only have to go to the definition of SnowflakeLoadFromHiveTSVTask to insert it.

edx/analytics/tasks/util/s3_util.py
pwnage101: super will pull it out of kwargs!

https://github.com/edx/luigi/blob/master/luigi/contrib/s3.py#L78

In python, explicit keyword args are sort of both positional args and keyword args at the same time, and it's kind of weird:

```python
>>> def foo(kwarg1=None, **kwargs):
...     print kwarg1
...     print kwags
... 
>>> foo(**{'kwarg1': 'hello', 'kwgarg2': 'world'})
hello
{'kwgarg2': 'world'}
>>> foo('hello', kwgarg2='world')
hello
{'kwgarg2': 'world'}
```

brianhw: But at this point, you no longer have either the key_id or the access_key in your **kwargs.   So if you pass **kwargs on to super, it won't include those values at all, if set.  You either have to put the values into **kwargs, or pass them positionally.   I'd vote for the latter.

pwnage101: Should I assume there are callers of ScalableS3Client that are passing aws_access_key_id and aws_secret_access_key positionally?  If not, then I'm pretty sure I have all the cases covered:

Case 1: We do not pass aws_access_key_id or aws_secret_access_key to ScalableS3Client at all:

1. ScalableS3Client's `__init__()` will call S3Client's `__init__()` with the following kwargs:
    * `{'host': 's3.amazonaws.com'}`
1. In S3Client's `__init__()`:
    * None will be assigned to both aws_access_key_id and aws_secret_access_key.
    * kwargs will contain only a single "host" key which will be merged into `options`.

So far so good...

Case 2: We DO pass aws_access_key_id and aws_secret_access_key to ScalableS3Client:

1. ScalableS3Client's `__init__()` will call S3Client's `__init__()` with the following kwargs:
    * `{'aws_access_key_id': 'foo', 'aws_secret_access_key': 'bar', 'host': 's3.amazonaws.com'}`
1. In S3Client's `__init__()`:
    * The variables aws_access_key_id and aws_secret_access_key will be assigned to "foo" and "bar" respectively
    * kwargs will contain only a single "host" key which will be merged into `options`.

That's also good.

pwnage101: Maybe what I'm missing is that that code was only there to preserve original function signature, which DID support positional arguments?  If that's the case then I can just throw in `*args`!

brianhw: Sorry, I think I didn't notice that you had also changed the __init__ signature, removing line 131 and adding line 133.    If you do that, then I guess you can remove line 140 and add line 137 and get the right functionality.   I thought you were just doing the latter without the former, and that wouldn't be quite correct.   Either way lines 133-136 can be removed because they exist in super().    So whichever you want to do.

brianhw (approved):
All of my remaining suggestions are optional.   👍

pwnage101:
it's actually good that I added the canonicalize_s3_url unit test because it caught what would have been a major flaw: it would have rejected "s3" as a URL scheme due to a code typo.

PR #724, opened 2019-05-09
Retry google bigquery API.

HassanJaveed84:


brianhw (approved):
LGTM.  👍 

Only question I have is if we need to add some kind of sleep statement in there, of like half a second or something.   If we had only five retries, then I'd imagine some kind of exponential backoff, but with 500 retries, we presumably don't need that.

brianhw:
Weird.  I don't see what you would have changed in this that would cause the quality job to fail.   Perhaps some global rules changed?

bmedx (approved):

bmedx:
For what it's worth, we've used the `backoff` package for this kind of thing in the past. This change is probably simple enough that it's not necessary, but backoff has a lot of features if we end up needing to do something more complicated elsewhere: https://pypi.org/project/backoff/

PR #727, opened 2019-05-21
Parse unicode chars in GPG Recipient name

hunytalk:
---

### DIFF
--- a/edx/analytics/tasks/util/encrypt.py
+++ b/edx/analytics/tasks/util/encrypt.py
@@ -106,7 +106,7 @@ def _import_key_files(gpg_instance, key_file_targets, hadoop_counter_incr_func=D
                                      "'%s' and recipient '%s' will expire in the next week",
                                      key_fingerprint, test_key["uids"])
                             for recipient in test_key["uids"]:
-                                hadoop_counter_incr_func("GPG Key for {} is near expiry".format(recipient))
+                                hadoop_counter_incr_func("GPG Key for {} is near expiry".format(recipient.encode("ascii", "ignore")))
                                 hadoop_counter_incr_func("Keys expiring")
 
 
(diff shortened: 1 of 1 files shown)


edx/analytics/tasks/util/encrypt.py
```diff
@@ -106,7 +106,7 @@ def _import_key_files(gpg_instance, key_file_targets, hadoop_counter_incr_func=D
                                      "'%s' and recipient '%s' will expire in the next week",
                                      key_fingerprint, test_key["uids"])
                             for recipient in test_key["uids"]:
-                                hadoop_counter_incr_func("GPG Key for {} is near expiry".format(recipient))
+                                hadoop_counter_incr_func("GPG Key for {} is near expiry".format(recipient.encode("ascii", "ignore")))
```
pwnage101: How about 'replace' instead of 'ignore' so that the length stays the same?

pwnage101 (approved):
Just one non-blocking comment.

edx/analytics/tasks/util/encrypt.py
hunytalk: By using replace, it replaces the unencodable unicode to a question mark ? whereas with ignore it simply ignores.

bmedx (approved):

doctoryes (approved):

edx/analytics/tasks/util/encrypt.py
brianhw: Is it possible to just put a 'u' in front of the format string -- that's what has been usually done in this code base.   So it would be:  hadoop_counter_incr_func(u"GPG Key for {} is near expiry".format(recipient))

Also, it looks like if we have the potential for recipient to contain unicode, we should fix this on line 102 as well.

brianhw (approved):
LGTM.   👍

edx/analytics/tasks/util/encrypt.py
hunytalk: @brianhw it's possible, new changes as per your recommendations.

brianhw:
(don't forget to squash... : )

PR #731, opened 2019-06-04
Added google spreadsheet loading workflow.

HassanJaveed84:
Added tasks to load data from a single google sheet to vertica. Expects a sheet key and worksheet name as parameters. Uses gspread which leverages Sheets API v4.

#### Speadsheet and Worksheet
We can fetch the spreadsheet using complete url, key(part of the url) or the Title. The google sheet needs to be shared with the service account email that is being used in order for it to be able to access. 

A spreadsheet can have multiple worksheets, we can access it through name or index. Currently, the task only pulls a single worksheet out of a spreadsheet into a vertica table named according to the worksheet name. 

#### Table name and schema
The task expects the first row of the worksheet to be the header with a skip_rows parameter to skip rows leading to the header if any.
Worksheet name is used as the table name in vertica. As a safety net, all characters which are not lower/upper alphabets or digits are replaced with underscore.

There seems to be no straightforward way of inferring column types. Pandas has a convert_objects method but it infers something like 5565105529276667903236 as a numpy float, does not infer datetime unless we pass in the column indexes which hold datetime values. 

We could manually analyze one row and deduce column types, but it won't work for edge cases where
the column in the first row has an int, and rest of the rows also include floats. 

Currently, everything is stored as a VARCHAR.

#### Format considerations
* Worksheet should be properly structured, including a header.  
* The worksheet name should only include characters that are allowed in a table name.
* Column names should be properly defined, only allowable characters. 
* It might make sense to include a data type along side the column name to help with sql schema.

### DIFF
--- a/edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
+++ b/edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
@@ -1,0 +1,230 @@
+"""
+Tasks to load google spreadsheet data into the warehouse.
+
+The spreadsheet should have the following appearance if column_types_row=True:
+
+            A                 B          C        D
+  +----------------------+-----------+--------+-------+-------------+
+1 |      timestamp       | full_name | foo    | bar   | record_date |
+  +----------------------+-----------+--------+-------+-------------+
+2 |      datetime        | string    | float  | int   |    date     |
+  +----------------------+-----------+--------+-------+-------------+
+3 | 2019-06-13T10:36:20Z | Blah Blah | 123.45 | 12345 |  2019-06-13 |
+  +----------------------+-----------+--------+-------+-------------+
+4 | 2019-06-14T00:13:01Z | Blah Blah | 12.345 | 23456 |  2019-06-14 |
+  +----------------------+-----------+--------+-------+-------------+
+...
+
+Where the first row contains column titles, the second row contains column
+types, and the third row begins the data. If column_types_row=False, the
+second row is omitted, and the data begins on the second row, and types are
+assumed to be 'string'. Possible values include:
+    * integer/int
+    * string
+    * boolean
+    * float
+    * decimal
+    * date
+    * datetime
+
+The worksheet titles are used as table names in Vertica, they can be
+128 characters long, beginning with an upper/lower alphabet or underscore,
+subsequent characters can include upper/lower alphabets, underscores and digits.
+"""
+import datetime
+import json
+import logging
+
+import luigi
+from google.auth.transport.requests import AuthorizedSession
+from google.oauth2 import service_account
+from gspread import client
+
+from edx.analytics.tasks.common.vertica_load import VerticaCopyTask, VerticaCopyTaskMixin
+from edx.analytics.tasks.util.hive import HivePartition, WarehouseMixin
+from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
+from edx.analytics.tasks.util.url import ExternalURL, get_target_from_url, url_path_join
+
+log = logging.getLogger(__name__)
+
+
+# Provides a mapping between the simple types provided by the spreadsheet author in a second header row, and the types
+# as they should be loaded into Vertica.
+DATA_TYPE_MAPPING = {
+    'integer': 'INT',
+    'int': 'INT',
+    'string': 'VARCHAR(500)',
+    'boolean': 'BOOLEAN',
+    'float': 'FLOAT',
+    'decimal': 'DECIMAL(12,2)',
+    'date': 'DATE',
+    'datetime': 'DATETIME',
+}
+
+def create_google_spreadsheet_client(credentials_target):
+    with credentials_target.open('r') as credentials_file:
+        json_creds = json.load(credentials_file)
+        credentials = service_account.Credentials.from_service_account_info(
+            json_creds,
+            scopes=['https://www.googleapis.com/auth/drive.readonly']
+        )
+
+    authed_session = AuthorizedSession(credentials)
+    return client.Client(None, authed_session)
+
+
+class PullWorksheetMixin(OverwriteOutputMixin):
+    date = luigi.DateParameter(default=datetime.datetime.utcnow().date())
+    google_credentials = luigi.Parameter(description='Path to the external access credentials file.')
+    # spreadsheet key/id is a 44 characters alphanumeric including hyphen and underscore extracted from the URL.
+    spreadsheet_key = luigi.Parameter(description='Google sheets key.')
+    worksheet_name = luigi.Parameter(description='Worksheet name.', default=None)
+    column_types_row = luigi.BoolParameter(default=False, description='Whether there is a row with column types.')
+
+
+class PullWorksheetDataTask(PullWorksheetMixin, WarehouseMixin, luigi.Task):
+    """
+    Task to read data from a google worksheet and write to a tsv file.
+    """
+
+    def requires(self):
+        return {
+            'credentials': ExternalURL(url=self.google_credentials),
+        }
+
+    def run(self):
+        self.remove_output_on_overwrite()
+        credentials_target = self.input()['credentials']
+        gs = create_google_spreadsheet_client(credentials_target)
+        sheet = gs.open_by_key(self.spreadsheet_key)
+        worksheet = sheet.worksheet(self.worksheet_name)
+        all_values = worksheet.get_all_values()
+        # Remove the header/column names.
+        self.header = [v.strip() for v in all_values.pop(0)]
+        if self.column_types_row:
+            self.types = [v.strip() for v in all_values.pop(0)]
+        else:
+            self.types = ['string'] * len(self.header)
+
+        with self.output().open('w') as output_file:
+            for value in all_values:
+                output_file.write('\t'.join([v.encode('utf-8') for v in value]))
+                output_file.write('\n')
+
+    @property
+    def columns(self):
+        return self.header
+
+    @property
+    def column_types(self):
+        return self.types
+
+    def output(self):
+        partition_path_spec = HivePartition('dt', self.date).path_spec
+
+        output_worksheet_name = self.worksheet_name
+        output_url = url_path_join(self.warehouse_path, 'google_sheets', self.spreadsheet_key,
+                                   output_worksheet_name, partition_path_spec, '{}.tsv'.format(output_worksheet_name))
+        return get_target_from_url(output_url)
+
+
+class LoadWorksheetToWarehouse(PullWorksheetMixin, VerticaCopyTask):
+    """
+    Task to load data from a google sheet into the Vertica data warehouse.
+    """
+
+    @property
+    def copy_null_sequence(self):
+        """
+        The null sequence in the data to be copied. Empty string is Vertica's default.
+        """
+        return "''"
+
+    def create_table(self, connection):
+        # Drop the table in case of overwrite
+        if self.overwrite:
+            connection.cursor().execute("DROP TABLE IF EXISTS {schema}.{table}".format(
+                                        schema=self.schema, table=self.table))
+        super(LoadWorksheetToWarehouse, self).create_table(connection)
+
+    def init_copy(self, connection):
+        # We have already dropped the table, so we do away with the delete here.
+        self.attempted_removal = True
+
+    @property
+    def insert_source_task(self):
+        return PullWorksheetDataTask(
+            date=self.date,
+            google_credentials=self.google_credentials,
+            spreadsheet_key=self.spreadsheet_key,
+            worksheet_name=self.worksheet_name,
+            column_types_row=self.column_types_row,
+            overwrite=self.overwrite,
+        )
+
+    @property
+    def table(self):
+        return self.worksheet_name
+
+    @property
+    def auto_primary_key(self):
+        return None
+
+    @property
+    def default_columns(self):
+        return None
+
+    @property
+    def columns(self):
+        columns = self.insert_source_task.columns
+        column_types = self.insert_source_task.column_types
+        mapped_types = [DATA_TYPE_MAPPING.get(column_type) for column_type in column_types]
+        return zip(columns, mapped_types)
+
+
+class LoadGoogleSpreadsheetsToWarehouseWorkflow(luigi.WrapperTask):
+    """
+    Provides entry point for loading a google spreadsheet into the warehouse.
+    All worksheets within the spreadsheet are loaded as separate tables.
+    """
+
+    spreadsheets_config = luigi.DictParameter(
+        config_path={'section': 'google-spreadsheets', 'name': 'config'},
+        description='A dictionary containing spreadsheets config where a key is the spreadsheet key/id extracted from '
+                    'spreadsheet url, value is a dictionary containing atleast schema key/value pair which specifies '
+                    'the vertica schema for the spreadsheet tables. Can also specify column_types_row key where the '
+                    'value is either true or false specifying whether the worksheets in the spreadsheet contain a '
+                    'column types row as a second header row.'
+    )
+    google_credentials = luigi.Parameter(
+        description='Path to the external access credentials file.'
+    )
+    overwrite = luigi.BoolParameter(
+        default=False,
+        description='Whether or not to overwrite S3 outputs and the warehouse tables.',
+        significant=False
+    )
+    date = luigi.DateParameter(
+        default=datetime.datetime.utcnow().date()
+    )
+
+    def requires(self):
+        credentials_target = ExternalURL(url=self.google_credentials).output()
+        gs = create_google_spreadsheet_client(credentials_target)
+        for spreadsheet_key, config in self.spreadsheets_config.items():
+            schema = config['schema']
+            column_types_row = config.get('column_types_row', False)
+
+            spreadsheet = gs.open_by_key(spreadsheet_key)
+            worksheets = spreadsheet.worksheets()
+
+            for worksheet in worksheets:
+                yield LoadWorksheetToWarehouse(
+                    date=self.date,
+                    schema=schema,
+                    google_credentials=self.google_credentials,
+                    spreadsheet_key=spreadsheet_key,
+                    worksheet_name=worksheet.title,
+                    column_types_row=column_types_row,
+                    overwrite=self.overwrite,
+                )
--- a/requirements/default.in
+++ b/requirements/default.in
@@ -13,6 +13,7 @@ filechunkio==1.8	# MIT
 google-cloud-bigquery==0.27.0
 google-api-python-client==1.7.7
 graphitesend==0.10.0    # Apache
+gspread==3.1.0    # MIT
 html5lib==1.0b3 	# MIT
 isoweek==1.3.3		# BSD
 numpy==1.11.3 		# BSD
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -11,19 +11,19 @@
 ansible==1.4.5
 argparse==1.2.1
 asn1crypto==0.24.0
-azure-common==1.1.18      # via azure-storage-blob, azure-storage-common, snowflake-connector-python
+azure-common==1.1.21      # via azure-storage-blob, azure-storage-common, snowflake-connector-python
 azure-nspkg==3.0.2        # via azure-common, azure-storage-nspkg
-azure-storage-blob==1.5.0  # via snowflake-connector-python
-azure-storage-common==1.4.0  # via azure-storage-blob
+azure-storage-blob==2.0.1  # via snowflake-connector-python
+azure-storage-common==2.0.0  # via azure-storage-blob
 azure-storage-nspkg==3.1.0  # via azure-storage-common
 backports-abc==0.5        # via tornado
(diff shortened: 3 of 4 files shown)


edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
```diff
@@ -0,0 +1,109 @@
...
+from edx.analytics.tasks.util.url import ExternalURL, get_target_from_url, url_path_join
+from google.auth.transport.requests import AuthorizedSession
+from google.oauth2 import service_account
+from gspread import client
+
+log = logging.getLogger(__name__)
+
+
+def identifier_from_string(s):
+        return re.sub('[^a-zA-Z0-9_]+', '_', s) # Replace all other characters including spaces with underscore.
```
pwnage101: nits:
* This line is indented 8 spaces, but should only be 4.
* Put the comment in the function docstring, then also explain what this is useful for, e.g. "Google sheet titles and column names can have spaces in them, but we cannot use space characters in warehouse identifiers."

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
```diff
@@ -0,0 +1,109 @@
...
+
+class PullGoogleSheetMixin(WarehouseMixin, OverwriteOutputMixin):
+    google_credentials = luigi.Parameter(description='Path to the external access credentials file.')
+    spreadsheet_key = luigi.Parameter(description='Google sheets key.')
+    worksheet_name = luigi.Parameter(description='Worksheet name.')
```
pwnage101: Can you add a comment providing examples of what a google sheets key or a worksheet name looks like?  Are they 40 characters hexadecimal?  Are they unicode with spaces?

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
```diff
@@ -0,0 +1,109 @@
...
+            google_credentials=self.google_credentials,
+            spreadsheet_key=self.spreadsheet_key,
+            worksheet_name=self.worksheet_name,
+            skip_rows=self.skip_rows,
+            overwrite=self.overwrite,
+        )
+
+    @property
+    def table(self):
+        return identifier_from_string(self.worksheet_name)
```
pwnage101: Are worksheet names subject to renaming from the Google webUI?  Shouldn't the destination table name be configurable so that if the source table changes names, we can just update the configuration without renaming the warehouse table?

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
```diff
@@ -0,0 +1,109 @@
...
+
+        output_worksheet_name = identifier_from_string(self.worksheet_name)
+        output_url = url_path_join(self.warehouse_path, 'google_sheets', self.spreadsheet_key,
+                        output_worksheet_name, partition_path_spec, '{}.tsv'.format(output_worksheet_name))
+        return get_target_from_url(output_url)
+
+
+class LoadGoogleSheetToWarehouse(PullGoogleSheetMixin, VerticaCopyTask):
+    """
+    Task to load data from a google sheet into the Vertica datawarehouse.
```
doctoryes: Nit: `datawarehouse`

HassanJaveed84:
Ready for a final review.

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
```diff
@@ -0,0 +1,166 @@
...
+    'float': 'FLOAT',
+    'date': 'DATE',
+    'datetime': 'DATETIME',
+}
+
+
+class PullWorksheetMixin(WarehouseMixin, OverwriteOutputMixin):
+    google_credentials = luigi.Parameter(description='Path to the external access credentials file.')
+    # spreadsheet key/id is a 44 characters alphanumeric including hyphen and underscore extracted from the URL.
+    spreadsheet_key = luigi.Parameter(description='Google sheets key i.')
```
pwnage101: typo?: "key i."

pwnage101:
Also, there's no more `identifier_from_string()` function, are you still replacing characters with underscore in worksheet names?

HassanJaveed84:
According to the format shared in the JIRA story, the column names and worksheet names are already appropriately named, that is why I removed the identifier_from_string function. If we think we still need it, I can put it back but I thought it makes more sense to get an error incase something is not appropriately named and fix in the source(spreadsheet).

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
```diff
@@ -0,0 +1,166 @@
...
+from edx.analytics.tasks.util.hive import HivePartition, WarehouseMixin
+from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
+from edx.analytics.tasks.util.url import ExternalURL, get_target_from_url, url_path_join
+
+log = logging.getLogger(__name__)
+
+
+DATA_TYPE_MAPPING = {
+    'string': 'VARCHAR',
+    'float': 'FLOAT',
```
brianhw, line 58: I wonder if we might also need a DECIMAL, in addition to a FLOAT.   Decimal works better for money, while I expect that float would be needed for some of the percentages.     (An empirical question, of course.  But just while I'm thinking of it.)   Also I'm really surprised that INT isn't being used yet.

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
```diff
@@ -0,0 +1,166 @@
+"""
+Tasks to load google spreadsheet data into the warehouse.
+"""
```
pwnage101, line 33: Can you put an example of a spreadsheet in the docstring? e.g.
```python
"""
Tasks to load google spreadsheet data into the warehouse.

The spreadsheet should have the following appearance if column_types_row=True:

            A                 B          C        D
  +----------------------+-----------+--------+-------+
1 |      timestamp       | full_name | foo    | bar   |
  +----------------------+-----------+--------+-------+
2 |      datetime        | string    | float  | int   |
  +----------------------+-----------+--------+-------+
3 | 2019-06-13T10:36:20Z | Blah Blah | 123.45 | 12345 |
  +----------------------+-----------+--------+-------+
4 | 2019-06-14T00:13:01Z | Blah Blah | 12.345 | 23456 |
  +----------------------+-----------+--------+-------+
  ...

Where the first row contains column titles, the second row contains column
types, and the third row begins the data.  if column_types_row=False, the
second row is omitted, and the data begins on the second row.
"""
```

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def requires(self):
+        return {
+            'credentials': ExternalURL(url=self.google_credentials),
+        }
+
+    def run(self):
+        self.remove_output_on_overwrite()
+        credentials_target = self.input()['credentials']
+        gs = self.create_google_spreadsheet_cient(credentials_target)
```
brianhw: cient => client

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
```diff
@@ -0,0 +1,166 @@
...
+
+log = logging.getLogger(__name__)
+
+
+DATA_TYPE_MAPPING = {
+    'string': 'VARCHAR',
+    'float': 'FLOAT',
+    'date': 'DATE',
+    'datetime': 'DATETIME',
+}
```
pwnage101, line 62: Should this support integer too?

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+class PullWorksheetMixin(WarehouseMixin, OverwriteOutputMixin):
+    google_credentials = luigi.Parameter(description='Path to the external access credentials file.')
+    # spreadsheet key/id is a 44 characters alphanumeric including hyphen and underscore extracted from the URL.
+    spreadsheet_key = luigi.Parameter(description='Google sheets key i.')
+    worksheet_name = luigi.Parameter(description='Worksheet name.', default=None)
+    column_types_row = luigi.BoolParameter(default=False, description='Whether there is a row with column types.')
+
+    def create_google_spreadsheet_cient(self, credentials_target):
```
bmedx: Nit: typo "cient"

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
```diff
@@ -0,0 +1,166 @@
...
+    'datetime': 'DATETIME',
+}
+
+
+class PullWorksheetMixin(WarehouseMixin, OverwriteOutputMixin):
+    google_credentials = luigi.Parameter(description='Path to the external access credentials file.')
+    # spreadsheet key/id is a 44 characters alphanumeric including hyphen and underscore extracted from the URL.
+    spreadsheet_key = luigi.Parameter(description='Google sheets key i.')
+    worksheet_name = luigi.Parameter(description='Worksheet name.', default=None)
+    column_types_row = luigi.BoolParameter(default=False, description='Whether there is a row with column types.')
```
brianhw, line 82: I might mention here (or somewhere else?) that if it's not specified here, the default is to assume all String values.

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
brianhw: cient => client

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def create_google_spreadsheet_cient(self, credentials_target):
+        with credentials_target.open('r') as credentials_file:
+            json_creds = json.load(credentials_file)
+            credentials = service_account.Credentials.from_service_account_info(
+                json_creds,
+                scopes=['https://www.googleapis.com/auth/drive.readonly'])
```
brianhw: style nit: paren on the next line?

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
```diff
@@ -0,0 +1,166 @@
...
+    def default_columns(self):
+        return None
+
+    @property
+    def columns(self):
+        columns = self.insert_source_task.columns
+        column_types = self.insert_source_task.column_types
+        mapped_types = [DATA_TYPE_MAPPING.get(column_type) for column_type in column_types]
+        return zip(columns, mapped_types)
+        # return [(column,'VARCHAR') for column in self.insert_source_task.columns]
```
pwnage101: Nit: delete dead code

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+class LoadGoogleSpreadsheetToWarehouseWorkflow(PullWorksheetMixin, VerticaCopyTaskMixin, luigi.WrapperTask):
```
brianhw: I mentioned this in a related PR, but as add-on work, we might want to also define a LoadGoogleSpreadsheetsToWarehouseWorkflow, where spreadsheet_keys is a list.   It would just loop through and call this.   If we find that we need it, it can go in a separate PR.

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
```diff
@@ -0,0 +1,166 @@
...
+        # Remove the header/column names.
+        self.header = all_values.pop(0)
+        if self.column_types_row:
+            self.types = all_values.pop(0)
+        else:
+            self.types = ['string'] * len(self.header)
+
+        with self.output().open('w') as output_file:
+            for value in all_values:
+                output_file.write('\t'.join([v.encode('utf-8') for v in value]))
```
bmedx, line 111: Are all of the values here guaranteed to be strings?

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
```diff
@@ -0,0 +1,166 @@
...
+
+class LoadGoogleSpreadsheetToWarehouseWorkflow(PullWorksheetMixin, VerticaCopyTaskMixin, luigi.WrapperTask):
+    """
+    Provides entry point for loading a google spreadsheet into the warehouse.
+    All worksheets within the spreadsheet are loaded as separate tables.
+    """
+
+    def requires(self):
+        credentials_target = ExternalURL(url=self.google_credentials).output()
+        gs = self.create_google_spreadsheet_cient(credentials_target)
```
brianhw: cient => client

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
brianhw: nit: remove comment

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
bmedx: Nevermind, I see in the docs that they explicitly are. 👍

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
```diff
@@ -0,0 +1,166 @@
...
+
+from edx.analytics.tasks.common.vertica_load import VerticaCopyTask, VerticaCopyTaskMixin
+from edx.analytics.tasks.util.hive import HivePartition, WarehouseMixin
+from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
+from edx.analytics.tasks.util.url import ExternalURL, get_target_from_url, url_path_join
+
+log = logging.getLogger(__name__)
+
+
+DATA_TYPE_MAPPING = {
```
brianhw, line 53: I would put a comment here to indicate what this maps between.   In particular, it maps between the simple types provided by the Spreadsheet author in a second header row, and the types as they should be loaded into *Vertica*.

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
HassanJaveed84: Thank you.

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
```diff
@@ -0,0 +1,166 @@
...
+    'string': 'VARCHAR',
```
brianhw: I think I found that if I just define a column as VARCHAR in Vertica, it gets defaulted to a length of 80.    We might want the default to be more than this.   255?   Something even bigger?

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
```diff
@@ -0,0 +1,185 @@
...
+  +----------------------+-----------+--------+-------+-------------+
+3 | 2019-06-13T10:36:20Z | Blah Blah | 123.45 | 12345 |  2019-06-13 |
+  +----------------------+-----------+--------+-------+-------------+
+4 | 2019-06-14T00:13:01Z | Blah Blah | 12.345 | 23456 |  2019-06-14 |
+  +----------------------+-----------+--------+-------+-------------+
+...
+
+Where the first row contains column titles, the second row contains column
+types, and the third row begins the data.  if column_types_row=False, the
+second row is omitted, and the data begins on the second row.
```
brianhw: And the types are assumed to be "string".    It would be good to list the possible values for types as well.

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
brianhw: Ah, you added INT.

brianhw (approved):
Mostly nits and requests for documentation.   LGTM. 👍

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
brianhw: Also, BOOLEAN?

pwnage101 (approved):
LGTM, with brian's comments addressed.

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+class LoadGoogleSpreadsheetsToWarehouseWorkflow(luigi.WrapperTask):
+    """
+    Provides entry point for loading a google spreadsheet into the warehouse.
+    All worksheets within the spreadsheet are loaded as separate tables.
+    """
+
+    spreadsheets_config = luigi.DictParameter(
+        config_path={'section': 'google-spreadhsheets', 'name': 'config'},
```
brianhw: spreadhsheets -> spreadsheets

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
```diff
@@ -0,0 +1,205 @@
...
+        description='A dictionary containing spreadsheets config.'
```
brianhw: I think the dictionary is a good approach here.  It would be good to document it a little more, particularly with what it can and must contain.  Perhaps in this description string.

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
```diff
@@ -0,0 +1,205 @@
...
+    )
+    date = luigi.DateParameter(
+        default=datetime.datetime.utcnow().date()
+    )
+
+    def requires(self):
+        credentials_target = ExternalURL(url=self.google_credentials).output()
+        gs = self.create_google_spreadsheet_client(credentials_target)
+        for spreadsheet_key, config in self.spreadsheets_config.items():
+            title = config['title']
```
brianhw: This isn't used.   Is it required?   I think this fails if it's not present, since you're not using gets.

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
```diff
@@ -0,0 +1,205 @@
...
+            schema = config['schema']
+            column_types_row = config['column_types_row']
```
brianhw: same thing -- should this default to None if not defined?

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
```diff
@@ -0,0 +1,205 @@
...
+
+            spreadsheet = gs.open_by_key(spreadsheet_key)
+            worksheets = spreadsheet.worksheets()
```
brianhw, line 219: Ah, I see.  I was imagining that the worksheets-per-spreadsheet would be in a separate task that could also be called directly, if only so one could test it without needing config, and do one-offs and the like.   But I don't feel strongly about that.  One would need yet another client call in that task, and I have no idea if that's an important optimization to avoid.  As it is, we have that task at the worksheet level, and it just means that one needs to know the tab title.   Which is not unreasonable for such needs.  Okay.

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
```diff
@@ -0,0 +1,224 @@
...
+from edx.analytics.tasks.common.vertica_load import VerticaCopyTask, VerticaCopyTaskMixin
+from edx.analytics.tasks.util.hive import HivePartition, WarehouseMixin
+from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
+from edx.analytics.tasks.util.url import ExternalURL, get_target_from_url, url_path_join
+
+log = logging.getLogger(__name__)
+
+
+# Provides a mapping between the simple types provided by the spreadsheet author in a second header row, and the types
+# as they should be loaded into Veritca.
```
brianhw: nit: Veritca => Vertica.

edx/analytics/tasks/warehouse/load_google_sheet_to_warehouse.py
```diff
@@ -0,0 +1,224 @@
...
+  +----------------------+-----------+--------+-------+-------------+
+3 | 2019-06-13T10:36:20Z | Blah Blah | 123.45 | 12345 |  2019-06-13 |
+  +----------------------+-----------+--------+-------+-------------+
+4 | 2019-06-14T00:13:01Z | Blah Blah | 12.345 | 23456 |  2019-06-14 |
+  +----------------------+-----------+--------+-------+-------------+
+...
+
+Where the first row contains column titles, the second row contains column
+types, and the third row begins the data. If column_types_row=False, the
+second row is omitted, and the data begins on the second row, and types are
```
brianhw, line 20: I would mention here also that the workbook titles are used as table names, and so they cannot contain embedded spaces.   Better yet, it would be good to list what characters are permitted in table names/workbook titles.

PR #735, opened 2019-06-14
Add Affiliate Window pulls to finance reporting task

bmedx:
---

### DIFF
--- a/edx/analytics/tasks/warehouse/financial/affiliate_window.py
+++ b/edx/analytics/tasks/warehouse/financial/affiliate_window.py
@@ -39,8 +39,8 @@ class AffiliateWindowTaskMixin(OverwriteOutputMixin):
         description='The parent folder to write the Affiliate Window report data to.',
     )
     run_date = luigi.DateParameter(
-        default=datetime.datetime.utcnow().date() - datetime.timedelta(days=-1),
-        description='The date to generate a report for. Default is yesterday, UTC.',
+        default=datetime.datetime.utcnow().date(),
+        description='The date to generate a report for. Default is today, UTC.',
     )
     advertiser_id = luigi.Parameter(
         config_path={'section': 'affiliate_window', 'name': 'advertiser_id'},
@@ -65,9 +65,9 @@ class IntervalPullFromAffiliateWindowTask(AffiliateWindowTaskMixin, WarehouseMix
     Determines a set of dates to pull, and requires them.
     """
     interval_end = luigi.DateParameter(
-        default=datetime.datetime.utcnow().date() - datetime.timedelta(days=-1),
+        default=datetime.datetime.utcnow().date(),
         significant=False,
-        description='Default is yesterday, UTC.',
+        description='Default is today, UTC.',
     )
 
     # Overwrite parameter definition to make it optional.
@@ -137,7 +137,7 @@ class DailyPullFromAffiliateWindowTask(AffiliateWindowTaskMixin, luigi.Task):
 
     Pulls are made for only a single day.
     """
-    run_date = luigi.DateParameter(
+    query_date = luigi.DateParameter(
         default=datetime.date.today(),
         description='Default is today.',
     )
@@ -151,8 +151,8 @@ class DailyPullFromAffiliateWindowTask(AffiliateWindowTaskMixin, luigi.Task):
     @retry(should_retry=should_retry, timeout=DEFAULT_TIMEOUT_SECONDS)
     def fetch_report(self):
         params = (
-            ('startDate', self.run_date.strftime('%Y-%m-%dT00:00:00')),
-            ('endDate', self.run_date.strftime('%Y-%m-%dT23:59:59')),
+            ('startDate', self.query_date.strftime('%Y-%m-%dT00:00:00')),
+            ('endDate', self.query_date.strftime('%Y-%m-%dT23:59:59')),
             ('timezone', 'UTC'),
             ('accessToken', self.api_token),
         )
@@ -164,7 +164,7 @@ class DailyPullFromAffiliateWindowTask(AffiliateWindowTaskMixin, luigi.Task):
         return response.json()
 
     def run(self):
-        print("Fetching report from Affiliate Window for {}.".format(self.run_date))
+        print("Fetching report from Affiliate Window for {}.".format(self.query_date))
         transactions = self.fetch_report()
 
         # if there are no transactions in response something is wrong.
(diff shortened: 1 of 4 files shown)


edx/analytics/tasks/warehouse/financial/affiliate_window.py
```diff
@@ -0,0 +1,305 @@
...
+        # if there are no transactions in response something is wrong.
+        if not transactions:
+            raise Exception('No transactions to process.')
+
+        with self.output().open('w') as output_file:
+            json.dump(transactions, output_file)
+
+    def output(self):
+        """
+        Output is in the form {output_root}/affiliate_window/{CCYY-mm}/affiliate_window_{CCYYmmdd}.csv
```
pwnage101: Is it not a json file?

edx/analytics/tasks/warehouse/financial/fees.py
```diff
@@ -0,0 +1,100 @@
...
+        # Affiliate Window reports for each day are stored in dated directories.
+        # We want to be able to load all that data into Vertica in one go, hence we use
+        # a wildcard('*') here.
+        url = url_path_join(self.warehouse_path, 'fees', 'affiliate_window') + '/dt=*/'
+        return ExternalURL(url=url).output()
+
+
+class LoadFeesToWarehouse(WarehouseMixin, VerticaCopyTask):
+    """
+    Loads Affiliate Window table from Hive into the Vertica data warehouse.
```
pwnage101: Nit: the class name suggests a general purpose, but the docstring treats Affiliate Window as the main purpose.  Is the class correctly named?

edx/analytics/tasks/warehouse/financial/finance_reports.py
```diff
@@ -6,6 +6,7 @@
 from edx.analytics.tasks.warehouse.financial.ed_services_financial_report import (
     LoadInternalReportingEdServicesReportToWarehouse
 )
+from edx.analytics.tasks.warehouse.financial.fees import LoadFeesToWarehouse
```
brianhw: nit: This is unused.

edx/analytics/tasks/warehouse/financial/affiliate_window.py
```diff
@@ -0,0 +1,264 @@
...
+            dates = [path.rsplit('/', 2)[-1] for path in paths]
+            latest_date = sorted(dates)[-1]
+            latest_completion_date = datetime.datetime.strptime(latest_date, "dt=%Y-%m-%d").date()
+            self.interval_start = latest_completion_date + datetime.timedelta(days=1)
+            print("Found previous reports to {}".format(latest_date))
+        else:
+            # If this is the first run, start from the beginning
+            print("Couldn't find last completed date, defaulting to start date: {}".format(self.interval_start))
+
+        self.run_interval = date_interval.Custom(self.interval_start, self.interval_end)
```
brianhw, line 101: I'm assuming that this is a closed interval, since interval_end is defined as today's date minus one day.  Maybe just note that in the description of interval_end, because most of our intervals defined elsewhere are open on the end.   That is, we define the interval as ending with today's date, so that it processes events (mostly events) up to midnight as of today.

edx/analytics/tasks/warehouse/financial/affiliate_window.py
```diff
@@ -0,0 +1,264 @@
...
+
+class DailyPullFromAffiliateWindowTask(AffiliateWindowTaskMixin, luigi.Task):
+    """
+    A task that pulls from the Affiliate Window REST API and writes to a file.
+
+    Pulls are made for only a single day.
+    """
+    run_date = luigi.DateParameter(
+        default=datetime.date.today(),
+        description='Default is today.',
```
brianhw, line 142: nit: should default be yesterday?

edx/analytics/tasks/warehouse/financial/affiliate_window.py
```diff
@@ -0,0 +1,264 @@
...
+    return error_response.status_code in DEFAULT_RETRY_STATUS_CODES
+
+
+class DailyPullFromAffiliateWindowTask(AffiliateWindowTaskMixin, luigi.Task):
+    """
+    A task that pulls from the Affiliate Window REST API and writes to a file.
+
+    Pulls are made for only a single day.
+    """
+    run_date = luigi.DateParameter(
```
brianhw: nit:   it's not actually the run_date, that is, the date the pull was run.  It's the date being pulled.   Would "pull_date" or "query_date" be better?

edx/analytics/tasks/warehouse/financial/affiliate_window.py
```diff
@@ -0,0 +1,264 @@
...
+        response = requests.get(url, params=params)
+        response.raise_for_status()
+
+        return response.json()
+
+    def run(self):
+        print("Fetching report from Affiliate Window for {}.".format(self.run_date))
+        transactions = self.fetch_report()
+
+        # if there are no transactions in response something is wrong.
```
brianhw, line 170: Just curious -- is it possible that there are days with nothing to pull?  Actually, if there were no transactions, would this actually be a json structure that is empty, but for which "if not transactions" would somehow be false?   Or would response.json() really be empty?

edx/analytics/tasks/warehouse/financial/fees.py
```diff
@@ -0,0 +1,100 @@
...
+    """
+    run_date = luigi.DateParameter()
+
+    # Forcing this to overwrite now, as this script will duplicate data
+    # without it. This does not trickle down to the Affiliate Window calls.
+    # TODO: Make this optional once VerticaCopyTask handles multiple input files.
```
brianhw, line 48: nit: I understand that overwrite=True here means that the Vertica table is being overwritten each time this runs.  I don't understand what it means for VerticaCopyTask to handle multiple input files.   (And maybe I don't have to. : )

brianhw (approved):
Only nits -- looks great!   👍

edx/analytics/tasks/warehouse/financial/affiliate_window.py
bmedx: I went back and forth on this a bit as I was trying to figure out the Affiliate Window approval process, I think I can adjust all of this to just be open on the end so I'll make that change wherever necessary in this PR.

edx/analytics/tasks/warehouse/financial/affiliate_window.py
bmedx: There were no days with empty transactions going back to 2019-01-01 so I'm assuming that doesn't happen. The `if not` should catch any case where there is valid JSON that is an empty string, or empty list. If it's not valid JSON it will error getting it, and if it's something other than a list of dicts it should fail downstream.

If there really are no transactions the API returns an empty list, so that should mean we passed in an invalid date, or our advertiser id changed or some other weird situation like that.

edx/analytics/tasks/warehouse/financial/fees.py
bmedx: Yeah, the comment is just bad. I envisioned this being one place to load all of those random fees we're currently getting from pasted spreadsheets, as we become able to operationalise them. I'll fix the comment.

edx/analytics/tasks/warehouse/financial/affiliate_window.py
bmedx: The other open end changes should make this ok.

edx/analytics/tasks/warehouse/financial/fees.py
bmedx: It's entirely possible that I have a misapprehension about how it works, I've made a note to talk to you about this when I get back and either clear it up or make a ticket to make things work the way I think they should.

PR #736, opened 2019-06-17
Add new f_orderline_transactions table to finance-report workflow

brianhw:
The main differences of f_orderline_transactions versus f_orderitem_transactions is the addition of four new columns and the removal of one.  The additions are "order_product_course_id" (the value of course_id stored in the product, rather than as an attribute value), "order_lms_user_id" (as stored in ecommerce_user), "order_partner_sku" (as requested long ago), and "order_status".   Also, the set of orders that are included in the table are slightly expanded in two ways:  one is the inclusion of orderlines for products that don't have "parent" products (so these are standalone and missing products), and second is the inclusion of orders that don't have a status of "Complete".  (If you don't want them, you can filter them using the new "order_status" column.)

Also adds loads of the orders tables that go into the f_orderline_transactions tables, for better auditing (as two separate tables: "full_otto_orders" and "full_shoppingcart_orders").  

By piggy-backing on the finance-report workflow, no new DSL job needs to be created, or coordinated with the finance-report job.


brianhw:
This also contains a fix for an acceptance test I broke (when updating a fixture).

brianhw:
Financial_reports acceptance test passed, which includes running all of these additional tasks.

pwnage101 (commented):
I skimmed the whole thing, and didn't see any red flags.  Maybe add a comment where you're piggy-backing with the three extra tasks?

pwnage101 (approved):
meant to approve.

PR #740, opened 2019-06-19
Screen out duplicate transactions with the same ID

brianhw:
Also requires they have the same "payment_ref_id", since dupes should generally be identical.

Addresses DE-1570.


brianhw:
This ran the finance-reports successfully, and resulted in the duplicate transactions disappearing from f_orderline_transactions, both same-day and different-day duplicates.

HassanJaveed84 (approved):

PR #267, opened 2016-07-06
pin the version of setuptools

mulby:
@HassanJaveed84 and @brianhw this is the hotfix I applied over the weekend

PR #741, opened 2019-06-25
Remove some joins from orders for finance.

brianhw:
Joins on entitlements and orderdiscounts were resulting in duplicates orders, because those tables themselves contained duplicates.  Rather than trying to add deduplication logic here, move it instead into financial reporting.  The columns provide additional metadata about order_lines, but are not actually involved in mapping order_lines to transactions.

Note that this only affects f_orderline_transactions, and has no effect on the older f_orderitem_transactions table.


brianhw:
@mattdrayer @jfinkiel  FYI.

brianhw:
Finance-report jenkins run #267 succeeded, and columns are removed.   Please review.

pwnage101 (approved):
LGTM, I generally favor this approach since, as you say, the extra information is not required for mapping transactions and order lines anyway.

PR #743, opened 2019-07-10
WL-1947 | Learner Progress Report not updating enterprise_sso_uid field with new UID value

HammadAhmadWaqas:
updated query to get latest enterprise_sso_uid from social_auth_users socialauth corresponding to enterprise_user

---

### DIFF
--- a/edx/analytics/tasks/enterprise/enterprise_enrollments.py
+++ b/edx/analytics/tasks/enterprise/enterprise_enrollments.py
@@ -231,16 +231,15 @@ class EnterpriseEnrollmentDataTask(
                     ON enterprise_user.user_id = grades.user_id
                     AND enterprise_course_enrollment.course_id = grades.course_id
             LEFT JOIN (
-                    SELECT
-                        user_id,
-                        provider,
-                        MAX(uid) AS uid_full
-                    FROM
-                        social_auth_usersocialauth
-                    WHERE
-                        provider = 'tpa-saml'
-                    GROUP BY
-                        user_id, provider
+                    SELECT user_id, uid AS uid_full
+                    FROM social_auth_usersocialauth sauth_a
+                    INNER JOIN (
+                        SELECT MAX(id) AS id
+                        FROM social_auth_usersocialauth
+                        WHERE provider = 'tpa-saml'
+                        GROUP BY user_id
+                        ) sauth_b
+                            ON sauth_a.id = sauth_b.id
                 ) social_auth
                     ON enterprise_user.user_id = social_auth.user_id
             JOIN course_catalog course
--- a/edx/analytics/tasks/enterprise/enterprise_user.py
+++ b/edx/analytics/tasks/enterprise/enterprise_user.py
@@ -117,16 +117,15 @@ class EnterpriseUserDataTask(
             ) user_activity
                 ON enterprise_user.user_id = user_activity.user_id
             LEFT JOIN (
-                SELECT
-                    user_id,
-                    provider,
-                    MAX(uid) AS uid_full
-                FROM
-                    social_auth_usersocialauth
-                WHERE
-                    provider = 'tpa-saml'
-                GROUP BY
-                    user_id, provider
+                SELECT user_id, uid AS uid_full
+                    FROM social_auth_usersocialauth sauth_a
+                    INNER JOIN (
+                        SELECT MAX(id) AS id
+                        FROM social_auth_usersocialauth
+                        WHERE provider = 'tpa-saml'
+                        GROUP BY user_id
+                        ) sauth_b
+                            ON sauth_a.id = sauth_b.id
             ) social_auth
                 ON enterprise_user.user_id = social_auth.user_id
         """
--- a/edx/analytics/tasks/tests/acceptance/fixtures/input/enterprise/lms/social_auth_usersocialauth.sql
+++ b/edx/analytics/tasks/tests/acceptance/fixtures/input/enterprise/lms/social_auth_usersocialauth.sql
@@ -20,5 +20,6 @@ CREATE TABLE `social_auth_usersocialauth` (
 INSERT INTO `social_auth_usersocialauth` VALUES
   (1,11,'tpa-saml','ent1:harry', '{}'),
   (2,12,'tpa-saml','ent1:ron', '{}'),
-  (3,13,'tpa-saml','ent2:hermione', '{}'),
-  (4,15,'tpa-saml','ent1:ginny', '{}');
+  (3,13,'tpa-saml','ent2:ron', '{}'),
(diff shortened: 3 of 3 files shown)


brianhw:
Apologies for the drive-by, but there aren't too many PR's that come into this repo.  According to WL-1947, what's wanted is the latest uid value for a given user.  The existing code made a stab at doing this by doing a GROUP BY and then pulled out the MAX(uid) value.  That might work sometimes, but I assume the ordering of uid values is completely random, so it would often not work.  (Hence the ticket.)  However, at the moment the PR is resulting in only one user getting a uid at all, because of the LIMIT 1 statement -- only the user with the highest id value.   To do what you want, I think you may need to use a windowing analytics function -- you would probably need to calculate a rank for each uid, and then select the top rank.  

```
LEFT JOIN (
    SELECT
	user_id,
	uid
    FROM (
	SELECT
	    user_id,
	    uid,
            RANK () OVER (
                PARTITION BY user_id, provider
                ORDER BY id DESC
            ) AS uid_rank
	FROM
	    social_auth_usersocialauth
	WHERE
	    provider = 'tpa-saml'
        ) ranked_social_auth
    WHERE
	uid_rank = 1
) social_auth
ON enterprise_user.user_id = social_auth.user_id
```

HammadAhmadWaqas:
@brianhw thank you for looking into this matter.

Before reading your comment I had also written a new solution. I have pushed it now.
could you please review updated query and provide your reviews which solution is the best fit for this problem.

edx/analytics/tasks/enterprise/enterprise_user.py
```diff
@@ -117,18 +117,17 @@ def insert_query(self):
...
+                    from social_auth_usersocialauth sauth_a
+                    INNER JOIN (
+                            SELECT MAX(id) AS id
+                            FROM social_auth_usersocialauth
+                            WHERE provider='tpa-saml'
+                            GROUP BY user_id, provider
+                            ) sauth_b
+                                ON sauth_a.id=sauth_b.id
+                ) social_auth
+                    ON enterprise_user.user_id = social_auth.user_id
```
brianhw: nit: Note that when you cut-and-paste the first solution to here, you might want to adjust the indentation to match the surrounding code.

edx/analytics/tasks/enterprise/enterprise_enrollments.py
```diff
@@ -171,7 +171,7 @@ def insert_query(self):
                         ELSE 1
                     END AS has_passed,
                     grades.passed_timestamp,
-                    SUBSTRING_INDEX(social_auth.uid_full, ':', -1) AS enterprise_sso_uid,
+                    SUBSTRING_INDEX(social_auth.uid, ':', -1) AS enterprise_sso_uid,
```
brianhw: nit: is there a reason to change this from uid_full to uid?   I think the use of uid_full was an effort to indicate that the full value was greater than the ultimate uid extracted from it.   Just assuming it was added for a reason in the first place.

edx/analytics/tasks/enterprise/enterprise_enrollments.py
```diff
@@ -231,16 +231,15 @@ def insert_query(self):
...
-                    FROM
-                        social_auth_usersocialauth
-                    WHERE
-                        provider = 'tpa-saml'
-                    GROUP BY
-                        user_id, provider
+                    SELECT user_id, provider, uid
+                    from social_auth_usersocialauth sauth_a
+                    INNER JOIN (
+                            SELECT MAX(id) AS id
```
brianhw: Yes, this approach should also work, and I don't have a sense of which would be more performant.   And this code could be made to work if you add the user_id here as well.   Note that you do not need to group or output the provider, since you've specified that in the WHERE clause.   So this line should be "SELECT user_id, MAX(id) as id".

edx/analytics/tasks/enterprise/enterprise_enrollments.py
```diff
@@ -231,16 +231,15 @@ def insert_query(self):
...
-                        user_id,
-                        provider,
-                        MAX(uid) AS uid_full
-                    FROM
-                        social_auth_usersocialauth
-                    WHERE
-                        provider = 'tpa-saml'
-                    GROUP BY
-                        user_id, provider
+                    SELECT user_id, provider, uid
```
brianhw: this would be SELECT user_id, uid as uid_full,

edx/analytics/tasks/enterprise/enterprise_enrollments.py
```diff
@@ -231,16 +231,15 @@ def insert_query(self):
...
+                    from social_auth_usersocialauth sauth_a
+                    INNER JOIN (
+                            SELECT MAX(id) AS id
+                            FROM social_auth_usersocialauth
+                            WHERE provider='tpa-saml'
+                            GROUP BY user_id, provider
+                            ) sauth_b
+                                ON sauth_a.id=sauth_b.id
```
brianhw: This join would need to be on id and also on user_id.   The result should be the max id value for each user. 

nits on the indenting of this line, and the lowercase "from" on 235.

brianhw (commented):
Promising!

edx/analytics/tasks/enterprise/enterprise_enrollments.py
HammadAhmadWaqas: My assumption was alias uid_full used to save MAX of uid before. But your logic also makes sense. I am going to revert this change

edx/analytics/tasks/enterprise/enterprise_enrollments.py
HammadAhmadWaqas: Yes, not using `provider`. so removing it from SELECT

edx/analytics/tasks/enterprise/enterprise_enrollments.py
HammadAhmadWaqas: Yes, We have specified it in WHERE clause so removing it from GROUP BY too.

edx/analytics/tasks/enterprise/enterprise_enrollments.py
HammadAhmadWaqas: As per my understanding, we are using id to make the join which is a unique field. So why do we need to add another condition here?
also, we are using user_id in outer join in line 245.

HammadAhmadWaqas (commented):
@brianhw I have made suggested changes, Could you please review it again.

brianhw (approved):
I understand your query now -- very nice!  Using MAX() allows use of group by instead of rank, and as you say, the id is unique so it is enough to join on it.

LGTM.   👍

ziafazal (approved):
👍

PR #746, opened 2019-07-17
Add python version, update Makefile

hunytalk:
--- Python 3 code changes for DE Pipelines

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,9 @@ reset-virtualenv:
 	# without bash, environment variables are not available
 	bash -c 'virtualenv --clear ${ANALYTICS_PIPELINE_VENV}/analytics_pipeline'
 
+reset-virtualenv-py3:
+	bash -c 'virtualenv --clear --python=$$(which python3) ${ANALYTICS_PIPELINE_VENV}/analytics_pipeline'
+
 upgrade: ## update the requirements/*.txt files with the latest packages satisfying requirements/*.in
 	pip install -qr requirements/pip-tools.txt
 	CUSTOM_COMPILE_COMMAND="make upgrade" pip-compile --upgrade -o requirements/pip-tools.txt requirements/pip-tools.in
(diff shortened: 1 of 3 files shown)


edx/analytics/tasks/launchers/remote.py
```diff
@@ -45,6 +45,7 @@ def main():
     parser.add_argument('--workflow-profiler', choices=['pyinstrument'], help='profiler to run on the launch-task process', default=None)
     parser.add_argument('--wheel-url', help='url of the wheelhouse', default=None)
     parser.add_argument('--virtualenv-extra-args', help='additional arguments passed to virtualenv command when creating the virtual environment', default=None)
+    parser.add_argument('--python-version', help='Python version that will be used to run pipelines task', default=None)
```
pwnage101: The help text will need an example.  e.g., `3` or `3.6` or `/usr/bin/python3.6`?

hunytalk: Ok will add that.

Makefile
```diff
@@ -45,6 +45,9 @@ reset-virtualenv:
 	# without bash, environment variables are not available
 	bash -c 'virtualenv --clear ${ANALYTICS_PIPELINE_VENV}/analytics_pipeline'
 
+reset-virtualenv-py3:
+	bash -c 'virtualenv --clear --python=$$(which python3.5) ${ANALYTICS_PIPELINE_VENV}/analytics_pipeline'
```
doctoryes: Why the `python3.5` here instead of `python3`?

hunytalk: I'll revert to python3.

edx/analytics/tasks/launchers/remote.py
```diff
@@ -48,6 +48,9 @@ def main():
     parser.add_argument('--skip-setup', action='store_true', help='assumes the environment has already been configured and you can simply run the task')
     parser.add_argument('--package', action='append', help='pip install these packages in the pipeline virtual environment')
     parser.add_argument('--extra-repo', action='append', help="""additional git repositories to checkout on the cluster during the deployment""")
+    parser.add_argument('--python-version',
+                        help='Python version that will be used to run pipelines task e.g., 3 or 3.6 or /usr/bin/python3.6',
```
pwnage101: I was actually asking a question, is it 3 or 3.6 or the full path?  Whichever is correct, say that in the help text.

doctoryes: Isn't the only valid selection here `/usr/bin/python3.6`?

hunytalk: Ticket mentions for version no only, but i think complete path like `/usr/bin/python3.6` makes more sense here.

doctoryes (approved):

bmedx (approved):

brianhw:
I'm not sure I understand the intent of having py2 and py3 versions of makefile target.  Isn't it just a matter of setting up the virtualenv to either use python2 or python3 at the time the virtualenv is created (mkvirtualenv -p /path/to/python), and then all the other targets will do the right thing when they run python?

PR #748, opened 2019-07-18
WIP: Cybersource: Replaced deprecated servlets method with REST API.

HassanJaveed84:
Having trouble getting the requirements in, haven't used pip compile yet. 
Other than that, I've locally tested this code using python 3.

### DIFF
--- a/edx/analytics/tasks/warehouse/financial/cybersource.py
+++ b/edx/analytics/tasks/warehouse/financial/cybersource.py
@@ -1,11 +1,14 @@
 """Collect information about payments from third-party sources for financial reporting."""
 from __future__ import absolute_import
 
+import codecs
 import csv
 import datetime
 import logging
 import os
 
+from CyberSource import ReportDownloadsApi
+from CyberSource.rest import ApiException
 import luigi
 import requests
 from luigi import date_interval
@@ -38,9 +41,8 @@ class PullFromCybersourceTaskMixin(OverwriteOutputMixin):
 
         config = get_config()
         section_name = 'cybersource:' + self.merchant_id
-        self.host = config.get(section_name, 'host')
-        self.username = config.get(section_name, 'username')
-        self.password = config.get(section_name, 'password')
+        self.keyid = config.get(section_name, 'keyid')
+        self.secretkey = config.get(section_name, 'secretkey')
         self.interval_start = luigi.DateParameter().parse(config.get(section_name, 'interval_start'))
 
         self.merchant_close_date = None
@@ -71,26 +73,45 @@ class DailyPullFromCybersourceTask(PullFromCybersourceTaskMixin, luigi.Task):
     # This is the table that we had been using for gathering and
     # storing historical Cybersource data.  It adds one additional
     # column over the 'PaymentBatchDetailReport' format.
-    REPORT_NAME = 'PaymentSubmissionDetailReport'
+    REPORT_NAME = 'PaymentSubmissionDetailReport_Daily_Classic'
     REPORT_FORMAT = 'csv'
+    # Specify the production/live environemnt for report downloads.
+    ENVIRONMENT = 'CyberSource.Environment.PRODUCTION'
 
     def requires(self):
         pass
 
     def run(self):
         self.remove_output_on_overwrite()
-        auth = (self.username, self.password)
-        response = requests.get(self.query_url, auth=auth)
-        if response.status_code != requests.codes.ok:  # pylint: disable=no-member
-            msg = "Encountered status {} on request to Cybersource for {}".format(response.status_code, self.run_date)
+        merchant_config = {
+            'authentication_type': 'http_signature',
+            'merchantid': self.merchant_id,
+            'run_environment': self.ENVIRONMENT,
+            'merchant_keyid': self.keyid,
+            'merchant_secretkey': self.secretkey,
+            'enable_log': False,
+        }
+        report_download_obj = ReportDownloadsApi(merchant_config)
+
+        # With the REST API, to get the same report we got with the servlets method, we need to specify `date + 1`.
+        batch_date = self.run_date + datetime.timedelta(days=1)
+        try:
+            return_data, status, body = report_download_obj.download_report(
+                batch_date.isoformat(),
+                self.REPORT_NAME,
+                organization_id=self.merchant_id
+            )
+        except ApiException as e:
+            log.error("Exception while downloading report for date(%s): %s", batch_date.isoformat(), e)
+            raise
+
+        # Cybersource REST API returns status as either 200(OK), 400(Invalid request) or 404(No reports found).
+        if status != requests.codes.ok:
+            msg = "Encountered status {} on request to Cybersource for {}".format(status, batch_date)
             raise Exception(msg)
 
-        # if there are no transactions in response, there will be no merchant id.
-        if self.merchant_id.encode('utf8') not in response.content and not self.is_empty_transaction_allowed:
-            raise Exception('No transactions to process.')
-
         with self.output().open('w') as output_file:
-            output_file.write(response.content.decode('utf8'))
+            output_file.write(body.encode('utf-8'))
 
     def output(self):
         """Output is in the form {output_root}/cybersource/{CCYY-mm}/cybersource_{merchant}_{CCYYmmdd}.csv"""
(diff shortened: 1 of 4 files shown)


edx/analytics/tasks/warehouse/financial/cybersource.py
```diff
@@ -103,22 +116,11 @@ def output(self):
...
-            report_name=self.REPORT_NAME,
-            report_format=self.REPORT_FORMAT
-        )
-        return url
-
 
 TRANSACTION_TYPE_MAP = {
     'ics_bill': 'sale',
+    'ics_auth,ics_bill': 'sale',
+    'ics_bill, ics_auth': 'sale',
```
brianhw, line 132: This looks a little hacky.   Would it make sense to make this a method that looks for "ics_bill" in the row, rather than enumerating this odd set of transaction_type values in a map?

edx/analytics/tasks/warehouse/financial/cybersource.py
```diff
@@ -70,26 +71,38 @@ class DailyPullFromCybersourceTask(PullFromCybersourceTaskMixin, luigi.Task):
...
+        except ApiException as e:
+            log.error("Exception when calling ReportDownloadsApi->download_report: %s", e)
+            raise
 
-        # if there are no transactions in response, there will be no merchant id.
-        if self.merchant_id not in response.content and not self.is_empty_transaction_allowed:
-            raise Exception('No transactions to process.')
+        if status != requests.codes.ok:
+            msg = "Encountered status {} on request to Cybersource for {}".format(status, self.run_date)
+            raise Exception(msg)
```
brianhw, line 111: So status is the only useful thing that is returned when it's not okay?  That's not very helpful of them.   It would be nice if there were more of an error message in the return_data, rather than having to construct one from scratch.  Not saying it was any better before.   Unless those are all more commonly raised as ApiExceptions.

edx/analytics/tasks/warehouse/financial/cybersource.py
```diff
@@ -70,26 +71,38 @@ class DailyPullFromCybersourceTask(PullFromCybersourceTaskMixin, luigi.Task):
...
+        }
+        report_download_obj = ReportDownloadsApi(merchant_config)
+        try:
+            return_data, status, body = report_download_obj.download_report(
+                self.run_date.isoformat(),
+                self.REPORT_NAME,
+                organization_id=self.merchant_id
+            )
+        except ApiException as e:
+            log.error("Exception when calling ReportDownloadsApi->download_report: %s", e)
```
brianhw: maybe add the run_date argument here as well?

brianhw (commented):
LGTM -- just a few suggestions. 

Was there any unit test coverage for this? Not terribly surprised if there weren't.

edx/analytics/tasks/warehouse/financial/cybersource.py
HassanJaveed84: Makes sense, I was thinking of doing something similar. I'll make the change.

edx/analytics/tasks/warehouse/financial/cybersource.py
HassanJaveed84: According to the API [docs](https://developer.cybersource.com/api/reference/api-reference.html) yes, that's the only thing returned. 
It returns 200, 400(Invalid request) or 404(No reports found).

Manual testing also confirmed that it's the only thing returned.

edx/analytics/tasks/warehouse/financial/cybersource.py
```diff
@@ -71,26 +72,38 @@ class DailyPullFromCybersourceTask(PullFromCybersourceTaskMixin, luigi.Task):
...
         self.remove_output_on_overwrite()
-        auth = (self.username, self.password)
-        response = requests.get(self.query_url, auth=auth)
-        if response.status_code != requests.codes.ok:  # pylint: disable=no-member
-            msg = "Encountered status {} on request to Cybersource for {}".format(response.status_code, self.run_date)
-            raise Exception(msg)
+        merchant_config = {
+            'authentication_type': 'http_signature',
+            'merchantid': self.merchant_id,
+            'run_environment': 'CyberSource.Environment.PRODUCTION',
```
pwnage101: Nit: set this as a constant at the top of the file with a comment.

pwnage101 (approved):
I have nothing more to add, besides brian's comments and one nit.

edx/analytics/tasks/warehouse/financial/cybersource.py
brianhw: Looks like you no longer need this.

requirements/default.txt
```diff
@@ -46,50 +52,70 @@ googleapis-common-protos==1.6.0  # via google-cloud-core
...
-python-cjson==1.2.1 ; python_version <= "2.7"
+pynacl==1.3.0             # via cybersource-rest-client-python, paramiko
+pyopenssl==19.0.0         # via cybersource-rest-client-python, paypalrestsdk, snowflake-connector-python
+pypi==2.1                 # via cybersource-rest-client-python
+pyrsistent==0.15.4        # via jsonschema
 python-daemon==2.2.3
 python-dateutil==2.6.1
 python-gnupg==0.3.9
+python-mimeparse==1.6.0   # via cybersource-rest-client-python, testtools
+python-subunit==1.3.0     # via cybersource-rest-client-python
```
brianhw, line 94: Looks like the cybersource code is pulling in its testing infrastructure.   Pretty annoying.   Pulls this in, along with testtools, traceback2, unittest2, fixtures, extras.

requirements/default.txt
```diff
@@ -46,50 +52,70 @@ googleapis-common-protos==1.6.0  # via google-cloud-core
...
 pycrypto==2.6.1
-pycryptodomex==3.8.2      # via snowflake-connector-python
+pycryptodome==3.8.2       # via cybersource-rest-client-python
+pycryptodomex==3.8.2      # via cybersource-rest-client-python, snowflake-connector-python
 pygeoip==0.3.2
-pyjwt==1.7.1              # via snowflake-connector-python
+pyjwt==1.7.1              # via cybersource-rest-client-python, snowflake-connector-python
 pymongo==3.8.0            # via edx-opaque-keys
-pyopenssl==19.0.0         # via paypalrestsdk, snowflake-connector-python
-python-cjson==1.2.1 ; python_version <= "2.7"
```
brianhw, line 72: Does this need to be manually restored?

pwnage101: Yes, this exact line needs to go into `extra.txt`.

pwnage101: but, it doesn't matter if this branch will only run python 3...

requirements/default.txt
HassanJaveed84: I noticed that as well, pretty bizarre.

edx/analytics/tasks/warehouse/financial/cybersource.py
```diff
@@ -71,26 +73,44 @@ class DailyPullFromCybersourceTask(PullFromCybersourceTaskMixin, luigi.Task):
...
+            'merchantid': self.merchant_id,
+            'run_environment': self.ENVIRONMENT,
+            'merchant_keyid': self.keyid,
+            'merchant_secretkey': self.secretkey,
+            'enable_log': False,
+        }
+        report_download_obj = ReportDownloadsApi(merchant_config)
+
+        # With the REST API, to get the same report we got with the servlets method, we need to specify `date + 1`.
+        batch_date = self.run_date + datetime.timedelta(days=1)
```
HassanJaveed84, line 97: @brianhw @pwnage101 this is the cleanest/less dirty way I could think of in order to get the date offset to work(and the dated naming convention in output) . The daily pull will run with 'yesterday' as the date, but it will call the API with the correct date, date +1.

pwnage101: So basically the new API will internally subtract 1 day, so we negate that effect by adding 1 day?

edx/analytics/tasks/warehouse/financial/cybersource.py
brianhw: Great!   Maybe put that in a comment there.

edx/analytics/tasks/warehouse/financial/cybersource.py
```diff
@@ -157,16 +166,25 @@ def run(self):
...
+            reader = csv.DictReader(codecs.iterdecode(input_file, 'utf-8'), delimiter=',')
             with self.output().open('w') as output_file:
                 for row in reader:
                     # Output most of the fields from the original source.
                     # The values not included are:
                     #   batch_id: CyberSource batch in which the transaction was sent.
                     #   payment_processor: code for organization that processes the payment.
+
+                    # Cybersource servlets reports used to return a negative amount in case of a refund.
+                    # However, the REST API does not, so we manually append a minus(-).
```
brianhw: nit:   append -> prepend.

requirements/default.txt
brianhw: Once we merge this into the py3-in-finance branch, we can fix it there, as part of eventually getting pipeline to py3 (if we ever do).   In the near term, you're right.

brianhw (approved):
Looks good, only a few doc'n nits.  

But again, I suggest we not delete the branch after merging.

edx/analytics/tasks/warehouse/financial/cybersource.py
HassanJaveed84: Correct.

PR #750, opened 2019-07-25
Brian/de 1589 py3 in finance tasks

brianhw:
@pwnage101 These are the changes I've made relative to your branch.  Some of the changes hardwire in use of python3 on hadoop clusters.   This shouldn't affect unit tests, but it's still a question as to how to hand that in the branch.  If it looks okay, we can merge and go back to working on a single branch.


brianhw:
@HassanJaveed84 FYI.   We may want to move your PR into Troy's branch as well?

PR #754, opened 2019-08-27
Added Mysql Database to Snowflake data set workflow

HassanJaveed84:
Differentiate between Snowflake DB name and MySQL DB name.
Improve logging upon failed Snowflake query.
Insert a date into each stage name, so different stages can point to different dates of the same DB export.

Relies on an ExternalURL task.

---

### DIFF
--- a/edx/analytics/tasks/common/snowflake_load.py
+++ b/edx/analytics/tasks/common/snowflake_load.py
@@ -17,6 +17,15 @@ from edx.analytics.tasks.util.url import ExternalURL
 log = logging.getLogger(__name__)
 
 
+def _execute_query(connection, query):
+    try:
+        connection.cursor().execute(query)
+    except ProgrammingError as e:
+        # Display the query along with the stack trace.
+        import sys
+        raise type(e), type(e)(str(e) + "\nQuery: %s\n" % query), sys.exc_info()[2]
+
+
 class SnowflakeTarget(luigi.Target):
     """
     Target for a resource in Snowflake.
@@ -33,7 +42,7 @@ class SnowflakeTarget(luigi.Target):
             self.account = creds.get('account')
             self.aws_key_id = creds.get('aws_key_id')
             self.aws_secret_key = creds.get('aws_secret_key')
-            self.database = database
+            self.sf_database = database
             self.schema = schema
             self.table = table
             self.role = role
@@ -80,8 +89,7 @@ class SnowflakeTarget(luigi.Target):
         MIN_CLUSTER_COUNT = 1 MAX_CLUSTER_COUNT = 2 SCALING_POLICY = 'STANDARD'
         INITIALLY_SUSPENDED = TRUE
         """.format(warehouse=self.warehouse)
-        log.debug(query)
-        cursor.execute(query)
+        _execute_query(connection, query)
 
     def touch(self, connection):
         """
@@ -90,27 +98,16 @@ class SnowflakeTarget(luigi.Target):
 
         self.create_marker_table()
 
-        connection.cursor().execute(
-            """
-            INSERT INTO {database}.{schema}.{marker_table} (update_id, target_table)
-            VALUES (%s, %s)
-            """.format(
-                database=self.database,
-                schema=self.schema,
-                marker_table=self.marker_table,
-            ),
-            (
-                self.update_id,
-                "{database}.{schema}.{table}".format(
-                    database=self.database,
-                    schema=self.schema,
-                    table=self.table,
-                ),
-            )
+        query = """
+        INSERT INTO {database}.{schema}.{marker_table} (update_id, target_table)
+        VALUES ('{update_id}', '{database}.{schema}.{table}')""".format(
+            update_id=str(self.update_id),
+            database=self.sf_database,
+            schema=self.schema,
+            marker_table=self.marker_table,
+            table=self.table
         )
-
-        # make sure update is properly marked
-        assert self.exists(connection)
+        _execute_query(connection, query)
 
     def exists(self, connection=None):
         close_connection = False
@@ -131,7 +128,7 @@ class SnowflakeTarget(luigi.Target):
             FROM {database}.{schema}.{marker_table}
             WHERE update_id='{update_id}' AND target_table='{database}.{schema}.{table}'
             """.format(
-                database=self.database,
+                database=self.sf_database,
                 schema=self.schema,
                 marker_table=self.marker_table,
                 update_id=self.update_id,
@@ -154,7 +151,7 @@ class SnowflakeTarget(luigi.Target):
         try:
             cursor.execute("SHOW TABLES LIKE '{marker_table}' IN SCHEMA {database}.{schema}".format(
                 marker_table=self.marker_table,
-                database=self.database,
+                database=self.sf_database,
                 schema=self.schema,
             ))
             row = cursor.fetchone()
@@ -173,7 +170,6 @@ class SnowflakeTarget(luigi.Target):
         """
 
         connection = self.connect(autocommit=True)
-        cursor = connection.cursor()
         query = """
         CREATE TABLE IF NOT EXISTS {database}.{schema}.{marker_table} (
             id            INT AUTOINCREMENT,
@@ -182,10 +178,8 @@ class SnowflakeTarget(luigi.Target):
             inserted      TIMESTAMP DEFAULT CURRENT_TIMESTAMP()::timestamp_ntz,
             PRIMARY KEY (update_id, id)
         )
-        """.format(database=self.database, schema=self.schema, marker_table=self.marker_table)
-        log.debug(query)
-        cursor.execute(query)
-        connection.close()
+        """.format(database=self.sf_database, schema=self.schema, marker_table=self.marker_table)
+        _execute_query(connection, query)
 
     def clear_marker_table(self, connection):
         """
@@ -196,12 +190,12 @@ class SnowflakeTarget(luigi.Target):
             DELETE FROM {database}.{schema}.{marker_table}
             WHERE target_table='{database}.{schema}.{table}'
             """.format(
-                database=self.database,
+                database=self.sf_database,
                 schema=self.schema,
                 marker_table=self.marker_table,
-                table=self.table,
+                table=self.table
             )
-            connection.cursor().execute(query)
+            _execute_query(connection, query)
 
 
 class SnowflakeLoadDownstreamMixin(OverwriteOutputMixin):
@@ -210,10 +204,10 @@ class SnowflakeLoadDownstreamMixin(OverwriteOutputMixin):
     """
 
     credentials = luigi.Parameter(description='Path to the external access credentials file.')
-    database = luigi.Parameter(description='The name of the database to which to write.')
-    schema = luigi.Parameter(description='The name of the schema to which to write.')
-    warehouse = luigi.Parameter(description='Name of virtual warehouse to use.')
-    role = luigi.Parameter(description='User role used to execute DDL/DML statements')
+    sf_database = luigi.Parameter(description='Name of the Snowflake database to which to write.')
+    schema = luigi.Parameter(description='Name of the Snowflake schema to which to write.')
+    warehouse = luigi.Parameter(description='Name of Snowflake virtual warehouse to use.')
+    role = luigi.Parameter(description='Snowflake user role used to execute DDL/DML statements')
 
 
 class SnowflakeLoadTask(SnowflakeLoadDownstreamMixin, luigi.Task):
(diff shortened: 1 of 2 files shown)


edx/analytics/tasks/common/snowflake_load.py
```diff
@@ -269,26 +263,35 @@ def pattern(self):
...
+    def table_description(self):
+        """
+        Description of table containing various facts, such as import time and excluded fields.
+        """
+        return ''
+
     def create_database(self, connection):
-        cursor = connection.cursor()
-        cursor.execute("CREATE DATABASE IF NOT EXISTS {database}".format(database=self.database))
+        query = "CREATE DATABASE IF NOT EXISTS {database}".format(database=self.sf_database)
```
bmedx: Do we want to do this here if we're expecting Terraform to handle all object management at this level?

edx/analytics/tasks/common/snowflake_load.py
```diff
@@ -269,26 +263,35 @@ def pattern(self):
...
+        _execute_query(connection, query)
 
     def create_schema(self, connection):
-        cursor = connection.cursor()
-        cursor.execute(
-            "CREATE SCHEMA IF NOT EXISTS {database}.{schema}".format(
-                database=self.database, schema=self.schema,
-            )
+        query = "CREATE SCHEMA IF NOT EXISTS {database}.{schema}".format(
```
bmedx: Possibly the same thing here regarding doing this through Terraform?

edx/analytics/tasks/common/snowflake_load.py
brianhw: I'm of two minds.  On the one hand, I think it is more general to leave this here, as it will have no impact if the database does exist and if we are using this for a one-off writing to an arbitrary place (as I have done for loading historical versions of tables in the past), then it will do the right thing.  On the other, if we want this to fail if we're expecting the table to exist, then we could add a flag (defaulting in either direction) to only do this when indicated.

edx/analytics/tasks/common/snowflake_load.py
```diff
@@ -325,22 +329,25 @@ def init_copy(self, connection):
 
             # Historically we've used DELETE as TRUNCATE would cause an implicit commit.
             # But with Snowflake TRUNCATE doesn't seem to cause and implicit commit.
-            connection.cursor().execute("TRUNCATE TABLE {database}.{schema}.{table}".format(
```
brianhw, line 328: nit: "cause and implicit" on line above => cause an implicit....

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
```diff
@@ -605,3 +606,197 @@ def output(self):
...
+        if not self.table_fields:
+            results = get_mysql_query_results(self.db_credentials, self.database, 'describe {}'.format(self.table_name))
+            for result in results:
+                field_name = result[0].strip()
+                field_type = result[1].strip()
+                field_null = result[2].strip()
+
+                # Enclose any Snowflake-reserved keyword field names within double-quotes.
+                if field_name.upper() in SNOWFLAKE_RESERVED_KEYWORDS:
+                    field_name = '"{}"'.format(field_name)
```
brianhw: Please remind me again why we don't just quote everything?

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
```diff
@@ -605,3 +606,197 @@ def output(self):
...
+                    field_type = field_type.rsplit('(')[0]
+                elif field_type == 'longtext':
+                    field_type = 'VARCHAR'
+                elif field_type == 'longblob':
+                    field_type = 'BINARY'
+
+                if field_null == 'NO':
+                    field_type += ' NOT NULL'
+
+                if self.should_exclude_field(self.table_name, field_name):
```
brianhw, line 664: Should we determine the exclusion before we quote reserved keywords?

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
```diff
@@ -605,3 +606,197 @@ def output(self):
...
+                    self.table_fields.append((field_name, field_type))
+
+        return self.table_fields
+
+    @property
+    def insert_source_task(self):
+        """
+        We are already exporting mysql databases to S3 using SqoopImportFromMysql through ImportMysqlToVerticaTask
+        workflow, so we specify ExternalURL here instead. In the future we can change this to a
+        SqoopImportFromMysql task.
```
brianhw: Just trying to understand if performance slowdown is a big reason not to be using SqoopImportFromMysql here, instead of ExternalURL.  We presumably need Mysql credentials anyway, in order to get schema information from the Mysql database, so we don't simplify that.   And we don't have to worry about interactions with the Vertica loads.   But I wonder if it would be more convenient going forward to have this decoupled from Vertica, and if the Vertica load has run, this will discover the same file, but if not, this will generate the dump.   As you say, we can do this later, but just curious what the issues are.

setup.cfg
```diff
@@ -67,6 +67,7 @@ edx.analytics.tasks =
     load-internal-active-users = edx.analytics.tasks.warehouse.load_internal_reporting_active_users:LoadInternalReportingActiveUsersToWarehouse
     load-internal-course-structure = edx.analytics.tasks.warehouse.load_internal_reporting_course_structure:LoadCourseBlockRecordToVertica
     load-internal-database = edx.analytics.tasks.warehouse.load_internal_reporting_database:ImportMysqlToVerticaTask
+    load-mysql-database-snowflake = edx.analytics.tasks.warehouse.load_internal_reporting_database:ImportMysqlDatabaseToSnowflakeDatasetTask
```
brianhw: nit:  We don't actually need to add this -- for loading tasks into plugins, we only need one task per file.

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
```diff
@@ -605,3 +606,197 @@ def output(self):
...
+        return "Copy of '{}' table from '{}' MySQL database on {}.{}".format(self.table_name, self.database, self.date.isoformat(), optional)
+
+    @property
+    def table_friendly_name(self):
+        return '{} from {}'.format(self.table_name, self.database)
+
+
+class ImportMysqlDatabaseToSnowflakeDatasetTask(MysqlToSnowflakeTaskMixin, SnowflakeLoadDownstreamMixin, luigi.WrapperTask):
+    """
+    Provides entry point for importing a MySQL database into Snowflake as a single dataset.
```
brianhw: The "Dataset" in the task name and in this comment is presumably left over from BigQuery?   Does Snowflake have datasets?  Should this be "Schema" instead?

brianhw (commented):
Looks good! Just a few nits.

edx/analytics/tasks/common/snowflake_load.py
pwnage101: In snowflake, we will not even grant the ELT user permission to create databases, so no matter how this is written it will fail with the following error:

```SQL access control error: Insufficient privileges to operate on account 'EDX'```

with, or without the "IF NOT EXISTS".  We simply cannot execute this statement, or we need to swallow any such errors.

Comment (HassanJaveed84) [reply edx/analytics/tasks/common/snowflake_load.py]: I'm inclined to go with this as is for us to be able to use this ad-hoc. If anyone feels strongly otherwise, I can change it include flags to control the create database and schema calls.

File: edx/analytics/tasks/common/snowflake_load.py (hunk 3 above)
Comment (HassanJaveed84) [reply edx/analytics/tasks/common/snowflake_load.py:328]: Done.

File: edx/analytics/tasks/warehouse/load_internal_reporting_database.py (hunk 5 above)
Comment (HassanJaveed84) [reply edx/analytics/tasks/warehouse/load_internal_reporting_database.py:664]: Done.

File: edx/analytics/tasks/warehouse/load_internal_reporting_database.py (hunk 4 above)
Comment (HassanJaveed84) [reply edx/analytics/tasks/warehouse/load_internal_reporting_database.py]: I don't exactly recall. Should we ?

File: edx/analytics/tasks/warehouse/load_internal_reporting_database.py (hunk 6 above)
Comment (HassanJaveed84) [reply edx/analytics/tasks/warehouse/load_internal_reporting_database.py]: Replaced as per your suggestion.

File: setup.cfg (hunk 7 above)
Comment (HassanJaveed84) [reply setup.cfg]: Removed.

File: edx/analytics/tasks/warehouse/load_internal_reporting_database.py (hunk 8 above)
Comment (HassanJaveed84) [reply edx/analytics/tasks/warehouse/load_internal_reporting_database.py]: Renamed.

**Comment by HassanJaveed84** [comment]
This is ready for a final review, along with https://github.com/edx/jenkins-job-dsl/pull/792

File: edx/analytics/tasks/common/snowflake_load.py (hunk 1 above)
Comment (HassanJaveed84) [reply edx/analytics/tasks/common/snowflake_load.py]: @pwnage101 missed your comment due to timing. If that's the case, I'll just remove these.

File: edx/analytics/tasks/common/snowflake_load.py (hunk 2 above)
Comment (pwnage101) [reply edx/analytics/tasks/common/snowflake_load.py]: As-is the terraform that I've written grants the ELT user permission to read/write the contents of schemas, but not to actually create or drop the schemas themselves.  I would argue that if we had schema management creep outside of terraform, we would no longer be in control of the permission structure which jeopardizes our RAW/PII/SAFE scheme.

I would also argue to remove any "CREATE SCHEMA..." statements from here.

File: edx/analytics/tasks/common/snowflake_load.py (hunk 1 above)
Comment (HassanJaveed84) [reply edx/analytics/tasks/common/snowflake_load.py]: Removed.

File: edx/analytics/tasks/common/snowflake_load.py (hunk 2 above)
Comment (HassanJaveed84) [reply edx/analytics/tasks/common/snowflake_load.py]: Removed.

File: edx/analytics/tasks/warehouse/load_internal_reporting_database.py (hunk 4 above)
Comment (pwnage101) [reply edx/analytics/tasks/warehouse/load_internal_reporting_database.py]: I would say because double quoting everything makes everything case-sensitive.  Here we are essentially using the lowercase variant of the conflicting identifier, and the rest of the column identifiers will be case insensitive.

I haven't completely thought through the implications of how this affects how we write queries, though.

Comment (pwnage101) [reply edx/analytics/tasks/warehouse/load_internal_reporting_database.py]: if querying from a table that has mixed case-sensitive/case-insensitive column names, we would need to write the case-insensitive columns without quotes, and case-sensitive columns lowercase with quotes.

Comment (pwnage101) [reply edx/analytics/tasks/warehouse/load_internal_reporting_database.py]: I'm finally wrapping up my thought on this, and I think it's better to have mixed-case-sensitivity (**keep the code as-written**).  Hopefully, most of the time there will not be a keyword conflict, and so people who write ad-hoc queries don't need to double quote everything and also remember to use lowercase.

If a table `mytable` has a case-sensitive column `"columnname"`, then the following query would return an error:

`SELECT columnname FROM mytable`

returns:

`SQL compilation error: error line 1 at position 7 invalid identifier 'columnname'`

Comment (bmedx) [reply edx/analytics/tasks/warehouse/load_internal_reporting_database.py]: FWIW the current PR for the views (that people would actually use to query against) has to manually work around these quoted keywords, but in that process can also unquote the non-keyword column name that it replaces them with. No one should be querying against keyword columns, we should be cleaning them at the first dbt step. However, I believe that script currently quotes everything, which I can change if it means we're forcing everything to upper case.

File: edx/analytics/tasks/warehouse/load_internal_reporting_database.py [hunk 9]
```diff
@@ -605,3 +606,209 @@ def output(self):
...
+                field_name = result[0].strip()
+                field_type = result[1].strip()
+                field_null = result[2].strip()
+
+                if self.should_exclude_field(self.table_name, field_name):
+                    self.deleted_fields.append(field_name)
+                else:
+                    # Enclose any Snowflake-reserved keyword field names within double-quotes.
+                    if field_name.upper() in SNOWFLAKE_RESERVED_KEYWORDS:
+                        field_name = '"{}"'.format(field_name)
```
brianhw: If we're quoting these, should they be uppercased?

edx/analytics/tasks/common/snowflake_load.py
```diff
@@ -269,53 +263,53 @@ def pattern(self):
...
-        cursor = connection.cursor()
-        cursor.execute("CREATE DATABASE IF NOT EXISTS {database}".format(database=self.database))
-
-    def create_schema(self, connection):
-        cursor = connection.cursor()
-        cursor.execute(
-            "CREATE SCHEMA IF NOT EXISTS {database}.{schema}".format(
-                database=self.database, schema=self.schema,
-            )
-        )
```
brianhw, line 282: Is there nothing that references these functions?   They weren't being called?  I don't see anything in this PR that is removing these calls.

brianhw (approved):
LGTM.    Just two optional questions.   👍

edx/analytics/tasks/common/snowflake_load.py
HassanJaveed84: Removed them here:
https://github.com/edx/edx-analytics-pipeline/pull/754/files#diff-8b6bd93d58592084fc399f02ed3d4e03L349

brianhw: See that now.  Thanks!

edx/analytics/tasks/common/snowflake_load.py
```diff
@@ -269,53 +263,53 @@ def pattern(self):
...
-        The resulting file format name should be: {self.database}.{self.schema}.{self.file_format_name}
+        The resulting file format name should be: {self.sf_database}.{self.schema}.{self.file_format_name}
         """
         raise NotImplementedError
 
     def create_stage(self, connection):
         stage_url = canonicalize_s3_url(self.input()['insert_source_task'].path)
         query = """
-        CREATE OR REPLACE STAGE {database}.{schema}.{table}_stage
+        CREATE OR REPLACE STAGE {database}.{schema}.{table}_{date}_stage
```
doctoryes: This technique will litter the schema with a new stage for each DB load. But they should be easy to clean-up in a later batch.

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
pwnage101: I was actually just thinking this whole logic of SNOWFLAKE_RESERVED_KEYWORDS should be implemented at a lower level, in https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/common/snowflake_load.py#L284

Then, you don't have to strip the quotes later, and it's clearer which version (uppercase or lowercase) gets saved to snowflake.

pwnage101: but my answer to your question brian is that I agree we should probably use quoted uppercase while saving, but in the long run it should not matter since we will set QUOTED_IDENTIFIERS_IGNORE_CASE.

edx/analytics/tasks/common/snowflake_load.py
pwnage101: @HassanJaveed84 If this is necessary, please add a comment about it explaining why.

HassanJaveed84: I don't remember adding the date here, I'll remove it.

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
pwnage101: per our discussion during retro, just set this to uppercase per brian's suggestion and ignore my recommendation.

edx/analytics/tasks/common/snowflake_load.py
HassanJaveed84: Removed date.

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
HassanJaveed84: Changed to use upper case for quoted reserved keywords.

PR #756, opened 2019-08-29
Pull PayPal Chargeback report

hunytalk:
---

### DIFF
--- a/edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
+++ b/edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
@@ -1,0 +1,370 @@
+import csv
+import datetime
+import fnmatch
+import logging
+
+import luigi
+from paramiko import SFTPClient, Transport
+
+from edx.analytics.tasks.common.vertica_load import VerticaCopyTask
+from edx.analytics.tasks.util.hive import HivePartition, WarehouseMixin
+from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
+from edx.analytics.tasks.util.record import DateField, DateTimeField, IntegerField, Record, StringField
+from edx.analytics.tasks.util.url import get_target_from_url, url_path_join
+
+log = logging.getLogger(__name__)
+
+
+class PayPalCaseReportMixin(OverwriteOutputMixin):
+    """Parameter for Loading Case Report task into Vertica """
+
+    _connection = None
+    run_date = luigi.DateParameter(
+        default=datetime.datetime.now() - datetime.timedelta(days=1),
+        description='Date to generate a report for. Default is yesterday',
+    )
+    host = luigi.Parameter(
+        config_path={'section': 'paypal_sftp', 'name': 'host'}, significant=False,
+        description='Hostname for sFTP connection.',
+    )
+    port = luigi.Parameter(
+        config_path={'section': 'paypal_sftp', 'name': 'port'}, significant=False,
+        description='Port for sFTP connection.',
+    )
+    username = luigi.Parameter(
+        config_path={'section': 'paypal_sftp', 'name': 'user'}, significant=False,
+        description='Username for sFTP connection.',
+    )
+    password = luigi.Parameter(
+        config_path={'section': 'paypal_sftp', 'name': 'password'}, significant=False,
+        description='Password for sFTP connection.',
+    )
+    remote_path = luigi.Parameter(
+        config_path={'section': 'paypal_sftp', 'name': 'remote_path'},
+        description='A remote path on sFTP server.',
+    )
+    schema = luigi.Parameter(
+        description='The schema in vertica database to which to write.',
+    )
+
+
+class PayPalPullCaseReportTask(PayPalCaseReportMixin, WarehouseMixin, OverwriteOutputMixin, luigi.Task):
+    """
+    Run PayPal 'Case Report', gather data and output raw csv file in S3.
+
+    """
+
+    def __init__(self, *args, **kwargs):
+        super(PayPalPullCaseReportTask, self).__init__(*args, **kwargs)
+        target = self.output()
+        if target.exists():
+            target.remove()
+
+    def create_connection(cls, host, port, username, password):
+        """ Create PayPal SFTP connection """
+        transport = Transport(host, port)
+        transport.connect(username=username, password=password)
+        return SFTPClient.from_transport(transport)
+
+    def get_file(self, remote_path):
+        """ Get remote filename. Sample remote filename: DDR-20190822.01.008.CSV """
+        date_string = self.run_date.strftime('%Y%m%d')
+        pattern = ('DDR', date_string, 'CSV')
+        remote_filepattern = "*".join(pattern)
+        for file in self._connection.listdir(self.remote_path):
+            if fnmatch.fnmatch(file, remote_filepattern):
+                return file
+        return None
+
+    def report_record_count(self, readdict):
+        """ To verify report record count with generated output in Case Report """
+        sb_count, sf_count = 0, 0
+        for row in readdict:
+            if row['CH'] == 'SB':
+                sb_count += 1
+            if row['CH'] == 'SF':
+                sf_count = int(row['Case type'])
+                break
+        if sb_count != sf_count:
+            return False
+        return True
+
+    def run(self):
+        """
+        Read Report file and reformat output.
+
+        Here is file structure doc link.
+        https://www.paypalobjects.com/webstatic/en_US/developer/docs/pdf/PP_LRD_GenDisputeReport.pdf#page=13
+        """
+
+        self._connection = self.create_connection(
+            host=self.host,
+            port=self.port,
+            username=self.username,
+            password=self.password,
+        )
+        remote_filename = self.get_file(self.remote_path)
+        if remote_filename is not None:
+            self._connection.chdir(self.remote_path)
+            input_file = self._connection.open(remote_filename, 'rb')
+            file_read = input_file.readlines()[3:]
+            reader = csv.DictReader(file_read, delimiter=str(',').encode('utf-8'))
+            record_count = self.report_record_count(reader)
+            if not record_count:
+                raise Exception('There is mismatch in report record count. Check remote file: {0}'.format(remote_filename))
+            input_file.close()
+            input_file = self._connection.open(remote_filename, 'rb')
+            with self.output().open('w') as output_file:
+                for row in input_file:
+                    output_file.write(row)
+                input_file.close()
+                self._connection.close()
+        else:
+            self._connection.close()
+            raise Exception("Remote File Not found for date: {0}".format(self.run_date.strftime('%Y-%m-%d')))
+
+    def output(self):
+        """
+        Output is set up so it can be read in as a Hive table with partitions.
+
+        The form is {output_root}/paypal_casereport/paypal_ftpraw/dt={CCYY-mm-dd}/PayPalCaseReport.tsv
+        """
+        date_string = self.run_date.strftime('%Y-%m-%d')
+        partition_path_spec = HivePartition('dt', date_string).path_spec
+        filename = "PayPalCaseReport.tsv"
+        url_with_filename = url_path_join(self.warehouse_path, "paypal_casereport", "paypal_ftpraw", partition_path_spec, filename)
+        return get_target_from_url(url_with_filename)
+
+
+class PayPalProcessCaseReportTask(PayPalCaseReportMixin, WarehouseMixin, OverwriteOutputMixin, luigi.Task):
+    """
+    Run PayPal 'Case Report', read raw CSV file and format output as TSV file.
+
+    """
+
+    def __init__(self, *args, **kwargs):
+        super(PayPalProcessCaseReportTask, self).__init__(*args, **kwargs)
+        target = self.output()
+        if target.exists():
+            target.remove()
+
+    def requires(self):
+        return PayPalPullCaseReportTask(
+            run_date=self.run_date,
+            schema=self.schema,
+        )
+
+    def filtercsv_row(self, rows):
+        """
+        There are multiple row types in PayPal Case Report. Like File header(FH), Section body(SB), Section Footer(SF).
+        We are only interested in SB columns in processed file
+        """
+        return 'SB' in rows.values()
+
+    def amount_to_decimal(self, amount):
+        """
+        PayPal represents currency as whole numbers. Like $50 will be represented as 5000. We will divide by
+        100 to get dollar value amount.
+        """
+        if amount == '':
+            return None
+        return int(amount) / 100
+
+    def run(self):
+        """
+        Read Report file and reformat output
+        Skip initial 3 lines that provide information about file source and header
+        4th line should contain header information. We are only interested in "Section Body Data" in docs
+        In processed file we are saving "Section Body Data" only.
+        """
+        with self.input().open('r') as input_file:
+            file_read = input_file.readlines()[3:]
+            reader = csv.DictReader(file_read, delimiter=',')
+            date_time_field = DateTimeField()
+            with self.output().open('w') as output_file:
+                for row in filter(self.filtercsv_row, reader):
+                    if row['Response due date'] == '':
+                        response_date = None
+                    else:
+                        response_date = date_time_field.deserialize_from_string(row['Response due date'])
+
+                    record = PayPalCaseReportRecord(
+                        case_type=row['Case type'],
+                        case_id=row['Case ID'],
+                        original_transaction_id=row['Original transaction ID'],
+                        transaction_date=date_time_field.deserialize_from_string(row['Transaction date']),
+                        transaction_invoice_id=row['Transaction invoice ID'],
+                        card_type=row['Card Type'],
+                        case_reason=row['Case reason'],
+                        claimant_name=row['Claimant name'],
+                        claimant_email_address=row['Claimant email address'],
+                        case_filing_date=date_time_field.deserialize_from_string(row['Case filing date']),
+                        case_status=row['Case status'],
+                        response_due_date=response_date,
+                        disputed_amount=self.amount_to_decimal(row['Disputed amount']),
+                        disputed_currency=row['Disputed currency'],
+                        disputed_transaction_id=row['Disputed transaction ID'],
+                        money_movement=row['Money movement'],
+                        settlement_type=row['Settlement type'],
+                        seller_protection=row['Seller protection'],
+                        seller_protection_payout_amount=self.amount_to_decimal(row['Seller protection payout amount']),
+                        seller_protection_currency=row['Seller protection currency'],
+                        payment_tracking_id=row['Payment Tracking ID'],
+                        buyer_comments=row['Buyer comments'],
+                        store_id=row['Store ID'],
+                        chargeback_reason_code=row['Chargeback Reason Code'],
+                        outcome=row['Outcome'],
+                        report_generation_date=DateField().deserialize_from_string(self.run_date.strftime('%Y-%m-%d'))
+                    )
+                    output_file.write(record.to_separated_values())
+                    output_file.write('\n')
+
+    def output(self):
+        """
+        Output is set up so it can be read in as a Hive table with partitions.
+
+        The form is {output_root}/paypal_casereport/paypal_ftpprocessed/dt={CCYY-mm-dd}/PayPalCaseReport.tsv
+        """
+        date_string = self.run_date.strftime('%Y-%m-%d')
+        partition_path_spec = HivePartition('dt', date_string).path_spec
+        filename = "PayPalCaseReport.tsv"
+        url_with_filename = url_path_join(self.warehouse_path, "paypal_casereport", "paypal_ftpprocessed", partition_path_spec, filename)
+        return get_target_from_url(url_with_filename)
+
+
+class PayPalCaseReportRecord(Record):
+    """Record in PayPal Case Report """
+
+    case_type = StringField(
+        length=50, nullable=True,
+        description='Type of case made against the transaction for e.g. Chargeback, Dispute, Claim etc'
+    )
+    case_id = StringField(
+        length=18, nullable=True,
+        description='PayPal generated unique ID for Case'
+    )
+    original_transaction_id = StringField(
+        length=255, nullable=True,
+        description='PayPal generated ID of transaction against which Case was filed'
+    )
+    transaction_date = DateTimeField(
+        nullable=True,
+        description='Completion date of the transaction'
+    )
+    transaction_invoice_id = StringField(
+        length=255, nullable=True,
+        description='Invoice ID provided with transaction'
+    )
+    card_type = StringField(
+        length=255, nullable=True,
+        description='Credit Card type used for the transaction'
+    )
+    case_reason = StringField(
+        length=255, nullable=True,
+        description='Systematic reason for the case like Inquiry by PayPal, Item not received, Not as described etc'
+    )
+    claimant_name = StringField(
+        length=128, nullable=True,
+        description='Name of the claimant'
+    )
+    claimant_email_address = StringField(
+        length=128, nullable=True,
+        description='PayPal email address of the buyer'
+    )
+    case_filing_date = DateTimeField(
+        nullable=True,
+        description='Date that the case was originally filed with PayPal'
+    )
+    case_status = StringField(
+        length=255, nullable=True,
+        description='State or the status of case'
+    )
+    response_due_date = DateTimeField(
+        nullable=True,
+        description='Date by which filed case should be responded'
+    )
+    disputed_amount = IntegerField(
+        nullable=True,
+        description='Amount being disputed by the buyer in the original transaction'
+    )
+    disputed_currency = StringField(
+        length=255, nullable=True,
+        description='Currency of the disputed amount'
+    )
+    disputed_transaction_id = StringField(
+        length=255, nullable=True,
+        description='Transaction ID generated at the time of the money movement event'
+    )
+    money_movement = StringField(
+        length=255, nullable=True,
+        description='PayPal amount status like Credit, Debit, On temporary hold etc'
+    )
+    settlement_type = StringField(
+        length=255, nullable=True,
+        description='Mode to return money to buyer'
+    )
+    seller_protection = StringField(
+        length=255, nullable=True,
+        description='Specifies the Seller Protection status'
+    )
+    seller_protection_payout_amount = IntegerField(
+        nullable=True,
+        description='Amount PayPal paid on the behalf of the seller to the buyer as a result of the Seller Protection coverage'
+    )
+    seller_protection_currency = StringField(
+        length=255, nullable=True,
+        description='Seller protection currency'
+    )
+    payment_tracking_id = StringField(
+        length=255, nullable=True,
+        description='Unique ID to obtain information about payment or refund'
+    )
+    buyer_comments = StringField(
+        length=500, nullable=True,
+        description='Comments by buyer'
+    )
+    store_id = StringField(
+        length=255, nullable=True,
+        description='Merchant identifier of the store where the purchase occurred'
+    )
+    chargeback_reason_code = StringField(
+        length=32, nullable=True,
+        description='Unique identifier to distinguish chargeback nature / reason'
+    )
+    outcome = StringField(
+        length=3000, nullable=True,
+        description='Outcome'
+    )
+    report_generation_date = DateField(
+        length=10, nullable=False,
+        description='Report file generation date'
+    )
+
+
+class LoadPayPalCaseReportToVertica(PayPalCaseReportMixin, WarehouseMixin, VerticaCopyTask):
+    """
+    Task for loading PayPal Case report from Hive into Vertica
+    """
+
+    @property
+    def insert_source_task(self):
+        return PayPalProcessCaseReportTask(
+            run_date=self.run_date,
+            schema=self.schema,
+        )
+
+    @property
+    def table(self):
+        return 'paypal_case_report'
+
+    @property
+    def default_columns(self):
+        return None
+
+    @property
+    def auto_primary_key(self):
+        return None
+
+    @property
+    def columns(self):
+        return PayPalCaseReportRecord.get_sql_schema()
--- a/setup.cfg
+++ b/setup.cfg
@@ -85,6 +85,7 @@ edx.analytics.tasks =
     orders = edx.analytics.tasks.warehouse.financial.orders_import:OrderTableTask
     payment_reconcile = edx.analytics.tasks.warehouse.financial.reconcile:ReconcileOrdersAndTransactionsTask
     paypal = edx.analytics.tasks.warehouse.financial.paypal:PaypalTransactionsByDayTask
+    paypal-report = edx.analytics.tasks.warehouse.financial.paypal_ftpreport:LoadPayPalCaseReportToVertica
     affiliate_window = edx.analytics.tasks.warehouse.financial.fees:LoadFeesToWarehouse
 
     # export:


edx/analytics/tasks/warehouse/financial/paypal.py
```diff
@@ -753,3 +760,208 @@ def requires(self):
...
+
+    def run(self):
+        # Read Report file and reformat output
+        self.remove_output_on_overwrite()
+        remote_filename = self.get_file(self.remote_path)
+        if remote_filename is not None:
+            self._connection.chdir(self.remote_path)
+            input_file = self._connection.open(remote_filename, 'r')
+            try:
+                # Skip initial 3 lines that provide information about file source and header
```
macdiesel: Could you provide a link in the comments to the documentation for the file format?

macdiesel (dismissed):
I would like to suggest adding a method to double check the hive table row numbers match what is going in to Vertica.

edx/analytics/tasks/warehouse/financial/paypal.py
hunytalk: Yeah sure.

setup.cfg
```diff
@@ -85,6 +85,7 @@ edx.analytics.tasks =
     orders = edx.analytics.tasks.warehouse.financial.orders_import:OrderTableTask
     payment_reconcile = edx.analytics.tasks.warehouse.financial.reconcile:ReconcileOrdersAndTransactionsTask
     paypal = edx.analytics.tasks.warehouse.financial.paypal:PaypalTransactionsByDayTask
+    paypal-report = edx.analytics.tasks.warehouse.financial.paypal:PayPalCaseReportTask
```
brianhw: nit: this isn't necessary, as we just need one per file, and we have that with the "paypal" entry above.   That said, I would be fine (happier?) with all of this new code being moved into a separate file -- e.g. "paypal_ftp.py" or similar.

edx/analytics/tasks/warehouse/financial/paypal.py
```diff
@@ -753,3 +760,208 @@ def requires(self):
...
+
+    _connection = None
+    run_date = luigi.DateParameter(
+        default=datetime.datetime.utcnow().date(),
+        description='Date to generate a report for. Default is today UTC.',
+    )
+    warehouse_path = luigi.Parameter(
+        config_path={'section': 'hive', 'name': 'warehouse_path'},
+        description='A URL location of the data warehouse.',
+    )
```
brianhw: We have traditionally been defining this using WarehouseMixin.   Is there a reason to define this explicitly here, instead of including WarehouseMixin?

edx/analytics/tasks/warehouse/financial/paypal.py
```diff
@@ -753,3 +760,208 @@ def requires(self):
...
+
+    def __init__(self, *args, **kwargs):
+        super(PayPalCaseReportTask, self).__init__(*args, **kwargs)
+        configuration = get_config()
+        self.host = configuration.get('paypal_sftp', 'host')
+        self.port = configuration.get('paypal_sftp', 'port')
+        self.username = configuration.get('paypal_sftp', 'user')
+        self.password = configuration.get('paypal_sftp', 'password')
+        self.remote_path = configuration.get('paypal_sftp', 'remote_path')
```
brianhw: Note that these could be defined similarly to the way that warehouse_path is defined above -- that is, instead of accessing get_config.get(), you could just define a luigi.Parameter with config path with section/name as defined here.  Just curious about the shift to use get_config() instead...?

edx/analytics/tasks/warehouse/financial/paypal.py
```diff
@@ -753,3 +760,208 @@ def requires(self):
...
+            password=self.password)
+        for file in self._connection.listdir(self.remote_path):
+            if fnmatch.fnmatch(file, remote_filepattern):
+                return file
+        return None
+
+    def run(self):
+        # Read Report file and reformat output
+        self.remove_output_on_overwrite()
+        remote_filename = self.get_file(self.remote_path)
```
brianhw: nit: I just find it a little harder to read the run() method and see self._connection just show up below, and have to look elsewhere to see it's a side-effect of calling get_file.   Two possibilities:  you could write a self.get_connection() method that checks self._connection and creates it and caches it if missing -- and then call self.get_connection everywhere that self._connection appears.  Or you could make the create_connection() call here, and pass the resulting connection into get_file() and use that below.  (You could also make create_connection() not a classmethod, and just use there the instance values for host, etc.)   Just a suggestion.

edx/analytics/tasks/warehouse/financial/paypal.py
```diff
@@ -753,3 +760,208 @@ def requires(self):
...
+
+    @classmethod
+    def create_connection(cls, host, port, username, password):
+        # Create PayPal SFTP connection
+        transport = Transport(host, port)
+        transport.connect(username=username, password=password)
+        return SFTPClient.from_transport(transport)
+
+    def get_file(self, remote_path):
+        # Get remote filename
```
brianhw: Can you also provide an example of what remote filenames tend to look like?   I'm not sure what this pattern is matching.

edx/analytics/tasks/warehouse/financial/paypal.py
brianhw: Yes, I would find it helpful to see the doc before reviewing the code below in the reader loop.

edx/analytics/tasks/warehouse/financial/paypal.py
```diff
@@ -753,3 +760,208 @@ def requires(self):
...
+
+    @property
+    def insert_source_task(self):
+        return PayPalCaseReportTask(
+            run_date=self.run_date
+        )
+
+    @property
+    def table(self):
+        return 'paypalreport'
```
brianhw: Is this just a chargeback report?   Is it possible to have many different reports that we would be pulling from this?   Do we need to specify this somehow in the input to PayPalCaseReportTask?   Is it enough just to call this 'paypal_case_report' to distinguish it?

edx/analytics/tasks/warehouse/financial/paypal.py
```diff
@@ -19,11 +22,15 @@
...
 from luigi import date_interval
 from luigi.configuration import get_config
+from paramiko import SFTPClient, Transport
 
 from edx.analytics.tasks.common.pathutil import PathSelectionByDateIntervalTask, PathSetTask
-from edx.analytics.tasks.util.hive import WarehouseMixin
+from edx.analytics.tasks.common.vertica_load import VerticaCopyTask
+from edx.analytics.tasks.util.hive import HivePartition, WarehouseMixin
 from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
+from edx.analytics.tasks.util.record import DateTimeField, IntegerField, Record, StringField
```
brianhw: Were you intending to use these?   I don't see them mentioned, but I could imagine defining the report output using a Record.

brianhw (commented):
Looking good!

setup.cfg
hunytalk: sounds good. I'll move it in separate file.

edx/analytics/tasks/warehouse/financial/paypal.py
hunytalk: No, I'll make changes as per you suggestion.

edx/analytics/tasks/warehouse/financial/paypal.py
hunytalk: ok

edx/analytics/tasks/warehouse/financial/paypal.py
hunytalk: Yeah multiple reports have been enabled for our edX account. 
No we don't need to specify table name as input. 
Table name sounds more logical.

edx/analytics/tasks/warehouse/financial/paypal.py
hunytalk: Earlier i intended to use Record, but not anymore. I'll remove.

edx/analytics/tasks/warehouse/financial/paypal.py
hunytalk: got it.

setup.cfg
hunytalk: Done

edx/analytics/tasks/warehouse/financial/paypal.py
hunytalk: done

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
```diff
@@ -0,0 +1,239 @@
...
+        description='Port for sFTP connection.',
+    )
+    username = luigi.Parameter(
+        config_path={'section': 'paypal_sftp', 'name': 'user'},
+        description='Username for sFTP connection.',
+    )
+    password = luigi.Parameter(
+        config_path={'section': 'paypal_sftp', 'name': 'password'},
+        description='Password for sFTP connection.',
+    )
```
brianhw, line 41: As mentioned in chat, I would probably just make the four of these significant=False.    And leave run_date and remote_path as significant.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
```diff
@@ -0,0 +1,239 @@
...
+                sys.setdefaultencoding('utf-8')
+                # Setting default encoding as reports character encoding is UTF-8
+                # Here file structre doc link.
+                # https://www.paypalobjects.com/webstatic/en_US/developer/docs/pdf/PP_LRD_GenDisputeReport.pdf#page=13
+                # Skip initial 3 lines that provide information about file source and header
+                # 4th line should contain header information. We are only interested in "Section Body Data" in docs
+                # Last 5 lines (or 'CH' in ('SF', 'SC', 'RF', 'RC', 'FF') ) provide information about file footer and record counts.
+                file_read = input_file.readlines()[3:]
+                reader = csv.DictReader(file_read, delimiter=',')
+                with self.output().open('w') as output_file:
```
brianhw: On reflection, I would suggest setting this up like the Cybersource pull, in that I think it makes sense to store the local version of the file in S3 in one pass, and then convert the local file to the CSV we want.   We would then collect a record of the original file, which is important for auditing, and it would allow us to change what we pull from them in future -- such as the footer information.   I'll look at the spec next to try to answer if we want footer data.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
```diff
@@ -0,0 +1,239 @@
...
+    """
+    Task for loading PayPal Chargeback report from Hive into Vertica
+    """
+
+    run_date = luigi.DateParameter(
+        default=datetime.datetime.utcnow().date(),
+        description='Date to generate a report for. Default is today UTC.',
+    )
+    schema = luigi.Parameter(
+        default='reports',
```
brianhw: It might be good to remove the default here -- not sure what it should be, so we should probably require it.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
```diff
@@ -0,0 +1,239 @@
...
+                self._connection.close()
+        else:
+            self._connection.close()
+            raise Exception("Remote File Not found for date: {0}".format(self.run_date.strftime('%Y-%m-%d')))
+
+    def output(self):
+        """
+        Output is set up so it can be read in as a Hive table with partitions.
+
+        The form is {output_root}/paypalreport/dt={CCYY-mm-dd}/PayPalChargebackReport.tsv
```
brianhw: Just thinking it might be good to be consistent between use of "case report" and "chargeback report".   Are they somehow different?   (The doc may tell me, once I get there...)

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
```diff
@@ -0,0 +1,239 @@
...
+        """
+        date_string = self.run_date.strftime('%Y-%m-%d')
+        partition_path_spec = HivePartition('dt', date_string).path_spec
+        filename = "PayPalChargebackReport.tsv"
+        url_with_filename = url_path_join(self.warehouse_path, "paypalreport", partition_path_spec, filename)
+        return get_target_from_url(url_with_filename)
+
+
+class LoadPayPalReportToVertica(WarehouseMixin, VerticaCopyTask):
```
brianhw: Might want to indicate which report here as well....

brianhw (commented):
Great progress!   A few more comments.  I'll still look at the doc you pointed to....

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
hunytalk: ok

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
hunytalk: Actually we're directly reading file from ftp server, rather than creating a local copy of it. 
So i'll use this approach, store original file as is in S3, and make another copy of it which'll have footer data excluded (if we want to) and load this file in vertica. 

Also should we use same warehouse bucket or another?
@brianhw Is this correct?

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
hunytalk: OK makes sense.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
hunytalk: You're right. Report is actually case report and not chargeback.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
hunytalk: sure

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
```diff
@@ -0,0 +1,239 @@
...
+        transport = Transport(host, port)
+        transport.connect(username=username, password=password)
+        return SFTPClient.from_transport(transport)
+
+    def get_file(self, remote_path):
+        # Get remote filename. Sample remote filename: DDR-20190822.01.008.CSV
+        date_string = self.run_date.strftime('%Y%m%d')
+        pattern = ('DDR', date_string, 'CSV')
+        remote_filepattern = "*".join(pattern)
+        if self._connection is not None:
```
brianhw: What should happen if the connection is None?   And why would that be the case?   Should that be tested for and fail earlier/elsewhere?

nit: I would just pass in the connection.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
```diff
@@ -0,0 +1,239 @@
...
+        super(PayPalCaseReportTask, self).__init__(*args, **kwargs)
+
+    def create_connection(cls, host, port, username, password):
+        # Create PayPal SFTP connection
+        transport = Transport(host, port)
+        transport.connect(username=username, password=password)
+        return SFTPClient.from_transport(transport)
+
+    def get_file(self, remote_path):
+        # Get remote filename. Sample remote filename: DDR-20190822.01.008.CSV
```
brianhw: nit:  do we want to fail if somehow PayPal changes the format from version 8?   Just a thought....

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
```diff
@@ -0,0 +1,239 @@
...
+        )
+        remote_filename = self.get_file(self.remote_path)
+        if remote_filename is not None:
+            self._connection.chdir(self.remote_path)
+            try:
+                input_file = self._connection.open(remote_filename, 'r')
+                reload(sys)
+                sys.setdefaultencoding('utf-8')
+                # Setting default encoding as reports character encoding is UTF-8
+                # Here file structre doc link.
```
brianhw: nit: file structure

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
```diff
@@ -0,0 +1,239 @@
...
+            username=self.username,
+            password=self.password,
+        )
+        remote_filename = self.get_file(self.remote_path)
+        if remote_filename is not None:
+            self._connection.chdir(self.remote_path)
+            try:
+                input_file = self._connection.open(remote_filename, 'r')
+                reload(sys)
+                sys.setdefaultencoding('utf-8')
```
brianhw: Is there not a way to tell the DictReader what encoding to use, rather than doing this at a system level?

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
brianhw: For cybersource we had separate DailyPull task that did the local copy to a warehouse location (in "cybersource"), and DailyProcess task that processed from the local "raw" copy to the "payment" format (in "payments").   I could imagine a similar PayPalPullCaseReportTask that did the local copy to a warehouse location (e.g. "paypal_ftpraw" or something), and a PayPalProcessCaseReportTask that converted (and checked) the local copy to create a single TSV that can be uploaded.

And yes, I think the TSV that can be uploaded should not include the footer information.  That information should be filtered out, but should be used in checking the data read in.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
```diff
@@ -0,0 +1,239 @@
...
+                # Setting default encoding as reports character encoding is UTF-8
+                # Here file structre doc link.
+                # https://www.paypalobjects.com/webstatic/en_US/developer/docs/pdf/PP_LRD_GenDisputeReport.pdf#page=13
+                # Skip initial 3 lines that provide information about file source and header
+                # 4th line should contain header information. We are only interested in "Section Body Data" in docs
```
brianhw: Hmm. It took me a while to get my head around this.   Rather than just parsing each row by splitting on commas, you are loading in the fourth line specifically so that you can use their column header names to refer to the specific columns.   But then once it's loaded as a CSV, it presumably is not outputting the column headers in the CSV's rows.   But it means that you don't process each section body line on its own and accumulate them as they are in fact SB's.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
```diff
@@ -0,0 +1,239 @@
...
+                # Last 5 lines (or 'CH' in ('SF', 'SC', 'RF', 'RC', 'FF') ) provide information about file footer and record counts.
```
brianhw: In addition to confirming that all columns had been SBs until the end, I think it would be good to be checking the counts in SC and RC, and seeing that they match the number of SBs read in above.   I think that may be what Beggs was thinking originally in terms of checking record counts. And maybe also confirming you have SF, RF, FF (and that it's not a two-part document).  (It's a safe assumption, I agree, that we will be under 100,000 claims.)

We may also want to parse one of these other lines, or header lines, to get report information to include.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
```diff
@@ -0,0 +1,239 @@
...
+                file_read = input_file.readlines()[3:]
+                reader = csv.DictReader(file_read, delimiter=',')
+                with self.output().open('w') as output_file:
+                    for row in reader:
+                        row.update((k, '') for k, v in row.items() if v is None)
+                        result = [
+                            # Type of Row
+                            row['CH'],
```
brianhw: If we've confirmed this is "SB" for all entries we are keeping and uploaded, then I don't think we want to include this column.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
brianhw: Hmm.  I'm thinking we would want to have the various reports be in different root directories, and that all files under that root would be reloaded into a Warehouse as that table name.  So instead of {output_root}/paypalreport, it might be better to have {output_root}/paypal_casereport.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
```diff
@@ -0,0 +1,239 @@
...
+                            # Unique ID to obtain information about payment or refund
+                            row['Payment Tracking ID'],
+                            # Comments by buyer
+                            row['Buyer comments'],
+                            # Merchant identifier of the store where the purchase occurred
+                            row['Store ID'],
+                            # Unique identifier to distinguish chargeback nature / reason
+                            row['Chargeback Reason Code'],
+                            # Outcome
+                            row['Outcome'],
```
brianhw: Thinking that we might also want to include information about what report this came from.   Perhaps it would be enough to include just the name of the source file, though we could also or instead include the report generation date.  (And window?)   Thoughts?   Maybe filename would be enough, and leave the other details in the raw files.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
```diff
@@ -0,0 +1,239 @@
...
+
+class PayPalCaseReportTask(WarehouseMixin, OverwriteOutputMixin, luigi.Task):
+    """
+    Run PayPal 'Case Report', gather data and output as TSV file.
+
+    """
+
+    _connection = None
+    run_date = luigi.DateParameter(
+        default=datetime.datetime.utcnow().date(),
```
brianhw: Is today the right thing to run, in general?   Or is yesterday the more common default?   It may not matter, as long as we're always specifying it....

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
```diff
@@ -0,0 +1,239 @@
...
+        return "''"
+
+    @property
+    def columns(self):
+        return [
+            ('column_header', 'VARCHAR(4)'),
+            ('case_type', 'VARCHAR(50)'),
+            ('case_id', 'VARCHAR(18)'),
+            ('original_transaction_id', 'VARCHAR(255)'),
+            ('transaction_date', 'TIMESTAMP'),
```
brianhw: Curious if this does the right thing with respect to timezones.    The input data includes timezone offset.  What happens to that when it's written out?  (Same below with case_filing_date and response_due_date.)   

Note that this might be one of the benefits of using the Record approach, in that it might properly handle the to-UTC conversion needed to get a string that can be safely loaded into a warehouse.  Another reason to use the Record approach would be that converting it to write to Snowflake types can be done with a single method on Snowflake, rather than having to rewrite these column types for the new server.  (It is between Vertica and BigQuery, in that it accepts lengths for varchars but doesn't want or need them.)

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
```diff
@@ -0,0 +1,239 @@
...
+            ('transaction_invoice_id', 'VARCHAR(255)'),
+            ('card_type', 'VARCHAR(255)'),
+            ('case_reason', 'VARCHAR(255)'),
+            ('claimant_name', 'VARCHAR(128)'),
+            ('claimant_email_address', 'VARCHAR(128)'),
+            ('case_filing_date', 'TIMESTAMP'),
+            ('case_status', 'VARCHAR(255)'),
+            ('response_due_date', 'TIMESTAMP'),
+            ('disputed_amount', 'INTEGER'),
```
brianhw: We should probably convert this to a decimal, and divide by 100, before it's loaded.  I believe the amounts from PayPal are cents (when the currency is USD), and that will not be obvious to anyone.   Same with seller_protection_payout_amount, though it will likely not be used very much.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
```diff
@@ -0,0 +1,239 @@
...
+            ('disputed_currency', 'VARCHAR(255)'),
+            ('disputed_transaction_id', 'VARCHAR(255)'),
+            ('money_movement', 'VARCHAR(255)'),
+            ('settlement_type', 'VARCHAR(255)'),
+            ('seller_protection', 'VARCHAR(255)'),
+            ('seller_protection_payout_amount', 'INTEGER'),
+            ('seller_protection_currency', 'VARCHAR(255)'),
+            ('payment_tracking_id', 'VARCHAR(255)'),
+            ('buyer_comments', 'VARCHAR(255)'),
```
brianhw: Maybe bump this length up?  (No idea what the actual length is.)

brianhw (commented):
Progress!   I finally read through the spec, and added a few more comments/thoughts based on that.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
hunytalk: ok

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
hunytalk: You meant to say if paypal changes remote file name or file format itself?

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
hunytalk: We're using csv Dictreader to refer each column of csv. Before any rows of body data in the report, a column header row lists the name of each of the fields in each body data row but only SB has column header and no column headers for header or footer section. 
Since we don't need to output column headers in TSV, that's why we've skipped outputting column headers.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
hunytalk: right

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
hunytalk: What could we benefit from saving source filename, if we're already saving output in partitioned date (we can easily check for source file from partition date)? Besides on ftp server we've a retention policy of 45 days since file generation. 
Yes it would be helpful to save report generation date and time, so i was thinking i'll save raw file in S3 with header and footer info.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
hunytalk: Yesterday. That's why we're always passing date through jenkins job. 
I'll change default to yesterday.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
hunytalk: No length mentioned in report docs. How about 500?

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
hunytalk: Ok

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
brianhw: Or you can also just remove the default if you'd like.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
brianhw: In the sample filename above, the 01 is the sequence number -- i.e. the first file for the date.   As long as there are <100000 entries, there will be only one file.    The 008 represents the PayPal version of the file.  If they change the format, they increment the version number.  Currently the report is at version 8.  One could enforce that by changing the pattern below to use '008.CSV' in the matching.   But again, just a thought.    And probably not that important since it's been at 8 since March, 2016.   Just thinking that if it changed out from under us and we didn't notice and it was likely to cause problems but not failures, then we'd want to check.   Anyway, never mind.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
brianhw: Yes, saving raw file in S3 with header and footer info makes a lot of sense.    (That was my comment about splitting into separate tasks.)   That especially makes sense because the retention is *only* 45 days, and we may need this to accumulate for years.  But you say that we save the output in partitioned date.  However, that information is not appearing in the resulting table in the Warehouse.   You can choose to add the partition date as a column, or you could add the original filename, but I think it would be helpful to have the information about the data source in the final warehouse table, which is presumably going to accumulate these reports over years.  (Hopefully we're not deleting the existing table and replacing with only one day's worth of data.)  Another reason to add this column would be to actually implement an overwrite functionality, where we delete all entries with the partition being written first, and reload the data.  You have added the overwrite parameter through the OverwriteOutputMixin mixin, but haven't actually implemented the delete.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
brianhw: Sure.  I've got really no idea.

brianhw (commented):
Some responses.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
```diff
@@ -0,0 +1,372 @@
...
+    username = luigi.Parameter(
+        config_path={'section': 'paypal_sftp', 'name': 'user'}, significant=False,
+        description='Username for sFTP connection.',
+    )
+    password = luigi.Parameter(
+        config_path={'section': 'paypal_sftp', 'name': 'password'}, significant=False,
+        description='Password for sFTP connection.',
+    )
+    remote_path = luigi.Parameter(
+        config_path={'section': 'paypal_sftp', 'name': 'remote_path'}, significant=False,
```
brianhw: nit:  this might be considered significant.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
```diff
@@ -0,0 +1,372 @@
...
+                self._connection.close()
+        else:
+            self._connection.close()
+            raise Exception("Remote File Not found for date: {0}".format(self.run_date.strftime('%Y-%m-%d')))
+
+    def output(self):
+        """
+        Output is set up so it can be read in as a Hive table with partitions.
+
+        The form is {output_root}/paypal_casereport/dt={CCYY-mm-dd}/paypal_ftpraw/PayPalChargebackReport.tsv
```
brianhw: nit: this comment is not accurate.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
```diff
@@ -0,0 +1,372 @@
...
+            if not record_count:
+                raise Exception('There is mismatch in report record count. Check remote file: {0}'.format(remote_filename))
+            input_file.close()
+            input_file = self._connection.open(remote_filename, 'rb')
+            reader = csv.reader(input_file, delimiter=str(',').encode('utf-8'))
+            with self.output().open('w') as output_file:
+                for row in reader:
+                    output_file.write(','.join(row))
+                    output_file.write('\n')
+                input_file.close()
```
brianhw, line 120: I believe that the preference of finance is that we make a copy of the original file, without any processing.   That makes it more likely you can guarantee that it's the original file, and not changed somehow by parsing it as a CSV.   That's what the ProcessCaseReportTask would do.    The record counting would then also be done as part of that second task.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
```diff
@@ -0,0 +1,372 @@
...
+        Skip initial 3 lines that provide information about file source and header
+        4th line should contain header information. We are only interested in "Section Body Data" in docs
+        In processed file we are saving "Section Body Data" only.
+        """
+        with self.input().open('r') as input_file:
+            file_read = input_file.readlines()[3:]
+            reader = csv.DictReader(file_read, delimiter=',')
+            date_time_field = DateTimeField()
+            with self.output().open('w') as output_file:
+                for row in filter(self.filtercsv_row, reader):
```
brianhw, line 185: you could just do the "for row in readdict" here that is currently in report_record_count.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
```diff
@@ -0,0 +1,372 @@
...
+    )
+    chargeback_reason_code = StringField(
+        length=32, nullable=True,
+        description='Unique identifier to distinguish chargeback nature / reason'
+    )
+    outcome = StringField(
+        length=3000, nullable=True,
+        description='Outcome'
+    )
+    report_generation_date = StringField(
```
brianhw: nit: you could make this a DateField, I believe.

brianhw (commented):
More progress!  Sorry if I wasn't clear about the copying of the original.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
hunytalk: @brianhw 
We are saving copy of original file as is in PayPalPullCaseReportTask once we've verified that row counts have matched here. 
I intentionally didn't put record count in second task as if there's a scenario that file is produced on server but somehow there's mismatch in rows, then we could check and identify it before even loading raw csv file. 
DictReader in PayPalPullCaseReportTask is just for record count verification purpose only. 
Does it makes sense!

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
hunytalk: In report_record_count, we're passing Dictreader object whereas here we only meant to filter 'SB' rows.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
hunytalk: Ok

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
brianhw: My concern here is that you're parsing the original into a CSV, and then writing out the CSV, as opposed to just copying the original content as-is.   It would be better to not use the csv.reader here, but just read the input_file and write to the output_file.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
brianhw: This is getting into nit territory, so my apologies, but just so I can explain myself:   you can do something like: 

```
        for row in reader: 
            if row['CH'] == 'SB':
                sb_count += 1
                self.output_row(row, output_file)
            if row['CH'] == 'SF':
                sf_count = int(row['Case type'])
        if sb_count != sf_count:
            raise Exception('There is mismatch in report record count. Check remote file: {0}'.format(remote_filename))
```

But what you have will also work fine.  So just change the output to output the raw file.

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
```diff
@@ -0,0 +1,372 @@
...
+                        report_generation_date=self.run_date.strftime('%Y-%m-%d')
+                    )
+                    output_file.write(record.to_separated_values())
+                    output_file.write('\n')
+
+    def output(self):
+        """
+        Output is set up so it can be read in as a Hive table with partitions.
+
+        The form is {output_root}/paypal_casereport/paypal_ftpprocessed/dt={CCYY-mm-dd}/PayPalChargebackReport.tsv
```
brianhw: oops, this comment is also not quite right:   PayPalCaseReport.tsv....

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
hunytalk: ok

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
hunytalk: ok

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
hunytalk: ok

edx/analytics/tasks/warehouse/financial/paypal_ftpreport.py
hunytalk: @brianhw Changes done.

brianhw (approved):
LGTM.   Thanks for making all the changes!   👍

PR #757, opened 2019-09-04
ENT-1588: Logic to determine value of `has_passed` column changed

ziafazal:
Logic to determine value of `has_passed` column should take into consideration value of letter_grade. There are some case when we have value in `passed_timestamp` but  value in `letter_grade` is empty these cases should have a False value in `has_passed` column.
ENT-1588

---


HammadAhmadWaqas (changes requested):
Can we add a test to support our new logic?
we have SQL fixture for `grades` in `edx/analytics/tasks/tests/acceptance/fixtures/input/load_grades_persistentcoursegrade.sql`
and the test is `edx/analytics/tasks/tests/acceptance/test_enterprise_enrollments.py`

HammadAhmadWaqas (approved):
:+1:

georgebabey (approved):

PR #760, opened 2019-09-16
ENT-1588: Logic to determine value of `has_passed` column changed

ziafazal:
Logic to determine value of has_passed column should take into consideration value of letter_grade. There are some case when we have value in passed_timestamp but value in letter_grade is empty these cases should have a False value in has_passed column.
ENT-1588
---


HassanJaveed84 (approved):
LGTM 👍

PR #764, opened 2019-09-24
Refactor Vertica exports

brianhw:
Reorganize Vertica exports to separate the writing to S3 from the loading into BigQuery and Snowflake.  We will write separate Jenkins jobs to do the export versus the loading, so that the loading into BQ and Snowflake can be done in parallel.

### DIFF
--- a/edx/analytics/tasks/common/vertica_export.py
+++ b/edx/analytics/tasks/common/vertica_export.py
@@ -78,7 +78,7 @@ def get_vertica_table_schema(credentials, schema_name, table_name):
         field_type = row[1].lower()
         nullable = row[2]
 
-        if field_type.rsplit('(')[0] in ['long varbinary']:
+        if field_type.split('(')[0] in ['long varbinary']:
             log.error('Error Vertica jdbc tool is unable to export field type \'%s\'.  This field will be '
                       'excluded from the extract.', field_type)
         else:
@@ -123,7 +123,7 @@ class VerticaTableToS3Mixin(VerticaTableFromS3Mixin):
 
 
 class VerticaExportMixin(WarehouseMixin):
-    """Information about Vertica warehouse, schema, and credentials are needed by all classes involved in exporting and loading."""
+    """Information about Vertica that is needed by all classes involved in exporting data and also in loading it elsewhere."""
 
     vertica_warehouse_name = luigi.Parameter(
         default='warehouse',
@@ -136,6 +136,21 @@ class VerticaExportMixin(WarehouseMixin):
         config_path={'section': 'vertica-export', 'name': 'credentials'},
         description='Path to the external Vertica access credentials file.',
     )
+    intermediate_warehouse_path = luigi.Parameter(
+        description='A path specifying the S3 root for storing dumps from Vertica.',
+        default=None,
+    )
+
+    def __init__(self, *args, **kwargs):
+        super(VerticaExportMixin, self).__init__(*args, **kwargs)
+        # If no explicit value is passed in, use the default.
+        if self.intermediate_warehouse_path is None:
+            self.intermediate_warehouse_path = self.default_intermediate_warehouse_path
+
+    @property
+    def default_intermediate_warehouse_path(self):
+        """Define root URL under which data should be written to or read from in S3."""
+        return url_path_join(self.warehouse_path, 'import/vertica/sqoop/')
 
 
 class VerticaTableExportMixin(VerticaExportMixin):
@@ -148,16 +163,9 @@ class VerticaTableExportMixin(VerticaExportMixin):
     table_name = luigi.Parameter(
         description='The Vertica table being dumped to S3.'
     )
-    intermediate_warehouse_path = luigi.Parameter(
-        description='A path specifying the S3 root for storing dumps from Vertica.',
-        default=None,
-    )
 
     def __init__(self, *args, **kwargs):
         super(VerticaTableExportMixin, self).__init__(*args, **kwargs)
-        # If no explicit value is passed in, use the same default that schema-level classes would have used.
-        if self.intermediate_warehouse_path is None:
-            self.intermediate_warehouse_path = url_path_join(self.warehouse_path, 'import/vertica/sqoop/')
         self.vertica_source_table_schema = None
 
     @property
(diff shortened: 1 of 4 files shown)


edx/analytics/tasks/warehouse/load_vertica_schema_to_snowflake.py
```diff
@@ -113,65 +62,45 @@ def columns(self):
...
+                warehouse=self.warehouse,
+                role=self.role,
+                sf_database=self.sf_database,
+                schema=self.schema,
+                table_name=table_name,
+                vertica_schema_name=self.vertica_schema_name,
+                vertica_warehouse_name=self.vertica_warehouse_name,
+                vertica_credentials=self.vertica_credentials,
+                sqoop_null_string=self.sqoop_null_string,
+                sqoop_fields_terminated_by=self.sqoop_fields_terminated_by,
```
brianhw, line 102: Once Julia's PR merges, this will also need to pass through the scratch_schema and run_id parameters to use.

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -54,343 +61,260 @@ def get_vertica_results(credentials, query):
...
 
     for row in rows:
         column_name = row[0]
+        field_type = row[1].lower()
         nullable = row[2]
-        field_type = row[1]
-        field_type = field_type.rsplit('(')[0]
 
-        if field_type.lower() in ['long varbinary']:
+        if field_type.rsplit('(')[0] in ['long varbinary']:
```
pwnage101: Just curious, why not split?  Isn't the behavior identical if there's no second parameter (maxsplit)?

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -54,343 +61,260 @@ def get_vertica_results(credentials, query):
...
-    'timestamp': 'datetime'
-}
-
-
 def get_vertica_table_schema(credentials, schema_name, table_name):
     """
-    Returns the Vertica schema in the format (column name, column type, nullable?) for the indicated table.
+    Returns the Vertica schema in the format (column name, column type, is_nullable) for the indicated table.
+
+    The column_type value is lowercased, but the size information is not stripped.
```
HassanJaveed84, line 68: Hopefully this won't cause any issues. I remember stripping the size info for mysql schema to vertica schema as some were incompatible.

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -54,343 +61,260 @@ def get_vertica_results(credentials, query):
...
 @workflow_entry_point
-class VerticaSchemaToBigQueryTask(luigi.WrapperTask):
+class ExportVerticaSchemaToS3Task(VerticaSchemaExportMixin, VerticaTableToS3Mixin, luigi.WrapperTask):
     """
-    A task that copies all the tables in a Vertica schema to S3.
+    A task that copies all the tables in a Vertica schema to S3, so other jobs can load from there.
 
     Reads all tables in a schema and, if they are not listed in the `exclude` parameter, schedules a
-    LoadVerticaToS3TableTask task for each table.
+    VerticaTableToS3Task task for each table.
```
pwnage101: maybe you meant "ExportVerticaTableToS3Task"?

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -54,343 +61,260 @@ def get_vertica_results(credentials, query):
...
                     'Vertica.'
     )
-    column_list = luigi.ListParameter(
-        description='The column names being extracted from this table.'
-    )
-    warehouse_name = luigi.Parameter(
+
+
+class VerticaExportMixin(WarehouseMixin):
+    """Information about Vertica warehouse, schema, and credentials are needed by all classes involved in exporting and loading."""
```
pwnage101: Nit: "and loading that exported data".  I initially interpreted "loading" as loading into vertica.

edx/analytics/tasks/warehouse/load_vertica_schema_to_bigquery.py
```diff
@@ -0,0 +1,149 @@
+"""
+Supports exporting data from Vertica to S3, for loading into other databases.
```
pwnage101: seems like this docstring is a copypasta?

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -54,343 +61,260 @@ def get_vertica_results(credentials, query):
...
     )
 
     def __init__(self, *args, **kwargs):
-        super(VerticaTableToS3Task, self).__init__(*args, **kwargs)
-        self.required_tasks = None
-        self.table_schema = None
+        super(VerticaTableExportMixin, self).__init__(*args, **kwargs)
+        # If no explicit value is passed in, use the same default that schema-level classes would have used.
+        if self.intermediate_warehouse_path is None:
+            self.intermediate_warehouse_path = url_path_join(self.warehouse_path, 'import/vertica/sqoop/')
```
pwnage101: Seems like this logic could be consolidated in the shared VerticaExportMixin class, or at least use a module constant to define "import/vertica/sqoop/"?

pwnage101 (approved):
pretty much only nits and typos.

edx/analytics/tasks/common/vertica_export.py
brianhw: This should not cause any change in behavior.  I moved the stripping that was here out to the bigquery loading, so that this code could also be used by the Snowflake loading as well.  The snowflake stuff only have the 'numeric(40' stripping in it.

edx/analytics/tasks/common/vertica_export.py
brianhw: done.

edx/analytics/tasks/common/vertica_export.py
brianhw: yup.

edx/analytics/tasks/warehouse/load_vertica_schema_to_bigquery.py
brianhw: not only that, it was the wrong copypasta.

edx/analytics/tasks/common/vertica_export.py
brianhw: Not sure.   I just kept what was there already.  But the doc agrees with you!   This seems to have migrated from the code loading MySQL read replicas into Vertica from years ago.

edx/analytics/tasks/common/vertica_export.py
brianhw: I got put off by the colliding names, I think, but if I separate the default property from the luigi parameter, then what you say works great.  Done.

brianhw (commented):
Thanks -- updated with fixes.

PR #765, opened 2019-09-24
Support DDL changes in MySQL->Snowflake loading

doctoryes:
---
https://openedx.atlassian.net/browse/DE-1712

Change the loading of MySQL tables into Snowflake via Sqoop by loading the table data first temporarily into a "scratch" Snowflake schema and then swapped/moved over into its target schema. This change supports DDL changes in existing tables over time.

I've tested this change via an import of the credentials DB using the Jenkins debug-pipeline-job job. I'm currently performing more testing, importing each of the four MySQL DBs we store in Snowflake.

The scratch schemas were created by Terraform in this PR: https://github.com/edx/terraform/pull/1871


doctoryes:
Successful imports so far:
http://jenkins.analytics.edx.org/job/debug-pipeline-job/312/   (credentials)
http://jenkins.analytics.edx.org/job/debug-pipeline-job/314/   (discovery)

doctoryes:
Currently, *no* tests exist for Snowflake loading - hence the coverage failure. I'll work on adding a few in a separate commit.

edx/analytics/tasks/common/snowflake_load.py
```diff
@@ -26,14 +26,36 @@ def _execute_query(connection, query):
...
+
+def qualified_marker_table_name(database, schema, marker_table):
+    """
+    Fully qualified marker table name.
+    """
+    return "{database}.{schema}.{marker_table}".format(
+        database=database,
+        schema=schema,
+        marker_table=marker_table
+    )
```
brianhw: nit:  is there a reason to have a separate method for marker tables?  It looks like you could just use qualified_table_name() for all cases.

brianhw (approved):
LGTM.  👍

doctoryes:
Successful ecommerce import here:
http://jenkins.analytics.edx.org/job/debug-pipeline-job/315/

edx/analytics/tasks/common/snowflake_load.py
```diff
@@ -281,14 +301,46 @@ def qualified_stage_name(self):
...
+            scratch_schema=self.scratch_schema,
+            table=self.table,
+            run_id=self.run_id,
+        )
+
+    def create_scratch_table(self, connection):
+        coldefs = ','.join(
+            '{name} {definition}'.format(name=name, definition=definition) for name, definition in self.columns
+        )
+        query = "CREATE TABLE IF NOT EXISTS {scratch_table} ({coldefs}) COMMENT='{comment}'".format(
```
bmedx: I think that in the event of an error we may find ourselves with an old version of a scratch table lying around, either one that got swapped back but not dropped, or one created and not yet swapped, which may have a different definition than what we're operating on here. Would it make sense to force a drop of the old swap table here too instead of trying to work with it?

edx/analytics/tasks/common/snowflake_load.py
```diff
@@ -281,14 +301,46 @@ def qualified_stage_name(self):
...
+            scratch_table=self.qualified_scratch_table_name,
+            coldefs=coldefs,
+            comment=self.table_description.replace("'", "\\'")
+        )
+        _execute_query(connection, query)
+
     def create_table(self, connection):
```
bmedx: Is this used anymore? If not, would it make sense to remove it so that people don't accidentally use it instead of the swap method?

edx/analytics/tasks/common/snowflake_load.py
doctoryes: Good point - I'll consolidate.

edx/analytics/tasks/common/snowflake_load.py
doctoryes: That condition *shouldn't* happen since, at @macdiesel 's suggestion, the code appends the Jenkins `BUILD_ID` in `self.run_id` to the table name, guaranteeing that each Jenkins run will create new scratch tables.

edx/analytics/tasks/common/snowflake_load.py
doctoryes: I agree and I'll remove it.

edx/analytics/tasks/common/snowflake_load.py
```diff
@@ -362,6 +417,8 @@ def run(self):
             cursor.execute("BEGIN")
             self.init_copy(connection)
             self.copy(connection)
+            self.swap(connection)
```
bmedx, line 420: I'm not sure off hand how each task will be called, but I thought we'd discussed doing all of the table swaps at the end of the whole database import at one point, to try to get things as consistent as we could. Is that just not feasible right now?

edx/analytics/tasks/common/snowflake_load.py
bmedx: That makes sense. I'm just curious why we would want `IF NOT EXISTS` here, then. It seems like one of those "this should never happen" scenarios that we'd want to error out in the unlikely event that it did.

doctoryes: Great point - I'll remove the `IF NOT EXISTS` here.

edx/analytics/tasks/common/snowflake_load.py
doctoryes: Hmm - I think it'd take breaking this task into two tasks - a loading-into-scratch task and a swap-and-drop task. That change would require some refactoring all around. While I'm all for writing a ticket for that work *and* I think it's an improvement that should be done, I'd prefer to get the loading working as-is, then move to that refactor work. What's implemented in this PR is no worse that the way the current loading into Vertica works. What do you think?

brianhw: I confess I was never clear on the decision about using scratch at the schema level or the table level.  The implementations are quite different.  For the schema level, none of these changes would be made at the snowflake_load.py level, but would be at the higher level where the tables per schema are looped through.   One would just do the load into the scratch using the run_id, and then once that succeeded, do a loop to swap or move into the target schema for all those tables.  One question I have is whether there are workflows for which we would prefer the table-level use of scratch.   It can be used as Julia wrote it here, and it would just work.  I'm again thinking this might be appropriate for copying large numbers of tables that take a long time, like BI, or possibly LMS.   And one could then do the upper-level task workflow for schema-level stuff for most of the smaller loads, and optimize the loads into the scratch schema by setting the scratch_schema and target_schema to be the same (scratch schema).

doctoryes: @brianhw I'm not exactly understanding the difference between schema-level and table-level scratch as you describe it above.

@bmedx What are your thoughts about proceeding with this PR as-is - and splitting the loading and swapping as follow-on work?

brianhw: Sorry not to be clearer.   By schema-level, I mean that tables are written to a scratch schema, and then the swapping is done at the schema level, once it's populated.  I guess I was still assuming that the tables in the target schema are replaced by those in the scratch schema, but it may also be possible to just swap the schemas with each other.  I'm not sure how that interacts with the "table_updates" table that might be stored in one or the other schema.  By table-level, I mean that a table is written to a scratch schema, and then the swapping is done at the table level, as you have implemented.   By doing both, I mean using the table-level code to write to the scratch schema with no swapping, and the schema-level code to then swap the schema or tables after the schema is loaded.

pwnage101: It definitely feels like we should try to merge this logic as-written (always swap tables after each table load), and offline or in a disco ticket evaluate whether efforts should go towards making small schema loads actually atomic or some other future streaming solution.

edit: whoops, lots of comments posted since my last refresh...

pwnage101 (approved):

edx/analytics/tasks/common/snowflake_load.py
brianhw: I think this code should work if the scratch_schema and target schema are the same.   No reason they need to be different schema.   Even better.  Anyway, my thumb still stands.

doctoryes: Thanks for the further explanation, @brianhw .

bmedx (approved):
Good to go as-as!

edx/analytics/tasks/common/snowflake_load.py
macdiesel: @bmedx and I were talking on site earlier and he has a good point that we could parallelize the loading in to snowflake much more than we could to say vertica. That said I would prefer that we don't undertake much refactoring work here in favor or fixing keyword issues in the source app DBs and getting stitch on board.

PR #770, opened 2019-10-30
ENT-2363 | updated Enterprise Enrollment Report query to support entitlements enrollments

HammadAhmadWaqas:
https://openedx.atlassian.net/browse/ENT-2363


edx/analytics/tasks/enterprise/enterprise_enrollments.py
```diff
@@ -305,6 +321,11 @@ def insert_query(self):
                         ON ecommerce_order_discount.offer_id = ecommerce_offer.id
                     LEFT JOIN offer_benefit ecommerce_benefit
                         ON ecommerce_offer.benefit_id = ecommerce_benefit.id
+                    LEFT JOIN entitlements_courseentitlement courseentitlement
+                        ON courseentitlement.course_uuid = REPLACE(productattributevalue.value_text,'-','')
+                        AND courseentitlement.order_number = ecommerce_order.number
```
brianhw, line 319: You may want to check whether this join introduces duplicates.   We found that with running on all entitlements that there were maybe a dozen duplicates that this introduces.

brianhw (commented):
Just one comment -- otherwise this looks fine.

georgebabey (approved):

edx/analytics/tasks/enterprise/enterprise_enrollments.py
HammadAhmadWaqas: @brianhw 
Thank you for reviewing my PR.
Yes, you are right, I was supposing that by combining `course_uuid` and `order_number` we will be having unique records for `entitlements_courseentitlement`. but unfortunately, that is not the case. As there are duplicate records in `entitlements_courseentitlement` that I think should not be there.

for example, if we run:
```
SELECT 
    course_uuid,
    order_number,
    enrollment_course_run_id,
    COUNT(*) no_of_records
FROM
    entitlements_courseentitlement entitlement
GROUP BY entitlement.course_uuid , entitlement.order_number
HAVING no_of_records > 1
    AND entitlement.course_uuid IS NOT NULL
    AND entitlement.order_number IS NOT NULL;
```
query in read-replicas we will find 100 such records.

to improve this query I have added another condition `AND courseentitlement.enrollment_course_run_id IS NOT NULL` (see line 327). 

Now if we run
```
SELECT 
    course_uuid,
    order_number,
    enrollment_course_run_id,
    COUNT(*) no_of_records
FROM
    entitlements_courseentitlement entitlement
GROUP BY entitlement.course_uuid , entitlement.order_number
HAVING no_of_records > 1
    AND entitlement.course_uuid IS NOT NULL
    AND entitlement.order_number IS NOT NULL
    AND entitlement.enrollment_course_run_id IS NOT NULL;
``` 
query, still there will be 25 such records that are douplicate.

In such a scenario what would you suggest? We should update `entitlements_courseentitlement` data (if you agree with my point) or we should add another subquery that will return distinct `entitlements_courseentitlement` records?

HammadAhmadWaqas: We have a story to remove this bad data.
https://openedx.atlassian.net/jira/servicedesk/projects/CR/queues/issue/CR-1514

edx/analytics/tasks/enterprise/enterprise_enrollments.py
```diff
@@ -300,6 +314,14 @@ def insert_query(self):
...
                     LEFT JOIN offer_benefit ecommerce_benefit
                         ON ecommerce_offer.benefit_id = ecommerce_benefit.id
+                    LEFT JOIN entitlements_courseentitlement courseentitlement
+                        ON courseentitlement.course_uuid = REPLACE(ecomm_order_product.course_uuid, '-', '')
+                        AND courseentitlement.order_number = ecommerce_order.number
+                        AND courseentitlement.enrollment_course_run_id IS NOT NULL
+                    LEFT JOIN student_courseenrollment student_course_enrollment
+                        ON student_course_enrollment.id = courseentitlement.enrollment_course_run_id
+                    WHERE ecommerce_catalogue_product.course_id = ecomm_order_product.course_id
+                            OR productattributevalue.value_text = ecomm_order_product.course_uuid
```
HammadAhmadWaqas, line 324: Added this `where` clause here instead of [line 304](https://github.com/edx/edx-analytics-pipeline/pull/770/files#diff-7ef89a9bf47ca9cc8336e883532141c7R304) because `OR` is not supported in `JOIN` in `hive`.

HammadAhmadWaqas:
@brianhw can you please review it again. I have added a test to support my changes.
In test I have added an entitlement enrollent.

edx/analytics/tasks/enterprise/enterprise_enrollments.py
HammadAhmadWaqas: @brianhw these duplicate records have been fixed now. could you please review it again.

macdiesel (approved):

PR #771, opened 2019-11-12
Aggregate programs report MapReduce task (#767)

michaelroytman:
---

### DIFF
--- a/edx/analytics/tasks/programs/program_reports.py
+++ b/edx/analytics/tasks/programs/program_reports.py
@@ -3,56 +3,35 @@ import datetime
 import logging
 
 import luigi
+from luigi.util import inherits
 
-from edx.analytics.tasks.common.mapreduce import MultiOutputMapReduceJobTask
-from edx.analytics.tasks.common.vertica_export import ExportVerticaTableToS3Task, get_vertica_table_schema
+from edx.analytics.tasks.common.mapreduce import MapReduceJobTask, MultiOutputMapReduceJobTask
+from edx.analytics.tasks.common.vertica_export import VERTICA_EXPORT_DEFAULT_FIELD_DELIMITER, ExportVerticaTableToS3Task
 from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
-from edx.analytics.tasks.util.url import ExternalURL, get_target_from_url, url_path_join
+from edx.analytics.tasks.util.url import get_target_from_url, url_path_join
 
 log = logging.getLogger(__name__)
 
 
-class BaseProgramReportsTask(OverwriteOutputMixin, MultiOutputMapReduceJobTask):
-    """ Generates CSV reports on program enrollment """
-
-    credentials = luigi.Parameter(
-        config_path={'section': 'vertica-export', 'name': 'credentials'},
-        description='Path to the external access credentials file.',
-    )
+class ProgramsReportTaskMixin(object):
+    """
+    Provides Luigi parameters common to the BuildLearnerProgramReport task and the BuildAggregateProgramReport task.
+    Namely,
+    - output_root: URL pointing to the location reports should be stored
+    - overwrite: a boolean representing whether or not to overwrite the existing output for this task. Luigi tasks
+        are idempotent, so given numerous same inputs, a task will generate the same output for each set of inputs.
+        This may not be a desirable behavior when, say, the underlying data or the output of (an) upstream task(s) has
+        changed and we need a fresh run of the task
+    - overwrite_export: a boolean representing whether or not to overwrite the Sqoop database export. As above, Luigi
+        tasks are idempotent. This particular flag is used to indicate to the upstream
+        ExportVerticaTableToS3Task whetheror not to re-export the Vertica table to S3
+    - table_name: the table containing the underlying data, used by Sqoop in the upstream ExportVerticaTableToS3Task
+    - sqoop_null_string: the string used to replace any null values encountered by Sqoop in the
+        ExportVerticaTableToS3Task; we use "null" here because we want to display "null" in the output CSV anyway
+    - vertica_schema_name: the Vertica schema to run this task against
+    """
     output_root = luigi.Parameter(
         description='URL pointing to the location reports should be stored',
-    )
-    report_name = luigi.Parameter(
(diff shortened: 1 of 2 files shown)


edx/analytics/tasks/programs/program_reports.py
```diff
@@ -157,3 +156,433 @@ def get_column_names():
...
+            ValueError: if value does not represent a boolean
+    """
+    if value.lower() == 'true':
+        return True
+    elif value.lower() == 'false':
+        return False
+    else:
+        raise ValueError('{} does not represent a boolean value.'.format(value))
+
+# Luigi has a great decorator, inherits, which "copies parameters (and nothing else) from one task class to another,
```
michaelroytman, line 610: I'm going to put in a ticket with Data Engineering to see if we can pin our version of Luigi to the version that doesn't have this issue. For now, I'm leaving this comment. Please let me know if it's not clear.

edx/analytics/tasks/programs/program_reports.py
```diff
@@ -3,41 +3,35 @@
...
     )
-    report_name = luigi.Parameter(
-        description='Name of report file(s) to output'
-    )
-    date = luigi.Parameter(
-        default=datetime.datetime.utcnow().date(),
-        description='Current run date. Used to tag report date'
-    )
-    schema_name = luigi.Parameter(
+    vertica_schema_name = luigi.Parameter(
```
michaelroytman: Using the same parameter name as the upstream task makes it easier to pass values for shared parameters to the upstream task (especially with the `clones` Luigi decorator), so I have changed those here.

edx/analytics/tasks/programs/program_reports.py
```diff
@@ -157,3 +156,433 @@ def get_column_names():
...
+                - audit
+                - verified
+                - professional or no-id-professional
+                - masters
+
+        It also includes whether the learner has completed the program.
+    """
+
+    def requires(self):
+        return self.clone(CombineCourseEnrollmentsTask)
```
michaelroytman, line 294: You can read about this decorator [here](https://luigi.readthedocs.io/en/stable/api/luigi.util.html).

setup.cfg
```diff
@@ -110,6 +110,7 @@ edx.analytics.tasks =
 
     # programs
     program_reports = edx.analytics.tasks.programs.program_reports:BuildLearnerProgramReportTask
+    course_enrollments = edx.analytics.tasks.programs.program_reports:CombineCourseEnrollments
```
michaelroytman: I'm not really sure what this does...

edx/analytics/tasks/programs/program_reports.py
```diff
@@ -87,6 +91,9 @@ def get_column_names(self):
         """
         return []  # should be implemented by child
 
+    def multi_output_reducer(self, key, values, output_file):
+        pass  # should be implemented by child
```
schenedx: If this should be implemented by child classes, the convention is to throw `NotImplementedError` here

edx/analytics/tasks/programs/program_reports.py
```diff
@@ -157,3 +154,433 @@ def get_column_names():
...
+
+class CombineCourseEnrollmentsTask(OverwriteOutputMixin, RemoveOutputMixin, MapReduceJobTask):
+    """
+        A Map Reduce task that combines multiple course run enrollment records for a single course into a single
+        record per course. It does some aggregation on data that may differ across multiple enrollments in the same course.
+
+        The distinct enrollment tracks are joined together.
+        The year of the earliest enrollment in any run of a course is used as the learner's entry year in that course. If there is no
+        enrollment date information for all runs, the learner's entry year is none.
+        The course is considered completed by the user if at least one run of the course is completed.
```
schenedx: In the doc string here, can you explain why some outputs needs to be overridden?

edx/analytics/tasks/programs/program_reports.py
```diff
@@ -157,3 +154,433 @@ def get_column_names():
...
+    )
+
+    def requires(self):
+        return self.clone(ExportVerticaTableToS3Task, overwrite=(self.overwrite_export and self.overwrite))
+
+    def mapper(self, line):
+        """Yield a (key, value) tuple for each course run enrollment record."""
+        fields = line.split(self.sqoop_fields_terminated_by.encode('ascii'))
+
+        authoring_org = fields[0]
```
schenedx: #NIT I have a C background. Whenever I see these static numbers `0, 1, 2`, I always believe they can be coded by constants. You can do something like
```
AUTHORING_ORG_COLUMN_NUMBER=0
```
to help replace the column number, so in the future, when that column number changes, we can easily update that

schenedx (commented):
See my inline comments. If you ignore all my comments, I am fine with that.

edx/analytics/tasks/programs/program_reports.py
michaelroytman: I'm happy to add an explanation, but I don't think I understand what you're referring to. What is the output that is being overridden? Are you referring to why, for example, we're picking a single entry year instead of using all of them? Or are you referring to `RemoveOutputMixin?

schenedx: You got two override parameters in this function: `overwrite` and `overwrite_export`. Could you please explain what those are for and the mechanism associated with them are? I couldn't understand those from your docs here. That's why I suggested above. This is not blocking!

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -44,6 +44,11 @@ def get_vertica_results(credentials, query):
         cred = json.load(credentials_file)
 
     # Externalize autocommit and read timeout
+    # WE CHANGED DATABASE FROM WAREHOUSE TO DOCKER
```
zacharis278: I don't think we need to commit this commented out code block.

zacharis278 (approved):
code looks good to me 👍

edx/analytics/tasks/common/vertica_export.py
michaelroytman: Whoops. I keep removing that file, but it keeps sneaking back in. Thanks for flagging that.

PR #777, opened 2019-11-19
BOM-1041 - Test

mraarif:


brianhw:
The DE group has already made a significant effort to get the pipeline to run under Python 3, and failed to complete that work.   For details, see https://openedx.atlassian.net/wiki/spaces/DE/pages/1034518574/Upgrading+edx-analytics-pipeline+for+Python3.

PR #781, opened 2019-11-21
Map Reduce Tasks to Generate Program Analytics Reports

michaelroytman:
---

### DIFF
--- a/edx/analytics/tasks/programs/program_reports.py
+++ b/edx/analytics/tasks/programs/program_reports.py
@@ -1,0 +1,603 @@
+import csv
+import datetime
+import logging
+from collections import namedtuple
+
+import luigi
+from luigi.util import inherits
+
+from edx.analytics.tasks.common.mapreduce import MapReduceJobTask, MultiOutputMapReduceJobTask
+from edx.analytics.tasks.common.vertica_export import VERTICA_EXPORT_DEFAULT_FIELD_DELIMITER, ExportVerticaTableToS3Task
+from edx.analytics.tasks.util.overwrite import OverwriteOutputMixin
+from edx.analytics.tasks.util.url import get_target_from_url, url_path_join
+
+log = logging.getLogger(__name__)
+
+
+class ProgramsReportTaskMixin(object):
+    """
+    Provides Luigi parameters common to the BuildLearnerProgramReport task and the BuildAggregateProgramReport task.
+    Namely,
+    - output_root: URL pointing to the location reports should be stored
+    - overwrite: a boolean representing whether or not to overwrite the existing output for this task. Luigi tasks
+        are idempotent, so given numerous same inputs, a task will generate the same output for each set of inputs.
+        This may not be a desirable behavior when, say, the underlying data or the output of (an) upstream task(s) has
+        changed and we need a fresh run of the task.
+    - overwrite_export: a boolean representing whether or not to overwrite the Sqoop database export. As above, Luigi
+        tasks are idempotent. This particular flag is used to indicate to the upstream.
+        ExportVerticaTableToS3Task whether or not to re-export the Vertica table to S3
+    - table_name: the table containing the underlying data, used by Sqoop in the upstream ExportVerticaTableToS3Task
+    - sqoop_null_string: the string used to replace any null values encountered by Sqoop in the
+        ExportVerticaTableToS3Task; we use "null" here because we want to display "null" in the output CSV anyway.
+    - vertica_schema_name: the Vertica schema to run this task against
+    """
+    output_root = luigi.Parameter(
+        description='URL pointing to the location reports should be stored',
+    )
+    overwrite = luigi.BoolParameter(
+        default=True,
+        description='Whether or not to overwrite existing outputs',
+    )
+    overwrite_export = luigi.BoolParameter(
+        default=False,
+        description='Whether or not to overwrite existing database export'
+    )
+    table_name = luigi.Parameter(
+        default='learner_enrollments',
+        description='Table containing enrollment rows to report on',
+    )
+    sqoop_null_string = luigi.Parameter(
+        default='null',
+        description='A string replacement value for any (null) values encountered by Sqoop when exporting from Vertica.',
+    )
+    vertica_credentials = luigi.Parameter(
+        config_path={'section': 'vertica-export', 'name': 'credentials'},
+        description='Path to the external Vertica access credentials file.',
+    )
+    vertica_schema_name = luigi.Parameter(
+        default='programs_reporting',
+        description='Vertica schema containing reporting table',
+    )
+    vertica_warehouse_name = luigi.Parameter(
+        default='warehouse',
+        description='The Vertica warehouse that houses the schema being copied.'
+    )
+
+
+class RemoveOutputMixin(object):
+
+    def run(self):
+        """
+        Clear out output if overwrite requested.
+        """
+        self.remove_output_on_overwrite()
+        super(RemoveOutputMixin, self).run()
+
+
+class BuildLearnerProgramReport(OverwriteOutputMixin, ProgramsReportTaskMixin, RemoveOutputMixin, MultiOutputMapReduceJobTask):
+    """
+    Generates CSV reports on individual program enrollment.
+
+    The task accepts the parameters inherited from the ProgramsReportTaskMixin.
+    It also accepts the following parameters:
+        - date: the date the report is run, which is used in the report CSV name to tag the report
+        - report_name: the name of the report, which is used in the report CSV name
+    """
+    report_name = luigi.Parameter(
+        default='learner_report',
+        description='Name of report file(s) to output'
+    )
+    date = luigi.DateParameter(
+        default=datetime.datetime.utcnow().date(),
+        description='Current run date. Used to tag report date'
+    )
+
+    def __init__(self, *args, **kwargs):
+        super(BuildLearnerProgramReport, self).__init__(*args, **kwargs)
+        self.columns = self.get_column_names()
+
+    def requires(self):
+        return self.clone(ExportVerticaTableToS3Task, overwrite=(self.overwrite_export and self.overwrite))
+
+    @staticmethod
+    def get_column_names():
+        """
+        List names of columns as they should appear in the CSV.
+
+        This must match the order they are stored in the exported warehouse table
+        """
+        return [
+            'Authoring Institution',
+            'Program Title',
+            'Program UUID',
+            'Program Type',
+            'User ID',
+            'Username',
+            'Name',
+            'User Key',
+            'Course Title',
+            'Course Run Key',
+            'External Course Key',
+            'Track',
+            'Grade',
+            'Letter Grade',
+            'Date First Enrolled',
+            'Date Last Unenrolled',
+            'Currently Enrolled',
+            'Date First Upgraded to Verified',
+            'Completed',
+            'Date Completed'
+        ]
+
+    def mapper(self, line):
+        """
+        Group input by authoring institution and program
+        """
+        (org, _, program_uuid, _) = line.split(VERTICA_EXPORT_DEFAULT_FIELD_DELIMITER.encode('ascii'), 3)
+        yield (org, program_uuid), line
+
+    def output_path_for_key(self, key):
+        org_key, program_uuid = key
+        filename = u'{}__{}.csv'.format(self.report_name, self.date)
+        return url_path_join(self.output_root, org_key, program_uuid, filename)
+
+    def multi_output_reducer(self, key, values, output_file):
+        """
+        Map export values to report output fields and write to csv.  Drops any extra columns
+        """
+        writer = csv.DictWriter(output_file, self.columns)
+        writer.writerow(dict(
+            (k, k) for k in self.columns
+        ))
+
+        for content in values:
+            fields = content.split(VERTICA_EXPORT_DEFAULT_FIELD_DELIMITER.encode('ascii'))
+            row = {field_key: field_value for field_key, field_value in zip(self.columns, fields)}
+            writer.writerow(row)
+
+
+class CombineCourseEnrollmentsTask(OverwriteOutputMixin, ProgramsReportTaskMixin, RemoveOutputMixin, MapReduceJobTask):
+    """
+    A Map Reduce task that combines multiple course run enrollment records for a single course into a single
+    record per course. It does some aggregation on data that may differ across multiple enrollments in the same course.
+
+    The distinct enrollment tracks are joined together.
+    The year of the earliest enrollment in any run of a course is used as the learner's entry year in that course.
+    If there is no enrollment date information for all runs, the learner's entry year is null.
+    The course is considered completed by the user if at least one run of the course is completed.
+
+    The task accepts the parameters inherited from the ProgramsReportTaskMixin.
+    """
+    COMBINED_COURSE_ENROLL_FIELDS = [
+        'authoring_org', 'program_title', 'program_uuid', 'program_type',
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -108,6 +108,9 @@ edx.analytics.tasks =
     enterprise_enrollments = edx.analytics.tasks.enterprise.enterprise_enrollments:ImportEnterpriseEnrollmentsIntoMysql
     enterprise_user = edx.analytics.tasks.enterprise.enterprise_user:ImportEnterpriseUsersIntoMysql
 
+    # programs
+    program_reports = edx.analytics.tasks.programs.program_reports:BuildLearnerProgramReport
+
```
michaelroytman, line 113: I have a question about the purpose of these such lines in this file. What are they for? I was able to run the tasks without specifying entry points for said tasks. Are these declarations necessary?

brianhw: These are needed to get the tasks in the program_reports file to be loaded.   There only needs to be one such task specified here, and then Stevedore imports it, and then all tasks in that file are accessible.  If you had no program_reports entry in this table, then none of the tasks in that file should have worked.

edx/analytics/tasks/programs/program_reports.py
```diff
@@ -0,0 +1,626 @@
...
+    """
+    if value.lower() == 'true':
+        return True
+    elif value.lower() == 'false':
+        return False
+    else:
+        raise ValueError('{} does not represent a boolean value.'.format(value))
+
+
+# Luigi has a great decorator, inherits, which "copies parameters (and nothing else) from one task class to another,
```
michaelroytman, line 587: Alternatively, we could update the pinned version of our Luigi fork to the version in which this decorator does not have this problem or cherry pick it into the edx-patch branch. Is that something that's reasonable to do?

edx/analytics/tasks/programs/program_reports.py
```diff
@@ -0,0 +1,626 @@
...
+    Provides Luigi parameters common to the BuildLearnerProgramReport task and the BuildAggregateProgramReport task.
+    Namely,
+    - output_root: URL pointing to the location reports should be stored
+    - overwrite: a boolean representing whether or not to overwrite the existing output for this task. Luigi tasks
+        are idempotent, so given numerous same inputs, a task will generate the same output for each set of inputs.
+        This may not be a desirable behavior when, say, the underlying data or the output of (an) upstream task(s) has
+        changed and we need a fresh run of the task
+    - overwrite_export: a boolean representing whether or not to overwrite the Sqoop database export. As above, Luigi
+        tasks are idempotent. This particular flag is used to indicate to the upstream
+        ExportVerticaTableToS3Task whetheror not to re-export the Vertica table to S3
```
brianhw: nit:  "whether or"

edx/analytics/tasks/programs/program_reports.py
```diff
@@ -0,0 +1,626 @@
...
+        # so to keep the list of tracks together, use a comma separated string
+        tracks = ','.join(tracks)
+
+        # if all course run enrollments have null for their first enrollment time, use null
+        # find the minimum value for the entry year excluding None; if None is the only value,
+        # then entry year is 'null'
+        if len(entry_years) > 0:
+            entry_year = min(entry_years)
+        else:
+            entry_year = 'null'
```
brianhw: nit: should this use self.sqoop_null_string?

edx/analytics/tasks/programs/program_reports.py
```diff
@@ -0,0 +1,626 @@
...
+        num_course_run_enrollments = 0
+        num_completed_courses = 0
+        num_audit_enrollments = 0
+        num_verified_enrollments = 0
+        num_professional_enrollments = 0
+        num_masters_enrollents = 0
+
+        for value in values:
+            # although we originally split on the sqoop_fields_terminated_by parameter,
+            # the writer of the previous reduce task uses a tab delimeter
```
brianhw: nit: "delimiter" sp

edx/analytics/tasks/programs/program_reports.py
```diff
@@ -0,0 +1,626 @@
...
+        # although we originally split on the sqoop_fields_terminated_by parameter,
+        # the writer of the previous reduce task uses a tab delimeter
+        fields = line.split('\t')
+
+        authoring_org = fields[self.AUTHORING_ORG_INDEX]
+        program_type = fields[self.PROGRAM_TYPE_INDEX]
+        program_title = fields[self.PROGRAM_TITLE_INDEX]
+        program_uuid = fields[self.PROGRAM_UUID_INDEX]
+        user_id = fields[self.USER_ID_INDEX]
+        timestamp = fields[self.TIMESTAMP_INDEX]
```
brianhw: Too little, too late:  wondering if this would have been easier with a namedtuple.   Then referencing fields here could use the column names instead of indices.  It could also be used for the output on line 255, so the output and input are better coupled. 

E.g.
COMBINED_COURSE_ENROLL_FIELDS = [ 'authoring_org', 'program_type', ... , 'timestamp']
CombinedCourseEnrollEntry = namedtuple('CombinedCourseEnrollEntry', COMBINED_COURSE_ENROLL_FIELDS)
entry = CombinedCourseEnrollEntry(*fields)
yield ( entry.authoring_org, entry.program_type, .... entry.timestamp), line

edx/analytics/tasks/programs/program_reports.py
```diff
@@ -0,0 +1,626 @@
...
+        """Yield a (key, value) tuple for each course run enrollment record."""
+        fields = line.split(VERTICA_EXPORT_DEFAULT_FIELD_DELIMITER.encode('ascii'))
+
+        authoring_org = fields[self.AUTHORING_ORG_INDEX]
+        program_title = fields[self.PROGRAM_TITLE_INDEX]
+        program_uuid = fields[self.PROGRAM_UUID_INDEX]
+        program_type = fields[self.PROGRAM_TYPE_INDEX]
+        user_id = fields[self.USER_ID_INDEX]
+        course_key = fields[self.COURSE_KEY_INDEX]
+        timestamp = fields[self.TIMESTAMP_INDEX]
```
brianhw: Same belated comment as below about using a namedtuple.   The difference here is that you want to also tie strings that have embedded spaces, so not sure that they can be used as tuple "column" names.

edx/analytics/tasks/programs/program_reports.py
```diff
@@ -0,0 +1,626 @@
...
+
+class ProgramsReportTaskMixin(object):
+    """
+    Provides Luigi parameters common to the BuildLearnerProgramReport task and the BuildAggregateProgramReport task.
+    Namely,
+    - output_root: URL pointing to the location reports should be stored
+    - overwrite: a boolean representing whether or not to overwrite the existing output for this task. Luigi tasks
+        are idempotent, so given numerous same inputs, a task will generate the same output for each set of inputs.
+        This may not be a desirable behavior when, say, the underlying data or the output of (an) upstream task(s) has
+        changed and we need a fresh run of the task
```
brianhw: Just a random question:  the code is so consistent in leaving off trailing periods at the end of comments, whether they are multiple sentences or single clauses.   Is that a deliberate style thing?   I find it weird to read, but that's a personal preference.

edx/analytics/tasks/programs/program_reports.py
```diff
@@ -0,0 +1,626 @@
...
+            tracks = fields[self.TRACKS_INDEX].split(',')
+            for track in tracks:
+                if track == 'audit':
+                    num_audit_enrollments += 1
+                elif track == 'verified':
+                    num_verified_enrollments += 1
+                elif track == 'professional' or track == 'no-id-professional':
+                    num_professional_enrollments += 1
+                elif track == 'masters':
+                    num_masters_enrollents += 1
```
brianhw: num_masters_enrollments (sp).

Just out of curiosity, do you expect there to be more tracks?   What about 'credit' enrollments?  Is it assumed they don't show up in programs?   Also assuming that no 'honors' track appear in programs either?

edx/analytics/tasks/programs/program_reports.py
```diff
@@ -0,0 +1,626 @@
...
+
+            if fields[self.ENTRY_YEAR_INDEX] != 'null':
```
brianhw: nit: should this use self.sqoop_null_string?

edx/analytics/tasks/programs/program_reports.py
```diff
@@ -0,0 +1,626 @@
...
+            for num in range(num_verified_enrollments):
+                num_learners_in_verified[num] += 1
+
+            for num in range(num_professional_enrollments):
+                num_learners_in_professional[num] += 1
+
+            for num in range(num_masters_enrollments):
+                num_learners_in_masters[num] += 1
+
+            num_completed_courses = int(fields[7])
```
brianhw: Should this use NUM_COMPLETED_COURSES_INDEX?

edx/analytics/tasks/programs/program_reports.py
brianhw: It is doable, but it is work, and the biggest issue is that it would potentially require testing a lot more code.   Given where we are with this repo, I think that we don't want to do that work right now.

edx/analytics/tasks/programs/program_reports.py
```diff
@@ -0,0 +1,626 @@
...
+# https://luigi.readthedocs.io/en/stable/api/luigi.util.html#using-inherits-and-requires-to-ease-parameter-pain.
+# However, we use a version of Luigi that has an older implementation of this decorator, which creates a
+# new wrapper class that subclasses the wrapped class. Normally, this isn't a problem. But because we're calling
+# super() in the run method, we run into a problem. Let's say that the subclassed class has name A. When we call
+# the decorator on A, we are returned a wrapper version of A, which is assigned the name A. Therefore, in the MRO
+# of class A, A appears twice. When super() in called in the run method, A is returned, leading to infinite recursion.
+# In future versions of Luigi, this is fixed. This is a temporary workaround.
+# Note that we do not use this decorator with any task that is directly downstream of ExportVerticaTableToS3Task
+# due issues with the "inheritance" of the sqoop_fields_terminated_by parameter, whose default value is
+# u'\x01'; Jenkins fails with the following error:
```
brianhw: nit: I'd say that "jobs running on EMR fail" rather than Jenkins.  (Jenkins itself isn't failing.)

edx/analytics/tasks/programs/tests/test_program_reports.py
```diff
@@ -0,0 +1,393 @@
...
+            ('current_enrollment_is_active', 'False'),
+            ('first_verified_enrollment_time', None),
+            ('completed', 'False'),
+            ('date_completed', None),
+            ('completed_program', 'False'),
+            ('course_key', 'course-v1:edX+UoX+Test_Course'),
+            ('timestamp', None)
+        ])
+        for key, value in kwargs.items():
+            learner_enrollment_template[key] = value
```
brianhw, line 66: Again, this might have been a good place to make use of a namedtuple, if you had one.

edx/analytics/tasks/programs/program_reports.py
```diff
@@ -0,0 +1,626 @@
...
+# parameters between dependent tasks. Read more here:
+# https://luigi.readthedocs.io/en/stable/api/luigi.util.html#using-inherits-and-requires-to-ease-parameter-pain.
+# However, we use a version of Luigi that has an older implementation of this decorator, which creates a
+# new wrapper class that subclasses the wrapped class. Normally, this isn't a problem. But because we're calling
+# super() in the run method, we run into a problem. Let's say that the subclassed class has name A. When we call
+# the decorator on A, we are returned a wrapper version of A, which is assigned the name A. Therefore, in the MRO
+# of class A, A appears twice. When super() in called in the run method, A is returned, leading to infinite recursion.
+# In future versions of Luigi, this is fixed. This is a temporary workaround.
+# Note that we do not use this decorator with any task that is directly downstream of ExportVerticaTableToS3Task
+# due issues with the "inheritance" of the sqoop_fields_terminated_by parameter, whose default value is
```
brianhw: nit: "due to issues"...

edx/analytics/tasks/programs/program_reports.py
```diff
@@ -0,0 +1,626 @@
...
+# and avoids direct pythonic inheritance". This is great in that it avoids the difficulties of sharing common
+# parameters between dependent tasks. Read more here:
+# https://luigi.readthedocs.io/en/stable/api/luigi.util.html#using-inherits-and-requires-to-ease-parameter-pain.
+# However, we use a version of Luigi that has an older implementation of this decorator, which creates a
+# new wrapper class that subclasses the wrapped class. Normally, this isn't a problem. But because we're calling
+# super() in the run method, we run into a problem. Let's say that the subclassed class has name A. When we call
+# the decorator on A, we are returned a wrapper version of A, which is assigned the name A. Therefore, in the MRO
+# of class A, A appears twice. When super() in called in the run method, A is returned, leading to infinite recursion.
+# In future versions of Luigi, this is fixed. This is a temporary workaround.
+# Note that we do not use this decorator with any task that is directly downstream of ExportVerticaTableToS3Task
```
brianhw, line 597: Wait -- so the problem was with the decorator after all?

brianhw (approved):
I don't see any substantive issues with this PR.   There are a few nits in comments, and an observation about using namedtuples as an alternative implementation strategy, but it looks very good!

edx/analytics/tasks/programs/program_reports.py
michaelroytman: I believe I was trying to disambiguate between the representation Sqoop uses for nulls from the nulls we want in the final report, trying to avoid the case where the `sqoop_null_string` is different than 'null', for example. But now that I think about it, we're not doing any transformation from `self.sqoop_null_string` to `null`, otherwise, so using `self.sqoop_null_string` would be more consistent.

edx/analytics/tasks/programs/program_reports.py
michaelroytman: It's not deliberate. Most of the time, I omit periods from the ends of fragments because it feels wrong to leave a period at the end of something that isn't a sentence, but that seems to have accidentally (yet consistently) carried over to comments that do deserve periods. I'll take a quick scan and fix that up.

michaelroytman: Out of curiosity, would you expect a period here?

```
    Provides Luigi parameters common to the BuildLearnerProgramReport task and the BuildAggregateProgramReport task.
    Namely,
    - output_root: URL pointing to the location reports should be stored[HERE!]
```

Normally, I wouldn't, since `output_root: URL pointing to the location reports should be stored` isn't a complete sentence.

edx/analytics/tasks/programs/program_reports.py
michaelroytman: From a product perspective, we don't expect information about either of those tracks to be useful for the consumer(s) of these reports. I don't know much about the `credit` track, but there are only five program course enrollments with that track, so I'm assuming it's not useful for the majority of programs. Since we no longer use the `honor` track, I believe we're not including it here either.

edx/analytics/tasks/programs/program_reports.py
michaelroytman: I don't think that the decorator is the problem. I think that the problem is the declaration of a Luigi parameter with that value as its default (although, to be honest, I'm not really sure _why_ that's the issue). So the fix, like you suggested before, was not to copy the declaration of this parameter from the `ExportVerticaTableToS3Task ` to this task. Using this decorator, however, would copy that declaration to the task. From the [docs](https://luigi.readthedocs.io/en/stable/api/luigi.util.html#luigi.util.inherits), 
> the inherits class decorator in this module copies parameters (and nothing else) from one task class to another...

Does that make sense? Do you think that my comment is unclear?

PR #782, opened 2019-12-04
Numpy version update

AhtishamShahid:
**Changes** 
- updated `numpy` version to `1.16.5`.
- Implemented `round` function because of value difference because of `numpy`  version update
- `percentile()` returns 16 digits after decimal in updated numpy version `1.16.5` whereas earlier version `1.11.3` returned 17 digits after the decimal and this cause test to fail.
- Updated tests with new values.

@brianhw : Tagging you here since upgrading the lib needed to change the tests. Can you please look at the changes and suggest what should be the better way to handle the rounding?

### DIFF
--- a/edx/analytics/tasks/insights/tests/test_module_engagement.py
+++ b/edx/analytics/tasks/insights/tests/test_module_engagement.py
@@ -579,11 +579,11 @@ class ModuleEngagementSummaryMetricRangesDataTaskReducerTest(ReducerTestMixin, T
 
     def test_zeroes_are_normal(self):
         values = [1, 0, 0, 0]
-        self.assert_ranges(values, [('normal', 0.0, 0.55), ('high', 0.55, 'inf')])
+        self.assert_ranges(values, [('normal', 0.0, '0.5499999999999998'), ('high', '0.5499999999999998', 'inf')])
 
     def test_zeroes_are_low(self):
         values = [0, 0, 0] + ([1] * 10) + ([2] * 4)
-        self.assert_ranges(values, [('low', 0, 0.4), ('normal', 0.4, 2.0), ('high', 2.0, 'inf')])
+        self.assert_ranges(values, [('low', 0, '0.3999999999999999'), ('normal', '0.3999999999999999', 2.0), ('high', 2.0, 'inf')])
 
 
 @ddt
--- a/requirements/default.in
+++ b/requirements/default.in
@@ -16,7 +16,7 @@ graphitesend==0.10.0    # Apache
 gspread==3.1.0    # MIT
 html5lib==1.0.1 	# MIT
 isoweek==1.3.3		# BSD
-numpy==1.11.3 		# BSD
+numpy==1.16.5 		# BSD
 paypalrestsdk==1.9.0    # Paypal SDK License
 psycopg2==2.6.2     # LGPL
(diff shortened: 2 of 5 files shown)


brianhw:
[201712-pipcompile.txt](https://github.com/edx/edx-analytics-pipeline/files/3928348/201712-pipcompile.txt)

I'm attaching my notes from starting this work two years ago.   Some of the updates have since been done (in part), such as htmllib (and luigi as far as 2.7.1), but others like the google update have not yet been done and have a few challenges.  I think the branch on Python 3 discovery has included in it the work for upgrading ansible, so you could borrow those modifications from there.   

Good luck!

AhtishamShahid:
> [201712-pipcompile.txt](https://github.com/edx/edx-analytics-pipeline/files/3928348/201712-pipcompile.txt)
> 
> I'm attaching my notes from starting this work two years ago. Some of the updates have since been done (in part), such as htmllib (and luigi as far as 2.7.1), but others like the google update have not yet been done and have a few challenges. I think the branch on Python 3 discovery has included in it the work for upgrading ansible, so you could borrow those modifications from there.
> 
> Good luck!

I have updated PR , so now in this PR only numpy version is updated otherwise it will increase the scope of the issue.

edx/analytics/tasks/insights/module_engagement.py
```diff
@@ -689,7 +689,7 @@ def reducer(self, course_id, lines):
             values = metric_values[metric]
             normal_lower_bound, normal_upper_bound = numpy.percentile(  # pylint: disable=no-member
                 values, [self.low_percentile, self.high_percentile]
-            )
+            ).round(2)
```
awaisdar001: @AhtishamShahid have you figure out how we are getting 2 precisions in the test before updating to the current version?

AhtishamShahid: There is no explicit `round()` functionality but it seems when values are  type-casted  to `str` then it automatically round the values 

Screenshot 2019-12-06 at 4 02 18 PM

DawoudSheraz: @AhtishamShahid Can you verify if the type casting to `str` in latest numpy version doesn't round the values? Maybe add a screenshot like you have done earlier to clarify. If there are any doc references, do add them in the PR description so that anyone following up later can know why that change was necessary.

AhtishamShahid: > @AhtishamShahid Can you verify if the type casting to `str` in latest numpy version doesn't round the values? Maybe add a screenshot like you have done earlier to clarify. If there are any doc references, do add them in the PR description so that anyone following up later can know why that change was necessary.

Updated PR description and added reason behind adding `round()` function, but unable to find official documentation regarding this change in numpy  .

DawoudSheraz (dismissed):
Looks good to me. Please squash the commits before merging

brianhw:
Sorry for the delay.   Our Jenkins acceptance tests have not been running properly, so we've not been able to test PRs before merging to master.   I'm making progress on getting some subset of them working.  Once I confirm master is passing, then we can run acceptance tests for this branch.

brianhw (approved):
Okay, I was able to run acceptance tests, and they ran fine.

PR #783, opened 2019-12-04
requests package upgraded

adeelehsan:


AhtishamShahid:
I have been working on OEP-18 compliance, so the packages you wanted to update are updated in following PR: https://github.com/edx/edx-analytics-pipeline/pull/782
Many packages have been updated and some tests are fixed in this PR.

PR #785, opened 2019-12-19
Dynamic Program Course Count

zacharis278:
@edx/masters-devs 

Aggregate program report columns will now be based on the number of courses in the program.

---

### DIFF
--- a/edx/analytics/tasks/programs/program_reports.py
+++ b/edx/analytics/tasks/programs/program_reports.py
@@ -2,7 +2,7 @@ import csv
 import datetime
 import json
 import logging
-from collections import namedtuple, OrderedDict
+from collections import OrderedDict, namedtuple
 
 import luigi
 from luigi.util import inherits
@@ -211,7 +211,7 @@ class CombineCourseEnrollmentsTask(OverwriteOutputMixin, ProgramsReportTaskMixin
         num_course_run_enrollments = 0
         tracks = set()
 
-        authoring_org, program_uuid, user_id, course_key, timestamp = key
+        authoring_org, program_uuid, user_id, course_key, timestamp = key  # pylint: disable=unused-variable
 
         for value in values:
             num_course_run_enrollments += 1
@@ -407,10 +407,10 @@ class CountProgramCohortEnrollmentsTask(OverwriteOutputMixin, RemoveOutputMixin,
         a specific number of courses
         """
         for n in range(number_of_enrollments):
-                if n in counts.keys():
-                    counts[n] += 1
-                else:
-                    counts[n] = 1
+            if n in counts.keys():
+                counts[n] += 1
+            else:
+                counts[n] = 1
 
     def requires(self):
         return self.clone(CountCourseEnrollments)
@@ -468,11 +468,6 @@ class CountProgramCohortEnrollmentsTask(OverwriteOutputMixin, RemoveOutputMixin,
             self.aggregate_enrollment_totals(cnt_learners_in_masters, int(entry.num_masters_enrollments))
             self.aggregate_enrollment_totals(cnt_learners_in_completed_courses, int(entry.num_completed_courses))
 
-            num_audit_enrollments = int(entry.num_audit_enrollments)
-            num_verified_enrollments = int(entry.num_verified_enrollments)
-            num_professional_enrollments = int(entry.num_professional_enrollments)
-            num_masters_enrollments = int(entry.num_masters_enrollments)
-
         result = OrderedDict([
             ('authoring_org', authoring_org),
             ('program_uuid', program_uuid),
@@ -489,7 +484,6 @@ class CountProgramCohortEnrollmentsTask(OverwriteOutputMixin, RemoveOutputMixin,
         ])
 
         yield (json.dumps(result),)
-
 
     def output(self):
         return get_target_from_url(url_path_join(self.output_root, 'temp/CountProgramCohortEnrollments/'))
@@ -612,10 +606,10 @@ class BuildAggregateProgramReportTask(OverwriteOutputMixin, RemoveOutputMixin, M
 
         program_course_count = int(program.course_count)
         columns = self.get_column_names(program_course_count)
-        writer = csv.DictWriter(output_file, columns) 
+        writer = csv.DictWriter(output_file, columns)
         writer.writeheader()
 
-        for entry_year, cohort in sorted(cohorts_by_year.items(), key=lambda x:x[1]):
+        for entry_year, cohort in sorted(cohorts_by_year.items(), key=lambda x: x[1]):
 
             def append_counts(row, values):
                 for n in range(program_course_count):
@@ -639,7 +633,7 @@ class BuildAggregateProgramReportTask(OverwriteOutputMixin, RemoveOutputMixin, M
             append_counts(row_items, cohort.get('professional_enrollment_counts'))
             append_counts(row_items, cohort.get('masters_enrollment_counts'))
             append_counts(row_items, cohort.get('course_completion_counts'))
-            
+
             row_items.append(cohort.get('timestamp'))
 
             writer.writerow({field_key: field_value for field_key, field_value in zip(columns, row_items)})
--- a/edx/analytics/tasks/programs/program_reports.py
+++ b/edx/analytics/tasks/programs/program_reports.py
@@ -185,7 +185,7 @@ class CombineCourseEnrollmentsTask(OverwriteOutputMixin, ProgramsReportTaskMixin
         fields = line.split(VERTICA_EXPORT_DEFAULT_FIELD_DELIMITER.encode('ascii'))
         entry = self.CombinedCourseEnrollEntry(*fields)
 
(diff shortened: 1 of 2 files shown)


edx/analytics/tasks/programs/program_reports.py
```diff
@@ -210,7 +211,7 @@ def reducer(self, key, values):
         num_course_run_enrollments = 0
         tracks = set()
 
-        authoring_org, program_type, program_uuid, program_title, user_id, course_key, timestamp = key
+        authoring_org, program_uuid, user_id, course_key, timestamp = key  # pylint: disable=unused-variable
```
michaelroytman: Does replacing `course_key` with `_` to ignore that variable avoid flagging that particular PyLint rule?

edx/analytics/tasks/programs/program_reports.py
```diff
@@ -452,36 +462,28 @@ def reducer(self, key, values):
...
+            self.aggregate_enrollment_totals(cnt_learners_in_completed_courses, int(entry.num_completed_courses))
+
+        result = OrderedDict([
+            ('authoring_org', authoring_org),
+            ('program_uuid', program_uuid),
+            ('entry_year', entry_year),
+            ('total_learners', total_num_learners),
+            ('total_enrollments', total_num_run_enrollments),
+            ('total_completions', total_num_program_completions),
+            ('audit_enrollment_counts', cnt_learners_in_audit.values()),
```
michaelroytman: I don't believe that the `dict` type affords us any guarantee about the order of the values returned by `values`. I think that, in this case, it probably happens to work out that the order of the values corresponds to the numeric order of the corresponding keys, but I don't think we can confidently rely on that. I'm basing this on [these](https://docs.python.org/3.5/library/stdtypes.html#dictionary-view-objects) docs, 

> Keys and values are iterated over in an arbitrary order which is non-random, varies across Python implementations, and depends on the dictionary’s history of insertions and deletions

We can't make any assessments about how we're inserting into the dictionary either, because each call to `aggregate_enrollment_totals` will be called with an unpredictable number for the `number_of_enrollments` parameter, so I think the the order of items returned by `values` is unpredictable.

I just worry that there will be some case in which the order of insertions or modifications to the dict will result in an ordering that we don't expect, and that will be a hard bug to track down.

I wonder if something like this might ensure a consistent ordering to the values:

```
def _get_enrollment_count_values(cnt_enrollments):
    values = []
    for key in sorted(cnt_enrollments.keys())
         values.append(cnt_enrollments[key])
    return values
```

What do you think?

edx/analytics/tasks/programs/program_reports.py
```diff
@@ -510,36 +512,47 @@ class BuildAggregateProgramReportTask(OverwriteOutputMixin, RemoveOutputMixin, M
...
-        self.columns = self.get_column_names()
 
     def requires(self):
-        return self.clone(CountProgramCohortEnrollments)
+        yield self.clone(
+            ExportVerticaTableToS3Task,
+            table_name=self.programs_table,
+            overwrite=(self.overwrite_export and self.overwrite)
+        )
+        yield self.clone(CountProgramCohortEnrollments)
```
michaelroytman, line 544: I don't understand why we need to yield these two dependencies. Does it not work to just return them as

` return self.clone(ExportVerticaTableToS3Task, table_name=self.programs_table, overwrite=(self.overwrite_export and self.overwrite)), self.clone(CountProgramCohortEnrollments)`?

That's how I saw multiple dependencies being returned [here](https://luigi.readthedocs.io/en/stable/tasks.html#task-requires). Is there something about the use of `yield` that's significant?

edx/analytics/tasks/programs/program_reports.py
```diff
@@ -452,36 +462,28 @@ def reducer(self, key, values):
...
-        row.append(timestamp)
-
-        yield row
+            self.aggregate_enrollment_totals(cnt_learners_in_audit, int(entry.num_audit_enrollments))
+            self.aggregate_enrollment_totals(cnt_learners_in_verified, int(entry.num_verified_enrollments))
+            self.aggregate_enrollment_totals(cnt_learners_in_professional, int(entry.num_professional_enrollments))
+            self.aggregate_enrollment_totals(cnt_learners_in_masters, int(entry.num_masters_enrollments))
+            self.aggregate_enrollment_totals(cnt_learners_in_completed_courses, int(entry.num_completed_courses))
+
+        result = OrderedDict([
```
michaelroytman, line 485: This is so much nicer to read now! :)

edx/analytics/tasks/programs/program_reports.py
```diff
@@ -549,18 +562,81 @@ def output_path_for_key(self, key):
...
     def mapper(self, line):
-        authoring_institution, _, _, program_uuid, _ = line.split('\t', 4)
-        yield (authoring_institution, program_uuid), line
+        """
+        Merges cohort enrollment output and a matching program entry from the
+        program metadata table into a common data structure.
+        """
+        input_type = 'cohort_enrollments'
+        export_delimiter = VERTICA_EXPORT_DEFAULT_FIELD_DELIMITER.encode('ascii')
+        if export_delimiter in line:
```
michaelroytman, line 584: Could you add a comment explaining why this is necessary? Maybe something along the lines of `mapper` will receive lines from both tables, and the `if` block covers the `program_metadata` table export, and the `else` block handles the output of the `CountProgramCohortEnrollments` task?

michaelroytman (changes requested):

edx/analytics/tasks/programs/program_reports.py
zacharis278: ah yes good catch

edx/analytics/tasks/programs/program_reports.py
zacharis278: end result is the same for both, I just thought yield made it cleaner to read. Seems like both approaches are used throughout the repo

zacharis278:
I also added a new task to act as the main entry point to run both reports

https://github.com/edx/edx-analytics-pipeline/pull/785/files#diff-3613ee53c6cd2dbc3281be68b95033ccR676

edx/analytics/tasks/programs/program_reports.py
```diff
@@ -97,7 +103,11 @@ def __init__(self, *args, **kwargs):
         self.columns = self.get_column_names()
 
     def requires(self):
-        return self.clone(ExportVerticaTableToS3Task, overwrite=(self.overwrite_export and self.overwrite))
+        return self.clone(
+            ExportVerticaTableToS3Task,
+            table_name=self.enrollments_table,
```
pwnage101, line 108: How is it possible to use `table_name` here, given that it was formerly provided by ProgramsReportTaskMixin but removed in this PR?

edx/analytics/tasks/programs/program_reports.py
```diff
@@ -392,13 +406,25 @@ class CountProgramCohortEnrollmentsTask(OverwriteOutputMixin, RemoveOutputMixin,
...
         'num_audit_enrollments', 'num_verified_enrollments', 'num_professional_enrollments',
         'num_masters_enrollments', 'is_program_completed', 'timestamp',
     ]
     CountProgramEnrollmentsEntry = namedtuple('CountProgramEnrollmentsEntry', COUNT_PROGRAM_ENROLLMENTS_FIELDS)
 
+    @staticmethod
+    def aggregate_enrollment_totals(counts, number_of_enrollments):
+        """
+        Helper function to build up a dictionary mapping a count of enrollments in
+        a specific number of courses
```
pwnage101: Can you clarify here that `counts` is a mapping of number (or more) of enrollments (key) to number of users (value), if that's correct?  Based on the wording of the docstring and the source code, it still took me a long time to actually understand which was the key and which was the value.

edx/analytics/tasks/programs/program_reports.py
zacharis278: I had to remove it from the mixin since `table_name` isn't the same every time we require the `ExportVerticaTableToS3Task`.  This is just mapping the appropriate value into `table_name` for that export task.

edx/analytics/tasks/programs/program_reports.py
zacharis278: I changed the wording on this to (hopefully) make it more clear.

edx/analytics/tasks/programs/program_reports.py
pwnage101: Oh, whoops, i confused the parameters of this task and ExportVerticaTableToS3Task.

zacharis278:
@pwnage101 / @MichaelRoytman any further comments on this?  Can I get a thumb to merge?

pwnage101 (approved):
I reviewed based on how good this looks as a luigi pipeline, map/reduce logic, python style, but I'm less familiar with the actual business logic.

zacharis278:
👍 Makes sense thanks!  PR template mentioned needing an approval so figured I should wait for that.

PR #787, opened 2019-12-30
user location hotfix

HassanJaveed84:
events with user_id(non-int): anonede71904b92643beb3fe caused job failures. Not sure if there is anything else we can do.

The jenkins job is currently running off this branch.

### DIFF
--- a/edx/analytics/tasks/insights/location_per_course.py
+++ b/edx/analytics/tasks/insights/location_per_course.py
@@ -68,6 +68,12 @@ class LastDailyIpAddressOfUserTask(
         if not user_id:
             return
 
+        try:
+            user_id = int(user_id)
+        except ValueError:
+            self.incr_counter('User Location', 'Discard event with malformed user_id', 1)
+            return
+
         # Get timestamp instead of date string, so we get the latest ip
         # address for events on the same day.
         timestamp = eventlog.get_event_time_string(event)


estute:
how often do we see failure types like this? is this the first time this kind of error has occurred? If we always expect them to be ints, is it concerning that this one wasn't? Perhaps we don't want to mask this type of failure.

HassanJaveed84:
I don't remember seeing an error like this before, specifically in this workflow. We should get an int. 
Maybe we can log the number of non-int ids and skip them. I'm open to suggestions.

estute:
I suppose I was mostly worried that if this behavior is new, there might be an upstream change that has caused this and would be concerned about masking a new behavior. Logging sounds like a good start.

doctoryes:
I agree with @estute that we should have an idea of the frequency of this error before discarding. It *might* be caused an upstream change. Or some frequent change that discards lots of data. But - of course - this implies that we'll need to follow-up and check on the log frequency.

HassanJaveed84:
The job failed again, I've added a counter.

edx/analytics/tasks/insights/location_per_course.py
```diff
@@ -68,6 +68,12 @@ def mapper(self, line):
         if not user_id:
             return
 
+        try:
+            user_id = int(user_id)
+        except ValueError:
+            self.incr_counter('User Location', 'Discard event with malformed user_id', 1)
```
doctoryes, line 74: This looks great - where can one view these counters?

doctoryes (approved):

edx/analytics/tasks/insights/location_per_course.py
HassanJaveed84: You can view these on the console output, and I think on our graphite instance as well.

pwnage101 (approved):
I'm hesitant to approve, but I'll have to make an effort to look into this during my oncall.

PR #788, opened 2020-01-06
remove hard coded warehouse string

zacharis278:
Fix so the passed in warehouse string parameter is used instead of 'warehouse'.  This is an issue for anyone trying to use the 'docker' warehouse in the docker analytics stack.

---

### DIFF
--- a/edx/analytics/tasks/common/vertica_export.py
+++ b/edx/analytics/tasks/common/vertica_export.py
@@ -233,7 +233,7 @@ class VerticaSchemaExportMixin(VerticaExportMixin):
         if not self._table_names_list:
             query = "SELECT table_name FROM all_tables WHERE schema_name='{schema_name}' AND table_type='TABLE' " \
                     "".format(schema_name=self.vertica_schema_name)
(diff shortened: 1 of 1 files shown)


macdiesel:
I'm concerned that the change in the signature of this method will cause other jobs that use this to fail. @HassanJaveed84 could you take a look at this PR when you get in tomorrow?

zacharis278:
the mixin does have a default for that parameter (https://github.com/edx/edx-analytics-pipeline/blob/0bcd1c5bfd109f23654b8dfabd946bcf119d1b23/edx/analytics/tasks/common/vertica_export.py#L128)
However, I could see this causing issues if some different value is getting passed in by a var somewhere that I'm not seeing.

HassanJaveed84:
Also need to update the call here: https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/common/vertica_export.py#L233
I think this code was added recently.

HassanJaveed84 (approved):

PR #798, opened 2020-01-30
Add metadata to Sqoop warehouse dumps

brianhw:
When dumping Vertica schemas to S3 using Sqoop, write out metadata about each table along with the data dump, by expanding the _metadata file.   Relevant information added includes the table schema information, so that the data on S3 can be read in elsewhere.  A metadata file is also written out for the schema dump itself, so there exists a place that lists the tables that were dumped.  Note that this is additional information -- it should not affect any existing formats or existing loaders. 

Second part of this is to modify the code that reads data from S3 to Snowflake.  Instead of querying Vertica again (and possibly getting incompatible information), the loading into Snowflake can use the metadata files generated above so that it no longer accesses Vertica, finally decoupling the dump to S3 from the subsequent load.  These are currently separate Task classes, so the existing loaders are unchanged.   If we want to switch all loaders, then we can either rename the tasks or update the Jenkins jobs, and remove the old code.

Third part is to add metadata to Mysql dumps to Vertica.   This will provide metadata at database level (i.e. list of tables) to enable copying of dumps from one location in S3 to another.  However, It does not include any code for loading from S3 into Snowflake using the metadata.

### DIFF
--- a/edx/analytics/tasks/common/vertica_export.py
+++ b/edx/analytics/tasks/common/vertica_export.py
@@ -331,3 +331,78 @@ class ExportVerticaSchemaToS3Task(VerticaSchemaExportMixin, VerticaTableToS3Mixi
                 sqoop_fields_terminated_by=self.sqoop_fields_terminated_by,
                 sqoop_delimiter_replacement=self.sqoop_delimiter_replacement,
             )
+
+
+@workflow_entry_point
+class ExportVerticaSchemaToS3WithMetadataTask(VerticaSchemaExportMixin, VerticaTableToS3Mixin, luigi.Task):
+    """
+    A task that copies all the tables in a Vertica schema to S3, so other jobs can load from there.
+
+    Reads all tables in a schema and, if they are not listed in the `exclude` parameter, schedules a
+    ExportVerticaTableToS3Task task for each table.
+    """
+    overwrite = luigi.BoolParameter(
+        default=False,
+        significant=False,
+        description='Indicates if the target data sources should be removed prior to generating.'
+    )
+
+    # Collect metadata on when the schema export was started (as set by first call to requires()).
+    creation_time = None
+
+    def requires(self):
+        if self.creation_time is None:
+            self.creation_time = datetime.datetime.utcnow().isoformat()
+
+        yield ExternalURL(url=self.vertica_credentials)
+
+        for table_name in self.get_table_list_for_schema():
+            yield ExportVerticaTableToS3Task(
+                date=self.date,
+                overwrite=self.overwrite,
+                table_name=table_name,
+                intermediate_warehouse_path=self.intermediate_warehouse_path,
+                vertica_schema_name=self.vertica_schema_name,
+                vertica_warehouse_name=self.vertica_warehouse_name,
+                vertica_credentials=self.vertica_credentials,
+                sqoop_null_string=self.sqoop_null_string,
+                sqoop_fields_terminated_by=self.sqoop_fields_terminated_by,
+                sqoop_delimiter_replacement=self.sqoop_delimiter_replacement,
+            )
+
+    @property
+    def s3_location_for_schema(self):
+        """
+        Returns the URL for the location of S3 data for the given table and schema.
+
+        This logic is shared by classes that dump data to S3 and those that load data from S3, so they agree.
+        """
+        partition_path_spec = HivePartition('dt', self.date).path_spec
+        url = url_path_join(self.intermediate_warehouse_path,
+                            self.vertica_warehouse_name,
+                            self.vertica_schema_name,
+                            'dump_schema_metadata_output',
+                            partition_path_spec) + '/'
+        return url
+
+    def output(self):
+        """Return target to which metadata about the schema-level task execution can be written."""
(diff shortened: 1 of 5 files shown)


edx/analytics/tasks/common/vertica_export.py
```diff
@@ -152,14 +156,26 @@ def default_intermediate_warehouse_path(self):
         """Define root URL under which data should be written to or read from in S3."""
         return url_path_join(self.warehouse_path, 'import/vertica/sqoop/')
 
+    @property
+    def s3_location_for_schema(self):
+        """
+        Returns the URL for the location of S3 data for the given schema.
```
HassanJaveed84: Found this a little confusing. The URL returned by this method points to the S3 directory which would have the schema metadata. The docstring made me think that it points to the actual data for the schema.

HassanJaveed84: Does it make sense to change the method/property to reflect that it's the location of schema metadata ?

brianhw: It's only used in get_schema_metadata_target() , so I've removed the property altogether and folded it into the method.

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
```diff
@@ -814,3 +883,180 @@ def complete(self):
...
+        for field_name, field_type, field_null in self.mysql_table_schema:
+
+            # Enclose any Snowflake-reserved keyword field names within double-quotes.
+            if field_name.upper() in SNOWFLAKE_RESERVED_KEYWORDS:
+                field_name = '"{}"'.format(field_name.upper())
+
+            mysql_types_with_parentheses = ['smallint', 'int', 'bigint', 'datetime', 'varchar']
+            if field_type == 'tinyint(1)':
+                field_type = 'BOOLEAN'
+            elif any(_type in field_type for _type in mysql_types_with_parentheses):
```
pwnage101, line 933: Should we normalize field_type to be lowercase for this comparison?
```suggestion
            elif any(_type in field_type.lower() for _type in mysql_types_with_parentheses):
```

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
```diff
@@ -814,3 +883,180 @@ def complete(self):
...
+                field_type = field_type.rsplit('(')[0]
```
pwnage101, line 934: Conversely, should we uppercase the value we set?
```suggestion
                field_type = field_type.rsplit('(')[0].upper()
```

bmedx (approved):

pwnage101 (approved):

HassanJaveed84 (approved):
LGTM. 
Do we want to test the `additional_metadata` here https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/common/tests/test_sqoop.py#L229 ?

edx/analytics/tasks/common/vertica_export.py
```diff
@@ -152,14 +156,23 @@ def default_intermediate_warehouse_path(self):
...
         return url_path_join(self.warehouse_path, 'import/vertica/sqoop/')
 
+    def get_schema_metadata_target(self):
+        """Return target for reading or writing out schema-level metadata file."""
+        partition_path_spec = HivePartition('dt', self.date).path_spec
+        url = url_path_join(
+            self.intermediate_warehouse_path,
+            self.vertica_warehouse_name,
+            self.vertica_schema_name,
+            'dump_schema_metadata_output',
```
HassanJaveed84: nit: dump_schema_metadata_output -> schema_metadata_output or schema_metadata

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
brianhw: Hmm.   Yes and no.   Right now, all type information coming out of mysql is lowercase, so there's been no need to .lower() anything, so this should be a no-op.   Also note that the other comparisons also don't use .lower().    I'm going to skip this, if that's okay.

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
brianhw: Conversely, the same thing here.   Most field types are not being uppercased, but we are uppercasing the ones we substitute for their original value.   I'm assuming that has not been a problem, so not sure of the value of fixing it for some here.

edx/analytics/tasks/common/vertica_export.py
brianhw: Actually, maybe I should prepend an underscore, so that it's that much less likely to be confused with a regular table's output. Would that be too confusable with the _metadata file contained within?   Something like "_schema_metadata/dt=2020-01-31/_metadata" ...

brianhw:
Unit tests added.   Nice suggestion!

edx/analytics/tasks/common/vertica_export.py
brianhw: I'm going with "_metadata_export_schema" here, and "_metadata_export_database" for the Mysql dumps.

PR #803, opened 2020-02-11
Program Reports Cohort Aggregation

zacharis278:
@edx/masters-devs 

Updates aggregate report to determine cohort by month and course run start instead of by year and date of learner enrollment in that run.

Depends on: https://github.com/edx/edx-analytics-pipeline-scripts/pull/544

Test Run: http://jenkins.analytics.edx.org/job/debug-pipeline-job/538/

---

### DIFF
--- a/edx/analytics/tasks/programs/program_reports.py
+++ b/edx/analytics/tasks/programs/program_reports.py
@@ -17,7 +17,7 @@ log = logging.getLogger(__name__)
 
 class ProgramsReportTaskMixin(object):
     """
-    Provides Luigi parameters common to the BuildLearnerProgramReport task and the BuildAggregateProgramReport task.
+    Provides Luigi parameters common to the BuildLearnerProgramReport task and the BuildCohortProgramReport task.
     Namely,
     - output_root: URL pointing to the location reports should be stored
     - overwrite: a boolean representing whether or not to overwrite the existing output for this task. Luigi tasks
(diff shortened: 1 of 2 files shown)


edx/analytics/tasks/programs/program_reports.py
```diff
@@ -244,25 +245,25 @@ def reducer(self, key, values):
             is_course_run_completed = string_to_bool(entry.course_run_completed)
             completed = completed or is_course_run_completed
 
-            if entry.date_first_enrolled != self.sqoop_null_string:
-                year = datetime.datetime.strptime(entry.date_first_enrolled, '%Y-%m-%d %H:%M:%S.%f').year
-                entry_years.add(year)
+            if entry.course_run_start != self.sqoop_null_string:
```
schenedx, line 248: Would course runs have this date to be `null`? I think we do have those. If so, how do you handle `null` values?

edx/analytics/tasks/programs/tests/test_program_reports.py
```diff
@@ -126,7 +127,7 @@ def build_cohort_enrollment_output(
         out = OrderedDict([
             ('authoring_org', authoring_org),
             ('program_uuid', program_uuid),
-            ('entry_year', '2019'),
+            ('entry_cohort', '2019-01'),
```
schenedx, line 130: Should we have a test case where the month number is `null`?
Same feedback to all the rest of the test cases you updated.

schenedx (commented):
Very simple feedback. This sounds like a easy change overall.

edx/analytics/tasks/programs/program_reports.py
zacharis278: the only nulls I see are unpublished courses so this case is very much an edge with this update.  I figured I should keep the logic so all nulls get grouped together in the same cohort just in case.

edx/analytics/tasks/programs/tests/test_program_reports.py
zacharis278: The reducers having a code path for 'null' should already be covered from what I can tell (CombineCourseEnrollments and CountCourseEnrollments).

schenedx (approved):
👍

PR #810, opened 2020-03-12
mroytman/fix program reports pipeline bug last activity date

michaelroytman:
---

### DIFF
--- a/edx/analytics/tasks/common/vertica_export.py
+++ b/edx/analytics/tasks/common/vertica_export.py
@@ -42,7 +42,7 @@ def get_vertica_results(warehouse_name, credentials, query):
     cred = None
     with credentials_target.open('r') as credentials_file:
         cred = json.load(credentials_file)
-
+    # Externalize autocommit and read timeout
     connection = vertica_python.connect(user=cred.get('username'), password=cred.get('password'), host=cred.get('host'),
                                         port=cred.get('port'), database=warehouse_name, autocommit=False,
                                         read_timeout=None)
--- a/edx/analytics/tasks/common/vertica_export.py
+++ b/edx/analytics/tasks/common/vertica_export.py
@@ -43,7 +43,6 @@ def get_vertica_results(warehouse_name, credentials, query):
     with credentials_target.open('r') as credentials_file:
         cred = json.load(credentials_file)
 
-    # Externalize autocommit and read timeout
     connection = vertica_python.connect(user=cred.get('username'), password=cred.get('password'), host=cred.get('host'),
                                         port=cred.get('port'), database=warehouse_name, autocommit=False,
                                         read_timeout=None)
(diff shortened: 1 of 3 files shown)


edx/analytics/tasks/common/vertica_export.py
```diff
@@ -285,7 +284,7 @@ def sqoop_dump_vertica_table_task(self):
 
             if len(column_list) <= 0:
                 raise RuntimeError('Error Sqoop copy of {schema}.{table} found no viable columns!'.
-                                   format(schema=self.schema_name, table=self.table))
+                                   format(schema=self.vertica_schema_name, table=self.table_name))
```
michaelroytman, line 287: This is unrelated to this work, but, as part of my analytics pipeline setup, I hit this error... which threw an exception when trying to format the error.

edx/analytics/tasks/programs/program_reports.py
```diff
@@ -194,10 +195,34 @@ def requires(self):
             overwrite=(self.overwrite_export and self.overwrite),
         )
 
+    def _trim_fields(self, fields):
```
michaelroytman, line 198: This will decouple the SQL table from this task. I don't think that this task should fail if we neglect to include any new columns in the namedtuple definition, particularly since `last_activity_date` is not used in this report.

edx/analytics/tasks/programs/program_reports.py
```diff
@@ -184,6 +184,7 @@ class CombineCourseEnrollmentsTask(OverwriteOutputMixin, ProgramsReportTaskMixin
         'user_id', 'name', 'username', 'user_key', 'course_title', 'course_run_key', 'course_run_start', 'external_course_run_key',
         'track', 'grade', 'letter_grade', 'date_first_enrolled', 'date_last_unenrolled', 'current_enrollment_is_active',
         'first_verified_enrollment_time', 'course_run_completed', 'date_completed', 'program_completed', 'course_key', 'timestamp',
+        'last_activity_date',
```
michaelroytman, line 187: Note that I'm not including tests, since, there's no change in the result of this task. The output before and after this PR is the same.

zacharis278:
looks good.  I think we should add a test case that creates an enrollment input with extra fields just to cover the new case.  If you're planning to refactor this to be used by the learner report improvement as well then we can probably just hit it with that PR instead.

zacharis278 (approved):

michaelroytman:
@Zacharis278 Thanks for the feedback. I tried to do that originally re: testing, but because the output of both the `mapper` and `reducer` functions hasn't changed, it doesn't really matter what the enrollment input is; with or without this change, the output is the same. I could test `_trim_fields` by directly calling it, but I figured this was an implementation detail that doesn't affect the output of the task; it just prevents an exception. The test of whether this change works is whether it throws an exception. What do you think?

PR #821, opened 2020-07-02
Mysql to Snowflake: Added include list.

HassanJaveed84:
---

### DIFF
--- a/edx/analytics/tasks/warehouse/load_internal_reporting_database.py
+++ b/edx/analytics/tasks/warehouse/load_internal_reporting_database.py
@@ -823,6 +823,10 @@ class ImportMysqlDatabaseToSnowflakeSchemaTask(MysqlToSnowflakeTaskMixin, Snowfl
 
     Assumes that only one database is written to each schema.
     """
+    include = luigi.ListParameter(
+        default=(),
+        description='List of regular expressions matching database table names that should be imported from MySQL to Snowflake.'
+    )
     exclude = luigi.ListParameter(
         default=(),
         description='List of regular expressions matching database table names that should not be imported from MySQL to Snowflake.'
@@ -840,9 +844,12 @@ class ImportMysqlDatabaseToSnowflakeSchemaTask(MysqlToSnowflakeTaskMixin, Snowfl
         """
         Determines whether to exclude a table during the import.
         """
-        if any(re.match(pattern, table_name) for pattern in self.exclude):
-            return True
-        return False
+        if any(re.match(pattern, table_name) for pattern in self.include):
+            if any(re.match(pattern, table_name) for pattern in self.exclude):
+                return True
+            else:
+                return False
+        return True
 
     def requires(self):
         """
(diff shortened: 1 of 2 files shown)


edx/analytics/tasks/warehouse/load_internal_reporting_database.py
```diff
@@ -840,9 +844,12 @@ def should_exclude_table(self, table_name):
...
         """
-        if any(re.match(pattern, table_name) for pattern in self.exclude):
-            return True
-        return False
+        if any(re.match(pattern, table_name) for pattern in self.include):
+            if any(re.match(pattern, table_name) for pattern in self.exclude):
+                return True
+            else:
+                return False
+        return True
```
brianhw, line 852: Logic here doesn't quite look right to me.   If there is no INCLUDE being set, then I would want it to include everything, and only apply the exclude list.  That way we could merge this code and it wouldn't break the current settings. 

If it's easier, I'd also propose that we could ignore the EXCLUDE list if there is an INCLUDE list.   I'm not imagining that we really need to support both at the same time.  

So there would be an "if self.include:" clause, then an "elif self.exclude:" clause (or just a test of the "if any(re.match(pattern, table_name) for pattern in self.exclude):" directly).

brianhw (commented):
I was also hoping that we could switch to using ImportMysqlDatabaseFromS3ToSnowflakeSchemaTask instead of ImportMysqlDatabaseToSnowflakeSchemaTask.   That would be a change in the jobs themselves.

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
HassanJaveed84: Ah, I think I confused the requirements. I thought we wanted to limit it to whatever is in INCLUDE only.

HassanJaveed84: This is from the A/C:
Add a new includes feature that:
* If anything is in the INCLUDE list, only these tables are copied (INCLUDE list takes precedence over exclude list)
* If EXCLUDE_FIELD is has value(s) it should still be applied to the items in the INCLUDES list

So the second point does mention of still applying the exclude.

HassanJaveed84: But what you said makes sense. So this is what I'm going to do:

* Include everything if INCLUDE is empty. But still apply EXCLUDE.
* Include only INCLUDE if INCLUDE is set and ignore EXCLUDE altogether.

Is this what we want ?

brianhw: I think so.  Note that the EXCLUDE_FIELD mentioned in the A/C is different than EXCLUDE, in that it deals with columns and not tables.

HassanJaveed84: Thanks!
I think my brain casually chose to ignore _FIELD there in the A/C..

edx/analytics/tasks/warehouse/tests/test_internal_reporting_database.py
```diff
@@ -153,3 +153,37 @@ def test_table_schema(self, mysql_query_results_mock):
...
             self.assertDictEqual(actual_schema, expected_schema)
+
+@ddt
+class ImportMysqlToSnowflakeTaskTest(TestCase):
+    """Test for ImportMysqlDatabaseToSnowflakeSchemaTask."""
+
+    def setUp(self):
+        self.task = ImportMysqlDatabaseToSnowflakeSchemaTask(
+            include=('auth_userprofile$', 'auth_user', 'assessment_assessmentpart', 'edxval_video'),
+            exclude=('auth_user$', 'assessment_assessment'),
```
pwnage101, line 164: Do you think it would add value to have two more tests: one with an empty exclude list, and another with an empty include list?

PR #822, opened 2020-07-08
[BD-18] switch from nose to pytest

emZubair:
## Switch from nose to pytest

**Packages removed**

- nose

**Packages added**

- pytest-cov


openedx-webhooks:
Thanks for the pull request, @emzubair! I've created [BLENDED-416](https://openedx.atlassian.net/browse/BLENDED-416) to keep track of it in Jira.

When this pull request is ready, tag your edX technical lead.

PR #819, opened 2020-06-25
Removed all BigQuery related code.

HassanJaveed84:
Analytics Pipeline Pull Request
---

Make sure that the following steps are done before merging:

  - [ ] If you have a migration please contact data engineering team before merging.
  - [ ] Before merging run full acceptance tests suite and provide URL for the acceptance tests run.
  - [ ] A member of data engineering team has approved the pull request.

PR #824, opened 2020-07-09
Decouple sqoop and warehouse loads.

HassanJaveed84:
---


edx/analytics/tasks/warehouse/load_internal_reporting_database.py
```diff
@@ -86,81 +87,269 @@ def quote_character(self):
...
         return self.required_tasks
 
+    def output(self):
+        return get_target_from_url(self.s3_location_for_table)
+
+    def complete(self):
+        return self.sqoop_export_mysql_table_task.complete()
+
     def mysql_compliant_schema(self):
-        if not self.mysql_table_schema:
```
brianhw, line 126: Do you not think the caching is useful?   I thought that this method was being called more than once, so I added caching to avoid multiple queries being made.

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
```diff
@@ -86,81 +87,269 @@ def quote_character(self):
...
     """
-    Parameters for importing a mysql database into Vertica.
+    Parameters for exporting a mysql database to S3.
     """
 
     # Don't use the same source for BigQuery loads as was used for Vertica loads,
     # until their formats and exclude-field parameters match.
     warehouse_subdirectory = luigi.Parameter(
-        default='import_mysql_to_vertica',
+        default='import_mysql_to_s3',
```
brianhw, line 98: I guess it doesn't hugely matter, but all the past loads are in the Vertica directory.   But this is clearer going forward, so if we're going to go forward for a while, then this would be better.

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
```diff
@@ -86,81 +87,269 @@ def quote_character(self):
...
+            mysql_schema = self.mysql_compliant_schema()
+            column_names = [field_name for (field_name, _field_type, _field_null) in mysql_schema]
+            partition_path_spec = HivePartition('dt', self.date.isoformat()).path_spec
+            destination = url_path_join(
+                self.warehouse_path,
+                self.warehouse_subdirectory,
+                self.database,
+                self.table_name,
+                partition_path_spec
+            ) + '/'
```
brianhw: Is this different than what is produced by self.s3_location_for_table?

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
```diff
@@ -86,81 +87,269 @@ def quote_character(self):
...
+        description='List of regular expression patterns for matching the names of tables that should not be output.',
+    )
+
+    def should_exclude_table(self, table_name):
+        """Determines whether to exclude a table during the import."""
+
+        if self.include:
+            return not any(re.match(pattern, table_name) for pattern in self.include)
+        elif self.exclude:
+            return any(re.match(pattern, table_name) for pattern in self.exclude)
```
brianhw, line 224: Looks good!

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
```diff
@@ -86,81 +87,269 @@ def quote_character(self):
...
+            self.required_tasks = {
+                'credentials': ExternalURL(url=self.credentials),
+                'insert_source': self.insert_source_task,
+            }
+        return self.required_tasks
 
     def vertica_compliant_schema(self):
         """Transforms mysql table schema into a vertica compliant schema."""
 
-        if not self.table_schema:
```
brianhw, line 141: Again, did you find that caching wasn't useful?

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
```diff
@@ -86,81 +87,269 @@ def quote_character(self):
...
+class LoadMysqlTableFromS3ToVerticaTask(MysqlTableExportMixin, VerticaCopyTask):
+    """
+    Task to import a MySQL table from S3 into vertica.
+    """
+    table_name = luigi.Parameter(
+        description='The name of the table.',
+    )
+    # Some fields are no longer needed when reading from S3.
+    exclude_field = None
+    db_credentials = None
```
brianhw, line 311: Should we instead just create a more focused mixin class that doesn't include these?   Not sure how much work that would be, and whether it would be worth the effort, but it might be clearer to add these where needed rather than subtracting where they're not?

brianhw (approved):
No showstoppers -- looks good!   👍

pwnage101 (approved):
LGTM, besides what brian has brought up.

I want to say however that for me this is one of the more hard-to-read codebases at edx and takes the longest for me to review, in part because of the mixin hierarchy (too many layers of abstraction & never knowing what the parameters are), but also the general lack of comments and detailed docstrings. I know this codebase is going out the door, so I just us want to keep this in mind for prefect and other codebases that might grow to this size.

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
HassanJaveed84: This method gets called in the `sqoop_export_mysql_table_task` method, which is already cached. I though that would be enough to avoid repeated calls. Thoughts ?

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
HassanJaveed84: This method does not make any database calls as it uses the metadata from S3.

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
HassanJaveed84: Good catch.

HassanJaveed84:
Acceptance test run: http://jenkins-ci.analytics.edx.org/view/ad-hoc/job/edx-analytics-pipeline-acceptance-test-manual/1866

edx/analytics/tasks/warehouse/load_internal_reporting_database.py
HassanJaveed84: We're going to be removing BigQuery related code soon. I have added this as an additional A/C on that PR: 
https://github.com/edx/edx-analytics-pipeline/pull/819#issuecomment-656340242
I think it will make it easier once all the BQ related code is removed.

PR #833, opened 2020-09-03
Add 502 and 520 to the edxapp api client retry codes

pwnage101:
This seems to mitigate an issue we're having with the load-course-structure jenkins job where it gives up early due to a 502 response that never got retried: http://jenkins.analytics.edx.org/job/load-course-structure/108

Here's a successful run which took 28 hours instead of typical 20 hours: http://jenkins.analytics.edx.org/job/load-course-structure/109

We've seen _both_ 502 and 520, hence adding retries on both in this PR.

Related to OpsGenie alerts `#6440` and `#6242`


brianhw (approved):
LGTM.  👍

PR #836, opened 2020-10-01
Add extraction of course_id from a block-based URL.

doctoryes:
https://openedx.atlassian.net/browse/DENG-522

At some point in August, the format of most course-based URLs changed from:

`https://courses.edx.org/courses/course-v1:org+course+run/courseware/unit1/der_3-sequential/?activate_block_id=block-v1%3Aorg%2Bcourse%2Brun%2Btype%40sequential%2Bblock%40der_3-sequential`

to:

`https://courses.edx.org/xblock/block-v1:org+course+run+type@vertical+block@3848270e75f34e409eaad53a2a7f1da5?show_title=0&show_bookmark_button=0`

This PR adds the parsing of course_ids from the "xblock"-based course URLs which are in events. These course_ids are currently ending up as "" (empty string) and are being ignored by the pipeline code.

---

### DIFF
--- a/edx/analytics/tasks/insights/user_activity.py
+++ b/edx/analytics/tasks/insights/user_activity.py
@@ -55,9 +55,19 @@ class UserActivityTask(OverwriteOutputMixin, WarehouseMixin, EventLogSelectionMi
             log.error("User-Activity: event without user_id in context: %s", event)
             return
 
+        # Course user activity URLs have changed recently with the introduction of micro-frontends (MFEs).
+        # This code attempts to handle those URL changes with minimal diffences in the number of events processed/used.
+        #
+        # First, attempt to extract the course_id in the historical way - without attempting to extract it
+        # from the event URL.
         course_id = eventlog.get_course_id(event)
         if not course_id:
-            return
+            # If the first attempt fails to extract a course_id, then attempt to extract the course ID
+            # from the URL as well. (This logic is new, so only use it when the above fails.)
+            course_id = eventlog.get_course_id(event, from_url=True)
+            if not course_id:
+                # If a course_id *still* has not been extracted successfully, ignore this event.
+                return
 
         for label in self.get_predicate_labels(event):
             yield date_string, self._encode_tuple((str(user_id), course_id, date_string, label))
@@ -79,6 +89,12 @@ class UserActivityTask(OverwriteOutputMixin, WarehouseMixin, EventLogSelectionMi
         labels = [ACTIVE_LABEL]
 
         if event_source == 'server':
+
+            # Ignore any server-based implicit events for the purposes of user activity.
+            # All implicit events have a source of 'server' and have a type beginning with '/'.
+            if event_type.startswith('/'):
+                return []
+
             if event_type == 'problem_check':
                 labels.append(PROBLEM_LABEL)
 
--- a/edx/analytics/tasks/util/opaque_key_util.py
+++ b/edx/analytics/tasks/util/opaque_key_util.py
@@ -5,16 +5,20 @@ import re
 
 from opaque_keys import InvalidKeyError
 from opaque_keys.edx.keys import CourseKey
-from opaque_keys.edx.locator import CourseLocator
+from opaque_keys.edx.locator import BlockUsageLocator, CourseLocator
 
 log = logging.getLogger(__name__)
 
-
-# from lms/envs/common.py:
-COURSE_KEY_PATTERN = r'(?P<course_key_string>[^/+]+(/|\+)[^/+]+(/|\+)[^/]+)'
+# Regex to extract a course_id from a course URL.
+# from openedx/core/constants.py:
+COURSE_KEY_PATTERN = r'(?P<course_key_string>[^/+]+(/|\+)[^/+]+(/|\+)[^/?]+)'
 COURSE_ID_PATTERN = COURSE_KEY_PATTERN.replace('course_key_string', 'course_id')
-# from common/djangoapps/util/request.py:
-COURSE_REGEX = re.compile(r'^.*?/courses/{}'.format(COURSE_ID_PATTERN))
+# from openedx/core/lib/request_utils.py:
+COURSE_REGEX = re.compile(r'^(.*?/courses/)(?!v[0-9]+/[^/]+){}'.format(COURSE_ID_PATTERN))
+
+# Regex to extract a course_id from a block URL/block ID.
+BLOCK_ID_PATTERN = r'(?P<block_id>[^/+]+(/|\+)[^/+]+(/|\+)[^/?#]+)'
+BLOCK_REGEX = re.compile(r'^(.*?/xblock/){}'.format(BLOCK_ID_PATTERN))
 
 # Make sure that Opaque Keys' Stevedore extensions are loaded, this can sometimes fail to happen in EMR
 # and it is vitally important that they be there, or jobs will succeed but produce incorrect data. See
@@ -100,13 +104,22 @@ def get_course_key_from_url(url):
     """
     url = url or ''
 
-    match = COURSE_REGEX.match(url)
-    course_key = None
-    if match:
-        course_id_string = match.group('course_id')
+    # First, try to extract the course_id assuming the URL follows this pattern:
+    # https://courses.edx.org/xblock/block-v1:org+course+run+type@vertical+block@3848270e75f34e409eaad53a2a7f1da5?show_title=0&show_bookmark_button=0
+    block_match = BLOCK_REGEX.match(url)
+    if block_match:
+        block_id_string = block_match.group('block_id')
         try:
-            course_key = CourseKey.from_string(course_id_string)
+            return BlockUsageLocator.from_string(block_id_string).course_key
         except InvalidKeyError:
-            pass
+            return None
 
-    return course_key
+    # Second, try to extract the course_id assuming the URL follows this pattern:
+    # https://courses.edx.org/courses/course-v1:org+course+run/courseware/unit1/der_3-sequential/?activate_block_id=block-v1%3Aorg%2Bcourse%2Brun%2Btype%40sequential%2Bblock%40der_3-sequential
+    course_match = COURSE_REGEX.match(url)
+    if course_match:
+        course_id_string = course_match.group('course_id')
+        try:
+            return CourseKey.from_string(course_id_string)
+        except InvalidKeyError:
+            return None
--- a/edx/analytics/tasks/util/tests/test_opaque_key_util.py
+++ b/edx/analytics/tasks/util/tests/test_opaque_key_util.py
@@ -5,17 +5,19 @@ from unittest import TestCase
 
 from ccx_keys.locator import CCXLocator
 from ddt import data, ddt, unpack
-from opaque_keys.edx.locator import CourseLocator
+from opaque_keys.edx.locator import BlockUsageLocator, CourseLocator
 
 import edx.analytics.tasks.util.opaque_key_util as opaque_key_util
 
-VALID_COURSE_ID = unicode(CourseLocator(org='org', course='course_id', run='course_run'))
+VALID_COURSE_KEY = CourseLocator(org='org', course='course_id', run='course_run')
+VALID_COURSE_ID = unicode(VALID_COURSE_KEY)
 VALID_LEGACY_COURSE_ID = "org/course_id/course_run"
 INVALID_LEGACY_COURSE_ID = "org:course_id:course_run"
 INVALID_NONASCII_LEGACY_COURSE_ID = u"org/course\ufffd_id/course_run"
 VALID_NONASCII_LEGACY_COURSE_ID = u"org/cours\u00e9_id/course_run"
 VALID_CCX_COURSE_ID = unicode(CCXLocator(org='org', course='course_id', run='course_run', ccx='13'))
 COURSE_ID_WITH_COLONS = unicode(CourseLocator(org='org', course='course:id', run='course:run'))
+VALID_BLOCK_ID = BlockUsageLocator(course_key=VALID_COURSE_KEY, block_type='video', block_id='Welcome')
 
 
 @ddt
@@ -100,6 +102,15 @@ class CourseIdTest(TestCase):
         self.assertEquals(unicode(course_key), course_id)
 
     @data(
+        VALID_BLOCK_ID,
+    )
+    def test_get_course_key_from_url(self, block_id):
+        url = u"https://courses.edx.org/xblock/{block_id}?stuff=things".format(block_id=block_id)
+        print(url)
+        course_key = opaque_key_util.get_course_key_from_url(url)
+        self.assertEquals(unicode(course_key), VALID_COURSE_ID)
+
+    @data(
         INVALID_LEGACY_COURSE_ID,
         INVALID_NONASCII_LEGACY_COURSE_ID,
         None,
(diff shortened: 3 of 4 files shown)


edx/analytics/tasks/insights/user_activity.py
```diff
@@ -55,7 +55,7 @@ def mapper(self, line):
             log.error("User-Activity: event without user_id in context: %s", event)
             return
 
-        course_id = eventlog.get_course_id(event)
+        course_id = eventlog.get_course_id(event, from_url=True)
```
doctoryes: @brianhw What's the easiest way to test the impact of this change? To check numbers of events processed before it vs. after it?

brianhw: First of all, my major concern with making this change is that it will identify many more events than were being counted before.   I don't know that we ever made informed decisions about what should count and what shouldn't -- we just did stuff back in the dark ages, and have been afraid to make changes since.   Part of the reason is that it takes time and energy to explain why things change, and that's in part because it may take a lot of work to actually figure out what changed and by how much, and partly because Insights users would (rightly?) be concerned that something changed in their courses.  So there's great bias to not making changes.  But here our hand is forced, so I would look to minimize change.

To that end, I would change the logic here to first call get_course_id(event), and if that is None *and* event_source is 'browser', then call get_course_id(event, from_url=True).   That way, we specifically don't start parsing implicit events.   (I'm assuming here that implicit events (i.e. those whose event_type begin with a slash) have an event_source that is "server" instead of "browser" -- but we could put an explicit check for that, I guess.)

edx/analytics/tasks/util/opaque_key_util.py
```diff
@@ -5,16 +5,20 @@
...
+COURSE_KEY_PATTERN = r'(?P<course_key_string>[^/+]+(/|\+)[^/+]+(/|\+)[^/?]+)'
 COURSE_ID_PATTERN = COURSE_KEY_PATTERN.replace('course_key_string', 'course_id')
-# from common/djangoapps/util/request.py:
-COURSE_REGEX = re.compile(r'^.*?/courses/{}'.format(COURSE_ID_PATTERN))
+# from openedx/core/lib/request_utils.py:
+COURSE_REGEX = re.compile(r'^(.*?/courses/)(?!v[0-9]+/[^/]+){}'.format(COURSE_ID_PATTERN))
+
+# Regex to extract a course_id from a block URL/block ID.
+BLOCK_ID_PATTERN = r'(?P<block_id>[^/+]+(/|\+)[^/+]+(/|\+)[^/?]+)'
+BLOCK_REGEX = re.compile(r'^(.*?/xblock/){}'.format(BLOCK_ID_PATTERN))
```
brianhw, line 21: I don't have a great idea with regexps as to whether we're missing cases or not.   The original code in opaque keys is https://github.com/edx/opaque-keys/blob/36ae1b3732ccaa8a4d4a864c640051f2924991a6/opaque_keys/edx/locator.py#L100-L124, and it looks ... hairy.   Maybe this is good enough, especially since we also validate with opaque keys.

edx/analytics/tasks/insights/user_activity.py
doctoryes: I'll make the suggested change. 

@brianhw How do you suggest I test it on production data? Could I possibly run this branch in a user-quality-production pipeline on a day before the event decrease? And then a recent day of events? If so, what's the easiest way for me to modify this pipeline so that it'd save the transformed data to a location other than `production.f_user_activity`?

brianhw: As for testing -- it might be reasonable to run this kind of thing *outside* of production.  For some perspective on what this means, you could look at the old documentation on https://openedx.atlassian.net/wiki/spaces/AN/pages/154861660/Analytics%2BPipeline%2BRelease%2BValidation%2BNotes.   This talks about "stage" versus "prod", but we have jobs instead defined for "dev" versus "prod", to indicate it's prod data but not shipping to prod.   Unfortunately, there's actually not much there about user_activity, which surprises me.  But I could imagine running the branch using a "dev" build, and with some amount of history copied over from prod, and then looking at what it produces.

brianhw: The job that was intended for running 'dev' is actually http://jenkins.analytics.edx.org/view/Release/job/user-activity-release/.   It's never been run since being moved off of scheduler.   We've so rarely made changes to the pipeline in recent years.

brianhw: It uses analytics-tasks/dev-edx.cfg, so that runs with warehouse_path with dev instead of prod.   You can copy historic data from prod to dev for "user_activity_by_user" up to 8/15 or whatever (or copy it all and then delete), and then run with something that makes sure to recalculate going back far enough.  (That is, you might need to set overwrite-n-days in the EXTRA_ARGS to be big enough to fill in the gap on the user_activity_by_user table.   And then there's "course_activity" that aggregates this data, by week, over a period of 24 weeks by default.)

brianhw: The dev configuration also points to a different result store Mysql database.  I've no idea if that still works.  Different server, different db name.   (See [database-export] section of dev-edx.cfg versus prod-edx.)  The job does not write to Vertica -- that's currently done by load-warehouse.   Not sure if there's an easier way to just load prod and dev from S3 into Snowflake directly, and do comparisons there.

doctoryes: @brianhw Thanks for the testing info. I'll try to struggle through it all.

But I don't actually understand the synthetic event comment above. I looked at all the tracking log events on a particular day with a leading slash and saw this:
```
select event:event_source, count(*) from PROD.TRACKING_EVENTS.TRACKING_EVENTS
WHERE date(time) > '2020-08-23'
AND date(time) <= '2020-08-24'
AND regexp_like(event:event_type, '^/.*')
group by 1 order by 2 desc;
+--------------------+----------+
| EVENT:EVENT_SOURCE | COUNT(*) |
|--------------------+----------|
| "server"           | 16727411 |
| "browser"          |    66750 |
+--------------------+----------+
```
But there's only a single event type with a "browser" source on that day:
```
select event:event_source, event:event_type, count(*) from PROD.TRACKING_EVENTS.TRACKING_EVENTS
WHERE date(time) > '2020-08-23'
AND date(time) <= '2020-08-24'
AND regexp_like(event:event_type, '^/.*')
AND event:event_source = 'browser'
group by 1, 2 order by 3 desc
LIMIT 10000;
+--------------------+-------------------------------------------------------+----------+
| EVENT:EVENT_SOURCE | EVENT:EVENT_TYPE                                      | COUNT(*) |
|--------------------+-------------------------------------------------------+----------|
| "browser"          | "//courses.edx.org/asset-v1.undefined_undefined.hxjs" |    66750 |
+--------------------+-------------------------------------------------------+----------+
```
Should I put any logic in to explicitly check for and discard synthetic events?

doctoryes: Oops - I confuddled synthetic events with implicit events. I meant implicit events in the text above. Can you please explain again what implicit events are? Is there some doc I could read to understand them? Are all the events with an event_source of 'server' implicit events? If so, can I just filter them all out with a check in the code?

brianhw: A clarification on nomenclature:   a "synthetic" event is one that we make up based on enrollment states because we believe an actual event is missing.   What I was referring to were what I call "implicit" events, which are not generated by specific events happening in the software but rather are automatically emitted by the LMS on each request it receives.  

That said, the query information above is very encouraging.   If one filters on "browser" alone, then it will not include implicit events.   The only browser event you found that came in were in a class of "hxjs" events, and it looks like it's a bogus event when something is not defined.   I'm guessing that other .hxjs events come in without the leading slash.  But it doesn't matter, because it's not going to find a course_id in the URL anyway.    So you could filter events with leading slashes or not, and it wouldn't matter.   For fixing this bug.   That is, the bug that was uncovered was specifically with "browser" events, not with server events, so doing a fix limited to browser events seems reasonable.   But maybe there are server events that would also be uncovered by making this change -- in which case we would want to filter on leading slash *instead* of on source='browser'.  

Okay, having said (written) that, I went back to look at the eventing in `common/djangoapps/track/views/__init__.py`, and the server_track() method also gets course_id information from URL.  It wasn't as obvious as in user_track(), which has a call to `contexts.course_context_from_url(page)`.  But server_track() calls `context_override = _get_course_context(page)`.  It differs only in that it calls `course_context = contexts.course_context_from_url(page)` and cleans up after the zero-length strings.  Which is little different.  

Hmm.  And that opens a much broader Pandora's Box here.   Maybe it *is* including implicit events already, if it's able to get information from *its* page values.  And maybe this *does* include about-page visits by unenrolled users (as I had remembered), which arguably should not be counted as user activity.   It may be just that I had forgotten the mechanism by which it happens.  I thought it was the from_url=True mechanism, and was surprised that it was only used by event export.   But if it's in server_track() and user_track(), then a broader class of events are getting course_id values that then causes them to be included in user activity.

brianhw: (I should definitely refresh before responding -- that's twice that you've replied and I didn't see it until after I replied.... : )

brianhw: For the implementation of implicit events, look at edx-platform/common/djangoapps/track/middleware.py.

doctoryes: OK, I'm going to take a day's events and try to answer the question of whether implicit events have always been processed by user_activity, since it's relevant to how we filter events in a fix that we create to accommodate the new-style `xblock` URL.

doctoryes:
FYI: Soon I plan to execute the plan described here:

https://openedx.atlassian.net/browse/DENG-522?focusedCommentId=498316

where I'll test this branch on actual production data, producing actual production results.

The day I've chosen to re-run is 2020-08-24, which was right before the MFE URL began to be phased in. Depending on the results of that run, I'll make further changes -or- merge and run on more days.

edx/analytics/tasks/util/tests/test_opaque_key_util.py
```diff
@@ -110,3 +121,15 @@ def test_get_course_key_from_invalid_url(self, course_id):
...
         self.assertIsNone(course_key)
+
+    @data(
+        None,
+        '',
+        'Invalid+Block+ID',
+        '\n',
+    )
+    def test_get_course_key_from_invalid_block_url(self, block_id):
+        url = u"https://courses.edx.org/xblock/{block_id}/stuff=things".format(block_id=block_id)
```
pwnage101: For this test, are we testing the invalid data from the data-driven-test data in the decorator, or are we testing whether it can end with `?stuff=things` vs. `/stuff=things`?

edx/analytics/tasks/insights/user_activity.py
```diff
@@ -243,30 +253,31 @@ class CourseActivityPartitionTask(WeeklyIntervalMixin, UserActivityDownstreamMix
...
+        #     cal.iso_week_start,
+        #     cal.iso_week_end,
+        #     act.category;
+        # """.format(
+        #     database_name=hive_database_name(),
+        #     table=self.hive_table_task.table,
+        #     partition=self.partition,
+        #     interval_start=self.interval.date_a.isoformat(),
+        #     interval_end=self.interval.date_b.isoformat(),
+        # )
```
pwnage101: I assume this is temporarily commented for testing purposes.

doctoryes: Correct.

edx/analytics/tasks/util/tests/test_opaque_key_util.py
doctoryes: It's the former - and the URL confuses the issue. I'll change it to end as `?stuff=things`.

doctoryes:
@pwnage101 @brianhw I've removed my test code from this branch and I'm ready for a re-review.

edx/analytics/tasks/util/opaque_key_util.py
pwnage101: If nothing else, if this is just a backport of new edx-platform code then it seems like an incremental improvement, isn't it?

pwnage101 (approved):

edx/analytics/tasks/insights/user_activity.py
```diff
@@ -79,6 +89,12 @@ def get_predicate_labels(self, event):
         labels = [ACTIVE_LABEL]
 
         if event_source == 'server':
+
+            # Ignore any server-based implicit events for the purposes of user activity.
+            # All implicit events have a source of 'server' and have a type beginning with '/'.
+            if event_type.startswith('/'):
+                return []
```
brianhw, line 96: Not that I disagree with this step, but I'm surprised that adding it didn't have an impact on the results of user_activity counts from mid-August.  That is, I presumed that using `course_id = eventlog.get_course_id(event)` on implicit events would still have found (and included) events, and that this branch would have found pretty much the same implicit events with course_id values in URLs (put there by LMS), but then this clause would remove them.  Knowing a little more about the disposition of this would help me with my work on tracking_events_with_metadata.

edx/analytics/tasks/util/opaque_key_util.py
```diff
@@ -100,13 +104,22 @@ def get_course_key_from_url(url):
...
+    # https://courses.edx.org/xblock/block-v1:org+course+run+type@vertical+block@3848270e75f34e409eaad53a2a7f1da5?show_title=0&show_bookmark_button=0
+    block_match = BLOCK_REGEX.match(url)
+    if block_match:
+        block_id_string = block_match.group('block_id')
         try:
-            course_key = CourseKey.from_string(course_id_string)
+            return BlockUsageLocator.from_string(block_id_string).course_key
         except InvalidKeyError:
-            pass
+            return None
```
brianhw, line 115: Just curious if there are cases where it finds a block key, but it isn't valid, but that there is also a course-key as well?   Probably not common, but wondering if you know.   After all, the patterns are deliberately generous, and don't eliminate illegal characters that help flag xblock/x/x/x old-style sequences that are not actually block_id keys.   Did you happen to notice if this happened much? 

Also, is there a reason to look for a block-id before looking for a course_id?   Are most URLs now the former instead of the latter?

edx/analytics/tasks/util/opaque_key_util.py
brianhw: Hopefully these two anchors (`/courses/v../` and `/xblock/`) covers most of our cases with explicit events.  Implicit events may have other places where one might find a course_id in a URL, but we don't care about that for user activity.   But once this fix goes to master, it will hopefully also somewhat improve our event export as well.   

At first, I thought there might not besupport here for CCX-based patterns.  But it just keeps going with the third component until it reaches a slash or otherwise end of url.   So it should find those as candidates as well.  Very liberal.

brianhw (approved):
A few questions seeking background information, but it looks good!   And if it produced good results in testing, then  👍

edx/analytics/tasks/insights/user_activity.py
doctoryes: @brianhw In the analysis I did, I simply compared the old vs. new number of ACTIVE events here:

https://docs.google.com/spreadsheets/d/1gWnaul2ROlZX-D9SuejdmgA0BJyXAOO1ic2r0Z9VrvY/edit#gid=70028599

I didn't count the various code branches, seeing what implicit events were included before vs. after. My assumptions:
- implicit events were filtered by the old code just based on the existing format of the implicit events
- any new implicit events were the result of the parsing code changes

edx/analytics/tasks/util/opaque_key_util.py
doctoryes: I did not check or notice whether events that didn't have a block ID in the URL had a course ID instead.

The MFE URL changes placed a block ID in the url of the event, as shown in this PR's description. I didn't perform an exhaustive analysis of old vs. new counts of course IDs vs block IDs and their location within a tracking event.

PR #838, opened 2020-10-14
Unpin pip and setuptools, upgrade virtualenv

bmedx:
SSL errors with pip have forced these upgrades, this may be the last time we can do this kind of upgrade without also upgrading all of our EMR infrastructure / OS versions / Python version.

---

### DIFF
--- a/share/task.yml
+++ b/share/task.yml
@@ -135,8 +135,10 @@
       sudo: True
       when: ansible_distribution in common_redhat_variants
 
+    # 2020-10-14: Virtualenv needed to be upgraded to get a newer version of pip installed
+    # This version was the newest that would work with our version of setuptools.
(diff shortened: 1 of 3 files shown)


share/task.yml
```diff
@@ -136,7 +136,7 @@
       when: ansible_distribution in common_redhat_variants
 
     - name: virtualenv installed
-      pip: name=virtualenv version=1.10.1
+      pip: name=virtualenv version=16.7.10
```
brianhw, line 141: We might want to add a comment here about how this particular version was chosen, and why.

brianhw (approved):
I'm guessing that the virtualenv change might be the only change that we actually needed to make.   But I think the other changes should be okay as well.   If not, we can go back and roll those parts back.

PR #842, opened 2020-11-10
Remove implicit event exclusion and multiple calls to get_course_id().

doctoryes:
---

This PR rolls back two of the changes in this previous PR: https://github.com/edx/edx-analytics-pipeline/pull/836

I had incorrectly assumed that my changes enabled the counting of implicit ("server"-sourced) events, when they actually accounted for 51% of previously-counted ACTIVE user activity events. Also, I had added an unnecessary extra call to get_course_id() which I've removed here.

### DIFF
--- a/edx/analytics/tasks/insights/user_activity.py
+++ b/edx/analytics/tasks/insights/user_activity.py
@@ -58,16 +58,13 @@ class UserActivityTask(OverwriteOutputMixin, WarehouseMixin, EventLogSelectionMi
         # Course user activity URLs have changed recently with the introduction of micro-frontends (MFEs).
         # This code attempts to handle those URL changes with minimal diffences in the number of events processed/used.
         #
-        # First, attempt to extract the course_id in the historical way - without attempting to extract it
-        # from the event URL.
-        course_id = eventlog.get_course_id(event)
+        # Attempt to extract the course_id. The get_course_id() method will first look for an explicit course ID in the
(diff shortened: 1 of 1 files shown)


edx/analytics/tasks/insights/user_activity.py
```diff
@@ -58,16 +58,13 @@ def mapper(self, line):
...
+        course_id = eventlog.get_course_id(event, from_url=True)
         if not course_id:
-            # If the first attempt fails to extract a course_id, then attempt to extract the course ID
-            # from the URL as well. (This logic is new, so only use it when the above fails.)
-            course_id = eventlog.get_course_id(event, from_url=True)
-            if not course_id:
-                # If a course_id *still* has not been extracted successfully, ignore this event.
-                return
+            # If a course_id has not been extracted successfully, ignore this event.
+            return
```
macdiesel, line 67: Should we log or count these?

doctoryes: Yes - I should add a `self.incr_counter()` here, in case anyone wants to look at the EMR logs in the future to see how many were ignored. I'll do it in a subsequent PR - I'm going to merge this so that the overnight run uses this new code, as I've already fixed the historical data.

doctoryes: https://github.com/edx/edx-analytics-pipeline/pull/848

PR #846, opened 2020-11-24
Add URL parsing for course_id to tracking log events in Vertica

pwnage101:
DENG-628


doctoryes (approved):
I'll park my 👍 here for easy merging after any needed back-filling is done.

PR #857, opened 2021-01-26
Remove code referencing Vertica and delete deprecated jobs.

bmedx:
---


setup.cfg
```diff
@@ -56,36 +55,7 @@ edx.analytics.tasks =
     student_engagement = edx.analytics.tasks.data_api.student_engagement:StudentEngagementTask
 
     # warehouse:
-    event-type-dist = edx.analytics.tasks.warehouse.event_type_dist:PushToVerticaEventTypeDistributionTask
-    load-course-catalog = edx.analytics.tasks.warehouse.load_internal_reporting_course_catalog:PullDiscoveryCoursesAPIData
```
brianhw, line 60: I think you will still need this.   Or something in this file, so its tasks get loaded.

edx/analytics/tasks/warehouse/load_internal_reporting_course_catalog.py
```diff
@@ -499,28 +498,6 @@ def requires(self):
...
-    @property
-    def table(self):
-        return 'd_program_course'
-
-    @property
-    def columns(self):
-        return ProgramCourseRecord.get_sql_schema()
-
-
 class CourseSeatRecord(Record):
```
brianhw, line 501: So I'm guessing you made a conservative cut, but I expect that there is a subset of such Records and the related tasks to keep to support enrollment.  And that this is likely one of the ones that could be deleted.  The include in enrollment mentions 
```
from edx.analytics.tasks.warehouse.load_internal_reporting_course_catalog import (
    CoursePartitionTask, LoadInternalReportingCourseCatalogMixin, ProgramCoursePartitionTask
)
```
So we should only need the mixin, and the stuff related to the two tasks.   Should...    

Admitted each of the two has some upstream stuff in the file to keep.  

Anyway, we can clean this out later, if you prefer to do it in stages.

brianhw (commented):
Looks good -- only one blocker.

setup.cfg
bmedx: Good catch, added that back

edx/analytics/tasks/warehouse/load_internal_reporting_course_catalog.py
bmedx: I think I'll leave that for later and try to get this through acceptance tests as-is. If I still have time after that I'll take a swipe at it and try to whittle off a little more.

setup.cfg
```diff
@@ -56,36 +55,8 @@ edx.analytics.tasks =
...
-    load-d-user-course = edx.analytics.tasks.warehouse.load_internal_reporting_user_course:LoadUserCourseSummary
-    load-events = edx.analytics.tasks.warehouse.load_internal_reporting_events:TrackingEventRecordDataTask
-    load-f-user-activity = edx.analytics.tasks.warehouse.load_internal_reporting_user_activity:LoadInternalReportingUserActivityToWarehouse
-    load-ga-permissions = edx.analytics.tasks.warehouse.load_ga_permissions:LoadGoogleAnalyticsPermissionsWorkflow
-    load-google-sheet-vertica = edx.analytics.tasks.warehouse.load_google_sheet_to_warehouse:LoadGoogleSpreadsheetsToVerticaWorkflow
     load-google-sheet-snowflake = edx.analytics.tasks.warehouse.load_google_sheet_to_warehouse:LoadGoogleSpreadsheetsToSnowflakeWorkflow
-    load-insights= edx.analytics.tasks.warehouse.load_warehouse_insights:LoadInsightsTableToVertica
-    load-internal-active-users = edx.analytics.tasks.warehouse.load_internal_reporting_active_users:LoadInternalReportingActiveUsersToWarehouse
-    load-internal-database = edx.analytics.tasks.warehouse.load_internal_reporting_database:ImportMysqlDatabaseFromS3ToVerticaSchemaTask
```
HassanJaveed84, line 72: We would still need to keep an entry point to `load_internal_reporting_database` and retain that file with code necessary for sqooping out mysql database to S3 and loading into snowflake.

setup.cfg
```diff
@@ -56,36 +55,8 @@ edx.analytics.tasks =
...
-
-    # financial:
-    cybersource = edx.analytics.tasks.warehouse.financial.cybersource:DailyPullFromCybersourceTask
-    financial_reports  = edx.analytics.tasks.warehouse.financial.finance_reports:BuildFinancialReportsTask
-    orders = edx.analytics.tasks.warehouse.financial.orders_import:OrderTableTask
-    payment_reconcile = edx.analytics.tasks.warehouse.financial.reconcile:ReconcileOrdersAndTransactionsTask
-    paypal = edx.analytics.tasks.warehouse.financial.paypal:PaypalTransactionsByDayTask
-    paypal-report = edx.analytics.tasks.warehouse.financial.paypal_ftpreport:LoadPayPalCaseReportToVertica
-    affiliate_window = edx.analytics.tasks.warehouse.financial.fees:LoadFeesToWarehouse
+    load-course-catalog = edx.analytics.tasks.warehouse.load_internal_reporting_course_catalog:PullDiscoveryCoursesAPIData
```
HassanJaveed84: We probably don't need to keep this.

bmedx: I just re-added that because @brianhw said we did, so I'm tempted to just leave it for now. :)

setup.cfg
bmedx: Added that back in, also removed a bunch of lingering BigQuery references.

setup.cfg
```diff
@@ -56,36 +54,9 @@ edx.analytics.tasks =
...
+    load-internal-database = edx.analytics.tasks.warehouse.load_internal_reporting_database:LoadMysqlTableFromS3ToSnowflakeTask
```
bmedx, line 58: @HassanJaveed84 I've changed the entry point to this, but am not familiar with how entry points are being used here. Does this seem like a reasonable change?

setup.cfg
brianhw: Yeah, this is my fault.  We link to tasks that are defined in this file, and I was concerned that the tasks would not be found if not added to Stevedore.   But that's only needed for entry points, and we no longer use anything in this file as an entry point anymore.   It becomes like any of the utility files.   So I think Hassan is right.

brianhw (approved):
LGTM.   👍

setup.cfg
HassanJaveed84: This should suffice.

PR #872, opened 2021-11-01
BOM-2951: Replace Travis with GitHub CI

UsamaSadiq:
**Issue:** [BOM-2951](https://openedx.atlassian.net/browse/BOM-2951)

### Description
- Replace `Travis` with `GitHub CI`.


estute:
@UsamaSadiq I'm not quite sure yet, but even though the make target `test-docker` passes the `-t` flag to docker-run, the Github Actions environment is still complaining about it not being a real TTY. After some digging, I found this example, which seems to do the trick: https://github.com/gfx/example-github-actions-with-tty/blob/4c5f457c65dfe61e273e0470414699420be5e134/.github/workflows/test.yml#L18. It's a temporary fix to get this merged in. I will follow up with other teams to see if there is a better way of dealing with this issue.

estute (approved):

